PR #10, opened 2015-09-11
updated Cassandra to work with multiple outputs

ataylor4:


cassandra-plugins/src/test/java/co/cask/plugin/etl/test/ETLCassandraTest.java
```diff
@@ -181,6 +182,7 @@ public void afterTest() throws Exception {
 
   @Test
   public void testCassandra() throws Exception {
+    testCassandraRealtimeSink();
```
gokulavasan, line 184: @ataylor4 @albertshau Do you have a link to the JIRA?

gokulavasan:
LGTM :+1:

cassandra-plugins/src/test/java/co/cask/plugin/etl/test/ETLCassandraTest.java
ataylor4, line 184: @gokulavasan https://issues.cask.co/browse/CDAP-3252?jql=text%20~%20%22testBase%22

PR #11, opened 2015-09-11
Updated for compatibility with CDAP

ataylor4:
updated plugins
renamed CassandraBatchSource
changed package names in pom from cdap-etl-bacth-app to cdap-etl-batch
updated initialize methods to take BatchRuntimeContext


albertshau:
:+1:

PR #16, opened 2015-09-20
(CDAP-3738) Added Elasticsearch plugin as well as ui resources

bdmogal:


sreevatsanraman:
Can we add instructions as to where to copy this config? Would be cool to do a small script to build and copy json and jar, but that can be done later. For now would be great to add instructions

bdmogal:
Addressed comments. Merging.

elasticsearch-plugins/resources/elasticsearch-plugin-batch.json
```diff
@@ -0,0 +1,3 @@
+{
+  "parents" : [ "cdap-etl-batch[0,100]" ]
```
gokulavasan, line 2: For a 3.2 compatible branch shouldn't this be [3.2.0,3.3.0) ?

elasticsearch-plugins/resources/elasticsearch-plugin-realtime.json
```diff
@@ -0,0 +1,3 @@
+{
+  "parents" : [ "cdap-etl-realtime[0,100]" ]
```
gokulavasan, line 2: same here

elasticsearch-plugins/resources/elasticsearch-plugin-batch.json
bdmogal, line 2: I agree we should probably change it, since there is only 1 version of this artifact available right now. Needs to be changed everywhere? guides, et all?

gokulavasan, line 2: All the examples in cdap-guides/cdap-etl-adapter-guide depend upon plugins that are shipped with cdap and those already have the correct parent version.

bdmogal, line 2: These things - https://github.com/cdap-guides/cdap-etl-adapter-guide/blob/develop/DBTableToHBaseTable/mysql-connector-java-5.1.35.json, etc.

PR #20, opened 2015-09-22
Use 3.2.0-SNAPSHOT. Rename jsons to include versions so they have the…

bdmogal:
… same name as the jar files


gokulavasan:
LGTM :+1:

PR #23, opened 2015-09-22
Created an ESProperties class to define ElasticSearch properties, ins…

bdmogal:
…tead of relying on the one from the CDAP repo or redefinining it in every plugin.


gokulavasan:
LGTM :+1:

PR #25, opened 2015-09-23
CDAP-3819: omit spaces in column names and properties

rohitsinha54:


cassandra-plugins/src/main/java/co/cask/plugin/etl/batch/source/BatchCassandraSource.java
```diff
@@ -92,7 +93,8 @@ public void prepareRun(BatchSourceContext context) throws Exception {
 
     if (!Strings.isNullOrEmpty(config.properties)) {
       for (String pair : config.properties.split(",")) {
-        conf.set(pair.split(":")[0], pair.split(":")[1]);
+        conf.set(CharMatcher.WHITESPACE.trimFrom(pair.split(":")[0]),
```
rohitsinha54, line 96: I am removing only the leading and trailing here as I assume the value of some key might have spaces.

PR #28, opened 2015-09-25
MongoDB Batch Plugins

gokulavasan:
Build : http://builds.cask.co/browse/CPG-PLUG1

JIRA : https://issues.cask.co/browse/CDAP-3867
- [X] Add unit tests
- [x] Testing on Cluster


mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/sink/MongoDBBatchSink.java
```diff
@@ -0,0 +1,104 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * A {@link BatchSink} that writes data to MongoDB.
+ * This {@link MongoDBBatchSink} takes a {@link StructuredRecord} in,
+ * converts it to {@link BSONWritable}, and writes it to MongoDB.
+ */
+@Plugin(type = "batchsink")
+@Name("MongoDB")
+@Description("CDAP MongoDB Batch Sink converts {@link StructuredRecord} to {@link BSONWritable} and writes to MongoDB")
```
awholegunch: These should not have any links in them; they won't be resolved, and instead you'll just have the raw text...:

```
MongoDB Batch Sink converts a StructuredRecord to a BSONWritable and writes it to a MongoDB.
```

awholegunch:
I've made edits. Please check my changes...Maybe need to change back to using "CDAP MongoDB Batch Source" in the description so people know it is CDAP's...what do you think?

nitinmotgi:
@gokulavasan UI JSON file is missing with this PR. Would it be possible to add one.

gokulavasan:
@nitinmotgi That work is captured in this JIRA https://issues.cask.co/browse/CDAP-3869. I was planning to the UI JSON in a separate PR.

gokulavasan:
@nitinmotgi Added UI json.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+        return input;
+      case ARRAY:
+        return GSON.fromJson(input, List.class);
+      case MAP:
+        return GSON.fromJson(input, Map.class);
+      case UNION:
+        if (field.getSchema().isNullableSimple()) {
+          return convertValue(input, Schema.Field.of(field.getName(), field.getSchema().getNonNullable()));
+        } else {
+          return null;
```
albertshau: seems like this should error instead. We could check the schema during configure time

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+
+
+    @Name(Properties.INPUT_FIELDS)
+    @Nullable
+    @Description("A projection document limiting the fields that appear in each document.")
+    private String inputFields;
+
+    @Name(Properties.SPLITTER_CLASS)
+    @Nullable
+    @Description("The name of the Splitter class to use.")
```
albertshau: what happens if this is not given?

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+    @Name(Properties.INPUT_QUERY)
+    @Description("Optionally filter the input collection with a query. This query must be represented in JSON " +
+      "format, and use the MongoDB extended JSON format to represent non-native JSON data types.")
+    @Nullable
+    private String inputQuery;
+
+
+    @Name(Properties.INPUT_FIELDS)
+    @Nullable
+    @Description("A projection document limiting the fields that appear in each document.")
```
albertshau: not sure what a 'projection document' is, is this mongo terminology? Let's add to the description what happens if this is not given.

mongodb-plugins/src/main/java/co/cask/plugin/etl/realtime/sink/MongoDBRealtimeSink.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public int write(Iterable<StructuredRecord> iterable, DataWriter dataWriter) throws Exception {
+    int recordCount = 0;
+    MongoCollection<Document> collection = mongoDatabase.getCollection(config.collectionName);
+    List<Document> documentList = new ArrayList<>();
+    for (StructuredRecord record : iterable) {
+      Document document = new Document();
+      for (Schema.Field field : record.getSchema().getFields()) {
+       document.append(field.getName(), record.get(field.getName()));
```
albertshau: nit: add another space

mongodb-plugins/src/test/java/co/cask/plugin/etl/testclasses/DataGeneratorSource.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+/**
+ * Realtime TestSource that emits {@link StructuredRecord} objects as needed for testing.
+ */
+@Plugin(type = "realtimesource")
+@Name("DataGenerator")
+@Description("Source that can generate test data for Real-time Stream and Table Sinks.")
+public class DataGeneratorSource extends RealtimeSource<StructuredRecord> {
```
albertshau: is this added twice? seems to be 2 DataGeneratorSource

albertshau:
some small comments, otherwise lgtm

gokulavasan:
@albertshau Thanks for the review. Addressed comments. Please take a look.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+        return Float.valueOf(input);
+      case DOUBLE:
+        return Double.valueOf(input);
+      case BYTES:
+        return input.getBytes();
+      case STRING:
+      case ENUM:
+        return input;
+      case ARRAY:
+        return GSON.fromJson(input, List.class);
```
albertshau: this will ignore the component type of the array.  For example, array of ints and array of strings we get decoded in the same way.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+      case MAP:
+        return GSON.fromJson(input, Map.class);
```
albertshau: similar comment about maps, the key and value types are lost here.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+      case UNION:
+        if (field.getSchema().isNullableSimple()) {
```
albertshau: this means nullable arrays and maps are not allowed. Also, if input is null, this will end up in a NPE.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+          return convertValue(input, Schema.Field.of(field.getName(), field.getSchema().getNonNullable()));
```
albertshau: could change convertValue to take a Schema rather than a Field. Then you don't need to create a new object here.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+    throw new IOException(String.format("Unsupported schema : %s for field '%s'", field.getSchema(), field.getName()));
```
albertshau: we should do schema validation at configure time to make sure this never happens at runtime. For example, check union fields, check that there are no record fields.

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+    emitter.emit(builder.build());
+  }
+
+  private Object convertValue(String input, Schema.Field field) throws IOException {
+    switch (field.getSchema().getType()) {
+      case NULL:
+        return null;
+      case BOOLEAN:
+        return Boolean.valueOf(input);
```
albertshau: how come we need to convert everything to a string and then convert back? does BSONObject not return the correct type?

mongodb-plugins/src/main/java/co/cask/plugin/etl/batch/source/MongoDBBatchSource.java
albertshau: does mongo support array fields? If so, does it support arrays of arbitrary type?

albertshau:
missed some things on the first pass, questions about the record conversion logic.

albertshau:
there is a RecordConverter class in cdap-etl-lib which might be useful

gokulavasan:
@albertshau Please review. I am using the RecordConverter class from cdap-etl-common. I can move the class to regular package instead of under co.cask.cdap.etl.common once https://github.com/caskdata/cdap/pull/4403 is merged.

mongodb-plugins/src/main/java/co/cask/cdap/etl/common/BSONConverter.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException;
+
+/**
+ * Converts {@link BSONObject} to {@link StructuredRecord}.
+ */
+public class BSONConverter extends RecordConverter<BSONObject, StructuredRecord> {
+  private final Schema schema;
+  private final org.apache.avro.Schema avroSchema;
+
+  public BSONConverter(String schemaString) throws IOException {
```
albertshau: think its a bit cleaner to pass in the schema object rather than parse it here.  Will encourage better error handling and the constructor won't need an exception.

albertshau:
minor comment, plus it would be good to add test cases that use more complex data types.  arrays and maps are being handled differently now than earlier in the pr.

mongodb-plugins/src/main/java/co/cask/cdap/etl/common/BSONConverter.java
gokulavasan: I had to derive both our schema and avro schema. Hence I had passed in the schemaString.

gokulavasan:
@albertshau Addressed comments. Please take a look again! Thanks!

mongodb-plugins/src/test/java/co/cask/plugin/etl/test/MongoDBTest.java
```diff
@@ -0,0 +1,282 @@
...
+    Table outputTable = outputManager.get();
+
+    // Scanner to verify the data in the table
+    Scanner scanner = outputTable.scan(null, null);
+    Row row1 = scanner.next();
+    Assert.assertNotNull(row1);
+    scanner.close();
+    Assert.assertArrayEquals("AAPL".getBytes(), row1.getRow());
+    Assert.assertEquals(10, (int) row1.getInt("num"));
+    Assert.assertEquals(23.23, row1.getDouble("price"), 0.00001);
```
albertshau, line 320: should also add a check for the "agents" field

albertshau:
1 comment, then lgtm

mongodb-plugins/src/test/java/co/cask/plugin/etl/test/MongoDBTest.java
gokulavasan, line 320: RecordPutTransformer handles only simple types and "agents" field is not a simple type. So the sink schema does not contain that type.

albertshau, line 320: ok i see. is the array type tested at all?

mongodb-plugins/src/main/java/co/cask/cdap/etl/common/BSONConverter.java
```diff
@@ -0,0 +1,89 @@
...
+      if (!VALID_TYPES.contains(field.getSchema().getType())) {
+        throw new IllegalArgumentException(String.format("Unsupported Type : Field Name : %s; Type : %s",
+                                                         field.getName(), field.getSchema().getType()));
+      }
+    }
+  }
+
+  private Object extractValue(Object object, Schema schema) {
+    Schema.Type fieldType = schema.getType();
+    switch (fieldType) {
```
albertshau, line 73: what happened to handling of nullables?

mongodb-plugins/src/test/java/co/cask/plugin/etl/test/MongoDBTest.java
gokulavasan, line 320: Yes it is, otherwise conversion of BSONObject of StructuredRecord would have failed. I am adding another test from Mongo to Mongo copy that will provide more confidence.

albertshau, line 320: ah thats right, got it

albertshau:
one question about nullable types in the source, otherwise lgtm

gokulavasan:
@albertshau Thanks for the catch! Fixed it (also adjusted the unit test to have a nullable field). Please review.

albertshau:
one more comment, plus travis failed

gokulavasan:
@albertshau Thanks for the review. Addressed comment. Travis build is passing now.

nitinmotgi:
Please add README.rst format same as https://github.com/caskdata/hydrator-plugins/blob/develop/transform-plugins/README.rst before merging.

albertshau:
lgtm after adding a readme like nitin suggested

gokulavasan:
@albertshau Included README.rst. Please review. Thanks!

albertshau:
lgtm

PR #49, opened 2015-11-01
Fixing checkstyle

nitinmotgi:


nitinmotgi:
self approving it. LGTM :+1:

PR #51, opened 2015-11-02
Encoder Plugin

nitinmotgi:
Implementation of encoder plugin -- supports BASE64, BASE32, HEX, STRING_BASE64, STRING_BASE32.

### DIFF
--- a/transform-plugins/resources/ui/Encoder.json
+++ b/transform-plugins/resources/ui/Encoder.json
@@ -9,7 +9,7 @@
         "encode" : {
           "widget" : "textbox",
           "label" : "Encode Fields",
-          "description" : "Encode fields <field>:<encode-type>[, <field>:<encode-type]*"
+          "description" : "Encode fields <field>:<encode-type>[, <field>:<encode-type>]*"
         },
         "schema" : {
           "widget": "schema",
--- a/transform-plugins/resources/ui/Encoder.json
+++ b/transform-plugins/resources/ui/Encoder.json
@@ -1,0 +1,24 @@
+{
+  "id": "Encoder",
+  "groups": {
+    "position": [ "group1" ],
+    "group1": {
+      "display": "Encode",
+      "position": [ "encode", "schema" ],
+      "fields": {
+        "encode" : {
+          "widget" : "textbox",
+          "label" : "Encode Fields",
+          "description" : "Encode fields <field>:<encode-type>[, <field>:<encode-type]*"
+        },
+        "schema" : {
+          "widget": "schema",
+          "label": "Schema",
+          "description" : "Schema of output JSON",
+          "schema-types" : [ "string" ],
+          "schema-default-type" : "string"
+        }
+      }
+    }
+  }
+}
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
@@ -174,7 +174,7 @@ public final class Encoder extends Transform<StructuredRecord, StructuredRecord>
         if (outFieldType == Schema.Type.BYTES) {
           builder.set(name, outValue);
         } else if (outFieldType == Schema.Type.STRING) {
-          builder.set(name, new String(outValue));
+          builder.set(name, new String(outValue, "UTF-8"));
         }
       }
     }
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
@@ -1,0 +1,252 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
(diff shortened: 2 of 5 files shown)


transform-plugins/resources/ui/Encoder.json
```diff
@@ -0,0 +1,24 @@
...
+  "groups": {
+    "position": [ "group1" ],
+    "group1": {
+      "display": "Encode",
+      "position": [ "encode", "schema" ],
+      "fields": {
+        "encode" : {
+          "widget" : "textbox",
+          "label" : "Encode Fields",
+          "description" : "Encode fields <field>:<encode-type>[, <field>:<encode-type]*"
```
bdmogal: nit: missing _>_ after `encode-type`

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
```diff
@@ -0,0 +1,252 @@
...
+@Description("Encodes the input field(s) using Base64, Base32 or Hex")
+public final class Encoder extends Transform<StructuredRecord, StructuredRecord> {
+  private static final Logger LOG = LoggerFactory.getLogger(Encoder.class);
+  private final Config config;
+
+  // Output Schema associated with transform output.
+  private Schema outSchema;
+
+  // Mapping of input field to encoder type. 
+  private final Map<String, EncodeType> encodeMap = Maps.newTreeMap();
```
bdmogal: `new TreeMap<>()`, since that's the standard we're moving to now.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
```diff
@@ -0,0 +1,252 @@
...
+
+  // Encoder handlers.
+  private final Base64 base64Encoder = new Base64();
+  private final Base32 base32Encoder = new Base32();
+  private final Hex hexEncoder = new Hex();
+
+  // Output Field name to type map
+  private Map<String, Schema.Type> outSchemaMap = Maps.newHashMap();
```
bdmogal: new `HashMap<>()`

bdmogal: Also, final?

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
```diff
@@ -0,0 +1,252 @@
...
+
+      // If format is not right, then we throw an exception.
+      if (params.length < 2) {
+        throw new IllegalArgumentException("Configuration " + mapping + " is in-correctly formed. " +
+                                             "Format should be <fieldname>:<encoder-type>");
+      }
+
+      String field = params[0];
+      String type = params[1].toUpperCase();
+      EncodeType eType;
```
bdmogal: So `EncodeType eType = EncodeType.valueOf(type)` will not work?

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
```diff
@@ -0,0 +1,252 @@
...
+          outValue = hexEncoder.encode(obj);
+        }
+
+        // Depending on the output field type, either convert it to 
+        // Bytes or to String. 
+        if (outFieldType == Schema.Type.BYTES) {
+          builder.set(name, outValue);
+        } else if (outFieldType == Schema.Type.STRING) {
+          builder.set(name, new String(outValue));
+        }
```
bdmogal, line 178: should we throw Exception if it is other than Bytes or String?

transform-plugins/resources/ui/Encoder.json
nitinmotgi: done

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi: done

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi: done, done

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi: done! will work.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi, line 178: The issue is it should not happen and it should be caught in configure (which we cannot today as we don't know the schema of input). But, let's say we throw here then for every record we will emit a log line. It's not a good experience. It's a lot of noise.

nitinmotgi:
@bdmogal addressed your comments please review

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
```diff
@@ -0,0 +1,224 @@
...
+        } else if (type == EncodeType.HEX) {
+          outValue = hexEncoder.encode(obj);
+        }
+
+        // Depending on the output field type, either convert it to 
+        // Bytes or to String. 
+        if (outFieldType == Schema.Type.BYTES) {
+          builder.set(name, outValue);
+        } else if (outFieldType == Schema.Type.STRING) {
+          builder.set(name, new String(outValue));
```
bdmogal: `Bytes.toString(outValue)`, to specify encoding.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
bdmogal, line 178: I see, so in that situation, will it mean that the record gets emitted as is? If so, I think its ok. However, would perhaps be good to increment a metric in that case. Do we have that capability today?

bdmogal:
1 comment and 1 question. Other than that, LGTM

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi: I would prefer to not add dependency on Bytes. `new String(value, "UTF-8")` is good IMHO.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Encoder.java
nitinmotgi, line 178: What would that metric signify and what would it mean to collect that metric. The right thing is for system to stop processing. Don't want to lose data.

nitinmotgi:
@bdmogal addressed yours comments.

bdmogal:
LGTM

PR #52, opened 2015-11-02
Implementation of Decoder Plugin

nitinmotgi:


awholegunch:
Please see https://github.com/caskdata/hydrator-plugins/pull/55 as changes to the README.rst.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decompressor.java
```diff
@@ -164,7 +164,7 @@ public void transform(StructuredRecord in, Emitter<StructuredRecord> emitter) th
       } else {
         // Now, the input field should be of type byte[]
         byte[] obj = new byte[0];
-        if (field.getSchema().getType() == Schema.Type.BYTES){
+        if (field.getSchema().getType() == Schema.Type.BYTES) {
           obj = in.get(name);
         } else {
           LOG.error("Input field '" + name + "' should be of type BYTES to decompress. It's currently of type" +
```
awholegunch: This statement needs some spaces and closing quote (top line length is 116):

```
           LOG.error("Input field '" + name + "' should be of type BYTES to decompress. It is currently of type " +
                       "'" + field.getSchema().getType().toString()) + "'";
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  private void parseConfiguration(String config) throws IllegalArgumentException {
+    String[] mappings = config.split(",");
+    for (String mapping : mappings) {
+      String[] params = mapping.split(":");
+
+      // If format is not right, then we throw an exception.
+      if (params.length < 2) {
+        throw new IllegalArgumentException("Configuration " + mapping + " is in-correctly formed. " +
```
awholegunch: is incorrectly formed.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Decodes the input fields as BASE64, BASE32 or HEX.
+ * 
+ * Please note that Encoder and Decoder might look the same right now, but 
+ * in near future they will diverge.
+ */
+@Plugin(type = "transform")
+@Name("Decoder")
+@Description("Decodes the input field(s) using Base64, Base32 or Hex")
```
awholegunch: "Decodes the input field(s) using Base64, Base32, or Hex"

nitinmotgi: done

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
nitinmotgi: done!

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decompressor.java
nitinmotgi: done

transform-plugins/src/test/java/co/cask/hydrator/plugin/DecoderTest.java
```diff
@@ -0,0 +1,196 @@
...
+  private static final Schema OUTPUTSTR = Schema.recordOf("outputstr",
+                                                       Schema.Field.of("a", Schema.of(Schema.Type.STRING)),
+                                                       Schema.Field.of("b", Schema.of(Schema.Type.STRING)));
+
+  @Test
+  public void testBase64Decoder() throws Exception {
+    String test = "This is a test for testing base64 decoding";
+    Transform<StructuredRecord, StructuredRecord> encoder =
+      new Encoder(new Encoder.Config("a:BASE64", OUTPUT.toString()));
+    encoder.initialize(null);
```
anwar6953, line 52: Instead of passing null, pass in `MockTransformContext`, in case the decoder transform initialize method does something with the context or if the superclass `Transform` does something with the context passed into the initialize method, right?

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
```diff
@@ -0,0 +1,223 @@
...
+  // Output Field name to type map
+  private final Map<String, Schema.Type> outSchemaMap = new HashMap<>();
+
+  // This is used only for tests, otherwise this is being injected by the ingestion framework.
+  public Decoder(Config config) {
+    this.config = config;
+  }
+
+  private void parseConfiguration(String config) throws IllegalArgumentException {
+    String[] mappings = config.split(",");
```
anwar6953, line 75: Using the `com.google.common.base.Splitter` class would simplify this method, especially since it handles the error cases, such as incorrect formatting (missing a colon in one of the mappings) or duplicate entries.

``` java
Map<String, String> inputConfig = Splitter.on(",").withKeyValueSeparator(":").split(config);
for (Map.Entry<String, String> entry : inputConfig.entrySet()) {
  DecoderType decoderType = DecoderType.valueOf(entry.getValue().toUpperCase());
  decodeMap.put(entry.getKey(), decoderType);
}
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
```diff
@@ -0,0 +1,223 @@
...
+    Schema inSchema = in.getSchema();
+    List<Field> inFields = inSchema.getFields();
+
+    // Iterate through input fields. Check if field name is present 
+    // in the fields that need to be decoded, if it's not then write
+    // to output as it is. 
+    for (Field field : inFields) {
+      String name = field.getName();
+
+      // Check if output schema also have the same field name. If it's not
```
anwar6953: this comment doesn't seem match the code. its doing `continue` instead of throwing exception.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
```diff
@@ -0,0 +1,223 @@
...
+
+      // Check if the input field name is configured to be decoded. If the field is not
+      // present or is defined as none, then pass through the field as is. 
+      if (!decodeMap.containsKey(name) || decodeMap.get(name) == DecoderType.NONE) {
+        builder.set(name, in.get(name));
+      } else {
+        // Now, the input field could be of type String or byte[], so transform everything
+        // to byte[] 
+        byte[] obj = new byte[0];
+        if (field.getSchema().getType() == Schema.Type.STRING) {
```
anwar6953, line 153: in the beginning, should we verify that the schema type will either be STRING or BYTES?
or else, throw an exception in the else case?

transform-plugins/README.rst
```diff
@@ -4,61 +4,55 @@ Hydrator Transform Plugin Collection
...
-- Clone Record,
-- Stream Formatter,
-- Compressor, 
-- Decompressor and 
+This project is a collection of useful transformations of data. These plugins are currently available:
+
+- CSV Parser
+- CSV Formatter
+- JSON Parser
+- JSON Parser
```
poornachandra: `JSON Parser` repeated twice.

transform-plugins/README.rst
```diff
@@ -66,18 +60,17 @@ CSV Parser
   Batch and
   Realtime
 :Description:
-  Parses an input field as CSV Record into a Structured Record. Support multi-line CSV record parsing 
-  into multiple structured records. Different formats of CSV record can be parsed using this plugin. 
-  Following are different CSV record types that are supported by this plugin: DEFAULT, EXCEL, MYSQL, RFC4180 and TDF.
+  Parses an input field as a CSV Record into a Structured Record. Supports multi-line CSV Record parsing
+  into multiple Structured Records. Different formats of CSV Record can be parsed using this plugin.
```
poornachandra, line 63: "Different formats of CSV Record can be parsed using this plugin" is redundant when read with next sentence, right?

transform-plugins/README.rst
```diff
@@ -85,16 +78,15 @@ CSV Formatter
...
 :Description:
-  Formats a structured record as a CSV Record. CSV Record formats supported are DELIMITED, EXCEL, MYSQL, RFC4180 and TDF. When the format is DELIMITED one can specify different delimiters that a CSV record should use for separting fields. 
+  Formats a Structured Record as a CSV Record. Supported CSV Record formats are DELIMITED, EXCEL, MYSQL, RFC4180, and TDF. When the format is DELIMITED, one can specify different delimiters that a CSV Record should use for separating fields.
 :Configuration:
-    **format:** Specifies the format of the CSV record to be generated,
-    **delimiter:** Specifies the delimiter to be used to generate a CSV Record. This option is available when format is specified as DELIMITED and 
-    **schema:** Specifies the output schema. Output Schema should have only field of type String. 
+  - **format:** Specifies the format of the CSV Record to be generated.
+  - **delimiter:** Specifies the delimiter to be used to generate a CSV Record. This option is available when the format is specified as DELIMITED.
+  - **schema:** Specifies the output schema. Output schema should only have fields of type String.
```
poornachandra, line 84: I think an example of an input record getting transformed into an output record along with format, delimiter and schema used will make things clearer.

transform-plugins/README.rst
poornachandra, line 63: I think it would be better if we also define what a `Structured Record` is.

transform-plugins/README.rst
poornachandra, line 84: In general, having an example for each plugin would be better I think.

transform-plugins/README.rst
```diff
@@ -102,111 +94,124 @@ JSON Parser
...
+:ID:
+  **StreamFormatter**
 :Type:
   Transform
 :Mode:
   Batch and
   Realtime
 :Description:
-  Formats a structured record as Stream format. Plugin will convert the structured record to Stream format. It will include header configuration and body configuration. The body of Stream event can be of any of the two types : CSV or JSON.
+  Formats a Structured Record as Stream format. Plugin will convert the Structured Record to Stream format.
```
poornachandra, line 139: We should define what a `Stream` means here.

transform-plugins/README.rst
```diff
@@ -102,111 +94,124 @@ JSON Parser
...
+  It will include a header and body configurations. The body of the Stream event can be either type CSV or JSON.
 :Configuration:
-    **body:** Specifies the fields from input structured record that should be included in the body of Stream event. 
-    **header:** Specifies the fields from input structured record that should be included in the header of Stream event. 
-    **format:** Specifies the format of the body. Currently supported formats are JSON, CSV, TSV and PSV.
-    **schema:** Specifies the output schema. The output schema can have only two fields. One of type STRING and the other of type MAP<STRING, STRING>.
+  - **body:** Specifies the input Structured Record fields that should be included in the body of the Stream event.
+  - **header:** Specifies the input Structured Record fields that should be included in the header of the Stream event.
+  - **format:** Specifies the format of the body. Currently supported formats are JSON, CSV, TSV, and PSV.
+  - **schema:** Specifies the output schema. The output schema can have only two fields: one of type STRING and the other of type MAP<STRING, STRING>.
```
poornachandra, line 145: If the schema is constant, then there is no need for user to configure it, right?

transform-plugins/README.rst
```diff
@@ -102,111 +94,124 @@ JSON Parser
...
   Realtime
 :Description:
-  Compresses configured fields. Multiple fields can be specified to be compressed using different compression algorithms. Currently plugin supports SNAPPY, ZIP and GZIP types of compression of fields. 
+  Compresses configured fields. Multiple fields can be specified to be compressed using different compression algorithms.
+  Plugin supports SNAPPY, ZIP, and GZIP types of compression of fields.
 :Configuration:
-    **compressor:** Specifies the configuration for compressing fields. Following is how this can be specified in JSON configuration <field>:<compressor>[,<field>:<compressor>]*,
-    **schema:** Specifies the output schema. The fields that are compressed will have the same field name, but they will be of type BYTE array. 
+  - **compressor:** Specifies the configuration for compressing fields. In JSON configuration, this is specified as ``<field>:<compressor>[,<field>:<compressor>]*``.
+  - **schema:** Specifies the output schema. The fields that are compressed will have the same field name but they will be of type BYTE array.
```
poornachandra, line 161: Is the schema unmodifiable here? If so, there is nothing for the user to configure, right?

transform-plugins/README.rst
```diff
@@ -102,111 +94,124 @@ JSON Parser
...
+  Realtime
 :Description:
-    Decompresses configured fields. Multiple fields can be specified to be decompressed using different decompression algorithms. Currently plugin supports SNAPPY, ZIP and GZIP types of decompression of fields.
+    Decompresses configured fields. Multiple fields can be specified to be decompressed using different decompression algorithms.
+    Plugin supports SNAPPY, ZIP, and GZIP types of decompression of fields.
 :Configuration:
-    **decompressor:** Specifies the configuration for decompressing fields. Following is how this can be specified in JSON configuration <field>:<decompressor>[,<field>:<decompressor>]*,
-    **schema:** Specifies the output schema. The fields that are decompressed will have the same field name, but they will be of type BYTE array or STRING.
+  - **decompressor:** Specifies the configuration for decompressing fields. In JSON configuration, this is specified as ``<field>:<decompressor>[,<field>:<decompressor>]*``.
+  - **schema:** Specifies the output schema. The fields that are decompressed will have the same field name but they will be of type BYTE array or STRING.
```
poornachandra, line 177: Same here (unmodifiable schema).

transform-plugins/README.rst
```diff
@@ -102,111 +94,124 @@ JSON Parser
...
+  Transform
+:Mode:
+  Batch and
+  Realtime
+:Description:
+  Encodes configured fields. Multiple fields can be specified to be encoded using different encoding methods.
+  Available encoding methods are STRING_BASE64, BASE64, BASE32, STRING_BASE32, and HEX.
+:Configuration:
+  - **encode:** Specifies the configuration for encode fields. In JSON configuration, this is specified as ``<field>:<encoder>[,<field>:<encoder>]*``.
+  - **schema:** Specifies the output schema. The fields that are encoded will have the same field name but they will be of type BYTE array or STRING.
```
poornachandra, line 193: Same here (unmodifiable schema).

transform-plugins/src/test/java/co/cask/hydrator/plugin/DecoderTest.java
nitinmotgi, line 52: `MockTransformContext` is currently only in Python plugin. So, we should move it out to common module to be used in other places. But, in this case, we don't use anything from `TransformContext` - so it should be fine.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
nitinmotgi, line 75: Don't want some random message pushed back to the user. I want to control how error messages are reported back to user.

transform-plugins/README.rst
nitinmotgi: fixed

transform-plugins/README.rst
nitinmotgi, line 63: It should be global. Don't think we should define it in every plugin. Because what happens if we change the definition of it.

transform-plugins/README.rst
nitinmotgi, line 145: I want to use my field names rather than trying to understand what is the name I should be using.

transform-plugins/README.rst
nitinmotgi, line 161: name of the field should be something that user should have the ability to choose.

transform-plugins/README.rst
nitinmotgi, line 177: Same answer

transform-plugins/README.rst
nitinmotgi, line 193: same here

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decoder.java
nitinmotgi: done

nitinmotgi:
@anwar6953 and @poornachandra please review.

anwar6953:
lgtm. @poornachandra ?

poornachandra:
LGTM

PR #55, opened 2015-11-03
Update README.rst

awholegunch:


transform-plugins/README.rst
```diff
@@ -4,230 +4,213 @@ Hydrator Transform Plugin Collection
...
-:Id:
+:ID: 
   **CSVParser**
-:Type:
+:Type: 
   Transform
-:Mode:
-  Batch and
+:Mode: 
+  Batch and
```
wolf31o2: Are there supposed to be whitespace after these?

awholegunch: I don't think it matters as far as the parser is concerned. However, I can (and will) take them out. Thanks!

PR #56, opened 2015-11-04
bring back maven plugins which got removed during resolving conflicts

rohitsinha54:
Some required maven plugins got removed while resolving conflicts from develop where my branch and also the the develop got rid of bundle plugin in main pom

https://github.com/caskdata/hydrator-plugins/pull/28/files#diff-600376dffeb79835ede4a0b285078036L391
While resolving conflicts I removed the other new addition plugins which should be there. I am adding them back.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
   <packaging>pom</packaging>
 
   <name>Hydrator Plugin Collection</name>
-  
+
   <url>http://docs.cdap.io/cdap/current/en/included-applications/index.html</url>
 
   <licenses>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -390,22 +390,112 @@
...
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 
-  <profiles>
```
anwar6953, line 396: why remove this? your previous PR had accidentally brought it in?

rohitsinha54, line 396: https://github.com/caskdata/hydrator-plugins/pull/28/files#diff-600376dffeb79835ede4a0b285078036L550

We removed it here previously. I think we dont want to have default modules. 

@gokulavasan : Are we getting rid of default modules ?

anwar6953:
If Gokul's response to the question in the above comments is yes, then remainder LGTM.

pom.xml
gokulavasan: We need this modules isn't it? Did you try building without this profile?

anwar6953: Seems like you removed it in the link @rohitsinha54 provided above.
He unintentionally brought it back in a previous PR:
https://github.com/caskdata/hydrator-plugins/pull/43

gokulavasan, line 396: We can remove the default modules.

anwar6953:
LGTM

pom.xml
rohitsinha54, line 396: Seems like you removed it in the link @rohitsinha54 provided above.
He unintentionally brought it back in a previous PR:
https://github.com/caskdata/hydrator-plugins/pull/43

@gokulavasan : Please look at the link in which you removed it. I brought them back and added the new modules too because it looks ideal to me. I am reverting the main pom to the one before  mine and wondering why they were removed.

rohitsinha54, line 396: Okay. Cool.

PR #58, opened 2015-11-07
Feature/hive source

rohitsinha54:
Addressed comments from here: 

https://github.com/caskdata/hydrator-plugins/pull/43
- This does the output sink setting in OutputFormatProvider
- Handle nullable types
- Other minor comments.

### DIFF
--- a/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
+++ b/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
@@ -122,7 +122,7 @@ public class HiveSchemaConverter {
           "Table schema contains field '%s' with complex type %s. Only primitive types are supported.",
           name, field.getTypeString()));
       }
-      fields.add(Schema.Field.of(name, Schema.of(getType(name, field.getTypeInfo().getPrimitiveCategory()))));
+      fields.add(Schema.Field.of(name, getType(name, field.getTypeInfo().getPrimitiveCategory())));
     }
     return Schema.recordOf("record", fields);
   }
@@ -134,28 +134,30 @@ public class HiveSchemaConverter {
    * @param category the field's {@link PrimitiveObjectInspector.PrimitiveCategory}
    * @return the {@link Schema.Type} for this field
    */
-  private static Schema.Type getType(String name, PrimitiveObjectInspector.PrimitiveCategory category) {
+  private static Schema getType(String name, PrimitiveObjectInspector.PrimitiveCategory category) {
+    System.out.println("### primitive cat " + category);
     switch (category) {
       case BOOLEAN:
-        return Schema.nullableOf(Schema.of(Schema.Type.BOOLEAN)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.BOOLEAN));
       case BYTE:
-      case CHAR:
       case SHORT:
       case INT:
-        return Schema.nullableOf(Schema.of(Schema.Type.INT)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.INT));
       case LONG:
-        return Schema.nullableOf(Schema.of(Schema.Type.LONG)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.LONG));
       case FLOAT:
-        return Schema.nullableOf(Schema.of(Schema.Type.FLOAT)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.FLOAT));
       case DOUBLE:
-        return Schema.nullableOf(Schema.of(Schema.Type.DOUBLE)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.DOUBLE));
+      case CHAR:
       case STRING:
       case VARCHAR:
-        return Schema.nullableOf(Schema.of(Schema.Type.STRING)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.STRING));
       case BINARY:
-        return Schema.nullableOf(Schema.of(Schema.Type.BYTES)).getType();
+        return Schema.nullableOf(Schema.of(Schema.Type.BYTES));
+      // We can support VOID by having Schema type as null but HCatRecord does not support VOID and since we read
+      // write through HCatSchema and HCatRecord we are not supporting VOID too for consistent behavior.
       case VOID:
-        return Schema.Type.NULL;
       case DATE:
       case TIMESTAMP:
       case DECIMAL:
--- a/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
+++ b/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
@@ -49,6 +49,8 @@ public class HiveSchemaConverter {
    */
   public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
     List<HCatFieldSchema> fields = Lists.newArrayList();
+    //TODO: Finalize a consistent way to whether do projecttion on source or not while reading and if needed
+    // do type check here
     for (Schema.Field field : schema.getFields()) {
       String name = field.getName();
       try {
--- a/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
+++ b/hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
@@ -49,14 +49,14 @@ public class HiveSchemaConverter {
    */
   public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
     List<HCatFieldSchema> fields = Lists.newArrayList();
-    //TODO: Finalize a consistent way to whether do projecttion on source or not while reading and if needed
+    //TODO: Finalize a consistent way to whether do projection on source or not while reading and if needed
     // do type check here
     for (Schema.Field field : schema.getFields()) {
       String name = field.getName();
       try {
         // this field of the schema must exist in the table
         Preconditions.checkNotNull(tableSchema.get(name), "Missing field %s in table schema", name);
-        fields.add(new HCatFieldSchema(name, getType(name, field.getSchema().getType()), ""));
+        fields.add(new HCatFieldSchema(name, getType(name, field.getSchema()), ""));
       } catch (HCatException e) {
         LOG.error("Failed to create HCatFieldSchema field {} of type {} from schema", name,
                   field.getSchema().getType());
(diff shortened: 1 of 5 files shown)


hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -49,12 +49,14 @@
    */
   public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
     List<HCatFieldSchema> fields = Lists.newArrayList();
+    //TODO: Finalize a consistent way to whether do projecttion on source or not while reading and if needed
```
nitinmotgi: Let's say Hive tables has 5 fields A,B,C,D & E. If the hive source output schema contains only B,D & E then the record has to be projected.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -88,7 +91,7 @@ private static PrimitiveTypeInfo getType(String name, Schema.Type category) {
         return TypeInfoFactory.binaryTypeInfo;
       default:
         throw new IllegalArgumentException(String.format(
-          "Schema contains field '%s' with unsupported type %s", name, category.name()));
+          "Schema contains field '%s' with unsupported type %s", name, type));
```
nitinmotgi: What's the corrective action one can take if this problem happens and why would it happen.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -67,11 +69,12 @@ public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
...
    * @param name name of the field
-   * @param category {@link Schema.Type} of the field
+   * @param schema {@link Schema} of the field
    * @return {@link PrimitiveTypeInfo} for the given {@link Schema.Type} which is compatible with Hive.
    */
-  private static PrimitiveTypeInfo getType(String name, Schema.Type category) {
-    switch (category) {
+  private static PrimitiveTypeInfo getType(String name, Schema schema) {
+    Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+    switch (type) {
```
albertshau, line 82: are there plans to support record, array, and map types?

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -127,24 +130,25 @@ public static Schema toSchema(HCatSchema hiveSchema) {
...
     switch (category) {
       case BOOLEAN:
-        return Schema.Type.BOOLEAN;
+        return Schema.nullableOf(Schema.of(Schema.Type.BOOLEAN)).getType();
       case BYTE:
       case CHAR:
       case SHORT:
       case INT:
-        return Schema.Type.INT;
+        return Schema.nullableOf(Schema.of(Schema.Type.INT)).getType();
```
albertshau: do byte, char, and short fields work when we convert a HCatRecord into a StructuredRecord?

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -49,12 +49,14 @@
    */
   public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
     List<HCatFieldSchema> fields = Lists.newArrayList();
+    //TODO: Finalize a consistent way to whether do projection on source or not while reading and if needed
```
albertshau: how come this is a todo instead of doing it? Looks like it is doing a projection, but not checking types.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
```diff
@@ -49,12 +49,14 @@
...
+    // do type check here
     for (Schema.Field field : schema.getFields()) {
       String name = field.getName();
       try {
         // this field of the schema must exist in the table
         Preconditions.checkNotNull(tableSchema.get(name), "Missing field %s in table schema", name);
-        fields.add(new HCatFieldSchema(name, getType(name, field.getSchema().getType()), ""));
+        fields.add(new HCatFieldSchema(name, getType(name, field.getSchema()), ""));
```
albertshau: Rather than creating the TypeInfo from the field schema, should just take whatever is already in the tableSchema since that is guaranteed to be correct.  Also need to perform validation to catch mismatches between the passed in Schema vs the actual HCatSchema

hive-plugins/src/main/java/co/cask/plugin/etl/batch/sink/HiveBatchSink.java
```diff
@@ -127,8 +113,56 @@ public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable, HCa
   public static class SinkOutputFormatProvider implements OutputFormatProvider {
     private final Map<String, String> conf;
 
-    public SinkOutputFormatProvider() {
-      this.conf = new HashMap<>();
+    public SinkOutputFormatProvider(BatchSinkContext context, Job job, HiveSinkConfig config) throws IOException {
+      job.setOutputKeyClass(WritableComparable.class);
```
albertshau: are these required? doesn't seem correct for multiple sinks.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/sink/HiveBatchSink.java
```diff
@@ -127,8 +113,56 @@ public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable, HCa
...
+
+    private Map<String, String> getPartitions(HiveSinkConfig config) {
+      Map<String, String> partitionValues = null;
+      if (config.partitions != null) {
+        partitionValues = GSON.fromJson(config.partitions, STRING_MAP_TYPE);
+      }
+      return partitionValues;
+    }
+
+    private Map<String, String> getConfigurationAsMap (Configuration conf) {
```
albertshau: remove empty space between 'AsMap (Configuration'

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
rohitsinha54: The inconsistent/unknown part is whether we should take schema(field and type) from the user or just the fields (like database plugin). After the above comments I am assuming we want to take the schema so I will add the type check part.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
rohitsinha54: The inconsistent/unknown part is whether we should take schema(field and type) from the user or just the fields (like database plugin). After the above comment and @nitinmotgi  comment I am assuming we want to take the schema so I will add the type check part.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/sink/HiveBatchSink.java
```diff
@@ -127,8 +113,56 @@ public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable, HCa
...
+      HCatOutputFormat.setSchema(modifiedConf, hiveSchema);
+      conf = getConfigurationDiff(originalConf, modifiedConf);
+      HiveSchemaStore.storeHiveSchema(context, config.dbName, config.tableName, hiveSchema);
+    }
+
+    private Map<String, String> getConfigurationDiff(Configuration originalConf, Configuration modifiedConf) {
+      MapDifference<String, String> mapDifference = Maps.difference(getConfigurationAsMap(originalConf),
+                                                                    getConfigurationAsMap(modifiedConf));
+      // new keys in the modified configurations
+      Map<String, String> stringStringMap = mapDifference.entriesOnlyOnRight();
```
albertshau: nit: newEntries is a more descriptive name than stringStringMap

albertshau:
please include a test for a pipeline that writes to hive and something else.

hive-plugins/src/main/java/co/cask/plugin/etl/batch/commons/HiveSchemaConverter.java
albertshau: ok got it. On a somewhat related note, would it be possible to get the HCatSchema at configure time (configurePipeline) ? Or is it only possible at runtime (beforeSubmit)

rohitsinha54: With HCatalog to get  the schema of the table we need to first set the input by calling HCatInputFormat.setInput(Job, ....) This call sets properties in the hadoop configuration. Since, conf is not available during configurePipeline I don't think it will be possible.

hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/commons/HiveSchemaConverter.java
```diff
@@ -52,9 +52,17 @@ public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
...
+        HCatFieldSchema hCatFieldSchema = tableSchema.get(name);
+        Preconditions.checkNotNull(hCatFieldSchema, "Missing field %s in table schema", name);
+        PrimitiveTypeInfo hiveType = hCatFieldSchema.getTypeInfo();
+        PrimitiveTypeInfo type = getType(name, field.getSchema());
+        if (hiveType != type) {
+          LOG.warn("The given schema {} for the field {} does not match the schema {} from the table. " +
+                     "The schema {} for field {} be used.", type, name, hiveType, hiveType, name);
+          type = hiveType;
+        }
+        fields.add(new HCatFieldSchema(name, type, ""));
```
albertshau: could you add hCatFieldSchema directly or is this different?

hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/HiveBatchSink.java
```diff
@@ -127,8 +111,54 @@ public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable, HCa
...
+                                                                                          getPartitions(config)));
+
+      HCatSchema hiveSchema = HCatOutputFormat.getTableSchema(modifiedConf);
+      if (config.schema != null) {
+        // if the user did provide a sink schema to use then use that one
+        hiveSchema = HiveSchemaConverter.toHiveSchema(Schema.parseJson(config.schema), hiveSchema);
+      }
+      HCatOutputFormat.setSchema(modifiedConf, hiveSchema);
+      conf = getConfigurationDiff(originalConf, modifiedConf);
+      HiveSchemaStore.storeHiveSchema(context, config.dbName, config.tableName, hiveSchema);
```
albertshau: this should happen in the beforeSubmit method rather than in this constructor. Usually don't want to do stuff that hits external systems in a constructor.

albertshau:
minor comments, otherwise lgtm assuming the char,byte,short stuff works, and assuming there is a jira open for unit tests

rohitsinha54:
@albertshau : I  have addressed comment and improvised the conversion for source. Please have a look. I had to rebase with develop and force push to update the PR. Last 3 commits are new commits added after your last review.

We have a jira for unit test here : https://issues.cask.co/browse/CDAP-4224 I will add them in a separate PR. 
Thanks.

albertshau:
:+1:

PR #60, opened 2015-11-09
Implementation of Hasher

nitinmotgi:


transform-plugins/src/main/java/co/cask/hydrator/plugin/Hasher.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Hash field(s) values using one the the digest algorithms.
+ */
+@Plugin(type = "transform")
+@Name("Hasher")
+@Description("Encodes field values using one of the digest algorithms. MD2, MD5, SHA1, SHA256, " +
+  "SHA384 and SHA512 are the supported message digest algorithms.")
+public final class Hasher extends Transform<StructuredRecord, StructuredRecord> {
+  private final Config config;
+  private Map<String, Boolean> fieldsMap = new HashMap<>();
```
anwar6953: seems a Set<String> would be more appropriate.

nitinmotgi: done!

nitinmotgi:
@anwar6953 addressed your comments. Please review again.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Hasher.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    super.configurePipeline(pipelineConfigurer);
+    
+    // Checks if hash specified is one of the supported types. 
+    if (!config.hash.equalsIgnoreCase("md2") && !config.hash.equalsIgnoreCase("md5") &&
+        !config.hash.equalsIgnoreCase("sha1") && !config.hash.equalsIgnoreCase("sha256") &&
+        !config.hash.equalsIgnoreCase("sha384") && !config.hash.equalsIgnoreCase("sha512")) {
```
poornachandra, line 70: I think the above condition would be easier to read and verify, if all supported digest algorithms are added into a set, and a `Set.contains(lowercase(specified digest))` is done. We could then use the same set for the exception message below.

poornachandra:
@nitinmotgi Just one minor comment, please fix and merge.

transform-plugins/src/main/java/co/cask/hydrator/plugin/Hasher.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Hash field(s) values using one the the digest algorithms.
+ */
+@Plugin(type = "transform")
+@Name("Hasher")
+@Description("Encodes field values using one of the digest algorithms. MD2, MD5, SHA1, SHA256, " +
+  "SHA384 and SHA512 are the supported message digest algorithms.")
+public final class Hasher extends Transform<StructuredRecord, StructuredRecord> {
+  private final Config config;
+  private Set<String> fieldsMap = new HashSet<>();
```
anwar6953: nit: also, the variable name is inconsistent now.

anwar6953: remainder lgtm.

nitinmotgi: done!

PR #62, opened 2015-11-20
CDAP-4226 various changes to cleanup build

albertshau:
Changing the build to only produce a single jar instead of
batch and realtime jars, since the only difference between those
jars was their manifests. Instead, the single jar should be set
to extend both cdap-etl-batch and cdap-etl-realtime.

Also fixing widget configs to specify schema in the way the
3.3.x UI expects, and updating the cdap version to be
3.3.0-SNAPSHOT.

Finally, adding a build step that creates the config file for
the plugin, which specifies widget and doc information as
artifact properties.


elasticsearch-plugins/widgets/Elasticsearch.json
```diff
@@ -4,7 +4,7 @@
     "position": [ "group1" ],
     "group1": {
       "display" : "Elasticsearch Configuration",
-      "position" : [ "es.host", "es.transportAddresses", "es.cluster", "es.index", "es.type", "es.idField", "schema", "query" ],
```
gsps1, line 7: reason for this change?
edit : just read the explanation about ui requirement in the summary.

pom.xml
```diff
@@ -495,6 +506,97 @@
...
+              <target>
+                <script language="javascript"> <![CDATA[
+                  importClass(java.io.File);
+                  importClass(java.io.FileWriter);
+                  importClass(java.io.BufferedWriter);
+                  importClass(java.nio.file.Files);
+                  importClass(java.nio.charset.StandardCharsets);
+
+                  // for some reason, project.basedir evaluates to null,
+                  // so main.basedir is set to project.basedir and then used here
```
gsps1: may be add info about where we set `main.basedir = project.basedir` here..

pom.xml
```diff
@@ -495,6 +506,97 @@
...
+                  var version = project.getProperty("project.version");
+
+                  var cfgFile = new File(targetDir, artifactId + "-" + version + ".json");
+                  if (!cfgFile.exists()) {
+                    cfgFile.createNewFile();
+                  }
+
+                  var etlRange = project.getProperty("etl.versionRange");
+                  var config = {
+                    "parents": [ "system:cdap-etl-batch" + etlRange, "system:cdap-etl-realtime" + etlRange ],
```
gsps1, line 545: question: maybe a basic question, but we merged batch and realtime into single jar right, so do we need to have two parents here?

pom.xml
```diff
@@ -495,6 +506,97 @@
...
+                  if (widgetsDir.isDirectory()) {
+                    var widgetsFiles = widgetsDir.listFiles();
+                    for (i = 0; i < widgetsFiles.length; i++) {
+                      var widgetsFile = widgetsFiles[i];
+                      if (widgetsFile.isFile()) {
+                        var contents = new java.lang.String(Files.readAllBytes(widgetsFile.toPath()), StandardCharsets.UTF_8);
+                        var contentsAsJson = JSON.parse(contents);
+                        var propertyName = "widgets." + widgetsFile.getName();
+                        // if the filename ends with .json, strip the .json
+                        if (propertyName.indexOf(".json", propertyName.length - 5) !== -1) {
```
gsps1: wouldn't it be simpler to use `if(propertyName.endsWith(".json"))`

pom.xml
albertshau, line 545: we haven't merged the batch and realtime apps into a single jar, we just got rid of the batch and realtime plugin jars. Instead of having cdap-etl-lib-3.3.0-batch extend cdap-etl-batch and cdap-etl-lib-3.3.0-realtime extend cdap-etl-realtime, we just have cdap-etl-lib-3.3.0 extend both cdap-etl-batch and cdap-etl-realtime.

We could probably combine the etl batch and realtime jars into one as well... but that would be a code change.

pom.xml
```diff
@@ -495,6 +506,97 @@
...
+                        var contents = new java.lang.String(Files.readAllBytes(docFile.toPath()), StandardCharsets.UTF_8);
+                        var propertyName = "doc." + docFile.getName();
+                        // if the filename ends with .md or .html, strip the extension
+                        if (propertyName.indexOf(".md", propertyName.length - 3) !== -1) {
+                          propertyName = propertyName.slice(0, -3);
+                        }
+                        if (propertyName.indexOf(".html", propertyName.length - 5) !== -1) {
+                          propertyName = propertyName.slice(0, -5);
+                        }
+                        config.properties[propertyName] = contentsAsJson;
```
gsps1, line 581: if there are two files, `Readme.md and  Readme.html` in the same dir, then one of them will be set to config.properties, replacing the other.  is that okay?

pom.xml
gsps1, line 545: ok

gsps1:
just few minor questions, lgtm otherwise.

pom.xml
albertshau, line 581: hm that is true... can probably start with just markdown

gsps1:
LGTM

PR #64, opened 2015-11-25
fix failing mongodb tests caused due to ETL stage api change

rohitsinha54:


albertshau:
ah sorry about that, good catch. lgtm

PR #69, opened 2015-12-10
copying cdap-etl-lib from cdap to here as core-plugins

albertshau:
Also doing some pom cleanup to remove cdap-etl-lib and replace
with core-plugins.


gsps1:
using gitk for code review, as github is not able to render everything due to a lot of changes. so just adding comments here.

gsps1:
logback.xml in core-plugins still uses co.cask.cdap class names for appender and logger names.

albertshau:
good catch on logback, corrected the class names

gsps1:
one minor comments, LGTM otherwise

PR #70, opened 2015-12-10
fix build to work with java8

albertshau:
changing the script in the build to use fully qualified classnames
instead of importClass(), since importClass is not available in
the Java 8 Nashorn engine.


gsps1:
LGTM

PR #72, opened 2015-12-11
CDAP-4415 switch off cdap-etl-common to hydrator-common

albertshau:
Also moving some mocks that were in multiple modules into the
common module.


core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
```diff
@@ -30,11 +30,11 @@
...
-  private final Map<Integer, co.cask.cdap.api.data.schema.Schema> schemaCache = Maps.newHashMap();
+  private final Map<Integer, Schema> schemaCache = Maps.newHashMap();
 
   @Override
   public StructuredRecord transform(GenericRecord genericRecord) throws IOException {
-    Schema genericRecordSchema = genericRecord.getSchema();
+    org.apache.avro.Schema genericRecordSchema = genericRecord.getSchema();
 
     int hashCode = genericRecordSchema.hashCode();
     co.cask.cdap.api.data.schema.Schema structuredSchema;
```
gsps1: this can be changed to be just Schema

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/ProjectionTransformTest.java
```diff
@@ -21,7 +21,8 @@
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.etl.api.Transform;
 import co.cask.cdap.etl.api.TransformContext;
-import co.cask.hydrator.plugin.common.MockEmitter;
+import co.cask.hydrator.common.mock.MockEmitter;
```
gsps1: since we are already in "mock" package, might be better to name the classes just Emitter, etc. what do you think?

gsps1:
just one minor comment and question, otherwise lgtm.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/ProjectionTransformTest.java
albertshau: think ill rename the package to 'test' since there is a no-op thing in there and not just mocks

core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
albertshau: good catch

PR #73, opened 2015-12-11
Transform plugins schema validation

gsps1:
After CDAP-4229, Plugins can validate input schema and set output schema from that stage. Added implementation for the transforms to use them.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
@@ -26,6 +26,7 @@ import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.Transform;
+import com.google.common.base.Throwables;
 import net.sf.uadetector.ReadableUserAgent;
 import net.sf.uadetector.UserAgentStringParser;
 import net.sf.uadetector.service.UADetectorServiceFactory;
@@ -98,10 +99,26 @@ public class LogParserTransform extends Transform<StructuredRecord, StructuredRe
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
     super.configurePipeline(pipelineConfigurer);
     if (!S3_LOG.equals(config.logFormat) && !CLF_LOG.equals(config.logFormat) &&
-        !CLOUDFRONT_LOG.equals(config.logFormat)) {
+      !CLOUDFRONT_LOG.equals(config.logFormat)) {
       LOG.error("Log format not currently supported.");
       throw new IllegalStateException("Unsupported log format: " + config.logFormat);
     }
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      if (!inputSchema.getType().equals(Schema.Type.RECORD)) {
+        throw new IllegalArgumentException("Only Input Schema of type Schema.Type.RECORD is supported");
+      }
+      Schema.Field inputNameSchema = inputSchema.getField(config.inputName);
+      try {
+        if (inputNameSchema != null && inputNameSchema.getSchema() != null) {
+          validateInputSchemaType(inputNameSchema.getSchema().getType());
+        }
+      } catch (IllegalArgumentException e) {
+        LOG.error(e.getMessage());
+        throw Throwables.propagate(e);
+      }
+    }
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(LOG_SCHEMA);
   }
 
   @Override
@@ -177,10 +194,10 @@ public class LogParserTransform extends Transform<StructuredRecord, StructuredRe
     }
     Schema.Type inputType = inputSchema.getType();
 
-    if (!Schema.Type.STRING.equals(inputType) && !Schema.Type.BYTES.equals(inputType)) {
-      LOG.error("Unsupported inputType in schema, only Schema.Type.BYTES and Schema.Type.STRING are supported. " +
-                  "InputType: {}", inputType.toString());
-      return null;
+    try {
+      validateInputSchemaType(inputType);
+    } catch (Exception e) {
+      LOG.error(e.getMessage());
     }
 
     if (Schema.Type.STRING.equals(inputType)) {
@@ -195,6 +212,15 @@ public class LogParserTransform extends Transform<StructuredRecord, StructuredRe
         LOG.debug("Not a byte type, type is {}", data.getClass().toString());
         return null;
       }
+    }
+  }
+
+  private void validateInputSchemaType(Schema.Type inputSchemaType) {
+    if (!Schema.Type.STRING.equals(inputSchemaType) && !Schema.Type.BYTES.equals(inputSchemaType)) {
+      throw new IllegalArgumentException(String.format(
+        "Unsupported inputType in schema, only Schema.Type.BYTES and Schema.Type.STRING are supported " +
+          "InputType: %s", inputSchemaType.toString()));
+
     }
   }
 
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ScriptTransform.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ScriptTransform.java
@@ -119,6 +119,9 @@ public class ScriptTransform extends Transform<StructuredRecord, StructuredRecor
     // try evaluating the script to fail application creation if the script is invalid
     init(null);
 
+    // init intializes schema if present in the config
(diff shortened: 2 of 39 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
```diff
@@ -102,6 +102,15 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
       LOG.error("Log format not currently supported.");
       throw new IllegalStateException("Unsupported log format: " + config.logFormat);
     }
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      if (!validateInputSchemaType(inputSchema.getType())) {
+        throw new IllegalArgumentException(
```
albertshau: probably better for the validate method to throw this, since we already have the message, and because validate is the one that knows whats invalid.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
```diff
@@ -198,6 +205,17 @@ private String getLog(StructuredRecord input) {
     }
   }
 
+  private boolean validateInputSchemaType(Schema.Type inputSchemaType) {
```
albertshau: this should take a Schema, not a Schema.Type.  We expect all Schemas to be of type Record.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
```diff
@@ -102,6 +102,15 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
       LOG.error("Log format not currently supported.");
       throw new IllegalStateException("Unsupported log format: " + config.logFormat);
     }
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      if (!validateInputSchemaType(inputSchema.getType())) {
```
albertshau: we expect all input schemas to be of type RECORD. This should be checking the type of the field as configured by config.inputName.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
albertshau: should change this to throw an exception. The caller can either choose to catch it and log it (in getLog), or let it go up (configurePipeline)

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
```diff
@@ -198,6 +205,17 @@ private String getLog(StructuredRecord input) {
     }
   }
 
+  private boolean validateInputSchemaType(Schema.Type inputSchemaType) {
+    if (!Schema.Type.STRING.equals(inputSchemaType) && !Schema.Type.BYTES.equals(inputSchemaType)) {
+      String message =
+        String.format("Unsupported inputType in schema, only Schema.Type.BYTES and Schema.Type.STRING are supported " +
+                        "InputType: %s", inputSchemaType.toString());
+      LOG.error(message);
```
albertshau: no need to log it here, caller should log it.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ScriptTransform.java
```diff
@@ -119,6 +119,8 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
     // try evaluating the script to fail application creation if the script is invalid
     init(null);
 
+    Schema outputSchema = (schema == null) ? pipelineConfigurer.getStageConfigurer().getInputSchema() : schema;
```
albertshau, line 123: schema will always be null here as it is set in the init method. You need to parse the schema from the config (if its not null) and then do this logic.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/MockTestConfigurer.java
```diff
@@ -0,0 +1,127 @@
...
+import co.cask.cdap.api.plugin.PluginSelector;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.StageConfigurer;
+
+import javax.annotation.Nullable;
+
+/**
+ * Mock test configurer used to test configure pipeline's validation of input schema and setting of output shcema.
+ */
+public class MockTestConfigurer implements PipelineConfigurer {
```
albertshau: MockPipelineConfigurer

albertshau: if you review https://github.com/caskdata/hydrator-plugins/pull/72, there will be a mock package in hydrator-common that we can put this in so it can be used by any of the other modules

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/MockTestConfigurer.java
```diff
@@ -0,0 +1,127 @@
...
+  Schema outputSchema;
+
+  public MockTestConfigurer(Schema inputSchema) {
+    this.inputSchema = inputSchema;
+  }
+
+  @Nullable
+  public Schema getOutputSchema() {
+    return outputSchema;
+  }
```
albertshau: nit: add newline

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/MockTestConfigurer.java
albertshau: actually, ignore the review comment, this is pretty specific to schema validation and should be kept in this package.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/LogParserTransformTest.java
```diff
@@ -47,6 +47,33 @@
...
+    Schema.Field.of("uri", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("ip", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("browser", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("device", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("httpStatus", Schema.of(Schema.Type.INT)),
+    Schema.Field.of("ts", Schema.of(Schema.Type.LONG)));
+
+  @Test
+  public void testConfigurePipelineSchemaValidation() throws Exception {
+    MockTestConfigurer mockConfigurer = new MockTestConfigurer(Schema.of(Schema.Type.BYTES));
```
albertshau: we expect our schemas to be of type RECORD.  This is a good test case to have though, should check that this throws an exception.

albertshau:
are the other transforms (Validator, ScriptFilter, StructuredRecordToGenericRecord, Python) coming in another pr?

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ScriptTransform.java
albertshau, line 123: please add a test case for this

albertshau:
also, please add tests to check schemas are set correctly.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
gsps1: its also called from transform, where we don't want to throw exception. that's why i return a boolean from that validate method and throw exception here.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/LogParserTransform.java
gsps1: okay

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ScriptTransform.java
gsps1, line 123: we call init before this, and that initializes schema. i can add a comment here.

albertshau, line 123: sorry didn't see that. ignore the comment then.

albertshau:
one small comment about making a string a constant, please fix then merge.

other than that, all comments are about performing additional schema validation.  But I think we can push that off to another PR, just need to be sure it's not dropped.

gsps1:
@albertshau good catch, i noticed that at some places we are doing field checks, like in `configurePipeline` decompressor, but not in few places like `Decoder`, i will fix those.

gsps1:
build green, rebasing and merging. have opened a JIRA to fix the configure plugin field name checks here : https://issues.cask.co/browse/CDAP-4458

PR #74, opened 2015-12-11
Add support for secure hive

rohitsinha54:


hive-plugins/src/main/java/co/cask/plugin/etl/batch/sink/HiveBatchSink.java
```diff
@@ -132,6 +143,18 @@ public HiveSinkOutputFormatProvider(Job job, HiveSinkConfig config) throws IOExc
       conf = getConfigurationDiff(originalConf, modifiedConf);
     }
 
+    private void addSecureHiveProperties(Configuration modifiedConf) throws IOException {
+      modifiedConf.set(HiveConf.ConfVars.METASTORE_USE_THRIFT_SASL.varname, "true");
+
+      UserGroupInformation ugi = UserGroupInformation.getCurrentUser();
```
poornachandra, line 151: Can you add comment as to why we need to do this workaround?

poornachandra:
LGTM

PR #75, opened 2015-12-14
Modifies the old config json files to the newer format

ajainarayanan:
Modifies the config json files from older format to new format (based on the spec).

Added the generated json and jar files to cdap standalone sdk and it worked fine in UI. Couldn't test all the plugins for its functionality. As far as rendering the plugin widgets in UI everything worked just fine.

### DIFF
--- a/cassandra-plugins/widgets/Cassandra.json
+++ b/cassandra-plugins/widgets/Cassandra.json
@@ -1,110 +1,127 @@
 {
-  "id": "Cassandra",
-  "groups": {
-    "position": [ "group1", "group2" ],
-    "group1": {
-      "display": "Cassandra Configuration",
-      "position": [ "initialAddress", "addresses", "port", "consistencyLevel", "compression", "username", "password", "properties" ],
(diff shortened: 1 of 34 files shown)


cassandra-plugins/widgets/Cassandra.json
```diff
@@ -1,110 +1,127 @@
...
-    "position": [ "group1", "group2" ],
-    "group1": {
-      "display": "Cassandra Configuration",
-      "position": [ "initialAddress", "addresses", "port", "consistencyLevel", "compression", "username", "password", "properties" ],
-      "fields": {
-        "initialAddress": {
-          "widget": "textbox",
+  "metadata": {
+    "spec-version": "1.0",
+    "artifact-version": "3.3.0-SNAPSHOT",
```
albertshau: thought we were removing these fields

albertshau:
files should be renamed too right?

albertshau:
:+1:

PR #76, opened 2015-12-16
CDAP-4454: Schema validation for sources

jwang47:
This currently contains no schema validation, only sets output schema to input schema for sources.

### DIFF
--- a/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
+++ b/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
@@ -24,6 +24,7 @@ import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.api.batch.BatchSourceContext;
 import com.datastax.driver.core.Row;
@@ -116,6 +117,15 @@ public class BatchCassandraSource extends BatchSource<Long, Row, StructuredRecor
       builder.set(field.getName(), extractValue(input.getValue(), field));
     }
     emitter.emit(builder.build());
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(inputSchema);
+    // TODO: validate schema
   }
 
   private Object extractValue(Row row, Schema.Field field) throws Exception {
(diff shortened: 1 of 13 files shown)


cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
```diff
@@ -118,6 +119,15 @@ public void transform(KeyValue<Long, Row> input, Emitter<StructuredRecord> emitt
     emitter.emit(builder.build());
   }
 
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(inputSchema);
```
albertshau, line 127: The input schema for a source will always be null.  Should be setting the schema based on what is set in the config.  Something like:

```
try {
  Schema schema = Schema.parseJson(config.schema);
  pipelineConfigurer.getStageConfigurer().setOutputSchema(schema);
} catch (Exception e) {
  throw new IllegalArgumentException("Invalid schema: " + e.getMessage(), e);
}
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/DBSource.java
```diff
@@ -72,11 +71,6 @@ public DBSource(DBSourceConfig dbSourceConfig) {
   }
 
   @Override
-  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
```
albertshau, line 75: shouldn't remove this

albertshau:
think there was some confusion, sources should be setting outputschema based on their config, not based on the getInputSchema method, which will always return null

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/KVTableSource.java
```diff
@@ -76,4 +77,11 @@ public KVTableSource(KVTableConfig kvTableConfig) {
   public void transform(KeyValue<byte[], byte[]> input, Emitter<StructuredRecord> emitter) throws Exception {
     emitter.emit(StructuredRecord.builder(SCHEMA).set("key", input.getKey()).set("value", input.getValue()).build());
   }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+
+    // TODO: validate schema
```
nitinmotgi, line 85: Why TODO ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/SnapshotFileBatchSource.java
```diff
@@ -51,10 +53,16 @@ public SnapshotFileBatchSource(SnapshotFileSetConfig config) {
...
+    super.configurePipeline(pipelineConfigurer);
+
     PartitionedFileSetProperties.Builder fileProperties = SnapshotFileSet.getBaseProperties(config);
     addFileProperties(fileProperties);
 
     pipelineConfigurer.createDataset(config.getName(), PartitionedFileSet.class, fileProperties.build());
+
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(inputSchema);
+    // TODO: validate schema
```
nitinmotgi, line 65: Remove TODO

albertshau:
closing for now, going to re-open with more complete fixes

PR #77, opened 2015-12-16
CDAP-4416 copy widget configs from cdap-etl-lib

albertshau:


albertshau:
this is a straight copy from cdap-etl-lib.  Follow up will fix things that are wrong.

ajainarayanan:
:+1: LGTM

PR #78, opened 2015-12-16
CDAP-4416 fix incorrect widget configs

albertshau:
A lot of configs include fields that should not be included or
are missing fields.


nitinmotgi:
lgtm :+1:

PR #80, opened 2015-12-16
moving docs from readmes into their own markdown files

albertshau:
this will allow automatic integration with the UI, as they will
be included as artifact properties.
Also fixed a bug with the build script to correctly set the
doc properties.


nitinmotgi:
LGTM :+1:

anwar6953:
https://github.com/caskdata/hydrator-plugins/pull/80/files#diff-65b1c16e2b1419c6f0f1afeed9b47002R55

The python transform `Known Issues` section now exists in two files (the one I linked as well as the .md file below it). Isn't it sufficient and better to be in only one file?

PR #81, opened 2015-12-16
fix pom dependencies

albertshau:
Several dependencies marked as provided should have been compile
or test. cdap-formats is compile, cdap-proto and etl-core are only
needed for test.


gokulavasan:
LGTM :+1:

PR #83, opened 2015-12-17
Update Hydrator Plugin Documentation

awholegunch:
When the doc files were converted to Markdown, the titles were stripped off. This causes problems when the documents are imported in Sphinx, as they have numerous levels that appear rather than a title with a set of lower level titles.
- Added titles to all files
- Changed all examples so that descriptive text precedes the example code rather than following it
- Removed reST formatting included in error
- Fixed formatting issues where noticed


albertshau:
lgtm

PR #84, opened 2015-12-18
Adding pipe delimiter to parsing record in CSVParser.

nitinmotgi:
Good to go!

### DIFF
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
@@ -73,7 +73,7 @@ public final class CSVParser extends Transform<StructuredRecord, StructuredRecor
   // Initialize Pipe Delimiter CSV Parser. 
   static {
     PDL = CSVFormat.DEFAULT.withDelimiter('|').withEscape('\\').withIgnoreEmptyLines(false)
-      .withAllowMissingColumnNames().withQuote((Character)null).withRecordSeparator('\n')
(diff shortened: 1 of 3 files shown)


transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -47,6 +47,7 @@
  *   <li>RFC4180</li>
  *   <li>MYSQL and</li>
```
anwar6953: the 'and' should probably go after `TDF` now

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -66,33 +67,43 @@
...
-                                           "DEFAULT, EXCEL, MYSQL, RFC4180 & TDF");
+                                           "DEFAULT, EXCEL, MYSQL, RFC4180, PDL & TDF");
     }
 
     // Check if schema specified is a valid schema or no.
     try {
-      Schema outputSchema = Schema.parseJson(config.schema);
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(outputSchema);
+      Schema outputSchema = Schema.parseJson(this.config.schema);
+      config.getStageConfigurer().setOutputSchema(outputSchema);
```
anwar6953: `configurer` seems like a more appropriate variable name than `config`

anwar6953:
all else LGTM!

PR #88, opened 2016-01-06
CDAP-4454 setting output schema for sources

gsps1:
set output schema for sources. 
https://issues.cask.co/browse/CDAP-4454

### DIFF
--- a/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
+++ b/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
@@ -24,6 +24,7 @@ import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.api.batch.BatchSourceContext;
 import com.datastax.driver.core.Row;
@@ -72,6 +73,17 @@ public class BatchCassandraSource extends BatchSource<Long, Row, StructuredRecor
 
   public BatchCassandraSource(CassandraSourceConfig config) {
     this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(config.schema), "Schema must be specified.");
+    try {
+      Schema schema = Schema.parseJson(config.schema);
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(schema);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Invalid output schema: " + e.getMessage(), e);
+    }
   }
 
   @Override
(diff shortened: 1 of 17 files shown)


cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/BatchCassandraSource.java
```diff
@@ -75,6 +76,17 @@ public BatchCassandraSource(CassandraSourceConfig config) {
   }
 
   @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Preconditions.checkArgument(!Strings.isNullOrEmpty(config.schema), "Schema must be specified.");
+    try {
+      Schema schema = Schema.parseJson(config.schema);
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(schema);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Invalid input schema: " + e.getMessage(), e);
```
albertshau: Invalid output schema

albertshau: similar comment for others

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
```diff
@@ -94,6 +94,10 @@
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     streamBatchConfig.validate();
     pipelineConfigurer.addStream(new Stream(streamBatchConfig.name));
+    // if no format is specified then default schema is used, if otherwise its based on input records.
+    if (streamBatchConfig.format == null) {
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(DEFAULT_SCHEMA);
+    }
```
albertshau, line 107: many formats will allow or require specifying a schema, which we can use to set the outputschema.

Can use `streamBatchConfig.getFormatSpec()`, then if the spec's format is non-null, set the output schema accordingly.  One thing that is a little tricky is that the spec's schema is the schema of the body, so you'd have to add ts and headers as fields before setting the output schema.

mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/MongoDBBatchSource.java
```diff
@@ -64,6 +64,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
     try {
       BSONConverter.validateSchema(Schema.parseJson(config.schema));
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(Schema.parseJson(config.schema));
     } catch (IOException e) {
       throw new IllegalArgumentException("Invalid schema", e);
```
albertshau: lets include e.getMessage() to be consistent with other plugins.

albertshau:
mostly the one comment about stream source. Other comments are minor.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
gsps1, line 107: yeah, have fixed it, the schema fields are added as separate fields and not as a single body field in transform, so doing the same.

gsps1:
have addressed the comments, thanks

albertshau:
:+1:

PR #89, opened 2016-01-07
CDAP-4456 validate schema for sinks (table, hbase)

gsps1:
In sinks with output schema, validate the input schema appropriately. 

Example : Table sink checks all the input schema fields are of simple type, rowkey field is present and they contain all output schema fields. 

JIRA : https://issues.cask.co/browse/CDAP-4456


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -53,10 +54,51 @@ public TableSink(TableSinkConfig tableSinkConfig) {
   }
 
   @Override
-  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
```
albertshau: dont need to have this throws statement

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -53,10 +54,51 @@ public TableSink(TableSinkConfig tableSinkConfig) {
...
     Preconditions.checkArgument(!Strings.isNullOrEmpty(tableSinkConfig.getRowField()),
                                 "Row field must be given as a property.");
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      validateInputAndOutputSchemaFields(inputSchema);
+
+      boolean rowFieldPresent = false;
+      for (Schema.Field field : inputSchema.getFields()) {
+        // simple type check for fields
+        if (!field.getSchema().getType().isSimpleType()) {
```
albertshau: a nullable string is represented as a union of null a string, which will fail this check. You can use `field.getSchema().isSimpleOrNullableSimple()`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -53,10 +54,51 @@ public TableSink(TableSinkConfig tableSinkConfig) {
...
+            String.format("Field '%s' is not of simple type, All fields for table sink should of simple type",
+                          field.getName()));
+        }
+        // check if this field is the row key field
+        if (field.getName().equals(tableSinkConfig.getRowField())) {
+          rowFieldPresent = true;
+        }
+      }
+
+      if (!rowFieldPresent) {
```
albertshau: a little simpler to just do `inputSchema.getField(...) == null`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -53,10 +54,51 @@ public TableSink(TableSinkConfig tableSinkConfig) {
...
+    try {
+      // check if input schema contains all the fields expected in the output schema (if output schema is provided)
+      if (tableSinkConfig.getSchemaStr() != null) {
+        Schema outputSchema = Schema.parseJson(tableSinkConfig.getSchemaStr());
+        for (Schema.Field field : outputSchema.getFields()) {
+          if (inputSchema.getField(field.getName()) == null) {
+            throw new IllegalArgumentException(
+              String.format("Field '%s' is present in output schema but not present in input schema",
+                            field.getName()));
+          }
```
albertshau: should also check that the fields' schemas are the same.

albertshau: seems like a common method, could put it in the common package somewhere as a `isSubset` method or something like that

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -53,10 +54,51 @@ public TableSink(TableSinkConfig tableSinkConfig) {
...
+      }
+
+      if (!rowFieldPresent) {
+        throw new IllegalArgumentException(
+          String.format("Row field key : '%s' is not present in the input schema",  tableSinkConfig.getRowField()));
+      }
+    }
+  }
+
+  private void validateInputAndOutputSchemaFields(Schema inputSchema) {
```
albertshau: seems like a common method, could put it in the common package somewhere as a `validateIsSubset` method or something like that

core-plugins/src/main/java/co/cask/hydrator/plugin/realtime/sink/RealtimeTableSink.java
```diff
@@ -58,11 +59,53 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     Preconditions.checkArgument(!Strings.isNullOrEmpty(tableSinkConfig.getName()), "Dataset name must be given.");
     Preconditions.checkArgument(!Strings.isNullOrEmpty(tableSinkConfig.getRowField()),
                                 "Field to be used as rowkey must be given.");
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      validateInputAndOutputSchemaFields(inputSchema);
+
+      boolean rowFieldPresent = false;
+      for (Schema.Field field : inputSchema.getFields()) {
```
albertshau: this logic also looks like common and repeated logic. Could also move it to common as `validateFieldsAreSimple` method or something like that.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLMapReduceTestRun.java
```diff
@@ -300,6 +300,133 @@ public void testTableToTableWithValidations() throws Exception {
     }
   }
 
+
+  @SuppressWarnings("ConstantConditions")
+  @Test
+  public void testInvalidTableConfigurationComplexType() throws Exception {
```
albertshau: These tests are fairly heavyweight and slow since they require hitting handlers and messing with jars, etc.

Since we're just testing schema validation, we should be able to write unit tests for just the sink. Like a TableSinkTest that creates the sink directly, sets the input schema, and calls configurePipeline directly.  That type of test would be much more lightweight and still test the same thing

albertshau:
we can also make changes so that a lot of these sinks don't require the schema property to be set anymore if there is an input schema. In those situations, just use the input schema as the output schema.

gsps1:
@albertshau have addressed your comments.

albertshau:
took a pass, mostly comments about validation logic

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -57,24 +58,29 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
     Preconditions.checkArgument(!Strings.isNullOrEmpty(tableSinkConfig.getRowField()),
                                 "Row field must be given as a property.");
+    Schema outputSchema =
+      SchemaValidator.validateOutputSchemaAndInputSchemaIfPresent(tableSinkConfig.getSchemaStr(),
+                                                                  tableSinkConfig.getRowField(), pipelineConfigurer);
+    // NOTE: this is done only for testing
```
albertshau: can you open a jira and mention it in this comment. Jira should have something about support for passing values calculated at configure time to the runtime context.

albertshau:
comment about a comment. lgtm

gsps1:
added JIRA details and squashed commits. merging.

PR #90, opened 2016-01-08
CDAP-3254 Adding unit tests to verify explore works for TPFSAvro/Parquet with map as one of the fields

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-3254

This issue doesn't seem to arise in unit tests. Enhanced tests to cover the case mentioned in the JIRA.
Will do testing on cluster to confirm that the issue is not reproducible and then close the JIRA.


albertshau:
:+1:

PR #91, opened 2016-01-11
bugfix : add missing test ETLMapreduceTestRun in the batch test suite

gsps1:
Added missing test and corresponding fixes for failing tests in that TestClass, 
this includes calling super class's configurePipeline and setting output shema for table sink in tests


albertshau:
how come the super.configurePipeline() calls are needed? I thought those are no-ops?

gsps1:
its needed for KVTableSource, as we create the dataset in BatchReadableSource#ConfigurePipeline, others are no-ops, i just included it for the purpose of consistency.

albertshau:
lgtm

PR #93, opened 2016-01-13
Feature/add schema validation kv table sink

gsps1:
Added schema validation and tests in KVTable sink and also a minor bugfix in StreamBatchSource.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
@@ -21,15 +21,18 @@ import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
 import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.hydrator.common.SchemaValidator;
 import co.cask.hydrator.plugin.common.Properties;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Maps;
 
 import java.nio.ByteBuffer;
+import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.Nullable;
 
@@ -80,8 +83,43 @@ public class KVTableSink extends BatchWritableSink<StructuredRecord, byte[], byt
   }
 
   @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    // validate that input and output fields are present
+    if (inputSchema != null) {
+      SchemaValidator.validateFieldsArePresentInSchema(inputSchema, kvTableConfig.keyField);
+      SchemaValidator.validateFieldsArePresentInSchema(inputSchema, kvTableConfig.valueField);
+      validateSchemaTypeIsStringOrBytes(inputSchema, kvTableConfig.keyField, false);
+      validateSchemaTypeIsStringOrBytes(inputSchema, kvTableConfig.valueField, true);
+    }
+  }
+
+  private void validateSchemaTypeIsStringOrBytes(Schema inputSchema, String fieldName, boolean isNullable) {
+    Schema fieldSchema = inputSchema.getField(fieldName).getSchema();
+    boolean fieldIsNullable = fieldSchema.isNullable();
+    if (!isNullable && fieldIsNullable) {
+      throw new IllegalArgumentException("Field " + fieldName + " cannot be nullable");
+    }
+    Schema.Type fieldType = fieldIsNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+
+    if (fieldType != Schema.Type.STRING && fieldType != Schema.Type.BYTES) {
+      throw new IllegalArgumentException(
+        String.format("Field name %s is of type %s, only types String and Bytes are supported for KVTable",
+                      fieldName, fieldType));
+    }
+  }
+
+  @Override
   protected Map<String, String> getProperties() {
-    Map<String, String> properties = Maps.newHashMap(kvTableConfig.getProperties().getProperties());
+
+    Map<String, String> properties;
+    // will be null only in tests
+    if (kvTableConfig.getProperties() == null) {
+      properties = new HashMap<>();
+    } else {
+      properties = new HashMap<>(kvTableConfig.getProperties().getProperties());
+    }
     properties.put(Properties.BatchReadableWritable.NAME, kvTableConfig.name);
     properties.put(Properties.BatchReadableWritable.TYPE, KeyValueTable.class.getName());
     return properties;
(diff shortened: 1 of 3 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -80,8 +82,34 @@ public KVTableSink(KVTableConfig kvTableConfig) {
   }
 
   @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    // validate that input and output fields are present
+    SchemaValidator.validateFieldsArePresentInSchema(inputSchema, kvTableConfig.keyField);
```
albertshau: only do this if the input schema is not null

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -80,8 +82,34 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
+
+  private void validateSchemaTypeIsStringOrBytes(Schema inputSchema, String fieldName) {
+    Schema.Type fieldType = inputSchema.getField(fieldName).getSchema().getType();
+    if (!fieldType.equals(Schema.Type.STRING) && (!fieldType.equals(Schema.Type.BYTES))) {
+      throw new IllegalArgumentException(
+        String.format("Field name %s is of type %s, only types String and Bytes are supported for KVTable",
+                      fieldName, fieldType));
+    }
+  }
+  @Override
```
albertshau, line 113: nit: add newline between methods

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -80,8 +82,34 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
 
   @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    super.configurePipeline(pipelineConfigurer);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    // validate that input and output fields are present
+    SchemaValidator.validateFieldsArePresentInSchema(inputSchema, kvTableConfig.keyField);
+    SchemaValidator.validateFieldsArePresentInSchema(inputSchema, kvTableConfig.valueField);
+    validateSchemaTypeIsStringOrBytes(inputSchema, kvTableConfig.keyField);
+    validateSchemaTypeIsStringOrBytes(inputSchema, kvTableConfig.valueField);
```
albertshau: the value can be nullable but the key cannot.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -91,9 +119,9 @@ public KVTableSink(KVTableConfig kvTableConfig) {
   public void transform(StructuredRecord input, Emitter<KeyValue<byte[], byte[]>> emitter) throws Exception {
     Object key = input.get(kvTableConfig.keyField);
     Preconditions.checkArgument(key != null, "Key cannot be null.");
-    byte[] keyBytes = key instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) key) : (byte[]) key;
+    byte[] keyBytes = key instanceof String ? Bytes.toBytes((String) key) : (byte[]) key;
```
albertshau: can't get rid of the ByteBuffer logic. The previous logic should be applied if the schema says the field's type is bytes.  If the schema says the field's type is string, then you can cast it directly to a String and just do Bytes.toBytes() on it.

gsps1:
@albertshau can you review again? addressed the comments. Thanks

gsps1:
@albertshau have updated the PR. please review. Thanks

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -91,9 +129,40 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
-    byte[] keyBytes = key instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) key) : (byte[]) key;
+    byte[] keyBytes;
+
+    Schema keyFieldSchema = input.getSchema().getField(kvTableConfig.keyField).getSchema();
+    if (keyFieldSchema.getType().equals(Schema.Type.STRING)) {
+      keyBytes = Bytes.toBytes((String) key);
+    } else if (keyFieldSchema.getType().equals(Schema.Type.BYTES)) {
+      keyBytes = key instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) key) : (byte[]) key;
+    } else if (keyFieldSchema.isNullable()) {
+      throw new Exception("Key is nullable in the input record : " + input);
```
albertshau: don't log the record, as StructuredRecord has not overridden toString() as anything useful.  Message should be something like "Key field %s cannot have schema %s. It must be a bytes or string.".

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -91,9 +129,40 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
+    } else {
+      throw new Exception(
+        "Key is of unsupported type in the record" + input + ", only String and Bytes are supported");
```
albertshau: don't log the record, as StructuredRecord has not overridden toString() as anything useful. Message should be something like "Key field %s cannot have schema %s. It must be a bytes or string.".

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
albertshau: "Key field %s cannot have nullable schema %s"

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -91,9 +129,40 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
+    }
+
+    Schema.Field valueFieldSchema = input.getSchema().getField(kvTableConfig.valueField);
+    if (valueFieldSchema == null) {
+      throw new Exception("Field " + kvTableConfig.valueField + " is missing in the input record");
```
albertshau: "Key field"

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KVTableSink.java
```diff
@@ -91,9 +129,40 @@ public KVTableSink(KVTableConfig kvTableConfig) {
...
+      Schema.Type valueFieldType =
+        valueFieldSchema.getSchema().isNullable() ? valueFieldSchema.getSchema().getNonNullable().getType() :
+          valueFieldSchema.getSchema().getType();
+      if (valueFieldType.equals(Schema.Type.STRING)) {
+        valBytes = Bytes.toBytes((String) val);
+      } else if (valueFieldType.equals(Schema.Type.BYTES)) {
+        valBytes = val instanceof ByteBuffer ? Bytes.toBytes((ByteBuffer) val) : (byte[]) val;
+      } else {
+        throw new Exception(
+          "Value is of unsupported type in the record, " + input + " only String and Bytes are supported");
```
albertshau: "Value field %s has unsupported schema %s. Only String and Bytes are supported"

albertshau:
remaining comments are only about exception messages, rest lgtm. Please fix then merge

PR #94, opened 2016-01-13
Fix/ui widgets json stream selector

ajainarayanan:
- Fixes Stream-batchsource.json to use `stream-selector` for Stream Name property.
- Removes `width` property from `widget-attribute` in config json as it is not being used in UI.


albertshau:
:+1:

PR #95, opened 2016-01-13
remove link annotation in Cassandra Batch source description

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-4596


albertshau:
:+1:

PR #96, opened 2016-01-13
Fix compilation errors.

anwar6953:
The RecordPutTransformer's package was moved a while back.


anwar6953:
Here's a build failure that shows the errors.
http://builds.cask.co/browse/CDAP-AP-7/log

albertshau:
lgtm

anwar6953:
Likely, these issues were not caught before because the old class was cached for people/builds.

PR #92, opened 2016-01-12
Feature/secure hive 3.3

rohitsinha54:
- Porting https://github.com/caskdata/hydrator-plugins/pull/74
- Updated the package name in export of  pom.

PR #98, opened 2016-01-14
Feature/secure hive 3.3

albertshau:


sagarkapare:
Tested against the original PR #92. Both Hive source and sink are working fine on the secure cluster. Merging this.

PR #100, opened 2016-01-15
CDAP-4619 fix db source to use the right input format class

albertshau:
DBSource was incorrectly calling DBInputFormat's setInput()
method instead of calling ETLDBInputFormat's setInput() method.
This results in mapreduce.job.inputformat.class=DBInputFormat.class
being set in the InputFormatProvider's configuration settings.
So the job will not use ETLDBInputFormat, which causes classloading
problems.

Also moving the unit tests to the correct directory so that they
actually get run.


gokulavasan:
LGTM :+1:

PR #102, opened 2016-01-15
table schema is not required.

albertshau:
If not specified, it will just try to write whatever records it gets


core-plugins/src/main/java/co/cask/hydrator/plugin/common/TableSinkConfig.java
```diff
@@ -36,6 +36,7 @@
   @Name(Properties.Table.PROPERTY_SCHEMA)
   @Description("schema of the table as a JSON Object. If the table does not already exist, one will be " +
     "created with this schema, which will allow the table to be explored through Hive.")
+  @Nullable
   private String schemaStr;
```
bdmogal, line 41: Should we specify the behavior when schema is absent in the description?

bdmogal:
1 question, rest LGTM

PR #103, opened 2016-01-15
Allow for null output schema.

anwar6953:
Related: https://github.com/caskdata/hydrator-plugins/pull/102
Failing test case: https://builds.cask.co/browse/CDAP-ITN9-DSTH23-48/test/case/37590364

If output schema is null, set it to the input schema. (also remove the `Preconditions#checkNotNull`).

### DIFF
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/SchemaValidator.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/SchemaValidator.java
@@ -21,6 +21,7 @@ import co.cask.cdap.etl.api.PipelineConfigurer;
 import com.google.common.base.Preconditions;
 
 import java.io.IOException;
+import javax.annotation.Nullable;
 
 /**
  * Utility class for with methods for validating schema fields and its types,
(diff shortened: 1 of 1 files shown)


albertshau:
:+1:

hydrator-common/src/main/java/co/cask/hydrator/common/SchemaValidator.java
```diff
@@ -29,33 +30,39 @@
...
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema == null && outputSchemaString == null) {
+       return null;
     }
 
-    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    // initialize output schema if present; otherwise, set it to input schema
+    Schema outputSchema;
+    if (outputSchemaString == null) {
+      outputSchema = inputSchema;
```
gsps1, line 51: @albertshau should we return null here? as we are setting output schema to be input schema and that get's propagated to the next stage.

but due to CDAP-4575, that's not actually true and in initialize output schema would still be null.

PR #105, opened 2016-01-16
Have a RecordPutTransformer which supports null schema and use it in …

anwar6953:
Have a RecordPutTransformer which supports null schema and use it in TableSink, RealtimeTableSink and HBaseSink.

The RecordPutTransformer class has been copied over from the CDAP repo.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
@@ -27,7 +27,7 @@ import co.cask.cdap.api.dataset.table.Table;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
-import co.cask.cdap.format.RecordPutTransformer;
+import co.cask.hydrator.common.RecordPutTransformer;
 import co.cask.hydrator.common.SchemaValidator;
 import co.cask.hydrator.plugin.common.Properties;
 import co.cask.hydrator.plugin.common.TableSinkConfig;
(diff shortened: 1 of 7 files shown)


gsps1:
fix checkstyle issues. changes lgtm, might be useful to copy the REcordputTrasnformerTest as well

anwar6953:
Fixed the checkstyle issue. Since this class is copied directly from cdap, and is only temporary, I'll copy just this class.

anwar6953:
RealtimeTableSinkTest is failing because one of the test cases expected a NPE when null schema is given.
https://travis-ci.org/caskdata/hydrator-plugins/builds/102738882

Updated it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -69,7 +69,13 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
     super.initialize(context);
-    Schema outputSchema = Schema.parseJson(tableSinkConfig.getSchemaStr());
+    Schema outputSchema = null;
```
bdmogal, line 72: any way this code can be refactored to a separate class? seems to be repeated

hydrator-common/src/main/java/co/cask/hydrator/common/RecordPutTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+import co.cask.cdap.api.dataset.table.Put;
+import com.google.common.base.Preconditions;
+
+import java.nio.ByteBuffer;
+import javax.annotation.Nullable;
+
+/**
+ * Transforms records into Puts.
+ */
+public class RecordPutTransformer {
```
bdmogal, line 32: can you add a comment + jira to remove this class?

bdmogal:
LGTM

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
anwar6953, line 72: I don't think its worth having this small (3-4) line refactored into its own class.

hydrator-common/src/main/java/co/cask/hydrator/common/RecordPutTransformer.java
anwar6953, line 32: Added.
https://issues.cask.co/browse/CDAP-4656

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
bdmogal, line 72: :)

PR #107, opened 2016-01-18
Bugfix/js transform

gsps1:
Had incorrect widget name, so we weren't getting JS editor. 
JIRA : https://issues.cask.co/browse/CDAP-4674


gsps1:
will open against release.

PR #109, opened 2016-01-19
CDAP-4672 - Added missing Widget config for few plugins

gsps1:
JIRA: https://issues.cask.co/browse/CDAP-4672

These plugins were missing in the widgets for core plugins. so the default formatting were used for these plugins.


albertshau:
are they missing their docs too then?  If so can you add them as well

gsps1:
docs are present.

albertshau:
:+1:

PR #111, opened 2016-01-19
fix transform plugins widgets Compressor, Decompressor, Encoder, Deco…

elfenheart:
…der to use keyvalue-dropdown

screen shot 2016-01-18 at 7 46 30 pm


ajainarayanan:
:+1: LGTM

PR #112, opened 2016-01-19
CDAP-4691 fix hbase and mongo plugins

albertshau:
include zookeeper in the hbase plugins jar since it is required
by TableOutputFormat. Also fix the package naming in the mongo
plugins so that the plugins are actually visible

### DIFF
--- a/hbase-plugins/pom.xml
+++ b/hbase-plugins/pom.xml
@@ -71,6 +71,12 @@
       <artifactId>hydrator-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!-- needed by TableOutputFormat, but TableOutputFormat is provided scope -->
+    <dependency>
+      <groupId>org.apache.zookeeper</groupId>
+      <artifactId>zookeeper</artifactId>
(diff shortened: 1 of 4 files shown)


hbase-plugins/pom.xml
```diff
@@ -71,6 +71,12 @@
       <artifactId>hydrator-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!-- needed by TableOutputFormat, but TableOutputFormat is provided scope -->
+    <dependency>
+      <groupId>org.apache.zookeeper</groupId>
+      <artifactId>zookeeper</artifactId>
+      <scope>compile</scope>
```
gsps1, line 78: why is this compile ? should it be provided instead ?

albertshau, line 78: no it has to be compile, zookeeper is not exposed by the cdap api.  The poms should be restructured, normally this would be pulled in by the hbase dependency but the parent pom has zookeeper at provided

sreevatsanraman:
LGTM

PR #113, opened 2016-01-19
fixing build by adding test dependency on hadoop-common:tests.

albertshau:
For some reason, ony some machines, this dependency wasn't getting
pulled in, so just declaring it explicitly.


anwar6953:
LGTM

PR #117, opened 2016-01-20
Cherry picking and fixing commit from PR https://github.com/caskdata/…

gokulavasan:
…hydrator-plugins/pull/29

Removed prepare call from write

Add back the logic to remove whitespace from column list


albertshau:
:+1:

PR #124, opened 2016-01-23
Add Project description, SCM URL, Developer information to pom.xml

anwar6953:
Add Project description, SCM URL, Developer information to pom.xml

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,10 @@
   <groupId>co.cask.hydrator</groupId>
   <artifactId>hydrator-plugins</artifactId>
   <version>1.2.0</version>
+  <name>Hydrator Plugin Collection</name>
+  <description>A collection of plugins for Cask Hydrator.</description>
+  <url>https://github.com/caskdata/hydrator-plugins</url>
+
   <modules>
     <module>hydrator-common</module>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -21,6 +21,10 @@
   <groupId>co.cask.hydrator</groupId>
   <artifactId>hydrator-plugins</artifactId>
   <version>1.2.0</version>
+  <name>Hydrator Plugin Collection</name>
+  <description>A collection of plugins for Cask Hydrator.</description>
+  <url>http://docs.cdap.io/cdap/current/en/included-applications/index.html</url>
```
albertshau: http://docs.cdap.io/cdap/current/en/cdap-apps/hydrator/index.html

albertshau: actually, shouldn't this be https://github.com/caskdata/hydrator-plugins?

anwar6953: I just moved the url line from a few lines below.
Want me to change it to `https://github.com/caskdata/hydrator-plugins`?

albertshau: yeah think that makes more sense

albertshau:
one comment then lgtm. fix then merge

PR #126, opened 2016-01-27
Fix usage of CDAP internal API Locations.toURI() which get removed fr…

chtyim:
…om cdap


bdmogal:
:+1:

chtyim:
Test passed. Merging.

PR #129, opened 2016-01-29
(CDAP-4801) Propagate fileSystemProperties while adding input paths a…

bdmogal:
…nd specifying input path filters too.


sreevatsanraman:
lgtm

bdmogal:
This is likely not needed. Will re-open if required.

PR #130, opened 2016-01-29
fixed case in AvroBatchSink, LogParser Transform, DataGenerator, JMS Realtime source

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-4727


albertshau:
lgtm

PR #132, opened 2016-02-01
CDAP-4656 Remove RecordPutTransformer from hydrator-plugins.

anwar6953:
There's already a class `RecordPutTransformer` in the CDAP repo, which I updated the usages to depend on that class now.
This class was pulled in to avoid having to wait for the changes in the CDAP repo to propagate; that need no longer exists.


albertshau:
:+1:

PR #133, opened 2016-02-01
(CDAP-4849) Use password widget for password field in teradata source

bdmogal:


anwar6953:
lgtm

gokulavasan:
LGTM :+1:

gokulavasan:
What about batch-sink?

bdmogal:
it was only missing in `Teradata-batchsource`. We don't have a `Teradata-batchsink`. Checked that `Database-batchsource` and `Database-batchsink` use password.

gokulavasan:
@bdmogal Oh ok right. Thanks!

PR #135, opened 2016-02-02
reducing the jar sizes

albertshau:
Moved the python evaluator transform into core-plugins since it
pulls in all of core-plugins anyway. Also moved the kafka source
out of core into the kafka module. These simple changes reduce
the total size of all jars by about 70mb.

Other changes are to remove the core-plugins-tests jar. Modules
should not be dependent on each other, even for tests.


database-plugins/src/test/java/co/cask/hydrator/plugin/DatabasePluginTestBase.java
```diff
@@ -53,32 +60,46 @@
 /**
  * Database Plugin Tests setup.
  */
-public class DatabasePluginTestBase extends ETLBatchTestBase {
-  public static boolean tearDown = true;
-
-  protected static final long CURRENT_TS = System.currentTimeMillis();
+public class DatabasePluginTestBase extends TestBase {
```
gsps1, line 63: why removing extending ETLBatchTestBase?

gsps1, line 63: to remove dependency on core-plugins ?

albertshau, line 63: to remove core-plugins test jar.  It also fails on my laptop, not sure why it doesn't fail on travis.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -110,7 +113,9 @@ public static void setupTest() throws Exception {
     addPluginArtifact(Id.Artifact.from(Id.Namespace.DEFAULT, "transforms", "1.0.0"), APP_ARTIFACT_ID,
                       ProjectionTransform.class, ScriptFilterTransform.class,
                       ValidatorTransform.class, CoreValidator.class,
-                      StructuredRecordToGenericRecordTransform.class);
+                      StructuredRecordToGenericRecordTransform.class,
+                      JavaScriptTransform.class,
+                      PythonEvaluator.class, PythonInterpreter.class);
```
gsps1: why do we need to add PythonInterpreter?

gsps1:
just couple of questions, lgtm otherwise.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
albertshau: hm good point. Not sure why I added this, ran tests without it and they all pass. Will remove.

gsps1:
there is one more place where PythonInterpreter is used.

PR #140, opened 2016-02-04
(CDAP-4870) Use select for column case in Teradata Source

bdmogal:
Jira: [CDAP-4870](https://issues.cask.co/browse/CDAP-4870)


database-plugins/widgets/Teradata-batchsource.json
```diff
@@ -65,6 +65,19 @@
           "widget-type": "textbox",
           "label": "Split-By Field Name",
           "name": "splitBy"
+        },
+        {
+          "widget-type": "select",
+          "label": "Column Case",
```
gsps1: Column name case

gsps1:
LGTM

PR #144, opened 2016-02-05
compile python script during configurePipeline stage

gsps1:
call init during configurePipeline stage.

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
@@ -132,6 +132,31 @@ public class PythonEvaluatorTest {
     Assert.assertEquals(expectedListField, output.get("arrayField"));
   }
 
+  @Test(expected = Exception.class)
+  public void testScriptCompilationValidation() throws Exception {
+    Schema outputSchema = Schema.recordOf(
+      "smallerSchema",
+      Schema.Field.of("intField", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("longField", Schema.of(Schema.Type.LONG)));
+
+    PythonEvaluator.Config config = new PythonEvaluator.Config(
+      "def transform(x, emitter, context):\n" +
+        " y = {}\n" +
+        "y['intField'] *= 1024\n" +
+        "  y['longField'] *= 1024\n" +
+        "  emitter.emit(y)",
+      outputSchema.toString());
+
+    Schema inputSchema = Schema.recordOf(
+      "biggerSchema",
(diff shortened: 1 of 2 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -132,6 +132,31 @@ public void testSimple() throws Exception {
...
+  @Test(expected = Exception.class)
+  public void testSchemaValidationInvalidScript() throws Exception {
+    Schema outputSchema = Schema.recordOf(
+      "smallerSchema",
+      Schema.Field.of("intField", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("longField", Schema.of(Schema.Type.LONG)));
+
+    PythonEvaluator.Config config = new PythonEvaluator.Config(
+      "def transform(x, emitter, context):\n" +
+        " y = {}\n" +
```
anwar6953, line 144: The transform method won't actually be executed, right?
This simply fails because the indentation of the script is wrong?

gsps1, line 144: yea, this would just catch compilation failures.

gsps1, line 144: any suggestions to make it more robust ?

anwar6953, line 144: That's fine. Just that the test case was misleading (implied catching errors due to invalid schema).

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -132,8 +132,8 @@ public void testSimple() throws Exception {
     Assert.assertEquals(expectedListField, output.get("arrayField"));
   }
 
-  @Test
-  public void testSchemaValidation() throws Exception {
+  @Test(expected = Exception.class)
+  public void testSchemaValidationInvalidScript() throws Exception {
```
albertshau: nit: name is misleading, since no schema validation is being tested. Can rename to something like testScriptCompilationValidation

albertshau:
minor nit, otherwise lgtm. fix then merge.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
gsps1: done.

PR #146, opened 2016-02-08
CDAP-4902 Add reference for emitting error from python transform

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-4902

### DIFF
--- a/core-plugins/docs/PythonEvaluator-transform.md
+++ b/core-plugins/docs/PythonEvaluator-transform.md
@@ -3,8 +3,10 @@
 
 Description
 -----------
-Executes user-provided Python code that transforms one record into another.
-
+Executes user-provided python code that transforms one record into zero or more records.
+Each input record is converted into a dictionary which can be directly accessed in
+python. The transform expects to receive a dictionary as input, which it can
+process and emit zero or more transformed dictionaries, or emit an error dictionary using the provided emitter object.
 
 Configuration
 -------------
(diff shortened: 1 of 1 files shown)


core-plugins/docs/PythonEvaluator-transform.md
```diff
@@ -25,35 +25,40 @@ schema is the same as the input schema.
...
-                    {\"name\":\"tax\",\"type\":\"double\"},
-                    {\"name\":\"total\",\"type\":\"double\"}
-                ]
-            }"
-        }
+      "name": "PythonEvaluator",
+      "properties": {
+        "script": "def transform(record, emitter, context):
+                     if (record['subtotal'] < 0):
+                      emitter.emitError({
```
albertshau: indent another space to be consistent

core-plugins/docs/PythonEvaluator-transform.md
```diff
@@ -25,35 +25,40 @@ schema is the same as the input schema.
...
+                        'errorCode' : 10,
+                        'errorMsg' : 'subtotal is less than 0',
+                        'invalidRecord' : record
+                      })
+                     else:
+                      tax = record['subtotal'] * 0.0975
```
albertshau: Use 2 spaces or 4 spaces for indentation instead of just 1.

core-plugins/docs/PythonEvaluator-transform.md
```diff
@@ -25,35 +25,40 @@ schema is the same as the input schema.
 
 Example
 -------
-The transform takes records that have a ``'subtotal'`` field, calculates ``'tax'`` and
-``'total'`` fields based on the subtotal, and then returns a record, as a Python dictionary,
-containing those three fields:
+The transform takes records that have a ``'subtotal'`` field, if ``'subtotal'`` is negative, it emits an error,
```
albertshau: It doesn't make sense to add error dataset into the example without describing what it is and how to use it.  Should add some description about it, perhaps when describing the script configuration setting.

core-plugins/docs/PythonEvaluator-transform.md
```diff
@@ -25,35 +25,40 @@ schema is the same as the input schema.
...
+else it calculates ``'tax'`` and ``'total'`` fields based on the subtotal, and then returns a record,
+as a Python dictionary, containing those three fields:
```
awholegunch: The transform checks each record's `'subtotal'` field: if the `'subtotal'` is negative, it emits an error;
else, it calculates the `'tax'` and `'total'` fields based on the `'subtotal'`, and then returns a record
as a Python dictionary containing those three fields:

awholegunch: I wrote the above because the script does not test for the existence of a subtotal field and then act; it assumes all records have these fields. If you mean for the script to only run this calculation for records with a 'subtotal' field, it would need to be written differently, right?

core-plugins/docs/PythonEvaluator-transform.md
gsps1: the error dataset tooltip has this information "Dataset that collects error messages from emitter. Please check reference section for usage."

gsps1:
@albertshau @awholegunch have addressed the comments.

core-plugins/docs/PythonEvaluator-transform.md
albertshau: There is nothing in the reference section about it though.  I can look at the reference and see an example, but I still don't know what it is.  This document should be good as a document on its own.  Somebody can look at it directly in the ui without reading the tooltip, somebody can look at it in github without deploying it to cdap, etc.

awholegunch:
I made minor edits; otherwise, text LGTM.

albertshau:
lgtm, please squash commits before merging

PR #147, opened 2016-02-09
(CDAP-4924) Moved the table existence check to job setup time from de…

bdmogal:
…ploy time.

Jira: [CDAP-4924](https://issues.cask.co/browse/CDAP-4924)

### DIFF
--- a/database-plugins/src/test/java/co/cask/hydrator/plugin/teradata/test/TeradataPluginTestRun.java
+++ b/database-plugins/src/test/java/co/cask/hydrator/plugin/teradata/test/TeradataPluginTestRun.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 Cask Data, Inc.
+ * Copyright © 2015-2016 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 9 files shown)


database-plugins/src/test/java/co/cask/hydrator/plugin/teradata/test/TeradataPluginTestRun.java
```diff
@@ -358,56 +358,22 @@ public void testNonExistentDBTable() throws Exception {
 
     ETLBatchConfig etlConfig = new ETLBatchConfig("* * * * *", sourceBadName, table, transforms);
     Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "dbSourceTest");
```
albertshau: should change the name of this app.  There are travis test failures that seem to indicate some other test uses the same app name.

albertshau:
comment about test failures, otherwise lgtm.  Verified on my cluster that pipelines using these plugins are ok.  Saw permgen get filled, then garbage collected successfully.

database-plugins/src/test/java/co/cask/hydrator/plugin/teradata/test/TeradataPluginTestRun.java
bdmogal: will try and make these tests more foolproof. Really need a better way of cleaning up when using TestBase - https://issues.cask.co/browse/CDAP-4925

bdmogal:
@albertshau fixed test failures. Please review again.

albertshau:
:+1:

PR #149, opened 2016-02-09
Merge release/1.2 to develop

awholegunch:
This branch was branched from release/1.2, and then develop was merged into it and the conflict in the pom.xml resolved in favor of develop.

PR #151, opened 2016-02-10
Remove python-evaluator-transform, left over from pull#149.

anwar6953:
Remove python-evaluator-transform, left over from https://github.com/caskdata/hydrator-plugins/pull/149.


albertshau:
:100:

PR #152, opened 2016-02-10
CDAP-4950 add a workaround for the hive jdbc driver

albertshau:
Added a workaround for drivers that dont support commit, like the
hive jdbc driver. Added a plugin property for enabling auto
commit, which the input format sets to false by default. If auto
commit is enabled, any Connection created by drivers will be
wrapped with a NoOpCommitConnection that doesn't do anything when
commit is called.


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/ETLDBInputFormat.java
```diff
@@ -39,17 +40,22 @@
  * Class that extends {@link DBInputFormat} to load the database driver class correctly.
  */
 public class ETLDBInputFormat extends DBInputFormat {
+  public static final String AUTO_COMMIT_ENABLED = "co.cask.hydrator.db.autocommit.enabled";
```
bdmogal: nit: although this is purely internal, any reason for using this long, almost package-name like parameter? Can we just use `autocommit.enabled`?

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/NoOpCommitConnection.java
```diff
@@ -0,0 +1,322 @@
...
+import java.sql.Struct;
+import java.util.Map;
+import java.util.Properties;
+import java.util.concurrent.Executor;
+
+/**
+ * A hack to work around jdbc drivers that create connections that don't support commit.
+ * This is true of the Hive jdbc driver. Delegates all operations except commit, which is a no-op.
+ */
+public class NoOpCommitConnection implements Connection {
```
bdmogal, line 42: does this have to be a public class? Do you think we'd use it in the ETLDBOutputFormat too?

bdmogal:
@albertshau just a couple minor comments.

However, this may be a generic config that the sink could use as well, right? Should we update `ETLDBOutputFormat` to use this? 

If we're updating  `ETLDBOutputFormat`,  can we move it to the right package as well? Its package is not consistent with `ETLDBInputFormat` right now

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/ETLDBInputFormat.java
albertshau: just trying to make sure there are no conflicts between config keys, had a bad experience with HCat and avro with overriding keys.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/NoOpCommitConnection.java
albertshau, line 42: doesn't have to be public, will restrict it

albertshau, line 42: though actually, like you mentioned, if its in output format it will have to be public

sreevatsanraman:
LGTM

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/ETLDBInputFormat.java
bdmogal: :+1:

PR #159, opened 2016-02-21
Changing 'ts' to 'offset' for each line read from file source

nitinmotgi:


bdmogal:
LGTM. Do we have to worry about backward-compatibility here though? I mean what if someone has a pipeline with a FileBatchSource that expects a `ts` field?

nitinmotgi:
I would be surprised if anyone would be using that. That field never made sense. 

###
Random auto-corrects and typos are my special gift to you. When I forward they are from others.

PR #160, opened 2016-02-24
CDAP-4908 remove table property from db source

albertshau:
The table name is not required at all. It makes especially little
sense when the import query contains multiple tables.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
@@ -95,7 +95,8 @@ public class DBSink extends BatchSink<StructuredRecord, DBRecord, NullWritable> 
     // make sure that the table exists
     try {
       Preconditions.checkArgument(
-        dbManager.tableExists(driverClass), "Table %s does not exist. Please check that the 'tableName' property " +
+        dbManager.tableExists(driverClass, dbSinkConfig.tableName),
+        "Table %s does not exist. Please check that the 'tableName' property " +
           "has been set correctly, and that the connection string %s points to a valid database.",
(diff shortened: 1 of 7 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
```diff
@@ -191,6 +192,9 @@ private void setResultSetMetadata() throws Exception {
   public static class DBSinkConfig extends DBConfig {
     @Description("Comma-separated list of columns in the specified table to export to.")
     public String columns;
+
+    @Description("Name of the database table to use as a source or a sink.")
```
gsps1: just sink right?

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
```diff
@@ -95,7 +95,8 @@ public void prepareRun(BatchSinkContext context) {
     // make sure that the table exists
     try {
       Preconditions.checkArgument(
-        dbManager.tableExists(driverClass), "Table %s does not exist. Please check that the 'tableName' property " +
+        dbManager.tableExists(driverClass, dbSinkConfig.tableName),
+        "Table %s does not exist. Please check that the 'tableName' property " +
```
gsps1, line 99: but tableName property is removed right? 

Edit : added for sink, alright.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
albertshau: good catch, will change.

gsps1:
lgtm

PR #162, opened 2016-02-25
CDAP-4143 handle semicolons at the end of a query.

albertshau:
We have noticed that many users put a semicolon at the end of the
query. The db input formats don't work in such cases, which is
not very friendly. So check for semicolon and strip it if it exists.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBConfig.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBConfig.java
@@ -65,6 +65,17 @@ public class DBConfig extends PluginConfig {
 
   protected String cleanQuery(String query) {
     query = query.trim();
-    return query.endsWith(";") ? query.substring(0, query.length() - 1) : query;
+    // find the last character that is not whitespace or a semicolon
+    int idx = query.length() - 1;
+    char currChar = query.charAt(idx);
+    while (idx > 0 && currChar == ';' || Character.isWhitespace(currChar)) {
+      idx--;
+      currChar = query.charAt(idx);
+    }
+    // why would somebody do this?
+    if (idx == 0) {
+      return "";
+    }
+    return query.substring(0, idx + 1);
   }
 }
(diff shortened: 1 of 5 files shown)


database-plugins/src/test/java/co/cask/hydrator/plugin/DBConfigTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * the License.
+ */
+
+package co.cask.hydrator.plugin;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ */
```
bdmogal, line 24: nit: empty javadoc

database-plugins/src/main/java/co/cask/hydrator/plugin/DBConfig.java
```diff
@@ -62,4 +62,9 @@
   public DBConfig() {
     jdbcPluginType = "jdbc";
   }
+
+  protected String cleanQuery(String query) {
+    query = query.trim();
+    return query.endsWith(";") ? query.substring(0, query.length() - 1) : query;
```
bdmogal: do you think we should handle multiple semi-colons? Or is that overkill?

Also, is the problem with only semicolons or with a set of characters at the end?

bdmogal:
couple questions, other than that LGTM

database-plugins/src/main/java/co/cask/hydrator/plugin/DBConfig.java
albertshau: eh might as well handle multiple semi-colons.  Not sure what you mean by set of characters though?

bdmogal: I meant do we need to handle any other special characters?

albertshau: i don't know of anything else

bdmogal: :+1: can do it if we see a use case

bdmogal:
:+1:

PR #164, opened 2016-02-25
CDAP-3749 db user does not require a password

albertshau:
Fix the database plugins so that specifying a user does not
require also specifying a password. It is completely valid,
though perhaps a bad idea, for a user to have no password.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBManager.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBManager.java
@@ -47,9 +47,6 @@ public class DBManager implements Destroyable {
     Preconditions.checkArgument(!(dbConfig.user == null && dbConfig.password != null),
                                 "user is null. Please provide both user name and password if database requires " +
                                   "authentication. If not, please remove password and retry.");
-    Preconditions.checkArgument(!(dbConfig.user != null && dbConfig.password == null),
-                                "password is null. Please provide both user name and password if database requires" +
-                                  "authentication. If not, please remove dbUser and retry.");
     Class<? extends Driver> jdbcDriverClass = pipelineConfigurer.usePluginClass(dbConfig.jdbcPluginType,
                                                                                 dbConfig.jdbcPluginName,
                                                                                 jdbcPluginId,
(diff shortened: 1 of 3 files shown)


database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/BatchETLDBTestRun.java
```diff
@@ -362,6 +346,17 @@ public void testUserNamePasswordCombinations() throws Exception {
...
+    Map<String, String> nullPassword = new HashMap<>(baseSourceProps);
+    emptyPassword.put(Properties.DB.USER, "emptyPwdUser");
+    database = new ETLStage("databaseSource", new Plugin("Database", nullPassword));
+    etlConfig = ETLBatchConfig.builder("* * * * *")
+      .setSource(database)
+      .addSink(table)
+      .addConnection(database.getName(), table.getName())
+      .build();
+    appRequest = new AppRequest<>(ETLBATCH_ARTIFACT, etlConfig);
+    TestBase.deployApplication(appId, appRequest);
```
bdmogal: for consistency, should we add a test using DBSink as well? Also, not sure why `deployApplication` in this class needs to be called with the class name

bdmogal:
minor comment. Also, do we have a test for empty password? Other than that, LGTM

database-plugins/src/main/java/co/cask/hydrator/plugin/DBManager.java
```diff
@@ -47,9 +47,6 @@ public void validateJDBCPluginPipeline(PipelineConfigurer pipelineConfigurer, St
     Preconditions.checkArgument(!(dbConfig.user == null && dbConfig.password != null),
                                 "user is null. Please provide both user name and password if database requires " +
                                   "authentication. If not, please remove password and retry.");
```
gsps1, line 49: looks like the username is always required, why can't we make that field non-nullable ?

albertshau:
yep empty password is tested right above that section

database-plugins/src/main/java/co/cask/hydrator/plugin/DBManager.java
albertshau, line 49: its not always required, you can connect to databases without a username

gsps1, line 49: oh, this check checks for password but no username, alright.

albertshau, line 49: yep

bdmogal:
:+1:

PR #165, opened 2016-02-25
CDAP-5038 LogParserTransform always has a fixed output schema

gsps1:
JIRA : https://issues.cask.co/browse/CDAP-5038


albertshau:
lgtm

PR #163, opened 2016-02-25
(CDAP-5047) Added a Batch Source for Azure Blob Store based on FileBa…

bdmogal:
…tchSource

Jira: [CDAP-5047](https://issues.cask.co/browse/CDAP-5047)

PR #166, opened 2016-02-25
Microsoft Azure Blob Storage Batch Source Plugin Documentation corrections

awholegunch:
- Fix typos. 
- Correct usage and explanation of the "account" property.
- Correct usage to include "Microsoft Azure" consistently.

See https://issues.cask.co/browse/CDAP-5047 and https://github.com/caskdata/hydrator-plugins/pull/163.

### DIFF
--- a/core-plugins/docs/AzureBlobStore-batchsource.md
+++ b/core-plugins/docs/AzureBlobStore-batchsource.md
@@ -32,7 +32,7 @@ Example
 This example connects to Microsoft Azure Blob Storage and reads in files found in the
 specified directory. This example uses the 'hydrator' container at the
 'hydratorstorage.blob.core.windows.net' endpoint on Microsoft Azure Storage, using the
-'hydratorstorage' account:
+'hydratorstorage' account name:
 
     {
         "name": "AzureBlobStore",
--- a/core-plugins/docs/AzureBlobStore-batchsource.md
+++ b/core-plugins/docs/AzureBlobStore-batchsource.md
@@ -1,34 +1,38 @@
-# Azure Blob Store Batch Source
+# Microsoft Azure Blob Storage Batch Source
 
 Description
 -----------
 
-Batch source to use Azure Blob Store as a source.
+Batch source to use Microsoft Azure Blob Storage as a source.
 
 Use Case
 --------
 
-This source is used whenever you need to read from a Azure Blob Store.
-For example, you may want to read in files from Azure Blob Storage, parse them and then store
-them in a Microsoft SQL Server Database.
+This source is used whenever you need to read from Microsoft Azure Blob Storage. For
+example, you may want to read in files from Microsoft Azure Blob Storage, parse them and
+then store them in a Microsoft SQL Server Database.
 
 Properties
 ----------
 
-**path:** The path on Azure Blob Storage to use as input.
+**path:** The path on Microsoft Azure Blob Storage to use as input.
 
-**account:** The Microsoft Azure Storage account to use.
+**account:** The Microsoft Azure Blob Storage endpoint, with the Microsoft Azure Storage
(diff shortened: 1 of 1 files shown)


core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -1,34 +1,38 @@
...
 
 Example
 -------
 
-This example connects to Azure Blob Store and reads in files found in the specified directory. This example uses
-the 'hydrator' container on the 'hydratorstorage.blob.core.windows.net' account on Azure Blob Store:
+This example connects to Microsoft Azure Blob Storage and reads in files found in the
+specified directory. This example uses the 'hydrator' container at the
+'hydratorstorage.blob.core.windows.net' endpoint on Microsoft Azure Storage, using the
+'hydratorstorage' account:
```
bdmogal: this adds to confusion now though - we call `hydratorstorage` as an account here, but refer to account differently above.

core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -1,34 +1,38 @@
...
+then store them in a Microsoft SQL Server Database.
 
 Properties
 ----------
 
-**path:** The path on Azure Blob Storage to use as input.
+**path:** The path on Microsoft Azure Blob Storage to use as input.
 
-**account:** The Microsoft Azure Storage account to use.
+**account:** The Microsoft Azure Blob Storage endpoint, with the Microsoft Azure Storage
```
bdmogal: is it ok if we don't make this change right now?

bdmogal: Ok, I just checked https://azure.microsoft.com/en-us/documentation/articles/hdinsight-hadoop-use-blob-storage/ as well, it seems to indicate that account name is just the first part like you call out. Since we expect the full name containing blob.core.windows.net, maybe lets just make sure that we convey that this is the full name, including blob.core.windows.net.

core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -1,34 +1,38 @@
...
+Must be a valid base64 encoded storage key provided by Microsoft Azure.
 
 Example
 -------
 
-This example connects to Azure Blob Store and reads in files found in the specified directory. This example uses
-the 'hydrator' container on the 'hydratorstorage.blob.core.windows.net' account on Azure Blob Store:
+This example connects to Microsoft Azure Blob Storage and reads in files found in the
+specified directory. This example uses the 'hydrator' container at the
+'hydratorstorage.blob.core.windows.net' endpoint on Microsoft Azure Storage, using the
```
bdmogal: endpoint seems like new terminology. Until we introduce it everywhere else, should we just call it `account`?

bdmogal:
@awholegunch just one comment. Although, now that you pointed out, we should create a jira to just accept an account name and suffix the endpoint in the code (if we're satisfied that there can only be a single endpoint for Azure).

core-plugins/docs/AzureBlobStore-batchsource.md
awholegunch: Well, that's what Microsoft calls it, so I followed their lead. But if you want to delete it, what about:

```
This example uses the 'hydrator' container at
'hydratorstorage.blob.core.windows.net' (Microsoft Azure Storage), using the...
```

awholegunch:
Created https://issues.cask.co/browse/CDAP-5127 "Simplify the Microsoft Azure Blob Storage Batch Source Plugin API".

core-plugins/docs/AzureBlobStore-batchsource.md
bdmogal: yes that's better

bdmogal:
LGTM

PR #172, opened 2016-03-16
getSchema and listTables plugin endpoint implementations in DBSource

gsps1:
Implementing listTables and getSchema endpoints in dbsource. 

REST call for this plugin method would look like,
`curl -v -X POST "<hostname>:10000/v3/namespaces/default/artifacts/database-plugins/versions/1.3.0-SNAPSHOT/plugintypes/batchsource/plugins/Database/methods/listTables?scope=system"  -d @dbconfig.json`

JIRA : https://issues.cask.co/browse/CDAP-5153
Blocked by : https://github.com/caskdata/cdap/pull/5264

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
@@ -16,6 +16,10 @@
 
 package co.cask.hydrator.plugin;
 
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import com.google.common.base.Throwables;
+import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -24,8 +28,13 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.sql.Driver;
 import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Types;
 import java.util.Hashtable;
 import java.util.List;
+import javax.annotation.Nullable;
 import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
 import javax.management.MBeanRegistrationException;
@@ -55,6 +64,122 @@ public final class DBUtils {
     }
     shutDownMySQLAbandonedConnectionCleanupThread(pluginClassLoader);
     unregisterOracleMBean(pluginClassLoader);
+  }
+
+  /**
+   * Ensures that the JDBC Driver specified in configuration is available and can be loaded. Also registers it with
+   * {@link DriverManager} if it is not already registered.
+   */
+  public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver> jdbcDriverClass,
+                                                          String connectionString,
+                                                          String jdbcPluginType, String jdbcPluginName)
+    throws IllegalAccessException, InstantiationException, SQLException {
+
+    try {
+      DriverManager.getDriver(connectionString);
+      return new DriverCleanup(null);
+    } catch (SQLException e) {
+      // Driver not found. We will try to register it with the DriverManager.
+      LOG.debug("Plugin Type: {} and Plugin Name: {}; Driver Class: {} not found. Registering JDBC driver via shim {} ",
+                jdbcPluginType, jdbcPluginName, jdbcDriverClass.getName(),
+                JDBCDriverShim.class.getName());
+      final JDBCDriverShim driverShim = new JDBCDriverShim(jdbcDriverClass.newInstance());
+      try {
+        DBUtils.deregisterAllDrivers(jdbcDriverClass);
+      } catch (NoSuchFieldException | ClassNotFoundException e1) {
+        LOG.error("Unable to deregister JDBC Driver class {}", jdbcDriverClass);
+      }
+      DriverManager.registerDriver(driverShim);
+      return new DriverCleanup(driverShim);
+    }
+  }
+
+  /**
+   * Given the result set, get the metadata of the result set and return
+   * list of {@link co.cask.cdap.api.data.schema.Schema.Field},
+   * where name of the field is same as column name and type of the field is obtained using {@link DBUtils#getType(int)}
+   * @param resultSet result set of executed query
+   * @return list of schema fields
+   * @throws SQLException
+   */
+  public static List<Schema.Field> getSchemaFields(ResultSet resultSet) throws SQLException {
+    List<Schema.Field> schemaFields = Lists.newArrayList();
+    ResultSetMetaData metadata = resultSet.getMetaData();
+    // ResultSetMetadata columns are numbered starting with 1
+    for (int i = 1; i <= metadata.getColumnCount(); i++) {
+      String columnName = metadata.getColumnName(i);
+      int columnSqlType = metadata.getColumnType(i);
+      Schema columnSchema = Schema.of(getType(columnSqlType));
+      if (ResultSetMetaData.columnNullable == metadata.isNullable(i)) {
+        columnSchema = Schema.nullableOf(columnSchema);
+      }
+      Schema.Field field = Schema.Field.of(columnName, columnSchema);
+      schemaFields.add(field);
+    }
+    return schemaFields;
+  }
+
+  // given a sql type return schema type
+  private static Schema.Type getType(int sqlType) throws SQLException {
+    // Type.STRING covers sql types - VARCHAR,CHAR,CLOB,LONGNVARCHAR,LONGVARCHAR,NCHAR,NCLOB,NVARCHAR
+    Schema.Type type = Schema.Type.STRING;
+    switch (sqlType) {
+      case Types.NULL:
+        type = Schema.Type.NULL;
+        break;
+
+      case Types.BOOLEAN:
+      case Types.BIT:
+        type = Schema.Type.BOOLEAN;
+        break;
+
+      case Types.TINYINT:
+      case Types.SMALLINT:
+      case Types.INTEGER:
+        type = Schema.Type.INT;
+        break;
+
+      case Types.BIGINT:
+        type = Schema.Type.LONG;
+        break;
+
+      case Types.REAL:
+      case Types.FLOAT:
+        type = Schema.Type.FLOAT;
+        break;
+
+      case Types.NUMERIC:
+      case Types.DECIMAL:
+      case Types.DOUBLE:
+        type = Schema.Type.DOUBLE;
+        break;
+
+      case Types.DATE:
+      case Types.TIME:
+      case Types.TIMESTAMP:
+        type = Schema.Type.LONG;
+        break;
+
+      case Types.BINARY:
+      case Types.VARBINARY:
+      case Types.LONGVARBINARY:
+      case Types.BLOB:
+        type = Schema.Type.BYTES;
+        break;
+
+      case Types.ARRAY:
+      case Types.DATALINK:
+      case Types.DISTINCT:
+      case Types.JAVA_OBJECT:
+      case Types.OTHER:
+      case Types.REF:
+      case Types.ROWID:
+      case Types.SQLXML:
+      case Types.STRUCT:
+        throw new SQLException(new UnsupportedTypeException("Unsupported SQL Type: " + sqlType));
+    }
+
+    return type;
   }
 
   /**
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DriverCleanup.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DriverCleanup.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright © 2015 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin;
+
+import co.cask.cdap.etl.api.Destroyable;
+import com.google.common.base.Throwables;
+
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import javax.annotation.Nullable;
+
+/**
+ * class to de-register driver
+ */
+public class DriverCleanup implements Destroyable {
+  private final JDBCDriverShim driverShim;
+
+  DriverCleanup(@Nullable JDBCDriverShim driverShim) {
+    this.driverShim = driverShim;
+  }
+
+  public void destroy() {
+    if (driverShim != null) {
+      try {
+        DriverManager.deregisterDriver(driverShim);
+      } catch (SQLException e) {
+        throw Throwables.propagate(e);
+      }
+    }
+  }
+}
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
@@ -20,8 +20,11 @@ import co.cask.cdap.api.annotation.Description;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.plugin.EndpointPluginContext;
 import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.api.plugin.PluginProperties;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
@@ -32,6 +35,7 @@ import co.cask.hydrator.plugin.DBConfig;
 import co.cask.hydrator.plugin.DBManager;
 import co.cask.hydrator.plugin.DBRecord;
 import co.cask.hydrator.plugin.DBUtils;
+import co.cask.hydrator.plugin.DriverCleanup;
 import co.cask.hydrator.plugin.FieldCase;
 import co.cask.hydrator.plugin.StructuredRecordUtils;
 import org.apache.hadoop.conf.Configuration;
@@ -41,7 +45,16 @@ import org.apache.hadoop.mapreduce.lib.db.DBConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.sql.Connection;
 import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.SQLSyntaxErrorException;
+import java.sql.Statement;
+import javax.annotation.Nullable;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.Path;
 
 /**
  * Batch source to read from a Database table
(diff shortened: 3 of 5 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     dbManager.validateJDBCPluginPipeline(pipelineConfigurer, getJDBCPluginId());
   }
 
+  class ListTablesRequest {
+
+    @Description("JDBC connection string including database name.")
```
albertshau: are these used by the platform?

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    public String password;
+
+    @Description("Name of the JDBC plugin to use. This is the value of the 'name' key defined in the JSON file " +
+      "for the JDBC plugin.")
+    public String jdbcPluginName;
+
+    @Description("Type of the JDBC plugin to use. This is the value of the 'type' key defined in the JSON file " +
+      "for the JDBC plugin. Defaults to 'jdbc'.")
+    @Nullable
+    public String jdbcPluginType = "jdbc";
```
albertshau, line 90: this does not work as setting a default because the object is created through deserialization.  You need a no-args constructor that sets all defaults.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  }
+
+  class GetSchemaRequest extends ListTablesRequest {
+    @Description("Select query for which schema has to be populated.")
+    public String importQuery;
```
albertshau: I think we should rename this.  We don't want people to confuse this with the importQuery for the source, as this should be as lightweight of a query as possible.

nitinmotgi:
This PR shouldn't be merged without considering the impact on this PR https://github.com/caskdata/hydrator-plugins/pull/174

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -58,6 +60,43 @@ public static void cleanup(Class<? extends Driver> driverClass) {
   }
 
   /**
+   * Ensures that the JDBC Driver specified in configuration is available and can be loaded. Also registers it with
+   * {@link DriverManager} if it is not already registered.
+   */
+  public static JDBCDriverShim ensureJDBCDriverIsAvailable(Class<? extends Driver> jdbcDriverClass,
```
chtyim: `@Nullable` since this method can return null. Also good to have it in Javadoc saying under what condition it will return `null`.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -58,6 +60,43 @@ public static void cleanup(Class<? extends Driver> driverClass) {
...
+      } catch (NoSuchFieldException | ClassNotFoundException e1) {
+        LOG.error("Unable to deregister JDBC Driver class {}", jdbcDriverClass);
+      }
+      DriverManager.registerDriver(driverShim);
+      return driverShim;
+    }
+    return null;
+  }
+
+  public static void deregisterShim(JDBCDriverShim driverShim) {
```
chtyim: Seems like the `driverShim` can be null from the caller (in DBManager class above). So this param should have `@Nullable` and null should be handled in the method.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
chtyim, line 90: It's ok if there is no constructor like the class in here. Java compiler will generate the no-arg default constructor and all field initializers will be part of it.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  public List<String> listTables(ListTablesRequest request,
+                                 EndpointPluginContext pluginContext) throws SQLException,
+    InstantiationException, IllegalAccessException {
+
+    JDBCDriverShim driverShim = loadPluginClassAndGetDriver(request, pluginContext);
+    List<String> tableNames = new ArrayList();
+
+    try {
+      // initialize connection
+      Connection connection = getConnection(request.connectionString, request.user, request.password);
```
chtyim: `Connection` is `AutoCloseable`. Use the `try (resource)` syntax instead.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+                                                 request.jdbcPluginType, request.jdbcPluginName);
+    } catch (IllegalAccessException | InstantiationException | SQLException e) {
+      LOG.error("Unable to load or register JDBC driver {} while getting the tables in a database.",
+                driverClass, e);
+      throw Throwables.propagate(e);
+    }
+  }
+
+  private Connection getConnection(String connectionString, String user, String password) throws SQLException {
+    if (user == null) {
```
chtyim, line 141: If `user` can be `null`, then it should be annotated with `@Nullable`.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +83,127 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  @Path("getSchema")
+  public Map<String, String> getSchema(GetSchemaRequest request,
+                                       EndpointPluginContext pluginContext)
+    throws SQLException, InstantiationException, IllegalAccessException {
+
+    JDBCDriverShim driverShim = loadPluginClassAndGetDriver(request, pluginContext);
+
+    try {
+      Connection connection = getConnection(request.connectionString, request.user, request.password);
+      Map<String, String> resultSchema = new HashMap<>();
```
chtyim: Rather then using a `HashMap`, it's better to use a `LinkedHashMap` so that the ordering of the insertion is maintained (which is the ordering of the fields in the table).

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1: yea, selectQuery would be a better name.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1: oh, you mean if we are storing these descriptions and returning ?, we aren't. will remove.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1: but that should be across config and these classes.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1: good point. have updated it.

gsps1:
@albertshau @chtyim thanks for review, have updated the PR.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -58,6 +61,50 @@ public static void cleanup(Class<? extends Driver> driverClass) {
   }
 
   /**
+   * Ensures that the JDBC Driver specified in configuration is available and can be loaded. Also registers it with
+   * {@link DriverManager} if it is not already registered.
+   * @return {@link JDBCDriverShim} returns null if we are able to locate an existing driver for the
+   * connection string using the DriverManager.
+   */
+  @Nullable public static JDBCDriverShim ensureJDBCDriverIsAvailable(Class<? extends Driver> jdbcDriverClass,
```
chtyim: Have the `@Nullable` on a new line.

``` java
@Nullable
public static ...
```

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -58,6 +61,50 @@ public static void cleanup(Class<? extends Driver> driverClass) {
   }
 
   /**
+   * Ensures that the JDBC Driver specified in configuration is available and can be loaded. Also registers it with
+   * {@link DriverManager} if it is not already registered.
+   * @return {@link JDBCDriverShim} returns null if we are able to locate an existing driver for the
```
chtyim: Seems really strange to return `null` to indicate if we are able to locate an existing driver. Why is it so? Is it purely just for `DBUtils.deregisterShim(driverShim)` call? An API design like this make classes tightly coupled, which is not good. If the return `driverShim` is actually not used for normal operation but just for resource free up, it's better for this method to return a `Cancellable` or `Closeable` (non-null), which can do the resource cleanup inside (or no-op, depending on the result).

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -58,6 +61,50 @@ public static void cleanup(Class<? extends Driver> driverClass) {
...
+   * deregister driver shim if its not null
+   * @param driverShim
+   */
+  public static void deregisterShim(@Nullable JDBCDriverShim driverShim) {
+    try {
+      if (driverShim != null) {
+        DriverManager.deregisterDriver(driverShim);
+      }
+    } catch (SQLException e) {
+      LOG.warn("Error while de-registering JDBC drivers in ETLDBOutputFormat.", e);
```
albertshau: this warning is inaccurate (not in ETLDBOutputFormat).  method should throw an exception and caller should handle it rather than logging, then propagating as runtime exception.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  private JDBCDriverShim loadPluginClassAndGetDriver(ListTablesRequest request, EndpointPluginContext pluginContext)
+    throws IllegalAccessException, InstantiationException, SQLException {
+    Class<? extends Driver> driverClass =
+      pluginContext.loadPluginClass(request.jdbcPluginType, request.jdbcPluginName, PluginProperties.builder().build());
+
+    try {
+      return DBUtils.ensureJDBCDriverIsAvailable(driverClass, request.connectionString,
+                                                 request.jdbcPluginType, request.jdbcPluginName);
+    } catch (IllegalAccessException | InstantiationException | SQLException e) {
+      LOG.error("Unable to load or register JDBC driver {} while getting the tables in a database.",
```
albertshau: in general, log then propagate is a sign of error handling happening in the wrong place.  This method says it throws these exceptions already, we should not wrap them in a runtime exception.

This error message is also wrong, its not getting database tables.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    public String user;
+    @Nullable
+    public String password;
+    public String jdbcPluginName;
+    @Nullable
+    public String jdbcPluginType = "jdbc";
+  }
+
+  class GetSchemaRequest extends ListTablesRequest {
+    public String importQuery;
```
albertshau: this should not be called importQuery. We don't want somebody to confuse this with the importQuery that the source takes

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+        ResultSetMetaData resultSetMetaData = statement.executeQuery(query).getMetaData();
+
+        // column index starts from 1 in ResultSetMetaData
+        for (int i = 1; i <= resultSetMetaData.getColumnCount(); i++) {
+          resultSchema.put(resultSetMetaData.getColumnName(i), resultSetMetaData.getColumnTypeName(i));
+        }
+      }
+      return resultSchema;
+    } catch (SQLException e) {
+      LOG.error("Exception while trying to get schema in a database for query {}.",
```
albertshau: nit: 'schema for query', don't need 'in a database'

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+                request.importQuery, e);
+
+      if (e instanceof SQLSyntaxErrorException) {
+        throw new BadRequestException(e.getMessage());
+      }
+      throw Throwables.propagate(e);
```
albertshau: can just `throw e` again.  Don't need to wrap it in a runtime exception.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+        for (int i = 1; i <= resultSetMetaData.getColumnCount(); i++) {
+          resultSchema.put(resultSetMetaData.getColumnName(i), resultSetMetaData.getColumnTypeName(i));
+        }
+      }
+      return resultSchema;
+    } catch (SQLException e) {
+      LOG.error("Exception while trying to get schema in a database for query {}.",
+                request.importQuery, e);
+
+      if (e instanceof SQLSyntaxErrorException) {
```
albertshau: this can just be another catch right?

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+  public Map<String, String> getSchema(GetSchemaRequest request,
+                                       EndpointPluginContext pluginContext)
+    throws SQLException, InstantiationException, IllegalAccessException, BadRequestException {
+
+    JDBCDriverShim driverShim = loadPluginClassAndGetDriver(request, pluginContext);
+    try {
+      Map<String, String> resultSchema = new LinkedHashMap<>();
+      try (Connection connection = getConnection(request.connectionString, request.user, request.password);) {
+        Statement statement = connection.createStatement();
+        String query  = request.importQuery;
```
albertshau: nit: don't need this, can just use request.importQuery

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+              tableNames.add(tableName);
+            }
+          }
+        }
+      }
+      return tableNames;
+    } catch (SQLException e) {
+      LOG.error("Exception while trying to get the tables in a database for connection {}.",
+                request.connectionString, e);
+      if (e instanceof SQLSyntaxErrorException) {
```
albertshau: this seems like it can be just another catch

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+        throw new BadRequestException(e.getMessage());
+      }
+      throw Throwables.propagate(e);
```
albertshau: In general, avoid using Throwables.propagate if you can, otherwise we start losing the ability to do any sane error handling and start getting exceptions wrapped in exceptions wrapped in exceptions.  Here you can just throw e.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+            String tableName = rs.getString(3);
+            if (tableName != null) {
+              tableNames.add(tableName);
+            }
+          }
+        }
+      }
+      return tableNames;
+    } catch (SQLException e) {
+      LOG.error("Exception while trying to get the tables in a database for connection {}.",
```
albertshau: does this show up as an error in app fabric logs?  If so, does app-fabric already log the exception this method throws?  If so we should remove this log message (and other similar log messages in this code)

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    @Nullable
+    public String jdbcPluginType = "jdbc";
+  }
+
+  class GetSchemaRequest extends ListTablesRequest {
+    public String importQuery;
+  }
+
+  /**
+   * Endpoint method to get the list of tables in a database
```
albertshau: nit: insert newline between description and params for consistent styling

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+        throw new BadRequestException(e.getMessage());
+      }
+      throw Throwables.propagate(e);
+    } finally {
+      DBUtils.deregisterShim(driverShim);
+    }
+  }
+
+  /**
+   * Endpoint method to get the output schema of a query.
```
albertshau, line 95: nit: insert newline between description and params for consistent styling

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+   * @param pluginContext context to create plugins
+   * @return Map of field name to field type.
+   * @throws SQLException
+   * @throws InstantiationException
+   * @throws IllegalAccessException
+   * @throws BadRequestException If executing the query threw {@link SQLSyntaxErrorException}
+   * we get the message and throw it as BadRequestException
+   */
+  @Path("getSchema")
+  public Map<String, String> getSchema(GetSchemaRequest request,
```
albertshau: this schema returned here is different than the schema we use everywhere else (avro style schema).  Can we just return a cdap Schema instead of a Map? Or Schema.toString() instead of the map? That way you can call this method, then plug it directly into a schema field.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -66,6 +85,135 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    InstantiationException, IllegalAccessException, BadRequestException {
+
+    JDBCDriverShim driverShim = loadPluginClassAndGetDriver(request, pluginContext);
+    List<String> tableNames = new ArrayList();
+
+    try {
+      // initialize connection
+      try (Connection connection = getConnection(request.connectionString, request.user, request.password)) {
+        DatabaseMetaData metadata = connection.getMetaData();
+        try (ResultSet rs = metadata.getTables(null, null, "%", null)) {
```
albertshau: do we get the table schemas back?  Would be nice to be able to know a table's schema as well as its name.

database-plugins/src/main/java/co/cask/hydrator/plugin/DriverCleanup.java
```diff
@@ -0,0 +1,44 @@
...
+  DriverCleanup(JDBCDriverShim driverShim) {
+    this.driverShim = driverShim;
+  }
+
+  public void destroy() {
+    if (driverShim != null) {
+      try {
+        DriverManager.deregisterDriver(driverShim);
+      } catch (SQLException e) {
+        throw Throwables.propagate(e);
```
gsps1, line 41: using Throwables.propagate, as using throw(e) would require changing destroy signature, which would also get propagated to change in usages of this.

gsps1:
@albertshau have addressed the comments. please take a look again.

database-plugins/src/main/java/co/cask/hydrator/plugin/DriverCleanup.java
albertshau, line 41: yup in these situations its fine

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -158,39 +104,26 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
    * we get the message and throw it as BadRequestException
    */
   @Path("getSchema")
-  public Map<String, String> getSchema(GetSchemaRequest request,
+  public Schema getSchema(GetSchemaRequest request,
```
albertshau, line 107: Schema will not get serialized correctly unless the Gson instance used to do it has the SchemaTypeAdapter registered.  So we either need to return Schema.toString() here, or make a change in cdap to support the Schema object as a return type.

database-plugins/src/main/java/co/cask/hydrator/plugin/DriverCleanup.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class DriverCleanup implements Destroyable {
+  private final JDBCDriverShim driverShim;
+
+  DriverCleanup(JDBCDriverShim driverShim) {
+    this.driverShim = driverShim;
+  }
+
+  public void destroy() {
+    if (driverShim != null) {
```
albertshau, line 37: if this can be null lets annotate with Nullable

albertshau:
please check if logging then throwing an exception results in a double log in app-fabric.  If so, need to change all those places to just throw the exception with a descriptive message and skip the log. otherwise lgtm

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1, line 107: not sure why this is in outdated diff.. 

we already have SchemaTypeAdapter registered for GSON in CDAP.
`private static final Gson GSON = new GsonBuilder()
    .registerTypeAdapter(Schema.class, new SchemaTypeAdapter())
    .registerTypeAdapter(PluginClass.class, new PluginClassDeserializer())
    .create();` in ArtifactHttpHandler

albertshau:
:+1:

PR #173, opened 2016-03-21
CDAP-5340 Modified specification version and also updated labels appropriately.

nitinmotgi:
Update the UI JSON Specification [here](https://wiki.cask.co/display/CE/UI+JSON+Specification)  and also updates the files to include right version and also updated labels appropriately. This change will not affect UI labels right now.


ajainarayanan:
We have to decide on how to fix this in UI without making too much calls to backend(for widget JSONs) before using Hydrator studio. Closing this PR for now to take it up in release 3.5 of CDAP.

PR #174, opened 2016-03-21
CDAP-4856 Replace DBSource with TeradataSource

nitinmotgi:
TeradataSource uses Data Driven ETL where instead of using LIMIT in the query, it uses the bounding query that provides the bounds for generating the MR splits. This PR includes the changes for the names of the files and also classes. UI widget configuration was also modified accordingly. All tests passed after the change.


gokulavasan:
This might have backward compatibility issues? Even if that is not an issue, the source should be named as DatabaseSource or DBSource or DataDrivenDBSource and not Teradata.

gsps1:
Agree it should be called database source and not tera data source. Also bounding should be an additional field and shouldn't replace limit query.

gsps1:
Also asking users to have $CONDITIONS as compulsory in import query just to be replaced by splitBy field is a overhead for other db sources. its hacky too, can this be made better?

nitinmotgi:
@gokulavasan @gsps1 What backward compatibility issues ? It should be called Database and I have modified it. Not sure what happened, because it was supposed to be Database.

nitinmotgi:
@gsps1 it's not hacky! it's something thats generic and better approach that LIMIT that we had. So, please suggest a better approach that would across all databases and not just some.

chtyim:
As long as it is released as a new artifact version, there shouldn't be any compatibility issue.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -44,34 +45,38 @@
 import java.sql.Driver;
 
 /**
- * Batch source to read from a Database table
+ * Batch source to read from a DB table
  */
 @Plugin(type = "batchsource")
 @Name("Database")
-@Description("Reads from a database using a configurable SQL query." +
+@Description("Reads from a database table using a configurable SQL query." +
```
albertshau: this could be misleading, as somebody can use an import query that reads from multiple tables

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -79,15 +84,17 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+      DBConfiguration.configureDB(hConf, driverClass.getName(), sourceConfig.connectionString,
+                                  sourceConfig.user, sourceConfig.password);
     }
-    ETLDBInputFormat.setInput(hConf, DBRecord.class, dbSourceConfig.getImportQuery(),
-                              dbSourceConfig.getCountQuery(), dbSourceConfig.getEnableAutoCommit());
-    context.setInput(new SourceInputFormatProvider(ETLDBInputFormat.class, hConf));
+    DataDrivenETLDBInputFormat.setInput(hConf, DBRecord.class,
+                                        sourceConfig.getImportQuery(), sourceConfig.getBoundingQuery(),
+                                        sourceConfig.getEnableAutoCommit());
+    job.getConfiguration().set(DBConfiguration.INPUT_ORDER_BY_PROPERTY, sourceConfig.splitBy);
```
albertshau: nit: `hConf.set`

database-plugins/widgets/Database-batchsource.json
```diff
@@ -33,6 +33,11 @@
         },
         {
           "widget-type": "textbox",
+          "label": "Driver Class",
+          "name": "driverClass"
```
albertshau: how come this was added?  Didn't see it as a plugin config property

albertshau:
Some things to consider if we remove the previous import query:

Compatibility is an issue in that pipelines using the 1.2.x db source cannot be upgraded to use 1.3.x db source.

There are certain (poorly designed) queries where the split key is not indexed that would have really bad performance.

If we are indeed removing it, we can also remove ETLDBInputFormat

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
nitinmotgi: Modified table to table(s). This makes it clear that it could possibly be reading from multiple tables.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
nitinmotgi: changed.

nitinmotgi:
Addressed all comments from @albertshau. Compatibility is not an issue if some one wants to use older version they can always use the previous version of JAR.

albertshau:
By compatibility, I meant a pipeline using cdap-etl-batch-3.3.x and dbsource in database-plugins-1.2.x will not be able to be upgraded to a pipeline using cdap-etl-batch-3.4.x and dbsource in database-plugins-1.3.x.  It also will not be able to be upgraded to a pipeline using cdap-etl-batch-3.4.x and db source in database-plugins-1.2.x, because 1.2.x only extends 3.3.x and not 3.4.x.

If that is fine, lgtm.

nitinmotgi:
I think that should be fine @albertshau. It will make it very complicated otherwise.

albertshau:
lgtm

PR #175, opened 2016-03-23
CDAP-5191 FTP Batch Source implementation

gokulavasan:
FTP Batch Source and Unit test

JIRA : https://issues.cask.co/browse/CDAP-5191


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FTPBatchSource.java
```diff
@@ -0,0 +1,99 @@
...
+
+  // Limit the number of splits to 1 since FTPInputStream does not support seek;
+  private static String limitSplits(@Nullable String fsProperties) {
+    Map<String, String> providedProperties;
+    if (fsProperties == null) {
+      providedProperties = new HashMap<>();
+    } else {
+      providedProperties = GSON.fromJson(fsProperties, MAP_STRING_STRING_TYPE);
+    }
+    providedProperties.put(FileInputFormat.SPLIT_MINSIZE, Long.toString(Long.MAX_VALUE));
```
gsps1: in the comment you say limiting the splits to 1, but the value is Long.MAX_VALUE, shouldn't it be 1 ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FTPBatchSource.java
```diff
@@ -0,0 +1,99 @@
...
+    public String fileRegex;
+
+    @Description(FileBatchSource.INPUT_FORMAT_CLASS_DESCRIPTION)
+    @Nullable
+    public String inputFormatClassName;
+
+    public FTPBatchSourceConfig() {
+      this.fileSystemProperties = GSON.toJson(ImmutableMap.<String, String>of());
+      this.fileRegex = ".*";
+    }
```
gsps1, line 88: can you extract these properties out of FileBatchConfig, so you don't have to pass null for those two params as well in constructor.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -88,7 +88,7 @@
     "2015-06-16-15 (June 16th 2015, 3pm), it will read in files that contain '2015-06-16-14' in the filename. " +
     "If the field 'timeTable' is present, then it will read in files that have not yet been read. Defaults to '.*', " +
     "which indicates that no files will be filtered.";
-  private static final String FILESYSTEM_PROPERTIES_DESCRIPTION = "A JSON string representing a map of properties " +
+  protected static final String FILESYSTEM_PROPERTIES_DESCRIPTION = "A JSON string representing a map of properties " +
```
gsps1, line 91: why this change?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLFTPTestRun.java
```diff
@@ -0,0 +1,124 @@
...
+ * FTP Source Test.
+ */
+public class ETLFTPTestRun extends ETLBatchTestBase {
+
+  private static final String USER = "ftp";
+  private static final String PWD = "abcd";
+  private static final String TEST_STRING = "Hello World";
+
+  @ClassRule
+  public static final TemporaryFolder TMP_FOLDER = new TemporaryFolder();
```
gsps1: can you reuse the Temp folder in test base ?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLFTPTestRun.java
```diff
@@ -0,0 +1,124 @@
...
+    ftpServer.setServerControlPort(0);
+
+    FileSystem fileSystem = new UnixFakeFileSystem();
+    fileSystem.add(new FileEntry(file.getAbsolutePath(), TEST_STRING));
+    ftpServer.setFileSystem(fileSystem);
+
+    ftpServer.addUserAccount(new UserAccount(USER, PWD, folder.getAbsolutePath()));
+    ftpServer.start();
+
+    Assert.assertTrue(ftpServer.isStarted());
```
gsps1: can this be false? if so, should we retry this check for it to be started

gsps1:
what would happen in failure scenario's ?  path is incorrect, access errors, etc? would we return appropriate error messages?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLFTPTestRun.java
gokulavasan: Yes I can. Will remove.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLFTPTestRun.java
gokulavasan: Yes, I will add a try mechanism to check

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
gokulavasan, line 91: So that we can access the property description in FTPBatchSourceConfig.

gokulavasan:
@gsps1 Regarding the question about failure scenarios, we delegate to FileBatchSource. We don't perform any additional checks.

gokulavasan:
Reopened here https://github.com/caskdata/hydrator-plugins/pull/182

PR #176, opened 2016-03-23
CDAP-5146 add a group by aggregator

albertshau:
The plugin can group by fields in a record, then calculate
count, sum, min, max, avg, first, and last on the grouped values.
Aggregate functions should eventually be turned into plugins.

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -61,6 +61,10 @@
     </dependency>
     <dependency>
       <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-etl-data-pipeline</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
       <artifactId>cdap-etl-realtime</artifactId>
     </dependency>
     <dependency>
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.StageConfigurer;
+import co.cask.cdap.etl.api.batch.BatchAggregator;
+import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.hydrator.plugin.batch.aggregator.function.AggregateFunction;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Batch group by aggregator.
+ */
+@Plugin(type = BatchAggregator.PLUGIN_TYPE)
+@Name("GroupByAggregate")
+@Description("Groups records by a set of fields, then performs aggregates on record fields.")
+public class GroupByAggregator extends BatchAggregator<StructuredRecord, StructuredRecord, StructuredRecord> {
+  private final GroupByConfig conf;
+  private Set<String> groupByFields;
+  private List<GroupByConfig.FunctionInfo> functionInfos;
+  private Schema outputSchema;
+  private Map<String, AggregateFunction> aggregateFunctions;
+
+  public GroupByAggregator(GroupByConfig conf) {
+    this.conf = conf;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Set<String> groupByFields = conf.getGroupByFields();
+    List<GroupByConfig.FunctionInfo> aggregates = conf.getAggregates();
+
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    // if null, the input schema is unknown, or its multiple schemas.
+    if (inputSchema == null) {
+      stageConfigurer.setOutputSchema(null);
+      return;
+    }
+
+    // otherwise, we have a constant input schema. Check that all the group by fields exist in the input schema,
+    List<Schema.Field> outputFields = new ArrayList<>(groupByFields.size() + aggregates.size());
+    for (String groupByField : groupByFields) {
+      Schema.Field field = inputSchema.getField(groupByField);
+      if (field == null) {
+        throw new IllegalArgumentException(String.format(
+          "Cannot group by field '%s' because it does not exist in input schema %s.",
+          groupByField, inputSchema));
+      }
+      outputFields.add(field);
+    }
+    // check that all fields needed by aggregate functions exist in the input schema.
+    for (GroupByConfig.FunctionInfo functionInfo : aggregates) {
+      // special case count(*) because we don't have to check that the input field exists
+      if (functionInfo.getField().equals("*")) {
+        AggregateFunction aggregateFunction = functionInfo.getAggregateFunction(null);
+        outputFields.add(Schema.Field.of(functionInfo.getName(), aggregateFunction.getOutputSchema()));
+        continue;
+      }
+
+      Schema.Field inputField = inputSchema.getField(functionInfo.getField());
+      if (inputField == null) {
+        throw new IllegalArgumentException(String.format(
+          "Invalid aggregate %s(%s): Field '%s' does not exist in input schema %s.",
+          functionInfo.getFunction(), functionInfo.getField(), functionInfo.getField(), inputSchema));
+      }
+      AggregateFunction aggregateFunction = functionInfo.getAggregateFunction(inputField.getSchema());
+      outputFields.add(Schema.Field.of(functionInfo.getName(), aggregateFunction.getOutputSchema()));
+    }
+    // propagate the schema, which is group by fields + aggregate fields
+    Schema outputSchema = Schema.recordOf(inputSchema.getRecordName() + ".agg", outputFields);
+    stageConfigurer.setOutputSchema(outputSchema);
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    groupByFields = conf.getGroupByFields();
+    functionInfos = conf.getAggregates();
+  }
+
+  @Override
+  public void groupBy(StructuredRecord record, Emitter<StructuredRecord> emitter) throws Exception {
+    // app should provide some way to make some data calculated in configurePipeline available here.
+    // then we wouldn't have to calculate schema here
+    StructuredRecord.Builder builder = StructuredRecord.builder(getGroupKeySchema(record.getSchema()));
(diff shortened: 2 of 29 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Sum.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  protected void updateInt(int val) {
+    intSum += val;
+  }
+
+  @Override
+  protected void updateLong(long val) {
+    longSum += val;
```
gsps1, line 61: this could cause overflow right?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByConfig.java
```diff
@@ -0,0 +1,224 @@
...
+      throw new IllegalArgumentException("The 'groupByFields' property must be set.");
+    }
+    return fields;
+  }
+
+  List<FunctionInfo> getAggregates() {
+    List<FunctionInfo> functionInfos = new ArrayList<>();
+    Set<String> aggregateNames = new HashSet<>();
+    for (String aggregate : Splitter.on(',').trimResults().split(aggregates)) {
+      int colonIdx = aggregate.indexOf(':');
```
gsps1, line 88: using regex would make this method much shorter.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByConfig.java
```diff
@@ -0,0 +1,224 @@
...
+  private final String groupByFields;
+
+  @Description("Aggregates to compute on grouped records. " +
+    "Supported aggregate functions are count, count(*), sum, avg, min, max, first, last. " +
+    "A function must specify the field it should be applied on, as well as the name it should be called. " +
+    "Aggregates are specified using syntax: \"name:function(field)[, other aggregates]\"." +
+    "For example, 'avgPrice:avg(price),cheapest:min(price)' will calculate two aggregates. " +
+    "The first will create a field called 'avgPrice' that is the average of all 'price' fields in the group. " +
+    "The second will create a field called 'cheapest' that contains the minimum 'price' field in the group")
+  private final String aggregates;
```
gsps1, line 53: nit : move this before groupByFields, as you refer to aggregates in that. It was easier for me to understand when i read this first before the `groupByFields` description.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Avg.java
```diff
@@ -0,0 +1,72 @@
...
+  private double count;
+
+  public Avg(String fieldName, Schema fieldSchema) {
+    this.fieldName = fieldName;
+    boolean isNullable = fieldSchema.isNullable();
+    Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+    if (!(fieldType == Schema.Type.INT || fieldType == Schema.Type.LONG ||
+      fieldType == Schema.Type.FLOAT || fieldType == Schema.Type.DOUBLE)) {
+      throw new IllegalArgumentException(String.format(
+        "Cannot compute avg on field %s because its type %s is not numeric", fieldName, fieldSchema));
```
gsps1: use `fieldType` rather than `fieldSchema` in the message.

core-plugins/pom.xml
```diff
@@ -61,6 +61,10 @@
     </dependency>
     <dependency>
       <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-etl-data-pipeline</artifactId>
```
nitinmotgi, line 64: it's not etl-data-pipeline right. We should call it cdap-batch-data-pipeline.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Avg.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public void update(StructuredRecord record) {
+    Object val = record.get(fieldName);
+    if (val == null) {
+      return;
+    }
+    sum += ((Number) val).doubleValue();
+    count++;
```
gsps1, line 55: can  we use a formula without count, something like this ?
https://commons.apache.org/proper/commons-math/apidocs/org/apache/commons/math3/stat/descriptive/moment/Mean.html

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Max.java
```diff
@@ -0,0 +1,95 @@
...
+  private Float maxFloat;
+  private Double maxDouble;
+
+  public Max(String fieldName, @Nullable Schema fieldSchema) {
+    super(fieldName, fieldSchema);
+  }
+
+  @Override
+  protected void startInt() {
+    maxInt = null;
```
gsps1, line 38: better to initialize this to `Integer.MinValue` so you don't have to check for null when you update.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Min.java
```diff
@@ -0,0 +1,95 @@
...
+  private Float minFloat;
+  private Double minDouble;
+
+  public Min(String fieldName, @Nullable Schema fieldSchema) {
+    super(fieldName, fieldSchema);
+  }
+
+  @Override
+  protected void startInt() {
+    minInt = null;
```
gsps1: `minInt = Integer.MaxValue`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Max.java
gsps1, line 38: is returning null when finish is called without any updates is important?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/NumberFunction.java
```diff
@@ -0,0 +1,224 @@
...
+    return typedDelegate.getOutputSchema();
+  }
+
+  protected abstract void startInt();
+
+  protected abstract void startLong();
+
+  protected abstract void startFloat();
+
+  protected abstract void startDouble();
```
gsps1, line 203: would it be better to have `start(NumericType type)` where NumericType can be Enum and we can set/update value based on that ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -0,0 +1,176 @@
...
+      }
+
+      Schema.Field inputField = inputSchema.getField(functionInfo.getField());
+      if (inputField == null) {
+        throw new IllegalArgumentException(String.format(
+          "Invalid aggregate %s(%s): Field '%s' does not exist in input schema %s.",
+          functionInfo.getFunction(), functionInfo.getField(), functionInfo.getField(), inputSchema));
+      }
+      AggregateFunction aggregateFunction = functionInfo.getAggregateFunction(inputField.getSchema());
+      outputFields.add(Schema.Field.of(functionInfo.getName(), aggregateFunction.getOutputSchema()));
```
gsps1: can you split these into separate methods, configurePipeline is pretty long.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Sum.java
albertshau, line 61: yes, but there is no way to get around that for a sum...

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByConfig.java
albertshau, line 88: i would not be able to give descriptive error messages though

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Avg.java
albertshau, line 55: you always need a count. Are you worried about overflow?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByConfig.java
gsps1, line 88: based on the regex groups, you can still provide messages right?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Avg.java
gsps1, line 55: yeah

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByConfig.java
albertshau, line 88: if it doesn't match the regex, you don't really know why, you just know it didn't match.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Max.java
albertshau, line 38: null seems to be the only sensible value to return in that case right?  min value would be wrong. It's also not just if update is never called.  Its if all the values in the group have a null value for that field.  For example, that int field could be a nullable field

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/NumberFunction.java
albertshau, line 203: would it be?  I think it would be more difficult, as the whole point of this class is to make it so child classes can just work with the right type.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -76,8 +80,12 @@
   @ClassRule
   public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", true);
 
-  protected static final Id.Artifact APP_ARTIFACT_ID = Id.Artifact.from(Id.Namespace.DEFAULT, "etlbatch", "3.2.0");
+  protected static final Id.Artifact ETLBATCH_ARTIFACT_ID =
+    Id.Artifact.from(Id.Namespace.DEFAULT, "etlbatch", "3.2.0");
+  protected static final Id.Artifact DATAPIPELINE_ARTIFACT_ID =
+    Id.Artifact.from(Id.Namespace.DEFAULT, "data-pipeline", "3.2.0");
   protected static final ArtifactSummary ETLBATCH_ARTIFACT = new ArtifactSummary("etlbatch", "3.2.0");
+  protected static final ArtifactSummary DATAPIPELINE_ARTIFACT = new ArtifactSummary("data-pipeline", "3.2.0");
```
anwar6953, line 88: nit - unused

anwar6953, line 88: nevermind; its used in subclass.

core-plugins/pom.xml
albertshau, line 64: thats what its called now.  Can change it in a separate cdap pr, then a pr here

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -0,0 +1,182 @@
...
+      aggregateFunctions.put(functionInfo.getName(), aggregateFunction);
+    }
+    outputSchema = Schema.recordOf(valueSchema.getRecordName() + ".agg", outputFields);
+  }
+
+  private Schema getGroupKeySchema(Schema inputSchema) {
+    List<Schema.Field> fields = new ArrayList<>();
+    for (String groupByField : conf.getGroupByFields()) {
+      fields.add(inputSchema.getField(groupByField));
```
gsps1: `inputSchema.getField(groupByField)` could be null right? is it okay to add that to the list?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Sum.java
gsps1, line 61: we should flag overflow, or at least emit a log.  otherwise there is no way to know if the value we emit is overflown one.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
albertshau: good catch. Will check for it and error

gsps1:
LGTM

PR #177, opened 2016-03-25
Cube widget improvements

gsps1:
Added improvements to cube widget, so the user doesn't have to provide JSON of properties, rather can provide the aggregation groups using key,value pairs. 

similarly can use key-value for adding widgets, with fixed measure type. Also made this measurement field as required. 

Added corresponding code changes, docs changes and tests in CubeSink(batch, realtime)

Having JSON and optional fields were causing issues with frequent misconfiguration.

After changes, the widget would look like below,
screen shot 2016-03-25 at 11 58 09 am

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCubeSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCubeSink.java
@@ -24,17 +24,26 @@ import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.cube.Cube;
 import co.cask.cdap.api.dataset.lib.cube.CubeFact;
 import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.hydrator.plugin.common.CubeSinkConfig;
+import co.cask.hydrator.plugin.common.CubeUtils;
 import co.cask.hydrator.plugin.common.Properties;
 import co.cask.hydrator.plugin.common.StructuredRecordToCubeFact;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Type;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -55,9 +64,6 @@ import java.util.Map;
 @Name("Cube")
 @Description("CDAP Cube Dataset Batch Sink")
 public class BatchCubeSink extends BatchWritableSink<StructuredRecord, byte[], CubeFact> {
-  private static final Gson GSON = new Gson();
-  private static final Type STRING_MAP_TYPE = new TypeToken<Map<String, String>>() { }.getType();
-
   private final CubeSinkConfig config;
 
   public BatchCubeSink(CubeSinkConfig config) {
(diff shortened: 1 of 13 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
```diff
@@ -25,7 +26,6 @@
  */
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
-  BatchCubeSinkTestRun.class,
```
gokulavasan, line 29: So the test is not run as part of this suite?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility class for usage by batch and real-time cube sinks.
+ */
+public class CubeUtils {
```
gokulavasan: final class?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
gsps1, line 29: yeah, i wanted it to be on the same package as the source class, so can call getProperties method for testing. so moved it.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeSinkConfig.java
```diff
@@ -46,6 +46,11 @@
     "http://docs.cask.co/cdap/current/en/developers-manual/building-blocks/datasets/cube.html#cube-configuration " +
     "for more information.")
   @Nullable
+  String aggregations;
```
gokulavasan, line 50: aggregations will still be Nullable?

gsps1, line 50: yeah, we can use an already existing cube. in that case we wouldn't need to provide aggregations.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCubeSink.java
```diff
@@ -69,24 +75,44 @@ public BatchCubeSink(CubeSinkConfig config) {
...
+                                                            return "dataset.cube.aggregation." + input + ".dimensions";
+                                                          }
+                                                        }));
+    }
+
+    // add measurements
+    if (!Strings.isNullOrEmpty(config.getMeasurements())) {
+      properties.remove(Properties.Cube.MEASUREMENTS);
+      properties.putAll(CubeUtils.parseAndGetProperties(Properties.Cube.MEASUREMENTS,
+                                                        config.getMeasurements(), ",", ":",
```
gokulavasan: Wondering if we can use ';' as the delimiter in both cases? Bigger change I guess?

core-plugins/widgets/Cube-realtimesink.json
```diff
@@ -35,6 +46,17 @@
       "label": "CubeFact Mapping",
       "properties": [
         {
+          "widget-type": "keyvalue-dropdown",
+          "label": "Measurements",
+          "name": "cubeFact.measurements",
+          "description": "Fields that are used as measurements and the measurment type.",
```
gokulavasan: Do we need more info about this? For example, say I add couple of aggregations each containing one field each. Now should I also provide measurements for each of those fields? What if I don't? What does it mean if I have measurements for some field and it is not present in aggregation?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
gokulavasan, line 29: Are you sure the test runs? I think we don't run tests whose classes end with 'Run' since they are assumed to be part of a TestSuite.

gokulavasan:
@gsps1 Thanks for making these changes! And for the screenshot. The user still needs to type out the name of the field right - prone to error? That, I guess, is common for all plugins? 

Few comments. Overall looks good.

gsps1:
yeah, i checked with @ajainarayanan, we don't have auto-complete with field values as suggestions, if we had that, that would make it easier for the users.

core-plugins/widgets/Cube-realtimesink.json
gsps1: measurement field and aggregation field doesn't have to mach. aggregation group fields become context, while measurement is like metric name.  
Let's say there are two aggregation groups
`[ {"agg1" : "remote_host"}, {"agg2" : "response_code"}]`and measurements [`{"response_count" : "COUNTER"}, {"total_count" : "COUNTER"}]`

if the input record has values for these fields `<remote_host, response_code, response_count>` we would emit 2 measures, one for each aggregation group with measurement name "response_count".

if input record has only these field values `<remote_host, response_code>` we wouldn't any measures, as none of the measurement fields are present in this record.

should we add this to reference doc?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
gsps1, line 29: i ran the package tests in IDE, but have removed the 'Run' suffix to keep the standard.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCubeSink.java
gsps1: yeah, done

gsps1:
thanks for the review @gokulavasan , have addressed the comments.  please take a look again, build running here : http://builds.cask.co/browse/HYP-DBT3-1

gokulavasan:
I think documentation can still be improved. But these changes, LGTM :+1:

gsps1:
build is green, squashed commits, merging.

PR #178, opened 2016-03-26
allow each module to specify their own parents

albertshau:
Since the data-pipeline app supports some plugin types that
the etl-batch app does not, we need a way for certain modules to
limit which apps they extend.


anwar6953:
lgtm

PR #181, opened 2016-03-27
CDAP-5322 Adding schema dataset property to filesets created by hydrator

poornachandra:
JIRA - https://issues.cask.co/browse/CDAP-5322
Build - http://builds.cask.co/browse/HYP-DBT4-1


albertshau:
lgtm

PR #189, opened 2016-03-29
CDAP-5042 add a setting for the number of mappers for db source

albertshau:
This will let users configure the parallelism of their query,
as well as control the number of output files.


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -210,20 +232,29 @@ private String getJDBCPluginId() {
...
     @Description("Bounding Query should return the min and max of the " +
-      "values of the 'splitBy' field. For example, 'SELECT MIN(id),MAX(id) FROM table'")
+      "values of the 'splitBy' field. For example, 'SELECT MIN(id),MAX(id) FROM table'. " +
+      "This is required unless numSplits is set to one.")
     String boundingQuery;
 
+    @Nullable
     @Name(SPLIT_BY)
-    @Description("Field Name which will be used to generate splits.")
+    @Description("Field Name which will be used to generate splits. This is not required if numSplits is set to one.")
```
gokulavasan: nit: May be we should use the same language as in other descriptions. 'This is required unless numSplits is set to one'.

database-plugins/widgets/Database-batchsource.json
```diff
@@ -75,6 +75,11 @@
           "name": "splitBy"
         },
         {
+          "widget-type": "textbox",
```
gokulavasan, line 78: Should this be a text box? Do we have any number type? @ajainarayanan

gokulavasan:
Couple of minor comments. Rest LGTM :+1:

PR #192, opened 2016-03-29
Revert "compile python script during configurePipeline stage"

anwar6953:
This reverts commit 7d4eb4d9cb445133b3277227a3d9abf896ce63b1.

This reverts https://github.com/caskdata/hydrator-plugins/pull/144
because calling init() leaks memory in the current implementation of PythonEvaluator,
so this avoids calling it in master.
Otherwise, you can only deploy a pipeline with this plugin a limited amount of times because master goes out of permgen space.
This revert will work around that issue.

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
@@ -132,31 +132,6 @@ public class PythonEvaluatorTest {
     Assert.assertEquals(expectedListField, output.get("arrayField"));
   }
 
-  @Test(expected = Exception.class)
-  public void testScriptCompilationValidation() throws Exception {
-    Schema outputSchema = Schema.recordOf(
-      "smallerSchema",
-      Schema.Field.of("intField", Schema.of(Schema.Type.INT)),
-      Schema.Field.of("longField", Schema.of(Schema.Type.LONG)));
-
-    PythonEvaluator.Config config = new PythonEvaluator.Config(
-      "def transform(x, emitter, context):\n" +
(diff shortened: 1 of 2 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -185,7 +160,7 @@ public void testSchemaValidation() throws Exception {
     // check if schema is null, input schema is used.
     config = new PythonEvaluator.Config(
       "def transform(x, emitter, context):\n" +
-        "  y = {}\n" +
+        " y = {}\n" +
```
gokulavasan, line 163: Space?

gokulavasan:
Any docs that needs to be fixed? Otherwise, LGTM :+1:

poornachandra:
JIRA?

anwar6953:
Two relevant JIRAs:
To fix memory leak - https://issues.cask.co/browse/CDAP-4222
To check script validity during deploy - https://issues.cask.co/browse/CDAP-4130

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
anwar6953, line 163: Yea, I just did a git revert of the referenced PR/commit.
We can revert this revert once the linked JIRA for memory leak is resolved.

poornachandra:
When will this revert get reverted?

anwar6953:
@poornachandra  Once CDAP-4222 is resolved.

@gokulavasan no docs need to be updated for this. We already list the referenced issue as a known issue in the docs for this plugin.

PR #193, opened 2016-03-30
Fix Travis test failure

gokulavasan:


anwar6953:
LGTM

gokulavasan:
@anwar6953 Please review again.

fileset-plugins-parquet/src/test/java/co/cask/hydrator/plugin/batch/file/ETLSnapshotTestRun.java
```diff
@@ -52,7 +54,8 @@
     Schema.Field.of("id", Schema.of(Schema.Type.STRING)),
     Schema.Field.of("price", Schema.of(Schema.Type.INT)));
 
-  //@Test
+  @Ignore
```
anwar6953, line 59: can you file a JIRA for this.
Other than that, LGTM

gokulavasan, line 59: @anwar6953 Should the JIRA say that the build failed because the Test annotation was not present and that this PR fixes it?

anwar6953, line 59: Just that this test case is ignored, and that it should it be fixed.

gokulavasan, line 59: @albertshau Any idea why this test was ignored?

gokulavasan:
@anwar6953 Created https://issues.cask.co/browse/CDAP-5488. Merging.

PR #200, opened 2016-04-05
Use _exportcontents instead of Export-Package

chtyim:
- _exportcontents won’t inspect the jar, hence is faster and works
  with all kinds of JAR.


albertshau:
lgtm

PR #205, opened 2016-04-08
fixing etl version range

russorat:


bdmogal:
Looks like there are merge conflicts. THe change LGTM though

bdmogal:
actually, I think we should keep it exclusive, and bump it to 3.6.0, so it works with 3.5.x as well.

russorat:
this PR is bunk. ignore. it was not merged and never will be.

PR #206, opened 2016-04-08
fixing parents

russorat:


bdmogal:
LGTM

gsps1:
noticed some pom files still having app.parents in them, 
https://github.com/caskdata/hydrator-plugins/blob/develop/hbase-plugins/pom.xml#L28 
should they be removed?

PR #207, opened 2016-04-08
fixing more parents

russorat:
missed a comment from the previous pr. this one has the extra fixes.

https://github.com/caskdata/hydrator-plugins/pull/206


gsps1:
just add a reference about this PR in the old PR, changes LGTM

PR #208, opened 2016-04-08
CDAP-5039 add macro support and change db plugins to use macros

albertshau:
Adding a MacroConfig class that has support for replacing certain
macros in string fields with some value based on the macro
function. Added runtime based functions that allow users to
calculate some date format using the logical start time of the run.

Currently only the database plugins make use of macros. There are
some limitations in the app that prevent transforms from using them,
namely the fact that the logical start time is not available to
transforms.


hydrator-common/src/main/java/co/cask/hydrator/common/macro/RuntimeMacro.java
```diff
@@ -0,0 +1,64 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import javax.annotation.Nullable;
+
+/**
+ * Runtime macros use the logical start time of a run to perform substitution.
+ * Runtime macros follow the syntax ${runtime(arguments)}. Arguments are expected to be either:
+ *
+ * empty string
+ * format
```
gokulavasan, line 30: What format is being referred to here? Link or example might be useful.

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSourceTestRun.java
```diff
@@ -49,6 +49,40 @@
...
+
+    Plugin sinkConfig = new Plugin("Table", ImmutableMap.of(
+      "name", "macroOutputTable",
+      Properties.Table.PROPERTY_SCHEMA, schema.toString(),
+      Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "ID"));
+
+    ApplicationManager appManager = deployETL(sourceConfig, sinkConfig);
+
+    Map<String, String> arguments = new HashMap<>();
+    arguments.put("logical.start.time", String.valueOf(CURRENT_TS));
```
gokulavasan, line 76: Why do we need to pass logical start time here? Isn't it set by the platform?

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
```diff
@@ -0,0 +1,158 @@
...
+
+  // parse an expression into an evaluation tree
+  private MathNode parseExpression(String expression) {
+    expression = expression.trim();
+    // we only support + and - right now so just evaluate from left to right
+    int idx = Math.max(expression.lastIndexOf('+'), expression.lastIndexOf('-'));
+
+    // if no operator, this should be a duration string
+    if (idx < 0) {
+      return "runtime".equals(expression) ? new ValueNode(runtime) : new ValueNode(parseDuration(expression));
```
gokulavasan, line 89: Should "runtime" be a static constant defined in a common place?

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
```diff
@@ -0,0 +1,158 @@
...
+    char operator = expression.charAt(idx);
+    if (idx == expression.length()) {
+      throw new IllegalArgumentException(
+        String.format("Invalid expression '%s'. Cannot end with %s.", expression, operator));
+    }
+
+    // if this in an expression like '-5d'
+    if (idx == 0) {
+      return new ValueNode(parseDuration(expression));
+      /*long dur = parseDuration(expression);
```
gokulavasan: Did you intend to include this multiline code comment?

gokulavasan: Hmm, looks like parseDuration doesn't handle '-' ?

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSourceTestRun.java
albertshau, line 76: I need to set it to what I want it to be in order for the test to work correctly, otherwise there are issues if the run happens to start on the time boundary.

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
albertshau: forgot to remove.  '-' is handled, there is a test case for it.

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
albertshau, line 89: its internal to this class and only used here.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBConfig.java
```diff
@@ -19,6 +19,7 @@
 import co.cask.cdap.api.annotation.Description;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.hydrator.common.macro.MacroConfig;
```
gokulavasan: unused?

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
gokulavasan, line 89: I see its also used in Macros class. Wondering if they both refer to the same thing.

hydrator-common/src/main/java/co/cask/hydrator/common/macro/MacroContext.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Context for macro substitution.
+ */
+public interface MacroContext {
+
+  long getLogicalStartTime();
+
+  @Nullable
+  TimeZone getTimeZone();
```
gokulavasan, line 37: We say Macros are generic but here it seems to be specific to Date/Time?

hydrator-common/src/test/java/co/cask/hydrator/common/macro/RuntimeMacroTest.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class RuntimeMacroTest {
+
+  @Test
+  public void testSubstitution() {
+    RuntimeMacro runtimeMacro = new RuntimeMacro();
+    MacroContext macroContext = new DefaultMacroContext(0, TimeZone.getTimeZone("UTC"));
+
+    Assert.assertEquals("1970-01-01", runtimeMacro.getValue("yyyy-MM-dd", macroContext));
+    Assert.assertEquals("1970-01-02", runtimeMacro.getValue("yyyy-MM-dd,-1d", macroContext));
```
gokulavasan, line 37: Can we also add a test case for '+1d'

gokulavasan, line 37: What happens if someone uses '0.5d' ?

gokulavasan:
Finished review.

hydrator-common/src/main/java/co/cask/hydrator/common/TimeParser.java
albertshau, line 89: they are different.  Here it is part of an expression, whereas Macros defines a macro type called 'runtime'.

hydrator-common/src/main/java/co/cask/hydrator/common/macro/MacroContext.java
albertshau, line 37: The context is meant to supply any information a macro may need.  It can ignore this completely.

hydrator-common/src/test/java/co/cask/hydrator/common/macro/RuntimeMacroTest.java
albertshau, line 37: partial units are not supported, which is why there is 'd', 'h', 'm', 's'

gokulavasan:
LGTM :+1:

hydrator-common/src/test/java/co/cask/hydrator/common/macro/RuntimeMacroTest.java
gokulavasan, line 37: ok thanks for confirming

PR #209, opened 2016-04-11
removing deprecated methods and duplicate classes in unit tests

albertshau:


albertshau:
its a lot of lines, but its just removing copied test classes in favor of mocks in hydrator-test module, and replacing deprecated classes with their new versions.

cassandra-plugins/src/test/java/co/cask/hydrator/plugin/test/ETLCassandraTest.java
```diff
@@ -360,7 +351,7 @@ private void testCassandraRealtimeSink() throws Exception {
       public Boolean call() throws Exception {
         CqlResult result = client.execute_cql3_query(ByteBufferUtil.bytes(cqlQuery),
                                                      Compression.NONE, ConsistencyLevel.ALL);
-        return result.rows.size() >= 2;
+        return result.rows.size() > 0;
```
anwar6953, line 354: we write different amount of data now?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
```diff
@@ -62,30 +64,36 @@ public void testMultiSnapshotOutput() throws Exception {
...
-        .put(Properties.Table.PROPERTY_SCHEMA, SCHEMA.toString())
-        .put(Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id")
-        .build()));
+      new ETLPlugin("Table", BatchSource.PLUGIN_TYPE,
+                    ImmutableMap.<String, String>builder()
+                      .put(Properties.Table.NAME, tableName)
+                      .put(Properties.Table.PROPERTY_SCHEMA, SCHEMA.toString())
+                      .put(Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id")
+                      .build(),
+                    null));
```
anwar6953, line 73: just a comment for a potential future change in CDAP - seems like it'd be useful to have a constructor without ArtifactSelector.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
```diff
@@ -141,27 +149,31 @@ public void testMultiSnapshotOutput() throws Exception {
...
     // run another pipeline that reads from avro dataset
     ETLStage source = new ETLStage(
       "source",
-      new Plugin(sourcePlugin, ImmutableMap.<String, String>builder()
-        .put(Properties.Table.NAME, sourceName)
-        .put(Properties.Table.PROPERTY_SCHEMA, SCHEMA.toString())
-        .put(Properties.Table.PROPERTY_SCHEMA_ROW_FIELD, "id")
-        .build()));
+      new ETLPlugin(sourceETLPlugin, BatchSource.PLUGIN_TYPE,
+                    ImmutableMap.<String, String>builder()
```
anwar6953, line 157: nit - seems simpler:

``` java
ImmutableMap.of(key1, value1,
                key2, value2,
                key3, value3);
```

cassandra-plugins/src/test/java/co/cask/hydrator/plugin/test/ETLCassandraTest.java
albertshau: not really sure why it was 2 before, since the test only checks one record

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
anwar6953, line 157: not a big deal - seems like we use the builder pattern everywhere else in these test cases.

cassandra-plugins/src/test/java/co/cask/hydrator/plugin/test/ETLCassandraTest.java
albertshau, line 354: the only reason it waited for 2 before was to check that the timestamp columns of both records were not equal (data generator source).  We shouldn't be doing that anyway, as that assumes both records are written with different timestamps.  Since its been moved to mock source, we can just check the timestamp column directly.

anwar6953, line 354: 👍

anwar6953:
LGTM

PR #214, opened 2016-04-12
CDAP-4955 move email action from cdap into this repo.

albertshau:


core-plugins/docs/Email-postaction.md
```diff
@@ -0,0 +1,57 @@
...
+
+**host:** The SMTP host to use. Defaults to localhost.
+
+**port:** The SMTP port to use. Defaults to 25.
+
+**includeWorkflowToken:** Whether to include the contents of the workflow token in the email message. Defaults to false.
+
+Example
+-------
+This example sends an email from 'team-ops@company.com' to 'team-alerts@company.com' whenever a run fails
```
gsps1: there's nothing in the config to make it execute only on run failure. change the description?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    if (Strings.isNullOrEmpty(config.username) ^ Strings.isNullOrEmpty(config.password)) {
+      throw new IllegalArgumentException(
+        String.format("You must either set both username and password " +
+                        "or neither username nor password. " +
+                        "Currently, they are username: \'%s\', and password: \'%s\'",
+                      config.username, config.password));
```
gsps1: i don't think its a good idea to include password in exception message

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    try {
+      new InternetAddress(config.senderEmailAddress);
```
gsps1: will be better to do 
`if (InternetAddress.parse(config.senderEmailAddress, true).length != 1) ) { throw exception(msg)}`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -0,0 +1,191 @@
...
+    } catch (AddressException e) {
+      throw new IllegalArgumentException(
+        String.format("%s is an invalid sender email address. Reason: %s",
+                      config.senderEmailAddress, e.getMessage()));
+    }
+
+    try {
+      new InternetAddress(config.recipientEmailAddress);
```
gsps1: same here

core-plugins/docs/Email-postaction.md
```diff
@@ -0,0 +1,57 @@
...
+the run failed for any reason.
+
+
+Properties
+----------
+**recipientEmailAddress:** The address to send the email to.
+
+**senderEmailAddress:** The address to send the email from.
+
+**message:** The message of the email.
```
gsps1, line 27: should we include message type in the config ?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLEmailActionTestRun.java
```diff
@@ -0,0 +1,98 @@
...
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "actionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+
+    WorkflowManager manager = appManager.getWorkflowManager("ETLWorkflow");
+    manager.start();
+    manager.waitForFinish(5, TimeUnit.MINUTES);
+
+    server.stop();
+
+    Assert.assertEquals(1, server.getReceivedEmailSize());
```
gsps1, line 91: there work after the server stop as well ?

gsps1:
few minor comments, couple of questions.

core-plugins/docs/Email-postaction.md
albertshau: weird thought I added that to the example. Going to add the settings for this.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLEmailActionTestRun.java
albertshau, line 91: guess so since the test passes. just copied this test

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
albertshau: hm though it may be a good idea to support multiple recipients.  Think I will change it so that is supported

core-plugins/docs/Email-postaction.md
albertshau, line 27: if we allow other types, then the message can't be a string and things get much more complex.  For example, what would we do for Partial type?  I think its best to keep it simple for now until we have a use case for more complex things.

core-plugins/docs/Email-postaction.md
```diff
@@ -0,0 +1,61 @@
...
+        "type": "postaction",
+        "properties": {
+            "recipientEmailAddress": "team-alerts@company.com",
+            "senderEmailAddress": "team-ops@company.com",
+            "subject": "Pipeline Failure ${runtime(yyyy-MM-dd)}",
+            "message": "The pipeline run failed.",
+            "includeWorkflowToken": "true",
+            "host": "smtp-server.com",
+            "port": "25",
+            "runOnFailure": "true",
```
gsps1: the field explanation above uses `runCondition`, should we have a single variable `runCondition` with 3 possible values, `onSuccess (or) onFailure (or) always`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -0,0 +1,206 @@
...
+          throw new IllegalArgumentException(String.format(
+            "%s is an invalid sender email address. Only one sender is supported.", sender));
+        }
+      } catch (AddressException e) {
+        throw new IllegalArgumentException(
+          String.format("%s is an invalid sender email address. Reason: %s", sender, e.getMessage()));
+      }
+
+      try {
+        InternetAddress.parse(sender);
```
gsps1: this should be `InternetAddress.parse(recipients)`

core-plugins/docs/Email-postaction.md
gsps1: based on widget config and code, you seem to use `runOnCondition`, so this config in properties is incorrect i guess ?

albertshau: agh oops. Had this originally then decided to change it to an enum and forgot to update this.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -0,0 +1,206 @@
...
+    @Nullable
+    @Description("The email protocol to use. smtp, smtps, and tls are supported. Defaults to SMTP.")
+    private String protocol;
+
+    @Nullable
+    @Description("The SMTP host to use. Defaults to localhost.")
+    private String host;
+
+    @Nullable
+    @Description("The SMTP port to use. Defaults to 25.")
```
gsps1: should we use port `587` as default ?

albertshau: seems like a better default, will change

core-plugins/widgets/Email-postaction.json
```diff
@@ -0,0 +1,83 @@
...
+    {
+      "label": "Distinct",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Sender",
+          "name": "sender"
+        },
+        {
+          "widget-type": "textbox",
```
gsps1: can you use `csv` widget type here, that will avoid user error and have better experience.
reference : https://github.com/caskdata/hydrator-plugins/blob/develop/core-plugins/widgets/Cube-batchsink.json#L20

gsps1:
just one minor comment, fix, squash and merge!

core-plugins/widgets/Email-postaction.json
albertshau: ah good point

PR #215, opened 2016-04-13
Adding a DatabaseQuery action that can run after a pipeline run.

albertshau:
This allows people to do things like run a cleanup query at the
end of a pipeline run.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TimePartitionedFileSetSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TimePartitionedFileSetSource.java
@@ -99,12 +99,7 @@ public abstract class TimePartitionedFileSetSource<KEY, VALUE> extends BatchSour
 
   @Override
   public final void prepareRun(BatchSourceContext context) {
-    Map<String, String> runtimeArgs = context.getRuntimeArguments();
     long runtime = context.getLogicalStartTime();
-    if (runtimeArgs.containsKey("runtime")) {
-      runtime = Long.parseLong(runtimeArgs.get("runtime"));
-    }
-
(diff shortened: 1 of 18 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TimePartitionedFileSetSource.java
```diff
@@ -99,12 +99,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
 
   @Override
   public final void prepareRun(BatchSourceContext context) {
-    Map<String, String> runtimeArgs = context.getRuntimeArguments();
     long runtime = context.getLogicalStartTime();
-    if (runtimeArgs.containsKey("runtime")) {
-      runtime = Long.parseLong(runtimeArgs.get("runtime"));
```
gsps1, line 105: why this change ?

gsps1:
lgtm, just one question

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TimePartitionedFileSetSource.java
gsps1, line 105: was this a hack before ? but seems we we used this as a way to process for a particular time, but now that has to be passed by reconfiguring the pipeline?

albertshau, line 105: yeah this was a workaround before, but now CDAP has the ability to set 'logical.start.time' as a runtime argument to override the normal logical start time.

gsps1, line 105: ok, got it.

PR #216, opened 2016-04-13
add an http callback post action

albertshau:
Adding a post action that makes an http call when a pipeline ends.

### DIFF
--- a/core-plugins/docs/HttpCallback-postaction.md
+++ b/core-plugins/docs/HttpCallback-postaction.md
@@ -1,4 +1,4 @@
-# Email Post Action
+# Http Callback Post Action
 
 
 Description
--- a/core-plugins/docs/HttpCallback-postaction.md
+++ b/core-plugins/docs/HttpCallback-postaction.md
@@ -1,0 +1,48 @@
+# Email Post Action
+
+
+Description
+-----------
+Performs an http request at the end of a pipeline run.
+
+
+Use Case
+--------
+This action can be used when you want to perform an http request at the end of a pipeline run.
+For example, you may want to configure a pipeline so that a request is made to an alerts endpoint
+if the pipeline run failed.
+
+
+Properties
+----------
+**runCondition:**" When to run the action. Must be 'completion', 'success', or 'failure'. Defaults to 'completion'.
+If set to 'completion', the action will be executed regardless of whether the pipeline run succeeded or failed.
+If set to 'success', the action will only be executed if the pipeline run succeeded.
+If set to 'failure', the action will only be executed if the pipeline run failed.
+
+**url:** The URL to call.
+
+**method:** The http request method.
+
+**body:** The optional request body.
+
+**connectionTimeoutMillis:** The connect timeout in milliseconds for the http request. Defaults to 0.
+A timeout of 0 is interpreted as an infinite timeout.
+
+**numRetries:** The number of times the request should be retried if the request fails. Defaults to 0.
+
+**requestProperties:** Properties to add to the request. Should be a JSONObject of string key values.
+For example: {\"Accept-Language\", \"en-US,en;q=0.5\"}.
+
+Example
+-------
+This example performs an HTTP POST to http://monitoring.com/alerts whenever a run fails:
+
+    {
+        "name": "HttpCallback",
+        "type": "postaction",
+        "properties": {
+            "url": "http://monitoring.com/alerts",
+            "method": "POST"
+        }
+    }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
@@ -1,0 +1,177 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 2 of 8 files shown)


core-plugins/docs/HttpCallback-postaction.md
```diff
@@ -0,0 +1,48 @@
+# Email Post Action
```
bdmogal: Http callback post action?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
```diff
@@ -0,0 +1,177 @@
...
+      Map<String, List<String>> headers = conf.parseRequestProperties();
+      try {
+        URL url = new URL(conf.url);
+        conn = (HttpURLConnection) url.openConnection();
+        conn.setRequestMethod(conf.method);
+        conn.setConnectTimeout(conf.connectTimeoutMillis);
+        for (Map.Entry<String, List<String>> propertyEntry : headers.entrySet()) {
+          String propertyKey = propertyEntry.getKey();
+          for (String propertyValue : propertyEntry.getValue()) {
+            conn.setRequestProperty(propertyKey, propertyValue);
```
gsps1: do you want to call `addRequestProperty` ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
```diff
@@ -0,0 +1,177 @@
...
+    conf.validate();
+  }
+
+  @SuppressWarnings("ConstantConditions")
+  @Override
+  public void run(BatchActionContext batchActionContext) throws Exception {
+    int retries = 0;
+    Exception exception = null;
+    do {
+      HttpURLConnection conn = null;
```
gsps1, line 69: can you use `HttpRequests` API from cdap-common ? that has a simpler api.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
```diff
@@ -0,0 +1,177 @@
...
+
+    if (exception != null) {
+      throw exception;
+    }
+  }
+
+  /**
+   * Config for the http callback action.
+   */
+  public static final class Conf extends PluginConfig {
```
gsps1: call it `HttpRequestConf` or something more specific.

core-plugins/widgets/HttpCallback-postaction.json
```diff
@@ -0,0 +1,67 @@
...
+          "label": "Connection Timeout (milliseconds)",
+          "name": "connectTimeoutMillis"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Number of Retries",
+          "name": "numRetries"
+        },
+        {
+          "widget-type": "textbox",
```
gsps1: csv type ?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HttpCallbackActionTestRun.java
```diff
@@ -0,0 +1,125 @@
...
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(ETLBATCH_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "actionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+
+    WorkflowManager manager = appManager.getWorkflowManager("ETLWorkflow");
+    manager.start();
+    manager.waitForFinish(5, TimeUnit.MINUTES);
+
+
```
gsps1: nit : remove extra lines

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
albertshau, line 69: I don't think its a good idea to pull in cdap-common just for this.  It would add unneeded dependencies, make the jar bigger, and set a bad example for people who are looking at hydrator-plugins by making it look like you need more than just the cdap api to write a plugin

core-plugins/widgets/HttpCallback-postaction.json
albertshau: its not a csv but a json.  Will change to json widget

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
gsps1, line 69: my bad, its from co.cask.common repo and not cdap-common. it should be light weight. https://github.com/caskdata/common

albertshau, line 69: i see.  I took a quick look, its really only simpler when you know the request details at compile time.  For example, I would need a series of if, else if on the request method, plus not all methods are in that library (only get,put,post,delete are there)

gsps1, line 69: okay

gsps1:
LGTM

PR #221, opened 2016-04-15
CDAP-5602 use jdbc as default plugin type if its not provided

gsps1:
JIRA https://issues.cask.co/browse/CDAP-5602


gsps1:
will open against release.

PR #227, opened 2016-04-18
CDAP-5648 : Dedup Aggregator Plugin

gokulavasan:
JIRA: https://issues.cask.co/browse/CDAP-5648

Note: Will add more tests (failure scenarios) and docs in a followup PR.


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
```diff
@@ -0,0 +1,101 @@
...
+ * Plugin Configuration
+ */
+public class DedupConfig extends AggregatorConfig {
+
+  @Description("Optional comma-separated list of fields to perform the distinct on. If none is given, each record " +
+    "will be taken as is. Otherwise, only fields in this list will be considered.")
+  @Nullable
+  private String uniqueFields;
+
+  @Description("Optional field that can be used to choose ")
```
albertshau: description is incomplete

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Plugin Configuration
+ */
+public class DedupConfig extends AggregatorConfig {
+
+  @Description("Optional comma-separated list of fields to perform the distinct on. If none is given, each record " +
+    "will be taken as is. Otherwise, only fields in this list will be considered.")
```
albertshau: I wrote this description, but its not very descriptive.  Lets change "each record will be taken as is" to "all record fields will be used."

albertshau: This description and property name are not enough for me to understand what this plugin does.  Please update this instead of copy-pasting from distinct.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
albertshau: seems like the value here will be something like "name:max(ts)"? If so, this should be changed, the name is not needed at all.  The description should include the syntax we are expecting here.

Also, the property name is misleading.  Based on the name I thought it would just be a field name, but this is more like a record selector or something like that.

albertshau: also, why is this optional?  What happens if it is not specified?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Nullable
+  DedupFunctionInfo getFilter() {
+    if (filterField == null) {
+      return null;
+    }
+
+    List<FunctionInfo> aggregates = parseAggregation(filterField);
+    if (aggregates.size() != 1) {
+      throw new IllegalArgumentException("Only one filter field is allowed!");
```
albertshau: nit: we don't use '!' anywhere else, lets replace with a '.'

albertshau: this is also just a confusing error condition. We shouldn't even be trying to parse it as multiple things.  Let's not re-use methods that dont make sense to re-use.  You can refactor to pull out the relevant parsing logic

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
```diff
@@ -0,0 +1,101 @@
...
+  static class DedupFunctionInfo extends FunctionInfo {
+
+    public DedupFunctionInfo(String name, String field, Function function) {
+      super(name, field, function);
+    }
+
+    public RecordAggregateFunction getAggregateFunction(Schema recordSchema) {
+      Schema.Field recordField = recordSchema.getField(getField());
+      Schema fieldSchema = recordField.getSchema();
+      switch (getFunction()) {
```
albertshau: function and field can be made protected

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RecordAggregateFunction.java
```diff
@@ -0,0 +1,34 @@
...
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+
+import javax.annotation.Nullable;
+
+/**
+ *
+ * @param <T> type of aggregate value
+ */
+public interface RecordAggregateFunction<T> extends AggregateFunction<T> {
```
albertshau: this is an odd name, its really more like a RecordSelector or a RecordFilter

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RecordAggregateFunction.java
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * @return {@link StructuredRecord} that is chosen based on the aggregate function.
+   */
+  @Nullable
+  StructuredRecord getChosenRecord();
```
albertshau: If we change this so that it returns a Collection, we can do things like TopK while still allowing min, max, etc.  We also wouldn't have to handle nulls

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RecordAggregateFunction.java
albertshau: It also shouldn't extend AggregateFunction, because then you require that it does both selection and aggregation which are completely separate things.  If you just need the start, update, finish methods from AggregateFunction, you can create a common interface they both extend.  Things like Max can implement both interfaces.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/NumberFunction.java
```diff
@@ -44,7 +45,10 @@ public void beginAggregate() {
         public void update(StructuredRecord record) {
           Number val = record.get(fieldName);
           if (val != null) {
-            updateDouble(val.doubleValue());
+            Double updateDouble = updateDouble(val.doubleValue());
+            if (val.doubleValue() == updateDouble) {
```
albertshau: I'm having trouble understanding this.  Is it to try and fit all AggregateFunctions into RecordAggregateFunctions?  Think we need some refactoring rather than trying to force things that don't fit together.

gokulavasan:
@albertshau Thanks for the comments. Addressed them. Please take a look again!

albertshau:
For future reference, it is easier to review if you rebase instead of merge, and squash post-review commits into a single commit.  Then the reviewer can just review that commit.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -0,0 +1,187 @@
...
+
+  private Schema getOutputSchema(Schema inputSchema) {
+    return Schema.recordOf(inputSchema.getRecordName() + ".dedup", inputSchema.getFields());
+  }
+
+  /**
+   * Endpoint request for output schema.
+   */
+  public static class GetSchemaRequest extends DedupConfig {
+    private String inputSchema;
```
albertshau: learned recently that this can actually be the 'Schema' object so that the plugin doesn't have to parse stringified json.  Should change the other aggregator to Schema as well

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -0,0 +1,187 @@
...
+
+      while (iterator.hasNext()) {
+        selectionFunction.operateOn(iterator.next());
+      }
+
+      selectionFunction.finishFunction();
+      List<StructuredRecord> outputRecords = selectionFunction.getSelectedRecords();
+      for (StructuredRecord outputRecord : outputRecords) {
+        Schema outputSchema = getOutputSchema(outputRecord.getSchema());
+        validateSchema(outputSchema, uniqueFields, filterFunction);
```
albertshau: unless I'm misunderstanding, I believe the schema can never be invalid at this point.  In order to get here, all unique fields had to have been present when grouping, and the function had to have worked.

If there is any validation needed, it should happen in the group method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/DedupConfig.java
```diff
@@ -0,0 +1,170 @@
...
+    "'fname, lname, item, cost' and if we want to deduplicate the records by name, then this property should be set " +
+    "to 'fname,lname'")
+  @Nullable
+  private String uniqueFields;
+
+  @Description("Optional property that can be set to predictably choose one or more records from the set of records " +
+    "that needs to be de-duplicated. This property takes in a field name and the logical operation that needs to be " +
+    "performed on that field on the set of records. The syntax is 'function(field)'. For example, if we want to " +
+    "choose the record with maximum cost for the records with schema 'fname, lname, item, cost', then this field " +
+    "should be set as 'max(cost)'. Note, only one pair of field and function is allowed.")
```
albertshau: Should add a note about what happens if this is not given.

albertshau: also should mention what functions are supported

gokulavasan:
@albertshau Thanks Albert. Will keep that in mind.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RecordFunctionLifecycle.java
```diff
@@ -0,0 +1,44 @@
...
+   * Called once for each {@link StructuredRecord} to operate up on. Calls to this method are made after the call to
+   * {@link #beginFunction}
+   * @param record {@link StructuredRecord} input record
+   */
+  void operateOn(StructuredRecord record);
+
+  /**
+   * Called once when the operation is complete.
+   */
+  void finishFunction();
```
albertshau: From what I saw, it seems like this is always a no-op, can we remove?

gokulavasan: We can but I thought it might be useful in case we make this into a plugin itself. I am ok removing it as well.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/NumberSelection.java
```diff
@@ -0,0 +1,222 @@
...
+
+  protected abstract void startInt();
+
+  protected abstract void startLong();
+
+  protected abstract void startFloat();
+
+  protected abstract void startDouble();
+
+  protected abstract void operateOnInt(StructuredRecord record);
```
albertshau: would be easier for subclasses if this method got both the int value and the record as arguments, since the first thing they will always do is look up the int value anyway.  Should be able to handle nulls in this class as well so that subclasses don't have to worry about it.

albertshau: similar comment for the other typed methods

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RecordFunctionLifecycle.java
albertshau: I think its best to remove until there is a use for it. Easier to introduce a method than take it away.

albertshau:
couple more comments

gokulavasan:
@albertshau Addressed comments. Please review the last commit. 
@awholegunch The last commit in this PR also contains a doc for the new Deduplicate plugin (Deduplicate-batchaggregator.md). Can you please review the same?

albertshau:
code lgtm

awholegunch:
I've edited the documentation. Please review my changes. Otherwise, LGTM.

gokulavasan:
Thanks @albertshau! Thanks @awholegunch! Doc fixes look good. Will merge after build passes.

gokulavasan:
Build passed : http://builds.cask.co/browse/HYP-DBT5-1

PR #229, opened 2016-04-18
CDAP-5645 http plugin cleanup

albertshau:
Renaming HttpCallback to HTTPCallback, and moving it into the
http-plugins module. Also renaming UrlFetch to HTTPPoller, so that
they have similar names, and consolidating common config logic for
the two plugins.


http-plugins/docs/HTTPPoller-realtimesource.md
```diff
@@ -19,28 +19,28 @@ Properties
...
 Example
 -------
 This example fetches data from a url every hour using a custom user agent:
 
     {
+        "name": "HTTPPoller",
         "type": "realtimesource",
-        "name": "UrlFetch",
         "properties": {
             "url": "http://api.example.com/sampleEndpoint",
```
awholegunch: Why not have the domain be "example.com"? Then the URL actually returns something, as that is what the `example.com` domain is intended for:

```
http://example.com/sampleEndpoint
```

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
```diff
@@ -114,14 +111,18 @@ public void run(BatchActionContext batchActionContext) throws Exception {
...
-    @Description("The URL to call.")
-    private String url;
+  public static final class HttpRequestConf extends HTTPConfig {
+    private static final Set<String> METHODS = ImmutableSet.of(HttpMethod.GET, HttpMethod.HEAD, HttpMethod.OPTIONS,
+                                                               HttpMethod.PUT, HttpMethod.POST, HttpMethod.DELETE);
 
     @Nullable
+    @Description("When to run the action. Must be 'completion', 'success', or 'failure'. Defaults to 'completion'. " +
+      "If set to 'completion', the action will be executed regardless of whether " +
+      "the pipeline run succeeded or failed." +
```
awholegunch: Needs to end with a space:

```
"the pipeline run succeeded or failed. " +
```

http-plugins/README.rst
```diff
@@ -5,31 +5,30 @@ URL Plugins
 Introduction
 ============
 
-A Hydrator Real Time Plugin that allows data to be fetched from a given URL at a specified interval.
+A Hydrator real-time plugin that allows data to be fetched from a given URL at a specified interval.
```
russorat: So i guess this needs to be rewritten a little bit since now this contains multiple plugins (and will contain more in the future). Maybe something like: 
This project is a collection of plugins to perform various HTTP actions in your pipeline. These plugins are currently available:
- HTTPPoller real-time Source
- HTTPCallback

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
```diff
@@ -72,31 +72,28 @@ public void run(BatchActionContext batchActionContext) throws Exception {
...
-          }
+        conn.setRequestMethod(conf.method.toUpperCase());
+        conn.setConnectTimeout(conf.getConnectTimeout());
+        for (Map.Entry<String, String> propertyEntry : headers.entrySet()) {
+          conn.addRequestProperty(propertyEntry.getKey(), propertyEntry.getValue());
         }
         if (conf.body != null) {
           conn.setDoOutput(true);
           try (OutputStream outputStream = conn.getOutputStream()) {
             outputStream.write(conf.body.getBytes(Charsets.UTF_8));
```
russorat, line 87: Should the charset be configurable?

http-plugins/src/main/java/co/cask/hydrator/plugin/config/HTTPConfig.java
```diff
@@ -0,0 +1,115 @@
...
+import com.google.common.base.Strings;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Config class for URL Fetch plugin.
```
russorat: update comment

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
albertshau, line 87: Since the body is a user supplied string, its being sent through rest as json and must be utf8.

http-plugins/src/main/java/co/cask/hydrator/plugin/config/HTTPConfig.java
```diff
@@ -0,0 +1,115 @@
...
+  private String requestHeaders;
+
+  @Name("followRedirects")
+  @Description("Whether to automatically follow redirects. Defaults to true.")
+  @Nullable
+  private Boolean followRedirects;
+
+  @Description("Sets the connection timeout in milliseconds. Set to 0 for infinite. Default is 60000 (1 minute).")
+  @Nullable
+  private Integer connectTimeout;
```
russorat, line 53: Missing @Name annotation if that matters.

http-plugins/src/main/java/co/cask/hydrator/plugin/config/HTTPConfig.java
```diff
@@ -0,0 +1,115 @@
...
+  private static final String DELIMITER = "\n";
+
+  @Description("The URL to fetch data from.")
+  private String url;
+
+  @Description("Request headers to set when performing the http request.")
+  @Nullable
+  private String requestHeaders;
+
+  @Name("followRedirects")
```
russorat: if at Name isn't required, this isn't needed either.

russorat:
awesome rewrite! just a few minor comments.

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
russorat, line 87: ok. is that a limit of hydrator or the plugin? maybe mention that it must be a json formatted string in the docs?

http-plugins/README.rst
albertshau: I may actually just remove this readme.  There is one at the top level with almost all of this information (only non-overlap is the intro)

http-plugins/src/main/java/co/cask/hydrator/plugin/config/HTTPConfig.java
albertshau: will remove, its not needed if its the same as the variable name.

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
albertshau, line 87: its just a property of json, since all the plugin configs are part of the app config in the create app request (http://docs.cdap.io/cdap/current/en/reference-manual/http-restful-api/lifecycle.html#create-an-application)

albertshau:
@russorat thanks for the review! Addressed the comments, please take another look

http-plugins/src/main/java/co/cask/hydrator/plugin/batch/HttpCallbackAction.java
russorat, line 87: ah, i keep forgetting this all gets translated into a json app config. sorry.

russorat:
👍  LGTM

PR #230, opened 2016-04-18
moving timezone into the runtime macro

albertshau:
addressing an earlier comment how the timezone is specific to
runtime macro. Moving it as an argument for that macro type rather
than having it across macro types. Also fixing post actions so that
they actually perform macro substitution.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
@@ -19,10 +19,10 @@ package co.cask.hydrator.plugin.batch.action;
 import co.cask.cdap.api.annotation.Description;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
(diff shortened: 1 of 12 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
```diff
@@ -63,6 +63,11 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   @SuppressWarnings("ConstantConditions")
   @Override
   public void run(BatchActionContext batchActionContext) throws Exception {
+    conf.substituteMacros(batchActionContext);
```
gokulavasan: Should we substituteMacros if we are not going to run the Batch Action?

gokulavasan:
One question. LGTM otherwise.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HttpCallbackAction.java
albertshau: true, it will be better to substitute after. will change.

PR #232, opened 2016-04-20
CDAP-5392 update stream source to support spark

albertshau:
Have to tweak the source for now, since mapreduce and spark return
a different key type. Also updating unit test to test stream source
with spark.


chtyim:
LGTM

PR #235, opened 2016-04-20
CDAP-5664 Use external dataset naming feature in hydrator-plugins for lineage/metadata etc

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-5664
Build : http://builds.cask.co/browse/HYP-DBT6

Created five classes : ReferencePluginConfig (which contains the referenceName property); ReferenceBatchSource, ReferenceBatchSink, ReferenceRealtimeSink, ReferenceRealtimeSource - these four classes validate whether the referenceName property satisfies the constraints for a dataset instance name. And in the case of realtime plugins, they also 'getDataset' to register lineage.

All the external source/sink plugins should extend one of the above four classes. Their PluginConfig classes should also extend ReferencePluginConfig class.

Unit Tests had to modified since referenceName is a non-nullable property and so it had to be included in the plugin properties.

Also, modified the widget JSONs to have the referenceName property.


core-plugins/widgets/AzureBlobStore-batchsource.json
```diff
@@ -8,6 +8,11 @@
       "properties": [
         {
           "widget-type": "textbox",
+          "label": "Name for Dataset",
```
poornachandra: What is label? It should not be "Name for Dataset", right? Should it be just "Reference Name"?

poornachandra:
Like we discussed, we need to make sure referenceName is a valid Id, right?

Other than that code changes look good.

core-plugins/widgets/AzureBlobStore-batchsource.json
gokulavasan: I was not sure what to use for label. Sure, I will use Reference Name.

gokulavasan:
@awholegunch We need to add the property 'referenceName' to all non-CDAP source/sink plugins. I didn't want to copy paste the description of the property to all .md files. Instead I will leave it to create a doc constant of some kind that we can refer to in all the plugin docs and then modify only that if required.

@poornachandra Thanks for the review Poorna. I am not sure if there is an easy way to make sure that referenceName is a valid CDAP Id. @elfenheart Does UI have such a logic? If yes, we can do that check in UI. Of course, that means the CLI, REST are prone to errors. @sreevatsanraman Any thoughts?

awholegunch:
I think the only solution is add this to the Markdown, as there is no pre-processing of these files.

poornachandra:
@gokulavasan Even if UI does the id name verification, I think the check should also be done in the plugins. Should we add a helper method in etl-common that can be used to check whether a given string is a valid id? 

@albertshau Any thoughts on how to do this?

gokulavasan:
@albertshau @poornachandra The deprecated Id class in cdap-proto has a 'isValidId(String)' method. I can't find an equivalent method in the new EntityId class. If we add one, then I think we can depend on cdap-proto and check if referenceName is a validId during the configurePipeline step and thus fail the creation of the pipeline itself, instead of failing during the MapReduce runtime.

poornachandra:
@gokulavasan +1 to the suggestion.

gokulavasan:
@poornachandra Can you take a look again? Addressed the comment about validating referenceName property.

poornachandra:
LGTM

gokulavasan:
Build passed : http://builds.cask.co/browse/HYP-DBT6

albertshau:
lgtm

gokulavasan:
@albertshau Please review the last commit. Unfortunately we need to invoke the lookup method on the LookupProvider for it to 'getDataset'. One problem with that is that an exception is thrown (and logged) and the tx is aborted. I will file a JIRA documenting this. This should not be a big overhead since this is done only when the pipeline is initialized. CC @poornachandra

gokulavasan:
@albertshau @poornachandra Or we can ignore the lookup that if it is an external dataset in the TxLookupProvider? Is it possible?

albertshau:
need to do the same thing for the sink.  Like you mentioned, we may be able to do some hack in the app to avoid the exception. lgtm after adding for sink

gokulavasan:
@albertshau @poornachandra Please review the second commit. Modified it to use reflection to invoke getDataset instead of relying on lookupProvider (since using the latter logs an error in the logs which is not ideal and will probably be misleading). Also opened a JIRA CDAP-5759 to fix it.

gokulavasan:
@awholegunch Please review the third commit. It contains the documentation changes related to the addition of the new referenceName property to external sources and sinks.

awholegunch:
@gokulavasan Third commit looks ok to me. I think maybe the order of the properties needs to be addressed, because I think they should bear some resemblance to the UI (and vice-versa) but that can be an issue for another day.

albertshau:
lgtm

PR #237, opened 2016-04-21
Modifies plugin-functions in database & groupByAggregate plugins

ajainarayanan:
- Modifies `plugin-functions` property passed in widget json
  - Renames `endpoint` to `widget` 
  - Adds `plugin-method` to differentiate between the widget and plugin method to invoke.


gokulavasan:
@ajainarayanan Can this be added to Deduplicate Batch Aggregator as well? What about Distinct Aggregator?

elfenheart:
Need to update the wiki: https://wiki.cask.co/display/CE/UI+JSON+Specification

elfenheart:
👍  LGTM

PR #239, opened 2016-04-21
log error when there is an exception in get schema db source

gsps1:


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -136,20 +136,24 @@ private String getJDBCPluginType() {
    */
   @Path("getSchema")
```
bdmogal, line 137: this is unrelated to this PR, but `getSchema` seems inappropriate for a REST like endpoint. Shouldn't it just be `schema`, and `get` is reflected in the Http Method?

albertshau:
👍

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
gsps1, line 137: ideally, but we don't have option to specify the httpMethod here. as all requests are POST currently, as we pass parameter for the method in Http request body.

PR #242, opened 2016-04-25
CDAP-5765 fix es batch sink logging dependency

albertshau:
Added a dependency to jcl-over-slf4j to fix class not found
errors on logging classes.

### DIFF
--- a/elasticsearch-plugins/pom.xml
+++ b/elasticsearch-plugins/pom.xml
@@ -88,7 +88,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>jcl-over-slf4j</artifactId>
-      <version>1.7.12</version>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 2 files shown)


elasticsearch-plugins/pom.xml
```diff
@@ -85,6 +85,11 @@
       <groupId>commons-httpclient</groupId>
       <artifactId>commons-httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>jcl-over-slf4j</artifactId>
+      <version>1.7.12</version>
```
gokulavasan: Wondering if we should consolidate version numbers in the main pom?

albertshau: ok can do.

albertshau:
@gokulavasan fixed, can you take another look?

gokulavasan:
LGTM 👍

PR #249, opened 2016-05-04
fix warnings in batch sources

albertshau:
Move off deprecated methods for setting input of a run. Also fixing
other assorted warnings.


gokulavasan:
LGTM 👍

PR #251, opened 2016-05-05
Set PluginClassLoader to be the context classloader when creating the Configuration object

gokulavasan:
When HBaseConfiguration tries to load resources (hbase-default.xml), it uses the context classloader which was initially set to CombineClassLoader which doesn't have access to hbase-default.xml. So switching the context class loader to PluginClassLoader and switching it back solved the problem.

Verified on SDK, Multinode clusters.

TODO: Add an integration test to catch this issue early on.

### DIFF
--- a/hbase-plugins/src/main/java/co/cask/hydrator/plugin/sink/HBaseSink.java
+++ b/hbase-plugins/src/main/java/co/cask/hydrator/plugin/sink/HBaseSink.java
@@ -69,10 +69,23 @@ public class HBaseSink extends ReferenceBatchSink<StructuredRecord, NullWritable
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    Job job = JobUtils.createInstance();
-    Configuration conf = job.getConfiguration();
+    Job job;
+    ClassLoader oldClassLoader = Thread.currentThread().getContextClassLoader();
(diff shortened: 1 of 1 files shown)


gokulavasan:
Closing the PR to find out the actual reason in the platform that exposes underlying HBase classes.

albertshau:
lgtm though it would be useful to add a comment about why its needed

hbase-plugins/src/main/java/co/cask/hydrator/plugin/sink/HBaseSink.java
```diff
@@ -69,10 +69,14 @@ public HBaseSink(HBaseSinkConfig config) {
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
+    ClassLoader oldClassLoader = Thread.currentThread().getContextClassLoader();
+    Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
     Job job = JobUtils.createInstance();
```
bdmogal: can there be a failure here? enclose in a `try...finally`?

gokulavasan:
@albertshau @bdmogal Addressed comments. Please take a look again. Thanks!

albertshau:
lgtm

bdmogal:
lgtm

PR #254, opened 2016-05-11
file batch source enhancements

albertshau:
Adding macro support to the FileBatchSource, and adding a field
to its output that contains the split path the record was read
from.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
@@ -22,7 +22,6 @@ import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.common.Bytes;
 import co.cask.cdap.api.data.batch.Input;
 import co.cask.cdap.api.data.format.StructuredRecord;
-import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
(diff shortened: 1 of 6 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -228,7 +222,7 @@ public FileBatchConfig() {
       super("");
       this.fileSystemProperties = GSON.toJson(ImmutableMap.<String, String>of());
       this.fileRegex = ".*";
-      this.inputFormatClass = CombineTextInputFormat.class.getName();
+      this.inputFormatClass = TextInputFormat.class.getName();
```
gokulavasan, line 225: Any reason behind this change? I guess it should be have TextInputFormat to begin with.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StructuredRecordFileInputFormat.java
```diff
@@ -0,0 +1,146 @@
...
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Wrapper around a FileInputFormat that attempts to capture the path each record was read from.
+ */
+public class StructuredRecordFileInputFormat extends FileInputFormat<NullWritable, StructuredRecord> {
+  public static final Schema SCHEMA = Schema.recordOf(
+    "fileRecord",
+    Schema.Field.of("offset", Schema.nullableOf(Schema.of(Schema.Type.LONG))),
```
gokulavasan, line 43: Do we document anywhere that this is the schema of the records emitted by this source?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StructuredRecordFileInputFormat.java
```diff
@@ -0,0 +1,146 @@
...
+      return NullWritable.get();
+    }
+
+    @Override
+    public StructuredRecord getCurrentValue() throws IOException, InterruptedException {
+      Object key = delegate.getCurrentKey();
+      Object val = delegate.getCurrentValue();
+      return StructuredRecord.builder(SCHEMA)
+        .set("path", path)
+        .set("offset", key instanceof LongWritable ? ((LongWritable) key).get() : null)
```
gokulavasan, line 131: Is the offset, the offset within that particular file or is it offset across all files put together?

gokulavasan:
Few comments. Can we make sure this doesn't break plugins that extend FileBatchSource, ex FTPSource etc?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StructuredRecordFileInputFormat.java
nitinmotgi, line 131: It's the offset in the file.

albertshau:
macros are being added as a generic capability for all of hydrator instead of using this approach. closing.

PR #256, opened 2016-05-11
Removing parquet.hadoop export

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-5768
Build: http://builds.cask.co/browse/HYP-DBT9-1

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -199,7 +199,6 @@
               co.cask.hydrator.plugin.validator.*;
               co.cask.hydrator.plugin.common;
               parquet.avro.*;
-              parquet.hadoop.*
             </_exportcontents>
             <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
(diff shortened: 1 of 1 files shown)


core-plugins/pom.xml
```diff
@@ -199,7 +199,6 @@
               co.cask.hydrator.plugin.validator.*;
               co.cask.hydrator.plugin.common;
               parquet.avro.*;
```
albertshau, line 201: can this be removed too?

rohitsinha54, line 201: I think the ParquetInputFormat comes from here

CuriousVini, line 201: I think we can't because of parquet.avro.AvroParquetInputFormat.

albertshau:
lgtm assuming writing to and reading from parquet has been tested on both mapreduce and spark

albertshau:
also, shouldn't this fix go to the 1.3 branch?

CuriousVini:
closing and reopening against release branch

PR #259, opened 2016-05-13
JIRA ID - CDAP-5880.Implement Row Denormalizer Plugin for Hydrator. Initial Draft

rashig:
RowDenormalizer Aggregator Plugin

### DIFF
--- a/core-plugins/docs/RowDenormalizer-batchaggregator.md
+++ b/core-plugins/docs/RowDenormalizer-batchaggregator.md
@@ -10,39 +10,60 @@ Use Case
 The transform takes input record that stores a variable set of custom attributes for an entity, denormalizes it on the basis of the key field, and then returns a denormalized table according to the output schema specified by the user.
 The denormalized data is easier to query.
 
+Conditions
+----------
+In case a field value is not present, then it will be considered as NULL.
+
+For Example,
+
+If keyfield('id') in the input record is NULL, then that particular row will not be considered.
+
+If fieldname('attribute') and fieldvalue('value') is not present, then denormalized output will have NULL for that
+field.
+
+If user provides output field which is not present in the input record, then it will be considered as NULL.
+
 Properties
 ----------
-**keyField:** Name of the column in the input record which will be used to group the raw data. For Example, KeyField.
+**keyField:** Name of the column in the input record which will be used to group the raw data. For Example, id.
 
-**fieldName:** Name of the column in the input record which contains the names of output schema columns. For example, input records have columns 'KeyField', 'FieldName', 'FieldValue'. 'FieldName' contains 'FirstName', 'LastName', 'Address'. "So output record will have column names as 'FirstName', 'LastName', 'Address'.
+**fieldName:** Name of the column in the input record which contains the names of output schema columns. For example,
+ input records have columns 'id', 'attribute', 'value' and the 'attribute' column contains 'FirstName', 'LastName',
+ 'Address'.
+  "So the output record will have column names as 'FirstName', 'LastName', 'Address'.
 
-**fieldValue:** Name of the column in the input record which contains the values for output schema columns. For example, input records have columns 'KeyField', 'FieldName', 'FieldValue' and the 'FieldValue' column contains 'John', 'Wagh', 'NE Lakeside'. So the output record will have values for columns 'FirstName', 'LastName', 'Address' as 'John', 'Wagh', 'NE Lakeside' respectively.
+**fieldValue:** Name of the column in the input record which contains the values for output schema columns. For
+example, input records have columns 'id', 'attribute', 'value' and the 'value' column contains 'John',
+'Wagh', 'NE Lakeside'. So the output record will have values for columns 'FirstName', 'LastName', 'Address' as 'John', 'Wagh', 'NE Lakeside' respectively.
 
-**outputFields:** List of the output fields with its alias to be included in denormalized output. This is a comma separated list of key-value pairs, where key-value pairs are separated by a colon ':'. For example, 'Firstname:fname,Lastname:lname,Address:addr'.
+**outputFields:** List of the output fields to be included in denormalized output.
 
-**numPartitions:** Number of partitions to use when grouping data. If not specified, the execution
-framework will decide on the number to use.
+**fieldAliases:** List of the output fields to rename. The key specifies the name of the field to rename, with its corresponding value specifying the new name for that field.
+
+**numPartitions:** Number of partitions to use when grouping data. If not specified, the execution framework will
+decide on the number to use.
 
 Example
 -------
-The transform takes input records that have columns keyfield, fieldname, fieldvalue, denormalizes it on the basis of keyfield, and then returns a denormalized table according to the output schema specified by the user.
+The transform takes input records that have columns id, attribute, value, denormalizes it on the basis of
+id, and then returns a denormalized table according to the output schema specified by the user.
 
     {
-          "name": "RowDenormalizer",
-          "type": "batchaggregator",
-          "properties": {
-            "outputFields": "Firstname:,Lastname:,Address:addr",
-            "keyField": "keyfield",
-            "fieldName": "fieldname",
-            "fieldValue": "fieldvalue"
-          }
-        }
-    }
+              "name": "RowDenormalizer",
+              "type": "batchaggregator",
+              "properties": {
+                "outputFields": "Firstname,Lastname,Address",
+                "fieldAliases": "Address:Office Address",
+                "keyField": "id",
+                "fieldName": "attribute",
+                "fieldValue": "value"
+              }
+     }
 
 For example, suppose the aggregator receives the input record:
 
     +======================================+
-    | keyfield  | fieldname   | fieldvalue |
+    | id        | attribute   | value      |
     +======================================+
     | joltie    | Firstname   | John       |
     | joltie    | Lastname    | Wagh       |
@@ -51,10 +72,10 @@ For example, suppose the aggregator receives the input record:
 
 Output records will contain all the output fields specified by user:
 
-    +=====================================================+
-    | keyfield  | Firstname   | Lastname   |  addr        |
-    +=====================================================+
-    | joltie    | John        | Wagh       |  NE Lakeside |
-    +=====================================================+
+    +=========================================================+
+    | id        | Firstname   | Lastname   |  Office Address  |
+    +=========================================================+
+    | joltie    | John        | Wagh       |  NE Lakeside     |
+    +=========================================================+
 
 
--- a/core-plugins/docs/RowDenormalizer-batchaggregator.md
+++ b/core-plugins/docs/RowDenormalizer-batchaggregator.md
@@ -14,12 +14,12 @@ Properties
 ----------
 **keyField:** Name of the column in the input record which will be used to group the raw data. For Example, id.
 
-**fieldName:** Name of the column in the input record which contains the names of output schema columns. For example,
+**nameField:** Name of the column in the input record which contains the names of output schema columns. For example,
  input records have columns 'id', 'attribute', 'value' and the 'attribute' column contains 'FirstName', 'LastName',
  'Address'.
   "So the output record will have column names as 'FirstName', 'LastName', 'Address'.
 
-**fieldValue:** Name of the column in the input record which contains the values for output schema columns. For
+**valueField:** Name of the column in the input record which contains the values for output schema columns. For
 example, input records have columns 'id', 'attribute', 'value' and the 'value' column contains 'John',
 'Wagh', 'NE Lakeside'. So the output record will have values for columns 'FirstName', 'LastName', 'Address' as 'John', 'Wagh', 'NE Lakeside' respectively.
 
@@ -36,10 +36,10 @@ In case a field value is not present, then it will be considered as NULL.
 
 For Example,
 
-If keyfield('id') in the input record is NULL, then that particular row will not be considered.
+If keyfield('id') in the input record is NULL, then that particular record will be filtered out.
 
-If fieldname('attribute') or fieldvalue('value') is not present for a particular keyfield('id') value, then the
-denormalized output value for that fieldname will be NULL.
+If namefield('attribute') or valuefield('value') is not present for a particular keyfield('id') value, then the
+denormalized output value for that namefield will be NULL.
 
 If user provides output field which is not present in the input record, then it will be considered as NULL.
 
@@ -55,8 +55,8 @@ id, and then returns a denormalized table according to the output schema specifi
          "outputFields": "Firstname,Lastname,Address",
          "fieldAliases": "Address:Office Address",
          "keyField": "id",
-         "fieldName": "attribute",
-         "fieldValue": "value"
+         "nameField": "attribute",
+         "valueField": "value"
        }
     }
 
--- a/core-plugins/docs/RowDenormalizer-batchaggregator.md
+++ b/core-plugins/docs/RowDenormalizer-batchaggregator.md
@@ -10,19 +10,6 @@ Use Case
 The transform takes input record that stores a variable set of custom attributes for an entity, denormalizes it on the basis of the key field, and then returns a denormalized table according to the output schema specified by the user.
 The denormalized data is easier to query.
 
-Conditions
-----------
-In case a field value is not present, then it will be considered as NULL.
-
-For Example,
-
-If keyfield('id') in the input record is NULL, then that particular row will not be considered.
-
-If fieldname('attribute') and fieldvalue('value') is not present, then denormalized output will have NULL for that
-field.
-
-If user provides output field which is not present in the input record, then it will be considered as NULL.
-
 Properties
 ----------
 **keyField:** Name of the column in the input record which will be used to group the raw data. For Example, id.
@@ -43,22 +30,35 @@ example, input records have columns 'id', 'attribute', 'value' and the 'value' c
 **numPartitions:** Number of partitions to use when grouping data. If not specified, the execution framework will
 decide on the number to use.
 
+Conditions
+----------
+In case a field value is not present, then it will be considered as NULL.
+
+For Example,
+
+If keyfield('id') in the input record is NULL, then that particular row will not be considered.
+
+If fieldname('attribute') or fieldvalue('value') is not present for a particular keyfield('id') value, then the
+denormalized output value for that fieldname will be NULL.
+
+If user provides output field which is not present in the input record, then it will be considered as NULL.
+
 Example
 -------
 The transform takes input records that have columns id, attribute, value, denormalizes it on the basis of
 id, and then returns a denormalized table according to the output schema specified by the user.
 
     {
-              "name": "RowDenormalizer",
-              "type": "batchaggregator",
-              "properties": {
-                "outputFields": "Firstname,Lastname,Address",
-                "fieldAliases": "Address:Office Address",
-                "keyField": "id",
-                "fieldName": "attribute",
-                "fieldValue": "value"
-              }
-     }
+      "name": "RowDenormalizer",
+      "type": "batchaggregator",
+      "properties": {
+         "outputFields": "Firstname,Lastname,Address",
+         "fieldAliases": "Address:Office Address",
+         "keyField": "id",
+         "fieldName": "attribute",
+         "fieldValue": "value"
+       }
+    }
 
 For example, suppose the aggregator receives the input record:
 
@@ -79,3 +79,28 @@ Output records will contain all the output fields specified by user:
     +=========================================================+
 
 
+Now, let's suppose the aggregator receives the input record with NULL values:
+
+    +======================================+
+    | id        | attribute   | value      |
+    +======================================+
+    | joltie    | Firstname   | John       |
+    | joltie    | Lastname    | Wagh       |
+    | joltie    | Address     | NE Lakeside|
+    | brett     | Firstname   | Brett      |
+    |           | Lastname    | Lee        |
+    | brett     | Address     | SE Lakeside|
+    | bob       | Firstname   | Bob        |
+    | bob       |             | Smith      |
+    | bob       | Address     |            |
+    +======================================+
+
+Output records will contain all the output fields specified by user:
+
+    +=========================================================+
+    | id        | Firstname   | Lastname   |  Office Address  |
+    +=========================================================+
+    | joltie    | John        | Wagh       |  NE Lakeside     |
+    | brett     | Brett       |            |  SE Lakeside     |
+    | bob       | Bob         |            |                  |
+    +=========================================================+
--- a/core-plugins/docs/RowDenormalizer-batchaggregator.md
+++ b/core-plugins/docs/RowDenormalizer-batchaggregator.md
@@ -1,0 +1,60 @@
+# RowDenormalizer
+
+
+Description
+-----------
+Converts raw data into denormalized data based on a key column. User is able to specify the list of fields that should be used in the denormalized record, with an option to use an alias for the output field name. For example, 'ADDRESS' in the input is mapped to 'addr' in the output schema. 
+
+Use Case
+--------
+The transform takes input record that stores a variable set of custom attributes for an entity, denormalizes it on the basis of the key field, and then returns a denormalized table according to the output schema specified by the user.
+The denormalized data is easier to query.
+
+Properties
+----------
+**keyField:** Name of the column in the input record which will be used to group the raw data. For Example, KeyField.
+
+**fieldName:** Name of the column in the input record which contains the names of output schema columns. For example, input records have columns 'KeyField', 'FieldName', 'FieldValue'. 'FieldName' contains 'FirstName', 'LastName', 'Address'. "So output record will have column names as 'FirstName', 'LastName', 'Address'.
+
+**fieldValue:** Name of the column in the input record which contains the values for output schema columns. For example, input records have columns 'KeyField', 'FieldName', 'FieldValue' and the 'FieldValue' column contains 'John', 'Wagh', 'NE Lakeside'. So the output record will have values for columns 'FirstName', 'LastName', 'Address' as 'John', 'Wagh', 'NE Lakeside' respectively.
+
+**outputFields:** List of the output fields with its alias to be included in denormalized output. This is a comma separated list of key-value pairs, where key-value pairs are separated by a colon ':'. For example, 'Firstname:fname,Lastname:lname,Address:addr'.
+
+**numPartitions:** Number of partitions to use when grouping data. If not specified, the execution
+framework will decide on the number to use.
+
+Example
+-------
+The transform takes input records that have columns keyfield, fieldname, fieldvalue, denormalizes it on the basis of keyfield, and then returns a denormalized table according to the output schema specified by the user.
+
+    {
+          "name": "RowDenormalizer",
+          "type": "batchaggregator",
+          "properties": {
+            "outputFields": "Firstname:,Lastname:,Address:addr",
+            "keyField": "keyfield",
+            "fieldName": "fieldname",
+            "fieldValue": "fieldvalue"
+          }
+        }
+    }
+
+For example, suppose the aggregator receives the input record:
+
+    +======================================+
+    | keyfield  | fieldname   | fieldvalue |
+    +======================================+
+    | joltie    | Firstname   | John       |
+    | joltie    | Lastname    | Wagh       |
+    | joltie    | Address     | NE Lakeside|
+    +======================================+
+
+Output records will contain all the output fields specified by user:
+
+    +=====================================================+
+    | keyfield  | Firstname   | Lastname   |  addr        |
+    +=====================================================+
+    | joltie    | John        | Wagh       |  NE Lakeside |
+    +=====================================================+
+
+
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
@@ -50,8 +50,8 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
   private Set<String> outputFields;
   private Schema outputSchema;
   private String keyField;
-  private String fieldName;
-  private String fieldValue;
+  private String nameField;
+  private String valueField;
 
   public RowDenormalizerAggregator(RowDenormalizerConfig conf) {
     this.conf = conf;
@@ -75,15 +75,16 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
     outputFields = conf.getOutputSchemaFields();
     outputMappings = conf.getFieldAliases();
     keyField = conf.getKeyField();
-    fieldName = conf.getFieldName();
-    fieldValue = conf.getFieldValue();
+    nameField = conf.getNameField();
+    valueField = conf.getValueField();
     outputSchema = initializeOutputSchema();
   }
 
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
     if (record.get(keyField) == null) {
-      if (!record.getSchema().getField(keyField).getSchema().isNullable()) {
+      if (record.getSchema().getField(keyField) != null && !record.getSchema().getField(keyField).getSchema()
+        .isNullable()) {
         throw new IllegalArgumentException("null value found for non-nullable field " + keyField);
       }
       return;
@@ -101,18 +102,16 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
     builder.set(conf.getKeyField(), groupKey);
     while (iterator.hasNext()) {
       StructuredRecord record = iterator.next();
-      String outputFieldName = record.get(fieldName);
-      String outputFieldValue = record.get(fieldValue);
+      String outputFieldName = record.get(nameField);
+      String outputFieldValue = record.get(valueField);
 
-      if (outputFieldName == null) {
-        if (!record.getSchema().getField(fieldName).getSchema().isNullable()) {
-          throw new IllegalArgumentException("null value found for non-nullable field " + fieldName);
-        }
+      if (outputFieldName == null && record.getSchema().getField(nameField) != null && !record.getSchema().getField
+        (nameField).getSchema().isNullable()) {
+        throw new IllegalArgumentException("null value found for non-nullable field " + nameField);
       }
-      if (outputFieldValue == null) {
-        if (!record.getSchema().getField(fieldValue).getSchema().isNullable()) {
-          throw new IllegalArgumentException("null value found for non-nullable field " + fieldValue);
-        }
+      if (outputFieldValue == null && record.getSchema().getField(valueField) != null && !record.getSchema().getField
+        (valueField).getSchema().isNullable()) {
+        throw new IllegalArgumentException("null value found for non-nullable field " + valueField);
       }
 
       outputFieldName = outputMappings.containsKey(outputFieldName) ? outputMappings.get(outputFieldName) :
@@ -139,36 +138,39 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
 
   /**
    * @param inputSchema Validates whether the keyfield, fieldname, and fieldvalue entered by the user is of type
-   *                    String and present in the input schema or not.
+   *                    String or Nullable String and present in the input schema or not.
    * @return true
    */
   private void validateInputFields(Schema inputSchema) {
 
     Schema.Field keyField = inputSchema.getField(conf.getKeyField());
-    Schema.Field fieldName = inputSchema.getField(conf.getFieldName());
-    Schema.Field fieldValue = inputSchema.getField(conf.getFieldValue());
+    Schema.Field nameField = inputSchema.getField(conf.getNameField());
+    Schema.Field valueField = inputSchema.getField(conf.getValueField());
 
     if (keyField == null) {
       throw new IllegalArgumentException(
         String.format("Keyfield '%s' does not exist in input schema %s", conf.getKeyField(), inputSchema));
-    } else if (fieldName == null) {
+    } else if (nameField == null) {
       throw new IllegalArgumentException(
-        String.format("Fieldname '%s' does not exist in input schema %s", conf.getFieldName(), inputSchema));
-    } else if (fieldValue == null) {
+        String.format("Namefield '%s' does not exist in input schema %s", conf.getNameField(), inputSchema));
+    } else if (valueField == null) {
       throw new IllegalArgumentException(
-        String.format("Fieldvalue '%s' does not exist in input schema %s", conf.getFieldValue(), inputSchema));
+        String.format("Valuefield '%s' does not exist in input schema %s", conf.getValueField(), inputSchema));
     } else if (!((keyField.getSchema().isNullable() ? keyField.getSchema().getNonNullable().getType() : keyField
       .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Keyfield '%s' in the input record must be of type Nullable String", conf.getKeyField()));
-    } else if (!((fieldName.getSchema().isNullable() ? fieldName.getSchema().getNonNullable().getType() : fieldName
+        String.format("Keyfield '%s' in the input record must be of type String or Nullable String",
+                      conf.getKeyField()));
+    } else if (!((nameField.getSchema().isNullable() ? nameField.getSchema().getNonNullable().getType() : nameField
       .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Fieldname '%s' in the input record must be of type Nullable String", conf.getFieldName()));
-    } else if (!((fieldValue.getSchema().isNullable() ? fieldValue.getSchema().getNonNullable().getType() : fieldValue
+        String.format("Namefield '%s' in the input record must be of type String or Nullable String",
+                      conf.getNameField()));
+    } else if (!((valueField.getSchema().isNullable() ? valueField.getSchema().getNonNullable().getType() : valueField
       .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Fieldvalue '%s' in the input record must be of type Nullable String", conf.getFieldValue()));
+        String.format("Valuefield '%s' in the input record must be of type String or Nullable String",
+                      conf.getValueField()));
     }
   }
 }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
@@ -71,14 +71,17 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
     outputFields = conf.getOutputFields();
-    outputMappings = conf.getOutputFieldMappings();
+    outputMappings = conf.getFieldAliases();
     keyField = conf.getKeyField();
     outputSchema = initializeOutputSchema();
   }
 
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
-    emitter.emit(record.get(keyField).toString());
+    if (record.get(keyField) == null) {
+      return;
+    }
+    emitter.emit((String) record.get(keyField));
   }
 
   @Override
@@ -114,8 +117,8 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
   }
 
   /**
-   * @param inputSchema Validates whether the keyfield, fieldname, and fieldvalue entered by the user is present in
-   *                    the input schema or not.
+   * @param inputSchema Validates whether the keyfield, fieldname, and fieldvalue entered by the user is of type
+   *                    String and present in the input schema or not.
    * @return true
    */
   private void validateInputFields(Schema inputSchema) {
@@ -128,6 +131,15 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
     } else if ((inputSchema.getField(conf.getFieldValue())) == null) {
       throw new IllegalArgumentException(
         String.format("Fieldvalue '%s' does not exist in input schema %s", conf.getFieldValue(), inputSchema));
+    } else if (!inputSchema.getField(conf.getKeyField()).getSchema().getType().equals(Schema.Type.STRING)) {
+      throw new IllegalArgumentException(
+        String.format("Keyfield '%s' must be of type String", conf.getKeyField()));
+    } else if (!inputSchema.getField(conf.getFieldName()).getSchema().getType().equals(Schema.Type.STRING)) {
+      throw new IllegalArgumentException(
+        String.format("Fieldname '%s' must be of type String", conf.getFieldName()));
+    } else if (!inputSchema.getField(conf.getFieldValue()).getSchema().getType().equals(Schema.Type.STRING)) {
+      throw new IllegalArgumentException(
+        String.format("Fieldvalue '%s' must be of type String", conf.getFieldValue()));
     }
   }
 }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
@@ -50,6 +50,8 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
   private Set<String> outputFields;
   private Schema outputSchema;
   private String keyField;
+  private String fieldName;
+  private String fieldValue;
 
   public RowDenormalizerAggregator(RowDenormalizerConfig conf) {
     this.conf = conf;
@@ -70,15 +72,20 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
 
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
-    outputFields = conf.getOutputFields();
+    outputFields = conf.getOutputSchemaFields();
     outputMappings = conf.getFieldAliases();
     keyField = conf.getKeyField();
+    fieldName = conf.getFieldName();
+    fieldValue = conf.getFieldValue();
     outputSchema = initializeOutputSchema();
   }
 
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
     if (record.get(keyField) == null) {
+      if (!record.getSchema().getField(keyField).getSchema().isNullable()) {
+        throw new IllegalArgumentException("null value found for non-nullable field " + keyField);
+      }
       return;
     }
     emitter.emit((String) record.get(keyField));
@@ -94,10 +101,24 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
     builder.set(conf.getKeyField(), groupKey);
     while (iterator.hasNext()) {
       StructuredRecord record = iterator.next();
-      String fieldName = record.get(conf.getFieldName());
-      fieldName = outputMappings.containsKey(fieldName) ? outputMappings.get(fieldName) : fieldName;
-      if (outputFields.contains(fieldName)) {
-        builder.set(fieldName, record.get(conf.getFieldValue()));
+      String outputFieldName = record.get(fieldName);
+      String outputFieldValue = record.get(fieldValue);
+
+      if (outputFieldName == null) {
+        if (!record.getSchema().getField(fieldName).getSchema().isNullable()) {
+          throw new IllegalArgumentException("null value found for non-nullable field " + fieldName);
+        }
+      }
+      if (outputFieldValue == null) {
+        if (!record.getSchema().getField(fieldValue).getSchema().isNullable()) {
+          throw new IllegalArgumentException("null value found for non-nullable field " + fieldValue);
+        }
+      }
+
+      outputFieldName = outputMappings.containsKey(outputFieldName) ? outputMappings.get(outputFieldName) :
+        outputFieldName;
+      if (outputFields.contains(outputFieldName)) {
+        builder.set(outputFieldName, outputFieldValue);
       }
     }
     emitter.emit(builder.build());
@@ -110,7 +131,7 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
     List<Schema.Field> fields = new ArrayList<>();
     // Input fields coming from source will be of type String; that's why we are building the output schema as String.
     fields.add(Schema.Field.of(conf.getKeyField(), Schema.of(Schema.Type.STRING)));
-    for (String outputField : conf.getOutputFields()) {
+    for (String outputField : conf.getOutputSchemaFields()) {
       fields.add(Schema.Field.of(outputField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
     }
     return Schema.recordOf("output.schema", fields);
@@ -122,24 +143,32 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
    * @return true
    */
   private void validateInputFields(Schema inputSchema) {
-    if ((inputSchema.getField(conf.getKeyField())) == null) {
+
+    Schema.Field keyField = inputSchema.getField(conf.getKeyField());
+    Schema.Field fieldName = inputSchema.getField(conf.getFieldName());
+    Schema.Field fieldValue = inputSchema.getField(conf.getFieldValue());
+
+    if (keyField == null) {
       throw new IllegalArgumentException(
         String.format("Keyfield '%s' does not exist in input schema %s", conf.getKeyField(), inputSchema));
-    } else if ((inputSchema.getField(conf.getFieldName())) == null) {
+    } else if (fieldName == null) {
       throw new IllegalArgumentException(
         String.format("Fieldname '%s' does not exist in input schema %s", conf.getFieldName(), inputSchema));
-    } else if ((inputSchema.getField(conf.getFieldValue())) == null) {
+    } else if (fieldValue == null) {
       throw new IllegalArgumentException(
         String.format("Fieldvalue '%s' does not exist in input schema %s", conf.getFieldValue(), inputSchema));
-    } else if (!inputSchema.getField(conf.getKeyField()).getSchema().getType().equals(Schema.Type.STRING)) {
+    } else if (!((keyField.getSchema().isNullable() ? keyField.getSchema().getNonNullable().getType() : keyField
+      .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Keyfield '%s' must be of type String", conf.getKeyField()));
-    } else if (!inputSchema.getField(conf.getFieldName()).getSchema().getType().equals(Schema.Type.STRING)) {
+        String.format("Keyfield '%s' in the input record must be of type Nullable String", conf.getKeyField()));
+    } else if (!((fieldName.getSchema().isNullable() ? fieldName.getSchema().getNonNullable().getType() : fieldName
+      .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Fieldname '%s' must be of type String", conf.getFieldName()));
-    } else if (!inputSchema.getField(conf.getFieldValue()).getSchema().getType().equals(Schema.Type.STRING)) {
+        String.format("Fieldname '%s' in the input record must be of type Nullable String", conf.getFieldName()));
+    } else if (!((fieldValue.getSchema().isNullable() ? fieldValue.getSchema().getNonNullable().getType() : fieldValue
+      .getSchema().getType()).equals(Schema.Type.STRING))) {
       throw new IllegalArgumentException(
-        String.format("Fieldvalue '%s' must be of type String", conf.getFieldValue()));
+        String.format("Fieldvalue '%s' in the input record must be of type Nullable String", conf.getFieldValue()));
     }
   }
 }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
@@ -83,11 +83,19 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
     if (record.get(keyField) == null) {
-      if (record.getSchema().getField(keyField) != null && !record.getSchema().getField(keyField).getSchema()
-        .isNullable()) {
-        throw new IllegalArgumentException("null value found for non-nullable field " + keyField);
+      if (record.getSchema().getField(keyField) == null) {
+        throw new IllegalArgumentException(
+          String.format("Keyfield '%s' does not exist in input schema %s", keyField, record.getSchema()));
       }
       return;
+    }
+    if (record.get(nameField) == null && record.getSchema().getField(nameField) == null) {
+      throw new IllegalArgumentException(
+        String.format("Namefield '%s' does not exist in input schema %s", nameField, record.getSchema()));
+    }
+    if (record.get(valueField) == null && record.getSchema().getField(valueField) == null) {
+      throw new IllegalArgumentException(
+        String.format("Valuefield '%s' does not exist in input schema %s", valueField, record.getSchema()));
     }
     emitter.emit((String) record.get(keyField));
   }
@@ -104,15 +112,6 @@ public class RowDenormalizerAggregator extends BatchAggregator<String, Structure
       StructuredRecord record = iterator.next();
       String outputFieldName = record.get(nameField);
       String outputFieldValue = record.get(valueField);
-
-      if (outputFieldName == null && record.getSchema().getField(nameField) != null && !record.getSchema().getField
-        (nameField).getSchema().isNullable()) {
-        throw new IllegalArgumentException("null value found for non-nullable field " + nameField);
-      }
-      if (outputFieldValue == null && record.getSchema().getField(valueField) != null && !record.getSchema().getField
-        (valueField).getSchema().isNullable()) {
-        throw new IllegalArgumentException("null value found for non-nullable field " + valueField);
-      }
 
       outputFieldName = outputMappings.containsKey(outputFieldName) ? outputMappings.get(outputFieldName) :
         outputFieldName;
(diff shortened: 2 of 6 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+    TimePartitionedFileSet fileSet = outputManager.get();
+    List<GenericRecord> outputRecords = readOutput(fileSet, outputSchema);
+
+    Set<String> key = new HashSet<>();
+    Set<String> firstName = new HashSet<>();
+    Set<String> lastName = new HashSet<>();
+    Set<String> address = new HashSet<>();
+    Set<String> salary = new HashSet<>();
+
+    for (GenericRecord record : outputRecords) {
```
anwar6953: why do we need a loop here, if we assert that its only length 1?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Batch RowDenormalizer Aggregator Plugin - It is used to de-normalize data based on the key column.
+ */
+@Plugin(type = BatchAggregator.PLUGIN_TYPE)
+@Name("RowDenormalizer")
+@Description("This plugin is used to convert the raw data into de-normalized data based on the key column." +
+  "User is able to specify the list of fields that should be considered in denormalized record. " +
+  "Also, users have option to put alias for output field name. " +
+  "For example - " +
+  "‘ADDRESS’ in input is mapped to ‘addr’ in output schema. This denormalized data would be easy to query.")
```
awholegunch:
```
@Description("This plugin is used to convert raw data into de-normalized data based on a key column. " +
  "User is able to specify the list of fields that should be used in the denormalized record, with " +
  "an option to use an alias for the output field name. For example: " +
  "'ADDRESS' in the input is mapped to ‘addr’ in the output schema. The denormalized data is easier to query.")
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+    emitter.emit(builder.build());
+  }
+
+  /**
+   * Builds the output schema, using output fields provided by user
+   */
+  private void initializeOutputSchema() {
+    List<Schema.Field> fields = new ArrayList<>();
+    // Input fields coming from source will be of String type, that's why we are building the output schema as String.
```
awholegunch: // Input fields coming from source will be of type String; that's why we are building the output schema as String.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    fields.add(Schema.Field.of(conf.getKeyField(), Schema.of(Schema.Type.STRING)));
+    for (String outputField : outputFields) {
+      fields.add(Schema.Field.of(outputField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    }
+    outputSchema = Schema.recordOf("group.key.schema", fields);
+  }
+
+  /**
+   * @param inputSchema Validates whether the keyfield, fieldname and fieldvalue entered by user is present in input
+   *                    schema or not.
```
awholegunch: - @param inputSchema Validates whether the keyfield, filename, and fieldvalue entered by the user is present in
-  the input schema or not.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Config for RowDenormalizer Aggregator Plugin
+ */
+public class RowDenormalizerConfig extends AggregatorConfig {
+
+  public static final String CHECK_ALIAS = "NO_ALIAS";
+  @Description("Key Field which will be used to de-normalize the raw data.")
```
awholegunch: field

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+  private final String keyField;
+  @Description("Name of the field in input record that contains output fields to be included in denormalized output.")
```
awholegunch:
```
"Name of the field in the input record that contains the output fields to be included in the denormalized output."
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+  private final String fieldName;
+  @Description("Name of the field in input record that contains values for output fields to be included " +
+    "in denormalized output.")
```
awholegunch:
```
@Description("Name of the field in the input record that contains the values for the output fields to be included " +
  "in the denormalized output.")
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+   * Returns name of the keyfield given as input by user.
+   *
+   * @return key field
+   */
+  String getKeyField() {
+    return keyField;
+  }
+
+  /**
+   * Returns name of the input field 'FieldName' given as input by user
```
awholegunch: user.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @return field name
+   */
+  String getFieldName() {
+    return fieldName;
+  }
+
+  /**
+   * Returns name of the input field 'FieldValue' given as input by user
```
awholegunch: user.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @return field value
+   */
+  String getFieldValue() {
+    return fieldValue;
+  }
+
+  /**
+   * Fetches the output fields or output field alias if provided, to build the output schema.
```
awholegunch: Fetches the output fields (or output field alias, if provided) to build the output schema.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+        outputFieldMappings.put(value[0], CHECK_ALIAS);
+      } else {
+        outputFieldMappings.put(value[0], value[1]);
+      }
+    }
+    return outputFieldMappings;
+  }
+
+  /**
+   * Parses whether the schema json is proper or not
```
awholegunch: JSON is proper or not.

awholegunch:
There should be a Markdown documentation file, and possibly a JSON for the UI, include in this PR.
Plus a `pom.xml` for building?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    outputFields = conf.getOutputFields();
+    outputMappings = conf.getOutputFieldMappings();
+  }
+
+  @Override
+  public void groupBy(StructuredRecord record, Emitter<StructuredRecord> emitter) throws Exception {
+    String keyField = conf.getKeyField();
+    Schema keyFieldSchema = Schema.recordOf("record", Schema.Field.of(keyField, Schema.of(Schema.Type.STRING)));
```
albertshau: you can set this in initialize instead of creating it for each groupBy call

albertshau: if the group key only comes from a single string field, you can just have the group key as a String instead of a StructuredRecord. It will perform better that way since we don't need to create records, and serialization/deserialization done by mapreduce will be simpler as well

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    outputSchema = Schema.recordOf("group.key.schema", fields);
+  }
+
+  /**
+   * @param inputSchema Validates whether the keyfield, fieldname and fieldvalue entered by user is present in input
+   *                    schema or not.
+   * @return true
+   */
+  private void validateInputFields(Schema inputSchema) {
+    if (!inputSchema.getFields().toString().contains(conf.getKeyField())) {
```
albertshau: you should not be checking if a field exists in the schema by doing .toString() on a list and checking the key field is part of that string, as you can get false positives that way.  Instead, you need to iterate through the list and check that the configured key field matches one of the field names.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+      throw new IllegalArgumentException(
+        String.format("Cannot denormalize using this field '%s' as it does not exist in input schema %s",
+                      conf.getKeyField(), inputSchema));
+    } else if (!inputSchema.getFields().toString().contains(conf.getFieldName())) {
```
albertshau: similar comment

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+      throw new IllegalArgumentException(
+        String.format("Cannot find field '%s' in input schema %s", conf.getFieldName(), inputSchema));
+    } else if (!inputSchema.getFields().toString().contains(conf.getFieldValue())) {
```
albertshau: similar comment

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+
+  /**
+   * @param inputSchema Validates whether the keyfield, fieldname and fieldvalue entered by user is present in input
+   *                    schema or not.
+   * @return true
+   */
+  private void validateInputFields(Schema inputSchema) {
+    if (!inputSchema.getFields().toString().contains(conf.getKeyField())) {
+      throw new IllegalArgumentException(
+        String.format("Cannot denormalize using this field '%s' as it does not exist in input schema %s",
```
albertshau: these error messages should indicate which config property is wrong. Key field '%s' does not exist in input schema %s

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+                      conf.getKeyField(), inputSchema));
+    } else if (!inputSchema.getFields().toString().contains(conf.getFieldName())) {
+      throw new IllegalArgumentException(
+        String.format("Cannot find field '%s' in input schema %s", conf.getFieldName(), inputSchema));
```
albertshau: fieldName '%s' does not exist in input schema %s

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    } else if (!inputSchema.getFields().toString().contains(conf.getFieldValue())) {
+      throw new IllegalArgumentException(
+        String.format("Cannot find field '%s' in input schema %s", conf.getFieldValue(), inputSchema));
```
albertshau: fieldValue '%s' does not exist in input schema %s

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: for consistent style, insert newlines between variables

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Config for RowDenormalizer Aggregator Plugin
+ */
+public class RowDenormalizerConfig extends AggregatorConfig {
+
+  public static final String CHECK_ALIAS = "NO_ALIAS";
+  @Description("Key Field which will be used to de-normalize the raw data.")
+  private final String keyField;
+  @Description("Name of the field in input record that contains output fields to be included in denormalized output.")
+  private final String fieldName;
```
albertshau: for consistent style, insert newlines after variables

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+  @Description("Name of the field in input record that contains values for output fields to be included " +
+    "in denormalized output.")
+  private final String fieldValue;
+  @Description("List of the output fields with its alias to be included in denormalized ouptut.")
+  private final String outputFields;
+  @Name("schema")
```
albertshau: name annotation is not needed when its the same as the variable name

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: I'm unable to determine what to set this property to based on this description. What does this mean?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: I also don't understand how to use this property either.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: Don't understand this description either

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+
+  public static final String CHECK_ALIAS = "NO_ALIAS";
+  @Description("Key Field which will be used to de-normalize the raw data.")
+  private final String keyField;
+  @Description("Name of the field in input record that contains output fields to be included in denormalized output.")
+  private final String fieldName;
+  @Description("Name of the field in input record that contains values for output fields to be included " +
+    "in denormalized output.")
+  private final String fieldValue;
+  @Description("List of the output fields with its alias to be included in denormalized ouptut.")
```
albertshau: same here. I also don't know how to specify the list. Is it comma separated? What's an alias?

albertshau: from the code I can tell this is a comma separated list of key-value pairs, where key-value pairs are separated by a colon ':'.  This needs to be mentioned in the description with an example.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Creates a map for output field and its alias, if present.
+   *
+   * @return Map of output fields and its alias
+   */
+  Map<String, String> getOutputFieldMappings() {
+    Map<String, String> outputFieldMappings = new HashMap<String, String>();
+    for (String field : Splitter.on(',').trimResults().split(outputFields)) {
+      String[] value = field.split(":");
+      if (value.length == 1) {
```
albertshau: should check the length of the array is what you're expecting. For example, what happens if there is more than one colon here?

albertshau: this is very confusing to me. So this property is a list of either single values or key-value pairs? If a field doesn't need to be aliased, shouldn't it just not be specified in this list?

albertshau: is an alias a rename of an input field?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+        outputFieldMappings.put(value[0], CHECK_ALIAS);
```
albertshau: you should avoid doing this type of thing where a special string is being reserved to have a specific meaning, as it will break if a user ever uses that special string.  Instead, can the absence of a field in this list be interpreted as there is no alias for that field?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: this needs a description. I don't understand how schema is being used.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+    return fields;
+  }
+
+  /**
+   * Returns the output schema
+   *
+   * @return schema
+   */
+  String getSchema() {
```
albertshau: why have both this and parseSchema?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+    return outputFieldMappings;
+  }
+
+  /**
+   * Parses whether the schema json is proper or not
+   *
+   * @return schema
+   */
+  Schema parseSchema() {
```
albertshau: methods should be annotated as `@Nullable` if they can return null.  What does a null schema mean?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      return Strings.isNullOrEmpty(schema) ? null : Schema.parseJson(schema);
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Invalid schema: " + e.getMessage());
+    }
+  }
+
+  /**
+   * Validates the config input coming from UI
```
albertshau: there should not be any mention of the UI in the backend code. Everything can be done without using the UI. This is validating config properties set by the user.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,161 @@
...
+   */
+  void validate() {
+    if (keyField.isEmpty()) {
+      throw new IllegalArgumentException("The 'Key Field' property must be set.");
```
albertshau: these error messages should use the name of the expected property. keyField

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    emitter.emit(builder.build());
+  }
+
+  @Override
+  public void aggregate(StructuredRecord groupKey, Iterator<StructuredRecord> iterator,
+                        Emitter<StructuredRecord> emitter) throws Exception {
+    if (!iterator.hasNext()) {
+      return;
+    }
+    if (outputSchema == null) {
```
albertshau: just initialize this in the initialize() method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+      initializeOutputSchema();
+    }
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    builder.set(conf.getKeyField(), groupKey.get(groupKey.getSchema().getFields().get(0).getName()));
```
albertshau: as mentioned above, this is much simpler if the group key is just a string. Then you could just set it here.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    while (iterator.hasNext()) {
+      StructuredRecord record = iterator.next();
+      String fieldName = record.get(conf.getFieldName());
+
+      if (outputFields.contains(fieldName)) {
+        builder.set(fieldName, record.get(conf.getFieldValue()));
+      } else {
+        if ((outputMappings.containsKey(fieldName)) && (!outputMappings.get(fieldName).
+          equals(RowDenormalizerConfig.CHECK_ALIAS))) {
```
albertshau: style: don't put method calls on a separate line if you can avoid it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    if (outputSchema == null) {
+      initializeOutputSchema();
+    }
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    builder.set(conf.getKeyField(), groupKey.get(groupKey.getSchema().getFields().get(0).getName()));
+    while (iterator.hasNext()) {
+      StructuredRecord record = iterator.next();
+      String fieldName = record.get(conf.getFieldName());
+
+      if (outputFields.contains(fieldName)) {
```
albertshau: this is a linear operation. Should use a Set instead of a List if you are using contains()

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+   * Builds the output schema, using output fields provided by user
+   */
+  private void initializeOutputSchema() {
+    List<Schema.Field> fields = new ArrayList<>();
+    // Input fields coming from source will be of String type, that's why we are building the output schema as String.
+    fields.add(Schema.Field.of(conf.getKeyField(), Schema.of(Schema.Type.STRING)));
+    for (String outputField : outputFields) {
+      fields.add(Schema.Field.of(outputField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    }
+    outputSchema = Schema.recordOf("group.key.schema", fields);
```
albertshau: this is a misleading record name. It is not the group key's schema. It is the schema of the output records.

albertshau: I'm very confused now.  The conf has a schema property, which is what configurePipeline() uses. But then this is the actual output schema that gets used. Seems like the conf should not have a schema property at all, and that this should be used to set the output schema in configurePipeline()

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: don't use special strings to have special meanings, as it will break if the user ever uses that string themselves.  Seems like it should be enough to check that the field name exists in the output mappings.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    while (iterator.hasNext()) {
+      StructuredRecord record = iterator.next();
+      String fieldName = record.get(conf.getFieldName());
+
+      if (outputFields.contains(fieldName)) {
+        builder.set(fieldName, record.get(conf.getFieldValue()));
+      } else {
+        if ((outputMappings.containsKey(fieldName)) && (!outputMappings.get(fieldName).
+          equals(RowDenormalizerConfig.CHECK_ALIAS))) {
+          builder.set(outputMappings.get(fieldName), record.get(conf.getFieldValue()));
```
albertshau: the output schema never looks at outputMappings, so won't this try and set a field that does not exist in the schema?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+@Plugin(type = BatchAggregator.PLUGIN_TYPE)
+@Name("RowDenormalizer")
+@Description("This plugin is used to convert the raw data into de-normalized data based on the key column." +
+  "User is able to specify the list of fields that should be considered in denormalized record. " +
+  "Also, users have option to put alias for output field name. " +
+  "For example - " +
+  "‘ADDRESS’ in input is mapped to ‘addr’ in output schema. This denormalized data would be easy to query.")
+public class RowDenormalizerAggregator extends RecordAggregator {
+  private final RowDenormalizerConfig conf;
+  private Map<String, String> outputMappings;
```
chtyim, line 49: Please group `final` fields together before non-final fields.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,135 @@
...
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    builder.set(conf.getKeyField(), groupKey.get(groupKey.getSchema().getFields().get(0).getName()));
+    while (iterator.hasNext()) {
+      StructuredRecord record = iterator.next();
+      String fieldName = record.get(conf.getFieldName());
+
+      if (outputFields.contains(fieldName)) {
+        builder.set(fieldName, record.get(conf.getFieldValue()));
+      } else {
+        if ((outputMappings.containsKey(fieldName)) && (!outputMappings.get(fieldName).
```
chtyim: combine the `else { if` to `else if {`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: seems like an alias is a rename of an input field.  The design for this property seems messed up to me.  Can't we just leave renaming to the projection transform? Why do we need to do multiple things in this aggregator.  It just makes this difficult to understand in my opinion.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: oh never mind, outputFields uses the value in outputMappings if it is there.  This is really confusing. Instead of this logic, just do:

```
String fieldName = record.get(conf.getFieldName());
fieldName = outputMappings.containsKey(fieldName) ? outputMappings.get(fieldName) : fieldName;
```

and you can remove this entire else clause

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfigTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testDormalizerConfWithNoFieldValue() {
+    RowDenormalizerConfig config = new RowDenormalizerConfig("KeyField", "FieldName", "", "Firstname:," +
+      "Lastname:lname," + "Address:addr", outputSchema.toString());
+    config.validate();
+  }
+}
+
+
```
albertshau, line 60: only need one newline at the end of a file

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test cases for {@link RowDenormalizerAggregator}.
+ */
+public class RowDenormalizerAggregatorTest extends ETLBatchTestBase {
+  private Schema inputSchema = Schema.recordOf(
```
albertshau: this can be a private static final

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+      Schema.Field.of("Firstname", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("Lastname", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("Address", Schema.nullableOf(Schema.of(Schema.Type.STRING)))
+    );
+
+    Map<String, String> configMap = new ImmutableMap.Builder<String, String>()
+      .put("keyField", "KeyField")
+      .put("fieldName", "FieldName")
+      .put("fieldValue", "FieldValue")
+      .put("outputFields", "Firstname:," + "Lastname:," + "Address:")
```
albertshau: there is no point in concatenating like this, just have a single string

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+
+    for (GenericRecord record : outputRecords) {
+      key.add(record.get("KeyField").toString());
+      firstName.add(record.get("Firstname").toString());
+      lastName.add(record.get("Lastname").toString());
+      address.add(record.get("Address").toString());
+    }
+
+    Assert.assertEquals(1, outputRecords.size());
+    Assert.assertEquals(ImmutableSet.of("A"), key);
```
albertshau: you don't need to maintain sets for each field. there is one expected record, just compare the record fields directly.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+    Set<String> address = new HashSet<>();
+
+    for (GenericRecord record : outputRecords) {
+      key.add(record.get("KeyField").toString());
+      firstName.add(record.get("Firstname").toString());
+      address.add(record.get("Addr").toString());
+    }
+
+    Assert.assertEquals(2, outputRecords.size());
+    Assert.assertEquals(ImmutableSet.of("A", "B"), key);
```
albertshau: these tests are not accurate enough.  You can't tell that the record with the 'A' key has the correct Firsname and Addr.  You should be comparing sets of records, not sets of field values.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+    Assert.assertEquals(ImmutableSet.of("A"), key);
+    Assert.assertEquals(ImmutableSet.of("ABC"), firstName);
+    Assert.assertEquals(ImmutableSet.of("XYZ"), lastName);
+    Assert.assertEquals(ImmutableSet.of("PQR place near XYZ"), address);
+  }
+
+  @Test
+  public void testDenormalizerWithSelectedAlias() throws Exception {
+    String inputDatasetName = "denormalizer_input3";
+    String outputDatasetName = "denormalizer-output3";
```
albertshau: don't use numbers in names. They are not descriptive and get messed up as the code evolves.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+
+    Assert.assertEquals(1, outputRecords.size());
+    Assert.assertEquals(ImmutableSet.of("A"), key);
+    Assert.assertEquals(ImmutableSet.of("ABC"), firstName);
+    Assert.assertEquals(ImmutableSet.of("XYZ"), lastName);
+    Assert.assertEquals(ImmutableSet.of("PQR place near XYZ"), address);
+  }
+
+  @Test
+  public void testDenormalizerWithSelectedAlias() throws Exception {
```
albertshau: this test seems to be a superset of testDenormalizerWithAllFields. If so, there is no point in having the other test.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    Assert.assertEquals(1, outputRecords.size());
+    Assert.assertEquals(ImmutableSet.of("A"), key);
+    Assert.assertEquals(ImmutableSet.of("ABC"), firstName);
+    Assert.assertEquals(ImmutableSet.of("PQR place near XYZ"), address);
+  }
+
+  @Test
+  public void testDenormalizerWithMultipleKeyFieldValues() throws Exception {
```
albertshau: this test seems to be a superset of the previous two. If so, there is no need for the previous two.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    Assert.assertEquals(2, outputRecords.size());
+    Assert.assertEquals(ImmutableSet.of("A", "B"), key);
+    Assert.assertEquals(ImmutableSet.of("ABC", "ABC1"), firstName);
+    Assert.assertEquals(ImmutableSet.of("PQR place near XYZ", "PQR1 place near XYZ1"), address);
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void testDenormalizerWithWrongKeyFieldName() throws Exception {
```
albertshau: we should not need to deploy a pipeline to test invalid configs. This should go in the config test.

albertshau:
there is no need to put 'JIRA ID - ' in the commit. Just use the jira.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
rashig: This is good idea. But, currently we are accepting alias for the output fields from user itself as per the requirement.
Let me know if this needs to be changed, accordingly i will make the changes in WIKI as well as code and documentation.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
rashig: RowDenormalizerAggregator class extends RecordAggreagtor which further extends BatchAggregator<StructuredRecord, StructuredRecord, StructuredRecord> class, so the overridden method requires the StructuredRecord as input.

If we need to change the method signature, then upper hierarchy needs to be changed.
Please let me know your inputs.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
rashig: Throwing this error when keyfield is empty. It seems like no point in replacing its value in error message.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
rashig: With this If-Else block, i am handling three conditions:
1. If output field is present, then there is no alias for this field and need to use as it is.
2. If else check for alias if present for that output field and use it.
3. If above both conditions fail then user does not want that field to be included in the denormalize output.

If I try to achieve above conditions using tertiary operator, then it will be unreadable.
Please let me know your inputs.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
rashig: RowDenormalizerAggregator class extends RecordAggreagtor which further extends BatchAggregator class, so the overridden method requires the StructuredRecord as input.

If we need to change the method signature, then upper hierarchy needs to be changed.
Please let me know your inputs.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: you can refactor RecordAggregator so that the group key is a generic. Or you could just not extend RecordAggregator, as it doesn't really do much.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: I mean the name of the property, not the value. 'keyField' instead of 'Key Field'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: the way it is now is confusing. How is

```
String fieldName = record.get(conf.getFieldName());
fieldName = outputMappings.containsKey(fieldName) ? outputMappings.get(fieldName) : fieldName;
if (outputFields.contains(fieldName)) {
  builder.set(fieldName, record.get(conf.getFieldValue()));
}
```

unreadable?

albertshau: this is a moot point though, as we should remove all the alias logic anyway.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: we should remove alias. It adds more confusion than value.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+ */
+public class RowDenormalizerConfig extends AggregatorConfig {
+
+  @Description("Name of the column in the input record which will be used to group the raw data. For Example: " +
+    "KeyField")
+  private final String keyField;
+
+  @Description("Name of the column in the input record which contains the names of output schema columns." +
+    "For example:" +
+    "Input Record has columns 'KeyField', 'FieldName', 'FieldValue'" +
```
albertshau: Add a period `'FieldValue'. "`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    "'FieldName' contains 'FirstName', 'LastName','Address'" +
+    "So Output Record will have column names as 'FirstName', 'LastName','Address'.")
```
albertshau: commas should be followed by a space `'LastName', 'Address'`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+public class RowDenormalizerConfig extends AggregatorConfig {
+
+  @Description("Name of the column in the input record which will be used to group the raw data. For Example: " +
+    "KeyField")
+  private final String keyField;
+
+  @Description("Name of the column in the input record which contains the names of output schema columns." +
+    "For example:" +
+    "Input Record has columns 'KeyField', 'FieldName', 'FieldValue'" +
+    "'FieldName' contains 'FirstName', 'LastName','Address'" +
```
albertshau: commas should be followed by a space 'LastName', 'Address'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: Only capitalize the first word in a sentence or proper nouns. Change to `input records have columns ...`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: `output record`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    "So Output Record will have column names as 'FirstName', 'LastName','Address'.")
+  private final String fieldName;
+
+  @Description("Name of the column in the input record which contains the values for output schema columns." +
```
albertshau: Periods should be followed by a space

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+ * Config for RowDenormalizer Aggregator Plugin.
+ */
+public class RowDenormalizerConfig extends AggregatorConfig {
+
+  @Description("Name of the column in the input record which will be used to group the raw data. For Example: " +
+    "KeyField")
+  private final String keyField;
+
+  @Description("Name of the column in the input record which contains the names of output schema columns." +
+    "For example:" +
```
albertshau: need a space after the colon.  Colon should also just be a comma.
`For example,`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    "Input Record has columns 'KeyField', 'FieldName', 'FieldValue'" +
+    "'FieldName' contains 'FirstName', 'LastName','Address'" +
+    "So Output Record will have column names as 'FirstName', 'LastName','Address'.")
+  private final String fieldName;
+
+  @Description("Name of the column in the input record which contains the values for output schema columns." +
+    "For example:" +
```
albertshau: For example, input records have columns ...

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    "Input Record has columns 'KeyField', 'FieldName', 'FieldValue' " +
+    "and 'FieldValue' column contains 'John', 'Wagh', 'NE Lakeside'" +
```
albertshau: and the 'FieldValue'

albertshau: Add a period and space after period. `'NE Lakeside'.`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,142 @@
...
+    "So Output Record will have values for columns 'FirstName', 'LastName','Address' as 'John', 'Wagh', 'NE " +
```
albertshau: the output record

albertshau: space after commas

albertshau:
mostly style comments, and a comment about removing aliasing.

rashig:
c321fe8 is pushed by user rashig. Due to incorrect machine settings, commit is appearing against unknown user. Hence putting a mention in the comment about actual user who has updated the PR.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -71,28 +77,24 @@ public void initialize(BatchRuntimeContext context) throws Exception {
   }
 
   @Override
-  public void groupBy(StructuredRecord record, Emitter<StructuredRecord> emitter) throws Exception {
-    Schema keyFieldSchema = Schema.recordOf("record", Schema.Field.of(keyField, Schema.of(Schema.Type.STRING)));
-    StructuredRecord.Builder builder = StructuredRecord.builder(keyFieldSchema);
-    builder.set(keyField, record.get(keyField));
-    emitter.emit(builder.build());
+  public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
+    emitter.emit(record.get(keyField).toString());
```
albertshau: don't call toString(). You should be validating that the key field is actually a string (at which point you can just cast it), and that it is not null before emitting anything. I'm seeing this across a bunch of commits, but toString() is usually meant for display purposes. If you are using it in the logic, it is probably wrong.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -0,0 +1,140 @@
...
+    "For example, " +
+    "input records have columns 'KeyField', 'FieldName', 'FieldValue' " +
+    "and the 'FieldValue' column contains 'John', 'Wagh', 'NE Lakeside'. " +
+    "So the output record will have values for columns 'FirstName', 'LastName', 'Address' as 'John', 'Wagh', 'NE " +
+    "Lakeside' respectively.")
+  private final String fieldValue;
+
+  @Description("List of the output fields with its alias to be included in denormalized output. This is a comma " +
+    "separated list of key-value pairs, where key-value pairs are separated by a colon ':'. For example, " +
+    "Firstname:fname,Lastname:lname,Address:addr")
```
albertshau: I thought we were removing all the alias logic?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return Schema.recordOf("output.schema", fields);
+  }
+
+  /**
+   * @param inputSchema Validates whether the keyfield, fieldname, and fieldvalue entered by the user is present in
+   *                    the input schema or not.
+   * @return true
+   */
+  private void validateInputFields(Schema inputSchema) {
```
albertshau, line 143: this should also validate that the key field is of type string.  You also need to define what should happen if the key is a nullable string. Do those records just get dropped in that case?  Or are nullable strings not allowed at all?  

Personally I think we should just drop records that have a null key.

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -0,0 +1,60 @@
...
+Example
+-------
+The transform takes input records that have columns keyfield, fieldname, fieldvalue, denormalizes it on the basis of keyfield, and then returns a denormalized table according to the output schema specified by the user.
+
+    {
+          "name": "RowDenormalizer",
+          "type": "batchaggregator",
+          "properties": {
+            "outputFields": "Firstname:,Lastname:,Address:addr",
+            "keyField": "keyfield",
```
albertshau: please use somewhat realistic values for the example.  It is confusing to have the value be just the lowercase of the key.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Builds the output schema, using output fields provided by user.
+   */
+  private Schema initializeOutputSchema() {
+    List<Schema.Field> fields = new ArrayList<>();
+    // Input fields coming from source will be of type String; that's why we are building the output schema as String.
+    fields.add(Schema.Field.of(conf.getKeyField(), Schema.of(Schema.Type.STRING)));
+    for (String outputField : conf.getOutputFields()) {
+      fields.add(Schema.Field.of(outputField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
```
albertshau, line 133: this is wrong, the field type should match whatever type the fieldValue field is

albertshau, line 133: which means you probably need to build the schema for every call to aggregate(), based on whatever comes from the iterator

albertshau, line 133: or if the contract is that these types must always be a string type, then that needs to be documented, that the field value will always be converted to a string.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param inputSchema Validates whether the keyfield, fieldname, and fieldvalue entered by the user is present in
+   *                    the input schema or not.
+   * @return true
+   */
+  private void validateInputFields(Schema inputSchema) {
+    if ((inputSchema.getField(conf.getKeyField())) == null) {
+      throw new IllegalArgumentException(
+        String.format("Keyfield '%s' does not exist in input schema %s", conf.getKeyField(), inputSchema));
+    } else if ((inputSchema.getField(conf.getFieldName())) == null) {
+      throw new IllegalArgumentException(
```
albertshau: need to check that the field is of type string. Similar thing, is this allowed to be null? If so, then a null means that it won't be included in the output record?

Whatever the case, it must be documented in the plugin so that the user knows what to expect.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau, line 143: similarly, need to check that the fieldName is of type string. Similar thing, is this allowed to be null? If so, then a null means that it won't be included in the output record?

Whatever the case, it must be documented in the plugin so that the user knows what to expect.

core-plugins/docs/RowDenormalizer-batchaggregator.md
albertshau: please use somewhat realistic values. It is confusing if the values are the the keys lowercased. For example, you could use 'id', 'attribute', and 'value'.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
albertshau: can we avoid specifying the output fields and their alias in the same property? For example, make this purely a list of output fields, and have an addition property called fieldAliases that is a list of key-value pairs. I think this will be much more clear.

core-plugins/widgets/RowDenormalizer-batchaggregator.json
```diff
@@ -0,0 +1,51 @@
...
+  ],
+  "outputs": [
+    {
+      "name": "schema",
+      "label": "schema",
+      "widget-type": "schema",
+      "widget-attributes": {
+        "schema-types": [
+          "string",
+          "map<string,string>"
```
albertshau: this should not be here, looks like only string types are supported by this plugin.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+      return;
+    }
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    builder.set(conf.getKeyField(), groupKey);
+    while (iterator.hasNext()) {
+      StructuredRecord record = iterator.next();
+      String fieldName = record.get(conf.getFieldName());
+      fieldName = outputMappings.containsKey(fieldName) ? outputMappings.get(fieldName) : fieldName;
+      if (outputFields.contains(fieldName)) {
+        builder.set(fieldName, record.get(conf.getFieldValue()));
```
albertshau: If every output field is supposed to be a string, need to check the type of this field and do the appropriate conversion to string.

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -10,39 +10,60 @@ Use Case
...
 
+Conditions
+----------
+In case a field value is not present, then it will be considered as NULL.
+
+For Example,
+
+If keyfield('id') in the input record is NULL, then that particular row will not be considered.
+
+If fieldname('attribute') and fieldvalue('value') is not present, then denormalized output will have NULL for that
```
albertshau: if the field name is not present, how can there be an output field? Do you mean if the fieldname is null the field will not exist in the output?

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -10,39 +10,60 @@ Use Case
 The transform takes input record that stores a variable set of custom attributes for an entity, denormalizes it on the basis of the key field, and then returns a denormalized table according to the output schema specified by the user.
 The denormalized data is easier to query.
 
+Conditions
+----------
```
albertshau: would be better to move this below the properties section, since you have to understand the properties to understand what this section talks about

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -10,39 +10,60 @@ Use Case
...
-          "type": "batchaggregator",
-          "properties": {
-            "outputFields": "Firstname:,Lastname:,Address:addr",
-            "keyField": "keyfield",
-            "fieldName": "fieldname",
-            "fieldValue": "fieldvalue"
-          }
-        }
-    }
+              "name": "RowDenormalizer",
```
albertshau: is there a reason everything was indented more?

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -51,10 +72,10 @@ For example, suppose the aggregator receives the input record:
 
 Output records will contain all the output fields specified by user:
 
-    +=====================================================+
-    | keyfield  | Firstname   | Lastname   |  addr        |
-    +=====================================================+
-    | joltie    | John        | Wagh       |  NE Lakeside |
-    +=====================================================+
+    +=========================================================+
+    | id        | Firstname   | Lastname   |  Office Address  |
```
albertshau, line 76: lets include the null scenarios in the example

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -128,6 +131,15 @@ private void validateInputFields(Schema inputSchema) {
     } else if ((inputSchema.getField(conf.getFieldValue())) == null) {
       throw new IllegalArgumentException(
         String.format("Fieldvalue '%s' does not exist in input schema %s", conf.getFieldValue(), inputSchema));
+    } else if (!inputSchema.getField(conf.getKeyField()).getSchema().getType().equals(Schema.Type.STRING)) {
```
albertshau: this should check that its a string or a nullable string.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -128,6 +131,15 @@ private void validateInputFields(Schema inputSchema) {
...
+      throw new IllegalArgumentException(
+        String.format("Keyfield '%s' must be of type String", conf.getKeyField()));
+    } else if (!inputSchema.getField(conf.getFieldName()).getSchema().getType().equals(Schema.Type.STRING)) {
```
albertshau: should check that its a string or a nullable string

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -128,6 +131,15 @@ private void validateInputFields(Schema inputSchema) {
...
+      throw new IllegalArgumentException(
+        String.format("Fieldname '%s' must be of type String", conf.getFieldName()));
+    } else if (!inputSchema.getField(conf.getFieldValue()).getSchema().getType().equals(Schema.Type.STRING)) {
```
albertshau: seems like the decision is that the field value will be changed to a string regardless of what its original type is, which means you shouldn't be checking that the original type is a string.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: nit: this is hard to read. Can we have a local variable for the Schema?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -112,12 +118,14 @@ String getFieldValue() {
...
-    for (String field : Splitter.on(',').trimResults().split(outputFields)) {
-      String[] value = field.split(":");
-      if (value.length == 2) {
-        outputFieldMappings.put(value[0], value[1]);
+    if (fieldAliases != null) {
+      for (String field : Splitter.on(',').trimResults().split(fieldAliases)) {
+        String[] value = field.split(":");
+        if (value.length == 2) {
+          outputFieldMappings.put(value[0], value[1]);
+        }
```
albertshau, line 134: else throw an exception with a descriptive message about whats wrong.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -96,12 +102,12 @@ String getFieldValue() {
    */
   Set<String> getOutputFields() {
     Set<String> fields = new HashSet<String>();
```
albertshau: this actually needs to be a LinkedHashSet, or you need to return a List.  Otherwise, when you build the output schema from this set, there is no guarantee what order the fields will come in. Different reducers could output different schemas.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: conversion to string still needs to be done.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
```diff
@@ -96,12 +102,12 @@ String getFieldValue() {
...
+    Map<String, String> outputFieldMappings = getFieldAliases();
     for (String field : Splitter.on(',').trimResults().split(outputFields)) {
-      String[] value = field.split(":");
-      if (value.length == 1) {
-        fields.add(value[0]);
+      if (outputFieldMappings.containsKey(field)) {
```
albertshau, line 109: so outputFields contains the pre-aliased field name? This should be mentioned in the description, otherwise its not clear if the pre-aliased or the aliased name should be used.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
ameyk: I believe this check is done for the input parameters (key, fieldvalue and fieldname). The decision of having String type is for the output that we are going to emit.
Here we are trying to ensure that input that plug-in is receiving is having String type.

albertshau: right but I see in the design that types are supposed to be converted to string, which means you could get an int or a long or something else here. Or is that not the case?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
rashig: As the input record will be of type String only, values coming will also be string.
I tried casting the record values as String while setting values in builder. It gives warning as 'redundant code'.
Let me know if I am missing anything.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerConfig.java
rashig, line 134: We are not throwing any exception because 'fieldAliases' is optional. So we are building map only when we will get the values from user, otherwise we are not doing anything.

albertshau, line 134: the else is for if you try and split on ':' and there are not 2 fields. This is an error.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -70,15 +72,20 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
     keyField = conf.getKeyField();
+    fieldName = conf.getFieldName();
+    fieldValue = conf.getFieldValue();
     outputSchema = initializeOutputSchema();
   }
 
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
     if (record.get(keyField) == null) {
+      if (!record.getSchema().getField(keyField).getSchema().isNullable()) {
```
albertshau: before doing this, you need to check that `record.getSchema().getField(keyField)` is not null. If it is, the key field does not exist in the input record.

core-plugins/docs/RowDenormalizer-batchaggregator.md
```diff
@@ -43,22 +30,35 @@ example, input records have columns 'id', 'attribute', 'value' and the 'value' c
 **numPartitions:** Number of partitions to use when grouping data. If not specified, the execution framework will
 decide on the number to use.
 
+Conditions
+----------
+In case a field value is not present, then it will be considered as NULL.
+
+For Example,
+
+If keyfield('id') in the input record is NULL, then that particular row will not be considered.
```
albertshau: 'will not be considered' is ambiguous. You can say 'then that particular record will be filtered out.'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -94,10 +101,24 @@ public void aggregate(String groupKey, Iterator<StructuredRecord> iterator,
...
       StructuredRecord record = iterator.next();
-      String fieldName = record.get(conf.getFieldName());
-      fieldName = outputMappings.containsKey(fieldName) ? outputMappings.get(fieldName) : fieldName;
-      if (outputFields.contains(fieldName)) {
-        builder.set(fieldName, record.get(conf.getFieldValue()));
+      String outputFieldName = record.get(fieldName);
+      String outputFieldValue = record.get(fieldValue);
+
+      if (outputFieldName == null) {
+        if (!record.getSchema().getField(fieldName).getSchema().isNullable()) {
```
albertshau: similar comment, should check that the field exists first, otherwise the user will only see a null pointer exception.

albertshau: also, whenever you see

```
if (x) {
  if (y) {
    do something
  }
}
```

it can be simplified to:

```
if (x && y) {
  do something
}
```

most IDEs will mark this as a warning.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -122,24 +143,32 @@ private Schema initializeOutputSchema() {
    * @return true
    */
   private void validateInputFields(Schema inputSchema) {
-    if ((inputSchema.getField(conf.getKeyField())) == null) {
+
+    Schema.Field keyField = inputSchema.getField(conf.getKeyField());
+    Schema.Field fieldName = inputSchema.getField(conf.getFieldName());
```
albertshau: I've never really liked the naming for 'fieldName' and 'fieldValue', as they sound like they are just the name and value of a field. Can we change to 'nameField' and 'valueField' (both here and in the plugin config as well)?  I think it is clearer that way, and it also makes it consistent with 'keyField'.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -75,15 +75,16 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
+    nameField = conf.getNameField();
+    valueField = conf.getValueField();
     outputSchema = initializeOutputSchema();
   }
 
   @Override
   public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exception {
     if (record.get(keyField) == null) {
-      if (!record.getSchema().getField(keyField).getSchema().isNullable()) {
+      if (record.getSchema().getField(keyField) != null && !record.getSchema().getField(keyField).getSchema()
```
albertshau: these conditions are different and should not result in the same error message. In the first case, the field is not in the schema at all. In the second, a non-nullable field has a null value.

The second doesn't actually need to be checked, as StructuredRecord.Builder does not let you set null for a non-nullable field.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -101,18 +102,16 @@ public void aggregate(String groupKey, Iterator<StructuredRecord> iterator,
...
-      String outputFieldName = record.get(fieldName);
-      String outputFieldValue = record.get(fieldValue);
+      String outputFieldName = record.get(nameField);
+      String outputFieldValue = record.get(valueField);
 
-      if (outputFieldName == null) {
-        if (!record.getSchema().getField(fieldName).getSchema().isNullable()) {
-          throw new IllegalArgumentException("null value found for non-nullable field " + fieldName);
-        }
+      if (outputFieldName == null && record.getSchema().getField(nameField) != null && !record.getSchema().getField
```
albertshau: same comment, you don't need to check if a non-nullable field has a null value. And actually, it is better to check for these fields in groupBy(), so that if they are missing it will fail earlier.

albertshau: when outputFieldName is null, you need to skip this record. The way it is now, it will result in a null pointer when you call `builder.set(outputFieldName, outputFieldValue)`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -139,36 +138,39 @@ private Schema initializeOutputSchema() {
...
+    } else if (nameField == null) {
       throw new IllegalArgumentException(
-        String.format("Fieldname '%s' does not exist in input schema %s", conf.getFieldName(), inputSchema));
-    } else if (fieldValue == null) {
+        String.format("Namefield '%s' does not exist in input schema %s", conf.getNameField(), inputSchema));
+    } else if (valueField == null) {
       throw new IllegalArgumentException(
-        String.format("Fieldvalue '%s' does not exist in input schema %s", conf.getFieldValue(), inputSchema));
+        String.format("Valuefield '%s' does not exist in input schema %s", conf.getValueField(), inputSchema));
     } else if (!((keyField.getSchema().isNullable() ? keyField.getSchema().getNonNullable().getType() : keyField
```
albertshau: this is really hard to read. Can you create a `Schema.Type keyFieldType` variable instead of doing all this in an if clause? You can also directly compare enums with '=='.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
albertshau: please make sure your tests cover this scenario.

rashig:
Tested RowDenormalizer flow for the case, when the input records have null values either for 3 input fields:
• KeyField
• NameField
• ValueField
Flow is working fine, without occurring null pointer exception.
Added the test case testDenormalizerWithNullValues() in RowDenormalizerAggregatorTest.java, for the same.

sreevatsanraman:
LGTM

PR #266, opened 2016-05-17
Feature/add label to plugin widget

ajainarayanan:
Reopening the same PR #173


elfenheart:
👍  LGTM  if we decides to move forward with this approach

ajainarayanan:
The corresponding CDAP PR was closed. Will close this one too as it depends on the same.

PR #268, opened 2016-05-18
CDAP-6041 fix classloading issues in hive source

albertshau:
Everything should be done using the plugin's classloader as the
context classloader. These hacks can be removed once CDAP-5950 is
fixed.


bdmogal:
lgtm

PR #269, opened 2016-05-18
CDAP-6044 correctly handle nullable fields in CSVParser

albertshau:
Fixed a bug where the CSVParser would always set the value of a
nullable field to null. Fixed the logic so that a nullable field
is only set to null if the parsed value is an empty string, and
the field is not a string or nullable string type.


transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -194,7 +193,25 @@ private StructuredRecord createStructuredRecord(CSVRecord record) {
...
+        Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+        // if the field is a string or a nullable string, set the value to the empty string
+        if (fieldType == Schema.Type.STRING) {
+          builder.set(field.getName(), "");
+        } else if (!isNullable) {
+          // otherwise, error out
+          throw new IllegalArgumentException(String.format(
+            "The %dth field named '%s' is of non-nullable type '%s', " +
+              "but was parsed as an empty string for CSV record '%s'",
+            i + 1, field.getName(), field.getSchema().getType(), record));
```
anwar6953: why i+1?

awholegunch: Why not (eliminate the usage of "th" with numbers, which probably isn't going to look very good):

```
"Field #%d (named '%s') is of non-nullable type '%s', " +
```

awholegunch:
Update copyrights to `2015-2016` as appropriate in the altered files?

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
albertshau: i like that format better, will change. I put i+1 because "0th" seemed weird, but `Field #0` may be better so there is no confusion if its 0 based or 1 based. Will change that as well.

sagarkapare:
LGTM 👍. squash and merge.

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -194,7 +193,25 @@ private StructuredRecord createStructuredRecord(CSVRecord record) {
...
+        Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+        // if the field is a string or a nullable string, set the value to the empty string
+        if (fieldType == Schema.Type.STRING) {
+          builder.set(field.getName(), "");
+        } else if (!isNullable) {
+          // otherwise, error out
+          throw new IllegalArgumentException(String.format(
+            "Field #%d (named '%s') is of non-nullable type '%s', " +
+              "but was parsed as an empty string for CSV record '%s'",
+            i, field.getName(), field.getSchema().getType(), record));
```
gsps1, line 210: field would be 0 indexed in the message. might be better to do `++i` after L#196

sagarkapare, line 210: I think its intentionally shown starting from 0 to avoid confusion. Please see albert's comment in this PR above.

gsps1, line 210: ok. since we also print name, its okay.

sagarkapare:
Build is green here - http://builds.cask.co/browse/HYP-DBT11-1. Merging.

PR #270, opened 2016-05-24
Support for Parsing complex Json using Json Path

ghost:
# JSON Parser Transform
## Description

Parses input Json event into a record. The input Json event could either a map of string field to values or it could be a complex nested Json structure. Plugin allows one to express Json paths for extracting fields from complex nested input Json.
## Parsing Simple Json

Simple Json (which is simply defined as a mapping from key to value) parsing is achieved by specifying the just the output schema fields. The field name in the output schema should be the same as the key in the input Json. The type of the field should also be the same as the input value. No implicit conversions are perform on the Json values.

Following is an example of event that is mapped to the output schema:

<pre>
  {
     "id" : 1000,
     "first" : "Joltie",
     "last" : "Neutrino",
     "email_id" : "joltie.neutrino@gmail.com",
     "address" : "666, Mars Street",
     "city" : "Marshfield",
     "state" : "MR",
     "country" : "Marshyland",
     "zip" : 34553423,
     "planet" : "Mars"
  }
</pre>

So, when output schema is specified it should be specified as follows

| Field | Type | Null |
| --- | --- | --- |
| id | int |  |
| first | string |  |
| last | string |  |
| email_id | string |  |
| address | string |  |
| city | string |  |
| state | string |  |
| country | string |  |
| zip | long |  |

> Note that the field "planet" has been excluded from the output schema meaning that the field would be ignored and
> not processed when the Json event is mapped.
## Nested Json

Parser also allows extracting fields from a complex nested Json. In order to extract fields it uses Json paths similar to XPath expression for XML. To extract fields using expression in Json, this plugin uses 'JsonPath' library. Plugin allows one to define a mapping from output field name to Json path expression that should be applied on input to extract value from the Json event. For example, let's say you have a Json that is as follows:

<pre>
  {
    "employee" : {
      "name" : {
        "first" : "Joltie",
        "last" : "Neutrino"
      },
      "email" : "joltie.neutrino@gmail.com",
      "address" :  {
        "street1" : "666, Mars Street",
        "street2" : "",
        "apt" : "",
        "city" : "Marshfield",
        "state" : "MR",
        "zip" : 34553423,
        "country" : "Marshyland"
      }
    }
  }
</pre>

then one can specify the mapping for extracting following fields from the input Json event
1. first,
2. last,
3. email,
4. street1,
5. city,
6. state, &
7. zip

Then the mappings in the plugin will look as follows:

| Output Field Name | Input Json Path Expression |
| --- | --- |
| first | $.employee.name.first |
| last | $.employee.name.last |
| email | $.employee.email |
| street | $.employee.address.street1 |
| city | $.employee.address.city |
| state | $.employee.address.state |
| zip | $.employee.address.zip |
| country | $.employee.address.country |
### Expression

The "root member object" for parsing any Json is referred to as `$` regardless of whether it's an array or an object. It also uses dot notation or bracket notation for defining the levels of parsing.
For example: `$.employee.name` or `$[employee][name]`
#### Supported Operators

| Operator | Description |
| --- | --- |
| `$` | The root element of the query. |
| `*` | Wildcard |
| `..` | Deep scan |
| `.<name>` | dot notation representing child |
| `[?(<expression>)]` | Filter expression, should be boolean result always |
#### Supported Functions

The functions perform aggregations at the tail end of the path. The functions take output of the expression as input for performing aggregations. The aggregation function can be only be applied where the expression results in an array.

| Function | Type | Desription |
| --- | --- | --- |
| min | double | min value of array of numbers |
| max | double | max value of array of numbers |
| avg | double | average value of array of numbers |
| stddev | double | standard deviation value of array of numbers |
| length | int | length of the array |
## Configuration

| Config | Description |
| --- | --- |
| **field** | Specifies the input field that should be parsed as a CSV Record. |
| **mapping** | Mapping specifying output field name to input Json path for extracting the field. |
| **schema** | Specifies the output schema for the JSON Record. |

### DIFF
--- a/transform-plugins/docs/JSONParser-transform.md
+++ b/transform-plugins/docs/JSONParser-transform.md
@@ -33,10 +33,10 @@ So, when output schema is specified it should be specified as follows
 
 | Field | Type   | Null |
 |-------|--------|------|
-| id    | int    |      |
-| first | string |      |
-| last  | string |      |
-| email_id | string |   |
+| id    | int    | |
+| first | string | |
+| last  | string | |
+| email_id | string | |
 | address | string | |
 | city | string | | 
 | state | string | | 
--- a/transform-plugins/docs/JSONParser-transform.md
+++ b/transform-plugins/docs/JSONParser-transform.md
@@ -129,8 +129,8 @@ an array.
 
 Configuration
 -------------
-| Field | Description |
-|---------------------|-------------|
+| Config | Description |
+|--------|-------------|
 |**field** | Specifies the input field that should be parsed as a CSV Record.|
 |**mapping** | Mapping specifying output field name to input Json path for extracting the field.|
 |**schema** | Specifies the output schema for the JSON Record.|
--- a/transform-plugins/docs/JSONParser-transform.md
+++ b/transform-plugins/docs/JSONParser-transform.md
@@ -1,15 +1,138 @@
 # JSON Parser Transform
 
+## Description
 
-Description
------------
-Parses an input field value as a JSON Object. Each record in the input is parsed as a JSON Object 
-and converted into a Structured Record. The Structured Record can specify particular fields that 
-it is interested in, making projections possible.
+Parses input Json event into a record. The input Json event could either a map of string field to values or it could 
+be a complex nested Json structure. Plugin allows one to express Json paths for extracting fields from complex nested 
+input Json.
 
+## Parsing Simple Json
+
+Simple Json (which is simply defined as a mapping from key to value) parsing is achieved by specifying the 
+just the output schema fields. The field name in the output schema should be the same as the key in the input Json. 
+The type of the field should also be the same as the input value. No implicit conversions are perform on the 
+Json values.
+
+Following is an example of event that is mapped to the output schema:
+<pre>
+  {
+     "id" : 1000,
+     "first" : "Joltie",
+     "last" : "Neutrino",
+     "email_id" : "joltie.neutrino@gmail.com",
+     "address" : "666, Mars Street",
+     "city" : "Marshfield",
+     "state" : "MR",
+     "country" : "Marshyland",
+     "zip" : 34553423,
+     "planet" : "Mars"
+  }
+</pre>
+
+So, when output schema is specified it should be specified as follows
+
+| Field | Type   | Null |
+|-------|--------|------|
+| id    | int    |      |
+| first | string |      |
+| last  | string |      |
+| email_id | string |   |
+| address | string | |
+| city | string | | 
+| state | string | | 
+| country | string | | 
+| zip | long | | 
+
+> Note that the field "planet" has been excluded from the output schema meaning that the field would be ignored and 
+not processed when the Json event is mapped. 
+
+## Nested Json
+
+Parser also allows extracting fields from a complex nested Json. In order to extract fields it uses Json paths similar
+to XPath expression for XML. To extract fields using expression in Json, this plugin uses 'JsonPath' library. Plugin
+allows one to define a mapping from output field name to Json path expression that should be applied on input to extract
+value from the Json event. For example, let's say you have a Json that is as follows:
+
+<pre>
+  {
+    "employee" : {
+      "name" : {
+        "first" : "Joltie",
+        "last" : "Neutrino"
+      },
+      "email" : "joltie.neutrino@gmail.com",
+      "address" :  {
+        "street1" : "666, Mars Street",
+        "street2" : "",
+        "apt" : "",
+        "city" : "Marshfield",
+        "state" : "MR",
+        "zip" : 34553423,
+        "country" : "Marshyland"
+      }
+    }
+  }
+</pre>
+
+then one can specify the mapping for extracting following fields from the input Json event
+ 
+1. first, 
+2. last, 
+3. email, 
+4. street1, 
+5. city, 
+6. state, &
+7. zip
+
+Then the mappings in the plugin will look as follows:
+
+| Output Field Name | Input Json Path Expression |
+|-------------------|----------------------------|
+| first             | $.employee.name.first |
+| last              | $.employee.name.last |
+| email             | $.employee.email |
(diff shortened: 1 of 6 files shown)


transform-plugins/docs/JSONParser-transform.md
```diff
@@ -1,15 +1,138 @@
...
+
+So, when output schema is specified it should be specified as follows
+
+| Field | Type   | Null |
+|-------|--------|------|
+| id    | int    |      |
+| first | string |      |
+| last  | string |      |
+| email_id | string |   |
+| address | string | |
```
anwar6953, line 40: you don't need the (empty) spacing in the last column for rows 5-10?

nitinmotgi:
@anwar6953 can you review it again -- thx

anwar6953:
:o Who is ghost?

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
```diff
@@ -86,12 +169,17 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
     @Description("Input Field")
     private String field;
 
+    @Name("mapping")
+    @Description("Mapping complex JSON to fields using JSON Path expressions")
+    private String mapping;
```
anwar6953, line 176: mark this @Nullable, to indicate that it is optional field? I think that is propagated to the UI then.
@gsps1, can you confirm? I remember you talking about that earlier this week.

transform-plugins/src/test/java/co/cask/hydrator/plugin/JSONParserTest.java
```diff
@@ -82,11 +133,58 @@ public void testSchemaInvalidSchema() throws Exception {
...
+  public void testComplexJSONParsing() throws Exception {
+    final String[] jsonPaths = {
+      "expensive:$.expensive",
+      "bicycle_color:$.store.bicycle.color",
+      "bicycle_price:$.store.bicycle.price",
+      "window:$.store.window"
+    };
+
+    MockEmitter<StructuredRecord> emitter = new MockEmitter<>();
+    JSONParser.Config config = new JSONParser.Config("body", com.google.common.base.Joiner.on(",").join(jsonPaths),
```
anwar6953: use import instead of specifying full package for the usages of the class.

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
```diff
@@ -61,21 +80,85 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
...
+    StructuredRecord.Builder builder = StructuredRecord.builder(outSchema);
+    for (Schema.Field field : fields) {
+      String name = field.getName();
+      if (mapping.containsKey(name)) {
+        String path = mapping.get(name);
+        try {
+          Object value = JsonPath.read(document, path);
+          builder.set(field.getName(), value);
+        } catch (PathNotFoundException e) {
+          LOG.error("Json path '" + path + "' specified for the field '" + name + "' doesn't exist. " +
```
anwar6953, line 149: Should we also include there variable `e` as the last parameter of the LOG.error method? So that the exception is also included in the error log.

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
```diff
@@ -61,21 +80,85 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
...
+    } else {
+      isSimple = false;
+      String[] pathMaps = config.mapping.split(",");
+      for (String pathMap : pathMaps) {
+        String[] mapParts = pathMap.split(":");
+        String field = mapParts[0];
+        String expression = mapParts[1];
+        if (field.isEmpty() && !expression.isEmpty()) {
+          throw new IllegalArgumentException("Json path expression '" + expression +
+                  "' has not output field specified");
```
anwar6953: not --> no

anwar6953:
Just had a few minor comments; remainder looked good to me.

@awholegunch, can you take a look at the docs for this?

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
nitinmotgi: fixed

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
nitinmotgi, line 149: That message will be very cryptic and not intuitive to the user. This was on purpose.

transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
nitinmotgi, line 176: done!

transform-plugins/src/test/java/co/cask/hydrator/plugin/JSONParserTest.java
nitinmotgi: done!

nitinmotgi:
@anwar6953 fixed issues reported. Please review again. Thanks.

transform-plugins/docs/JSONParser-transform.md
```diff
@@ -1,15 +1,138 @@
 # JSON Parser Transform
 
+## Description
 
-Description
------------
-Parses an input field value as a JSON Object. Each record in the input is parsed as a JSON Object 
-and converted into a Structured Record. The Structured Record can specify particular fields that 
-it is interested in, making projections possible.
+Parses input Json event into a record. The input Json event could either a map of string field to values or it could
```
anwar6953: Insert 'be' before 'either'?

anwar6953:
1 last doc update. Remainder LGTM

nitinmotgi:
merging!

PR #274, opened 2016-05-26
Added a keep feature (opposite of drop) to the Projection Transform: CDAP 6067

riwazp7:
The projection transform lets you specify fields to drop, which is nice if you are only dropping a few fields. Sometimes the input is a very wide record and you are only interested in keeping a few of the fields. In that scenario, its easier to specify a list of fields to keep rather than a list of fields to drop.

Added drop field to the projection transform

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
@@ -33,6 +33,7 @@ import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.HashBiMap;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -45,11 +46,11 @@ import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 /**
- * Projection transform that allows dropping, renaming, and converting field types.
+ * Projection transform that allows dropping, keeping, renaming, and converting field types.
  */
 @Plugin(type = "transform")
 @Name("Projection")
-@Description("The Projection transform lets you drop, rename, and cast fields to a different type.")
+@Description("The Projection transform lets you drop, keep, rename, and cast fields to a different type.")
 public class ProjectionTransform extends Transform<StructuredRecord, StructuredRecord> {
   private static final String DROP_DESC = "Comma-separated list of fields to drop. For example: " +
     "'field1,field2,field3'.";
@@ -64,6 +65,8 @@ public class ProjectionTransform extends Transform<StructuredRecord, StructuredR
     "long, float, double, bytes, string). Any simple type can be converted to bytes or a string. Otherwise, a type " +
     "can only be converted to a larger type. For example, an int can be converted to a long, but a long cannot be " +
     "converted to an int.";
+    private static final String KEEP_DESC = "Comma-separated list of fields to keep. For example: " +
+     "'field1,field2,field3'.";
 
   /**
    * Config class for ProjectionTransform
@@ -81,10 +84,14 @@ public class ProjectionTransform extends Transform<StructuredRecord, StructuredR
     @Nullable
     String convert;
 
-    public ProjectionTransformConfig(String drop, String rename, String convert) {
+    @Description(KEEP_DESC)
+    @Nullable
+    String keep;
+    public ProjectionTransformConfig(String drop, String rename, String convert, String keep) {
       this.drop = drop;
       this.rename = rename;
       this.convert = convert;
+      this.keep = keep;
     }
   }
 
@@ -96,6 +103,7 @@ public class ProjectionTransform extends Transform<StructuredRecord, StructuredR
 
   private static final Pattern fieldDelimiter = Pattern.compile("\\s*,\\s*");
   private Set<String> fieldsToDrop = Sets.newHashSet();
+  private Set<String> fieldsToKeep = Sets.newHashSet();
   private BiMap<String, String> fieldsToRename = HashBiMap.create();
   private Map<String, Schema.Type> fieldsToConvert = Maps.newHashMap();
   // cache input schema hash to output schema so we don't have to build it each time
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -127,10 +134,13 @@ public void transform(StructuredRecord valueIn, Emitter<StructuredRecord> emitte
     StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
     for (Schema.Field inputField : inputSchema.getFields()) {
       String inputFieldName = inputField.getName();
-      if (fieldsToDrop.contains(inputFieldName)) {
+      if (!fieldsToKeep.isEmpty()) {
+        if (!fieldsToKeep.contains(inputFieldName)) {
```
albertshau: can be simplified into one if condition, like you've done further down in the code.

albertshau:
nice job. We should also modify the doc at docs/Projection-transform.md and the UI widgets at widgets/Projection-transform.json to support the new config property.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -316,6 +336,8 @@ private Schema getOutputSchema(Schema inputSchema) {
       String inputFieldName = inputField.getName();
       if (fieldsToDrop.contains(inputFieldName)) {
         continue;
+      } else if (!fieldsToKeep.isEmpty() && !fieldsToKeep.contains(inputFieldName)) {
```
gsps1, line 335: so let's say there's a typo in the fields list, say the input schema has the fields "userid, score"  i want to keep score field, but have a typo "scorw".. so we have a field to keep which is not present in the input schema, should we throw an exception in this case?

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -151,10 +161,20 @@ public void transform(StructuredRecord valueIn, Emitter<StructuredRecord> emitte
...
+      throw new IllegalArgumentException("Cannot specify both drop and keep. One should be empty or null");
+    }
+
     if (!Strings.isNullOrEmpty(projectionTransformConfig.drop)) {
       for (String dropField : Splitter.on(fieldDelimiter).split(projectionTransformConfig.drop)) {
         fieldsToDrop.add(dropField);
       }
+    } else if (!Strings.isNullOrEmpty(projectionTransformConfig.keep)) {
+      for (String keepField : Splitter.on(fieldDelimiter).split(projectionTransformConfig.keep)) {
+        fieldsToKeep.add(keepField);
```
chtyim: A cleaner way to write it is:

``` java
Iterables.addAll(fieldsToKeep, Splitter.on(fieldDelimiter).split(projectionTransformConfig.keep));
```

Same for the for loop above.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -127,10 +134,11 @@ public void transform(StructuredRecord valueIn, Emitter<StructuredRecord> emitte
     StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
     for (Schema.Field inputField : inputSchema.getFields()) {
       String inputFieldName = inputField.getName();
-      if (fieldsToDrop.contains(inputFieldName)) {
+      if (!fieldsToKeep.isEmpty() && (!fieldsToKeep.contains(inputFieldName)) {
```
chtyim: Remove the necessary `()` in the second clause.

chtyim: hum... in fact, does it compile?

chtyim: BTW, shouldn't the logic simply be:

``` java
if (!fieldsToKeep.contains(inputFieldName) && fieldsToDrop.contains(inputFieldName)) {
  continue;
}
```

??

riwazp7: Hmm if fieldsToKeep is specified and a field is not in fieldsToKeep, fieldsToDrop wouldn't contain it (both keep and drop fields cannot be specified), so it wouldn't be dropped?

chtyim: Do you have the requirement spec?

riwazp7: There was a jira but it didn't specify clearly. I assume both drop and keep can be empty to not have the user add fields to keep everytime they're not using drop?

riwazp7: So what next?

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
riwazp7, line 335: Not sure. What should I do?

riwazp7:
The build fails in code other than the one I modified.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -64,6 +65,8 @@
     "long, float, double, bytes, string). Any simple type can be converted to bytes or a string. Otherwise, a type " +
     "can only be converted to a larger type. For example, an int can be converted to a long, but a long cannot be " +
     "converted to an int.";
+    private static final String KEEP_DESC = "Comma-separated list of fields to keep. For example: " +
```
sreevatsanraman: what happens if the same field is in drop and keep. which takes precedence?

sreevatsanraman, line 68: I see based on the code below, you can have keep or drop and not both. That should be reflected in the docs here and in drop_desc

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
sreevatsanraman, line 335: The right thing would be throw an exception during configure time, this will throw an error during deployment of the pipeline. This will be applied not only to keep, but also to drop and rename functionalities. @riwazp7 can you open a JIRA for schema validation for projection transform?

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
sreevatsanraman, line 68: I see the doc is updated in the doc update PR. 👍

sreevatsanraman:
👍

riwazp7:
https://issues.cask.co/browse/CDAP-6287

PR #278, opened 2016-06-06
CDAP-6140: binary field name seems to cause trouble with explore

gokulavasan:
`binary` is a reserved keyword. Added backticks to the query and the failing test ETLWorkerTest#testTableSink passes.

JIRA : https://issues.cask.co/browse/CDAP-6140


anwar6953:
LGTM

PR #279, opened 2016-06-06
CDAP-6140 binary is a reserved keyword. hence use backticks in explore

gokulavasan:
Using backticks for binary since it is a keyword

Cherry-picked from the PR merged to develop #278


rohitsinha54:
LGTM

PR #280, opened 2016-06-06
Make hydrator 1.3.2 use cdap 3.4.1 temporarily

rohitsinha54:
Should use cdap 3.4.1 until 3.4.2  is out.


anwar6953:
lgtm. We should bump it to 3.4.2 as a final change to this repo.

PR #271, opened 2016-05-24
Added ability to pass through input record fields into output if field

ghost:
if the names are the same and fix DB lower case issue with CDAP-6079.

PR #281, opened 2016-06-06
Fix conflicts

gokulavasan:
This will cause the build to fail since the TypeConvertor class is not found in release/1.3 and it was not part of the PR #271


gokulavasan:
Closing this since there changes that seems to have gone into release/1.3 and that have not been merged into develop (#269) which causes some Tests to fail in this branch since the logic merged to develop has not captured that (so simply cherry-picking that commit didn't work). Also note that the file `TypeConvertor.java` is present in develop but not in release/1.3. I have no explanation as to how that happened.

PR #284, opened 2016-06-08
GroupBy Aggregate with StdDev function.

nitinmotgi:


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -0,0 +1,79 @@
...
+  private double sum1;
+  private double sum2;
+  private int count;
+  private double stddev;
+
+  public Stddev(String fieldName, Schema fieldSchema) {
+    this.fieldName = fieldName;
+    boolean isNullable = fieldSchema.isNullable();
+    Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+    if (!(fieldType == Schema.Type.INT || fieldType == Schema.Type.LONG ||
```
Nishith: This looks like an often reused check, would it make sense to make it part of the type?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -0,0 +1,79 @@
...
+    sum2 += Math.pow(value, 2);
+    stddev = Math.sqrt(count*sum2 - Math.pow(sum1, 2));
+    count++;
+  }
+
+  @Override
+  public Double getAggregate() {
+    if (count == 0) {
+      // only happens if the field value was always null
+      return null;
```
Nishith: This looks dangerous, should the interface method return an Optional instead?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public void operateOn(StructuredRecord record) {
+    Object val = record.get(fieldName);
+    if (val == null) {
+      return;
+    }
+    double value = ((Number) val).doubleValue();
+    sum1 += value;
+    sum2 += Math.pow(value, 2);
+    stddev = Math.sqrt(count*sum2 - Math.pow(sum1, 2));
```
anwar6953: Why do we need to compute the `stddev` on each record?
Much more efficient to compute it only once (and cache it?) when it's used.

nitinmotgi: This technique allows you to come compute stddev in one pass, otherwise you first have to calculate mean and then do the sum of squared. The interface doesn't allow you to come that way. So, had to find techniques that allow you to compute stddev in a single pass over the numbers.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: What you are saying makes sense, but, that would need an interface change. I am sure @albertshau might have thought of this. So, I am sticking to the interface as it is designed right now.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: Schema doesn't have comparators for specific types right. It should a utility or common function. But, you are right that it gets used multiple times in multiple places. I will move it into a common function.

nitinmotgi:
@Nishith @anwar6953 please review again. Thanks.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator;
+
+import co.cask.cdap.api.data.schema.Schema;
+
+/**
+ * Common functions for aggregation related functionalities.
+ */
+public final class AggrCommon {
```
chtyim: Is this a common util class? If it is, call it `AggregationUtils` and add a private constructor.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Checks if the type of the field is Numeric.
+   * @param fieldType type of the field to be checked.
+   * @return true if numeric, false otherwise.
+   */
+  public static boolean isNumericType(Schema.Type fieldType) {
+    if (fieldType == Schema.Type.INT || fieldType == Schema.Type.LONG ||
```
Nishith: How about
public static boolean isNumericType(Schema.Type fieldType) {
    return (fieldType == Schema.Type.INT || fieldType == Schema.Type.LONG ||  fieldType == Schema.Type.FLOAT ||    fieldType == Schema.Type.DOUBLE);
}

You can keep what you have if you think that's more readable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
```diff
@@ -0,0 +1,39 @@
...
+      fieldType == Schema.Type.FLOAT || fieldType == Schema.Type.DOUBLE) {
+      return true;
+    }
+    return false;
```
chtyim: You can simplify it to `return (field == Schema.Type.INT) || ....` instead of `if`.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Calculates the Standard Deviation
+ */
+public class Stddev implements AggregateFunction<Double> {
+  private final String fieldName;
+  private final Schema outputSchema;
+  private double sum1;
+  private double sum2;
+  private int count;
```
chtyim: Should it be `long`?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -0,0 +1,79 @@
...
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.hydrator.plugin.batch.aggregator.AggrCommon;
+
+/**
+ * Calculates the Standard Deviation
+ */
+public class Stddev implements AggregateFunction<Double> {
+  private final String fieldName;
+  private final Schema outputSchema;
+  private double sum1;
```
chtyim: why the sum is double instead of long?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
chtyim: how about this? http://www.johndcook.com/blog/standard_deviation/

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
chtyim: The interface doesn't have `@Nullable`, so I don't think it's safe to return `null`. How about returning `0` instead?

Nishith: 0 Might be a valid value, perhaps Double.NaN?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
anwar6953: In the current implementation (lines 59-62), it seems like `stddev` should be able to be computed just once at the end.
Just build up `sum1` and `sum2` as you are now (they don't depend on `stddev`, and neither does `stddev` depend on its previous values). And then at the end, perform the statement on line 62.
Am I missing something?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
nitinmotgi: done!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
nitinmotgi: done!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: done!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: values could be real

Nishith:
LGTM.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: Using Double.NaN

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Stddev.java
nitinmotgi: Going to use the new implementation that @chtyim suggested.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggrCommon.java
nitinmotgi: done!

nitinmotgi:
@chtyim and @Nishith one last review please.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator.function;
+
+public final class RunningStats  {
```
anwar6953, line 24: javadoc probably needed, for checkstyle, right?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+    }  else  {
+      newM = oldM + (x - oldM) / entries;
+      newS = oldS + (x - oldM) * (x - newM);
+      // set up for next iteration
+      oldM = newM;
+      oldS = newS;
+    }
+  }
+
+  public long values() {
```
anwar6953: this is more like `numValues` or `numEntries`.
Method name like `values` makes it seem like it returns the actual values.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Variance.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public void beginFunction() {
+    stats = new RunningStats();
+  }
+
+  @Override
+  public void operateOn(StructuredRecord record) {
+    Object val = record.get(fieldName);
+    if (val == null) {
```
anwar6953, line 50: Would a user ever want this to be equivalent to a `push(0)`?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/Properties.java
```diff
@@ -156,7 +156,7 @@
   }
 
   /**
-   * AggrCommon properties for BatchWritable source and sinks
+   * AggregationUtils properties for BatchWritable source and sinks
```
anwar6953: this javadoc seems wrong/confusing.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggregationUtils.java
```diff
@@ -21,19 +21,18 @@
 /**
  * Common functions for aggregation related functionalities.
```
anwar6953: The function(s) defined in this utility class seem to be unrelated to Aggregation.
They just so happen to be used in/from an aggregation plugin.

Maybe not a great name, but `NumericUtils`, `NumberUtils`, or `SchemaUtils` seems more appropriate.
What do you think?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/Properties.java
```diff
@@ -156,7 +156,7 @@
   }
 
   /**
-   * Common properties for BatchWritable source and sinks
+   * AggregationUtils properties for BatchWritable source and sinks
```
anwar6953: this javadoc seems wrong/confusing.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/AggregationUtils.java
```diff
@@ -0,0 +1,38 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator;
+
+import co.cask.cdap.api.data.schema.Schema;
+
+/**
+ * Common functions for aggregation related functionalities.
```
anwar6953, line 22: The function(s) defined in this utility class seem to be unrelated to Aggregation.
They just so happen to be used in/from an aggregation plugin.

Maybe not a great name, but `NumericUtils`, `NumberUtils`, or `SchemaUtils` seems more appropriate.
What do you think?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator.function;
+
+public final class RunningStats  {
+  private long entries = 0l;
+  private double oldM = 0d;
```
Nishith: These names are very hard to read. Expand the abbreviation?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+  private double newM = 0d;
+  private double oldS = 0d;
+  private double newS = 0d;
+
+  public void push(double x) {
+    entries = entries + 1;
+    // See Knuth TAOCP vol 2, 3rd edition, page 232
```
Nishith: Might be helpful to point to an online source too. e.g. https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance  (Online algorithm)

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+  public long values() {
+    return entries;
+  }
+
+  public double mean() {
+    return (entries > 0) ? newM : 0.0;
+  }
+
+  public double variance() {
+    return ( (entries > 1) ? newS /(entries - 1) : 0.0 );
```
Nishith: Space after /

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public double stddev() {
+    return Math.sqrt(variance());
```
Nishith: Maybe add a note saying variance() can never return a negative value.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
nitinmotgi, line 24: done!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
nitinmotgi: removed method. Not needed.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/Variance.java
nitinmotgi, line 50: Fixed the algo and now it doesn't matter.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/Properties.java
nitinmotgi: fixed

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
nitinmotgi: If I knew what they were, I would have. Could you help with this ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
nitinmotgi: done!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
nitinmotgi: new algo, so this is gone.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
Nishith: M -> mean
S -> stddev

anwar6953:
Code changes look good.
There don't seem to be unit or integration test cases for this though.

nitinmotgi:
@anwar6953 in the next PR.

PR #288, opened 2016-06-09
Revert "Make hydrator 1.3.2 use cdap 3.4.1 temporarily"

rohitsinha54:
Reverts caskdata/hydrator-plugins#280

Since 3.4.2 is release now we can depend on 3.4.2


anwar6953:
lgtm

PR #289, opened 2016-06-10
[CDAP-4989] Adding Binary Format to Stream and Kafka format option

elfenheart:
Widget JSON update for https://github.com/caskdata/cdap/pull/5875


bdmogal:
I think this is ok for now, but seems like this should be abstracted somewhere? Shouldn't have to manually enter these options for each such plugin. Should we create an improvement?

bdmogal:
Nothing you can do now, this is ok to merge. Squash before you do that though.

bdmogal:
👍

PR #291, opened 2016-06-16
[CDAP-6123] TPFS sink and source to configure input and output schema…

anew:
… as dataset properties, to allow MapReduce and Spark to use these datasets (Previously, we set only the explore schema but not the schema properties expected by the input/output formats for parquet and avro). Also some refactoring to remove code duplication between avro/parquet and source/sink.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.common;
+
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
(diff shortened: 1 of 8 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
```diff
@@ -0,0 +1,157 @@
...
+    hConf.clear();
+    AvroParquetOutputFormat.setSchema(job, avroSchema);
+    for (Map.Entry<String, String> entry : hConf) {
+      properties.setOutputProperty(entry.getKey(), entry.getValue());
+    }
+  }
+
+  /**
+   * Configure a file set to use Avro file format with a given schema. The schema is parsed
+   * as an Avro schema, validated and converted into a Hive schema. The fileset is configured to use
```
awholegunch: The file set is configured to use [?]

anew: configured to use Avro key input and output format. 

There happens to be a proper name after the line break, but the sentence continues.

awholegunch: No, I was thinking you should use "The file set" instead of "The fileset"... sorry for the confusion!

anew: ok. done

core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Utilities for configuring file sets during pipeline configuration.
+ * TODO (CDAP-6211): Why do we lower-case the schema for Parquet but not for Avro?
+ */
+public class FileSetUtil {
+
+  /**
+   * Configure a file set to use Parquet file format with a given schema. The schema is lower-cased, parsed
+   * as an Avro schema, validated and converted into a Hive schema. The fileset is configured to use
```
awholegunch: file set

chtyim:
LGTM

anew:
Thanks. Squashing commits and merging.

PR #292, opened 2016-06-22
CDAP-6238 Use the configs for default file system and its implementation while configuring the job in plugin.

sagarkapare:
JIRA: https://issues.cask.co/browse/CDAP-6238

### DIFF
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/SourceInputFormatProvider.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/SourceInputFormatProvider.java
@@ -17,8 +17,10 @@
 package co.cask.hydrator.common;
 
 import co.cask.cdap.api.data.batch.InputFormatProvider;
+import co.cask.hydrator.common.batch.JobUtils;
 import com.google.common.collect.ImmutableMap;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.mapreduce.InputFormat;
 
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


sagarkapare:
@chtyim Please help review the PR. Thanks!

hydrator-common/src/main/java/co/cask/hydrator/common/batch/JobUtils.java
```diff
@@ -28,14 +29,24 @@
...
    * Creates a new instance of {@link Job}. Note that the job created is not meant for actual MR
    * submission. It's just for setting up configurations.
    */
   public static Job createInstance() throws IOException {
     Job job = Job.getInstance();
     Configuration conf = job.getConfiguration();
+    // Remember the values of the default filesystem and implementation of the filesystem
+    // for MapR clusters. The values are used by FileInputFormat class while configuring the job.
+    String fsDefaultURI = conf.get(FileSystem.FS_DEFAULT_NAME_KEY);
+    String maprfsImplValue = conf.get(MAPR_FS_IMPLEMENTATION_KEY);
```
nitinmotgi, line 44: MapR specific code ? Isn't this a problem.

chtyim, line 44: For non mapr cluster, the value would be null, and we won't be setting it down below.

BTW @sagarkapare I think it's better to clear these two options (probably through the configurer) when the input provider is set.

sagarkapare:
@chtyim Updated code. Please take a look again. Thanks!

hydrator-common/src/main/java/co/cask/hydrator/common/SourceInputFormatProvider.java
```diff
@@ -38,6 +40,12 @@ public SourceInputFormatProvider(Class<? extends InputFormat> inputFormatClass,
 
   public SourceInputFormatProvider(String inputFormatClassName, Configuration hConf) {
     Map<String, String> config = new HashMap<>();
+    // Unset the values of the default filesystem and implementation of the filesystem
```
anwar6953, line 43: Maybe we should have a method in JobUtils:
`public Map<String, String> getConfiguration(Job job);`
so that the unsetting logic is in the same class as the setting of these parameters.

sagarkapare, line 43: However this would mean changing the `SourceInputFormatProvider` constructor to accept the `Job` instance and this also seems like an additional layer to get the configuration as
`JobUtils.getConfiguration(job)`.
Don't have strong opinion on this. Let me know your thoughts.

sagarkapare:
Build is green here - http://builds.cask.co/browse/HYP-DBT13-1. @chtyim Please take a look. Thanks!

chtyim:
LGTM

sagarkapare:
Squashing and merging.

PR #116, opened 2016-01-20
Remove snapshot version: '1.2.0-SNAPSHOT' --> '1.2.0'

anwar6953:
To get this, I ran the command:
`mvn versions:set -DnewVersion=1.2.0 -DgenerateBackupPoms=false`

PR #293, opened 2016-06-22
XMLParser Transform plugin - CDAP-6177

ananya-bhattacharya:
The XMLParser transform plugin receives the structured record from XMLReader source plugin and creates structured record based on the xpath specified by the user for each field.

Wiki link for Requirements and Design : https://wiki.cask.co/display/CE/XML+Parser

Build status success:
[INFO] BUILD SUCCESS
[INFO] Total time: 2:42.291s
[INFO] Finished at: Wed Jun 22 14:12:20 IST 2016
[INFO] Final Memory: 67M/340M

### DIFF
--- a/transform-plugins/docs/XMLParser-transform.md
+++ b/transform-plugins/docs/XMLParser-transform.md
@@ -1,0 +1,93 @@
+# XML Parser Transform
+
+Description
+-----------
+The XML Parser Transform uses XPath to extract fields from a complex XML event. This plugin should generally be used
+in conjunction with the XML Reader Batch Source. The XML Reader will provide individual events to the XML Parser,
+which will be responsible for extracting fields from the events and mapping them to the output schema.
+
+
+Use Case
+--------
+The transform takes an input record that contain XML events or records, parses it using the specified XPaths and returns
+a structured record according to the specified schema. For example, this plugin can be used in conjunction with the XML
+Reader Batch Source to extract values from XMLNews documents and create structured records which are easier to query.
+
+
+Properties
+----------
+
+**input:** The field in the input record that is the source of the XML event or record.
+
+**encoding:** The source XML character set encoding (default UTF-8).
+
+**xPathMappings:** Mapping of the field names to the XPaths of the XML record. A comma-separated list, each element of
+which is a field name, followed by a colon, followed by an XPath expression. XPath location paths can include predicates
+and supports XPath 1.0.
+Example : ``<field-name>:<XPath expression>``
+
+**fieldTypeMapping:** Mapping of field names in the output schema to data types. Consists of a comma-separated list,
+each element of which is a field name followed by a colon and a type, where the field names are the same as used in the
+xPathMappings, and the type is one of: boolean, int, long, float, double, bytes, or string.
+Example : ``<field-name>:<data-type>``
+
+**processOnError:** The action to take in case of an error.
+                     - "Ignore error and continue"
+                     - "Exit on error" : Stops processing upon encountering an error
+                     - "Write to error dataset" :  Writes the error record to an error dataset and continues
+
+Example
+-------
+
+This example parses an XML record received in the "body" field of the input record following the *xPathMappings* for
+each field name. The output structured record will be created  using the type specified for each field in the
+"fieldTypeMapping". Only years and prices will be passed on for books with a price over 35.00:
+
+       {
+            "name": "XMLParser",
+            "plugin": {
+                "name": "XMLParser",
+                "type": "transform",
+                "label": "XMLParser",
+                "properties": {
+                    "input": "body",
+                    "encoding": "UTF-8",
+                    "xPathMappings": "category://book/@category,
+                                      title://book/title,
+                                      year:/bookstore/book[price>35.00]/year,
+                                      price:/bookstore/book[price>35.00]/price,
+                                      subcategory://book/subcategory",
+                    "fieldTypeMapping": "category:string,title:string,year:int,price:double,subcategory:string",
+                    "processOnError": "Ignore error and continue"
+                }
+            }
+       }
+
+For example, suppose the transform receives these input records:
+
+    +=========================================================================================================+
+    | offset   | body                                                                                         |
+    +=========================================================================================================+
+    | 1        | <bookstore><book category="cooking"><subcategory><type>Continental</type><genre>European     |
+    |          | cuisines</genre></subcategory><title lang="en">Everyday Italian</title><author>Giada De      |
+    |          | Laurentiis</author><year>2005</year><price>30.00</price></book></bookstore>                  |
+    | 2        | <bookstore><book category="children"><subcategory><type>Series</type><genre>fantasy          |
+    |          | literature</genre></subcategory><title lang="en">Harry Potter</title><author>J. K. Rowling   |
+    |          | </author><year>2005</year><price>49.99</price></book></bookstore>                            |
+    +=========================================================================================================+
+
+The output records will contain:
+
+    +=========================================================================================================+
+    | category  | title              | year   |  price  | subcategory                                         |
+    +=========================================================================================================+
+    | cooking   | Everyday Italian   | null   |  null   | <subcategory><type>Continental</type><genre>European|
+    |           |                    |        |         | cuisines</genre></subcategory>                      |
+    | children  | Harry Potter       | 2005   | 49.99   | <subcategory><type>Series</type><genre>fantasy      |
+    |           |                    |        |         | literature</genre></subcategory>                    |
+    +=========================================================================================================+
+
+Here, since the subcategory contains child nodes, the plugin will retrun the complete subcategory node (along with its
+child elements) as string as ``<subcategory><type>Continental</type><genre>European cuisines</genre></subcategory>`` .
+This is to ensure that the plugin returns a sigle XML event for a structured record intead of the two child events:
+ ``<type>Continental</type>`` and ``<genre>European cuisines</genre>``.
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
@@ -1,0 +1,287 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.InvalidEntry;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.Transform;
+import co.cask.cdap.etl.api.TransformContext;
+import com.google.common.base.Strings;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
+
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathFactory;
+
+/**
+ * Parses XML Event using XPath.
+ * This should generally be used in conjunction with the XML Reader Batch Source.
+ */
+@Plugin(type = Transform.PLUGIN_TYPE)
+@Name("XMLParser")
+@Description("Parse XML events based on XPath")
+public class XMLParser extends Transform<StructuredRecord, StructuredRecord> {
+
+  private static final String EXIT_ON_ERROR = "Exit on error";
+  private static final String WRITE_ERROR_DATASET = "Write to error dataset";
+  private final Config config;
+  private Schema outSchema;
+  private Map<String, String> xPathMapping = new HashMap<>();
+
+  // Required only for testing.
+  public XMLParser(Config config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    super.configurePipeline(pipelineConfigurer);
+    outSchema = config.getOutputSchema();
+    validateXpathAndSchema();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(outSchema);
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    outSchema = config.getOutputSchema();
+    xPathMapping = getXPathMapping();
+  }
+
+  /**
+   * Valid if xpathMappings and schema contain the same field names.
+   */
+  private void validateXpathAndSchema() {
+    xPathMapping = getXPathMapping();
+    List<Schema.Field> outFields = outSchema.getFields();
+    // Checks if all the fields in the XPath mapping are present in the output schema.
+    // If they are not a list of fields that are not present is included in the error message.
+    StringBuilder notOutput = new StringBuilder();
+    for (Schema.Field field : outFields) {
+      String fieldName = field.getName();
+      if (!xPathMapping.keySet().contains(field.getName())) {
+        notOutput.append(fieldName + ";");
+      }
+    }
+    if (notOutput.length() > 0) {
+      throw new IllegalArgumentException("Following fields are not present in output schema :" +
+                                           notOutput.toString());
+    }
+  }
+
+  private Map<String, String> getXPathMapping() {
+    Map<String, String> map = new HashMap<>();
+    String[] xpaths = config.xPathFieldMapping.split(",");
+    for (String xpath : xpaths) {
+      String[] xpathmap = xpath.split(":"); //name:xpath[,name:xpath]*
+      String fieldName = xpathmap[0].trim();
+      if (Strings.isNullOrEmpty(fieldName)) {
+        throw new IllegalArgumentException("Field name cannot be null or empty.");
+      } else if (xpathmap.length < 2 || Strings.isNullOrEmpty(xpathmap[1])) {
+        throw new IllegalArgumentException(String.format("XPath for field name %s cannot be null or empty.",
+                                                         fieldName));
+      }
+      map.put(fieldName, xpathmap[1].trim());
+    }
+    return map;
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) {
+    try {
+      InputSource source = new InputSource(new StringReader((String) input.get(config.inputField)));
+      source.setEncoding(config.encoding);
+      DocumentBuilderFactory builderFactory = DocumentBuilderFactory.newInstance();
+      DocumentBuilder documentBuilder = builderFactory.newDocumentBuilder();
+      Document document = documentBuilder.parse(source);
+      XPathFactory xpathFactory = XPathFactory.newInstance();
+      XPath xpath = xpathFactory.newXPath();
+      StructuredRecord.Builder builder = StructuredRecord.builder(outSchema);
+      for (Schema.Field field : outSchema.getFields()) {
+        String fieldName = field.getName();
+        //To evaluate a node, the type(Nodelist or Node) should be known before hand.
+        //Since, the type is not specified from user inputs, taking everything as NodeList and then evaluating.
+        NodeList nodeList = (NodeList) xpath.compile(xPathMapping.get(fieldName)).evaluate(document,
+                                                                                           XPathConstants.NODESET);
+        if (nodeList.getLength() > 1) {
+          throw new IllegalArgumentException("Cannot specify an XPath that is an array");
+        }
+        Node node = nodeList.item(0);
+        //Since all columns have nullable schema extracting not nullable type.
+        Schema.Type type = field.getSchema().getNonNullable().getType();
+        String value = getValue(node, type, fieldName);
+        if (value == null) {
+          builder.set(fieldName, null);
+        } else {
+          builder.set(fieldName, TypeConvertor.get(value, type));
+        }
+      }
+      emitter.emit(builder.build());
+    } catch (Exception e) {
+      switch (config.processOnError) {
+        case EXIT_ON_ERROR:
+          throw new IllegalStateException("Terminating process on error: " + e.getMessage());
+        case WRITE_ERROR_DATASET:
+          emitter.emitError(new InvalidEntry<>(31, e.getStackTrace()[0].toString() + " : " + e.getMessage(), input));
+          break;
+        default:
+          //ignore on error(case "Ignore error and continue")
+          break;
+      }
+    }
+  }
+
+  /**
+   * Get the node value to be parsed into the required format by parseValues().
+   *
+   * @param node      Node from which the text has to be extracted
+   * @param type      schema type to check if it is a nullable string, in case the xpath evaluates to node with children
+   * @param fieldName field name for which the type is to be evaluated
+   * @return node value as string
+   */
+  private String getValue(Node node, Schema.Type type, String fieldName) {
+    if (node != null) {
+      Node firstChild = node.getFirstChild();
+      //If the xpath evaluates to node which contains child element, the output will be an xml record
+      if (firstChild != null && (firstChild.getNodeType() == Node.ELEMENT_NODE || (firstChild.getNextSibling()
+        != null && (firstChild.getNextSibling().getNodeType() == Node.ELEMENT_NODE)))) {
+        if (!type.equals(Schema.Type.STRING)) {
+          throw new IllegalArgumentException(String.format("The xpath returned node which contains child nodes. " +
+                                                             "Cannot convert %s  to type %s", fieldName, type));
+        } else {
+          return nodeToString(node.cloneNode(true));
+        }
+      } else {
+        return node.getTextContent();
+      }
+    } else {
+      return null;
+    }
+  }
+
+  /**
+   * Convert node to string to be returned in the output, for cases which contains child elements.
+   *
+   * @param node node to be converted to string
+   * @return converted node as string
+   */
+  private String nodeToString(Node node) {
+    StringWriter stringWriter = new StringWriter();
+    try {
+      Transformer transformer = TransformerFactory.newInstance().newTransformer();
+      transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
+      transformer.setOutputProperty(OutputKeys.INDENT, "no");
+      transformer.transform(new DOMSource(node), new StreamResult(stringWriter));
+    } catch (TransformerException e) {
+      throw new IllegalArgumentException("Cannot convert node to string. Transformer exception ", e);
+    }
+    return stringWriter.toString();
+  }
+
+  /**
+   * Configuration for the XMLParser transform..
+   */
+  public static class Config extends PluginConfig {
+
+    @Name("input")
+    @Description("The field in the input record that is the source of the XML event or record.")
+    private final String inputField;
+
+    @Description("The source XML character set encoding (default UTF-8).")
+    private final String encoding;
+
+    @Name("xPathMappings")
+    @Description("Mapping of the field names to the XPaths of the XML record. A comma-separated list, each element " +
+      "of which is a field name, followed by a colon, followed by an XPath expression. XPath location paths can " +
+      "include predicates and supports XPath 1.0. Example : <field-name>:<XPath expression>")
+    private final String xPathFieldMapping;
+
+    @Description("Mapping of field names in the output schema to data types. Consists of a comma-separated list, " +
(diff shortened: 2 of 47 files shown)


transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,338 @@
...
+        .set("body", "<bookstore><book category=\"cooking\"><title lang=\"en\">Everyday Italian</title>" +
+          "<author>Giada De Laurentiis</author><year>2005</year><price>30.00</price></book>" +
+          "<book category=\"children\"><title lang=\"en\">Harry Potter</title><author>J K. Rowling</author>" +
+          "<year>2005</year><price>29.99</price></book></bookstore>").build()
+    );
+    MockSource.writeInput(inputManager, input);
+    try {
+      MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+      mrManager.start();
+      mrManager.waitForFinish(5, TimeUnit.MINUTES);
```
anwar6953: put an `Assert.fail(...);` call after the `waitForfinish`.
Currently, when the mapreduce completes (without throwing exception), the assertion on line 335 will not get executed, because no IllegalStateException is caught.

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,338 @@
...
+      Transform<StructuredRecord, StructuredRecord> transform = new XMLParser(config);
+      transform.initialize(new MockTransformContext());
+
+      StructuredRecord inputRecord = StructuredRecord.builder(INPUT)
+        .set("offset", 1)
+        .set("body", "<book category=\"COOKING\"><title lang=\"en\">Everyday Italian</title>" +
+          "<author>Giada De Laurentiis</author><year>2005</year><price>30.00</price></book>").build();
+
+      MockEmitter<StructuredRecord> emitter = new MockEmitter<>();
+      transform.transform(inputRecord, emitter);
```
anwar6953, line 265: put Assert.fail() after the transform call, since we expect that an IllegalStateException is thrown.

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,338 @@
...
+      .addConnection(source.getName(), transform.getName())
+      .addConnection(transform.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(ETLBATCH_ARTIFACT, etlConfig);
+    ApplicationId appId = NamespaceId.DEFAULT.app("XMLReaderTest");
+    try {
+      deployApplication(appId.toId(), appRequest);
+      Assert.fail();
+    } catch (IllegalStateException e) {
```
anwar6953: anything in the error message of this exception that we can check against?

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,314 @@
...
+      this.inputField = inputField;
+      this.encoding = encoding;
+      this.xpathFieldMapping = xpathFieldMapping;
+      this.fieldTypeMapping = fieldTypeMapping;
+      this.processOnError = processOnError;
+    }
+
+    /**
+     * Create output schema from the field name and type value coming from keyvalue-dropdown widget.
+     * Since the xpath acn evaluate to null(when no node is selected), creating nullable schema for all columns.
```
anwar6953: acn --> can

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,314 @@
...
+    String fieldName = field.getName();
+    Schema fieldSchema = field.getSchema();
+    if (Strings.isNullOrEmpty(value)) {
+      if (!fieldSchema.isNullable()) {
+        throw new IllegalArgumentException("NULL value found for non-nullable field : " + fieldName);
+      }
+      return null;
+    }
+    Schema.Type fieldType = fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+    switch (fieldType) {
```
anwar6953: Is this the same as functionality in `TypeConvertor`? If so, can we reuse that?

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,314 @@
...
+   */
+  private String nodeToString(Node node) {
+    StringWriter stringWriter = new StringWriter();
+    try {
+      Transformer transformer = TransformerFactory.newInstance().newTransformer();
+      transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
+      transformer.setOutputProperty(OutputKeys.INDENT, "no");
+      transformer.transform(new DOMSource(node), new StreamResult(stringWriter));
+    } catch (TransformerException te) {
+      System.out.println("nodeToString Transformer Exception");
```
anwar6953: don't write to std out.
use LOG.warn/error/info/debug, as appropriate.
Also, do we want to return empty string in the case that there's a TransformerException? Or should the exception be propagated?

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+    if (notOutput.length() > 0) {
+      throw new IllegalArgumentException("Following fields are not present in output schema :" +
+                                           notOutput.toString());
+    }
+  }
+
+  @Override
+  public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter) {
+    String fieldName;
```
anwar6953: I don't think it helps to keep these variable declarations separate from the assignments.

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,314 @@
...
+    super.configurePipeline(pipelineConfigurer);
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getOutputSchema());
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    outSchema = config.getOutputSchema();
+    validateXpathAndSchema();
+    mapErrorProcessing.put("Ignore error and continue", 1);
```
anwar6953: I think its better to have these three as String constants in this class, instead of encoded in a map like this.
It'll make the switch statement below easier to understand also.

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
anwar6953: Note that `waitForFinish` will return without throwing an exception, even if the launched mapreduce fails, because the program has indeed finished, just with a FAILED status.
You would probably need to check the status via the program's history, to assert that it has indeed failed.

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
ananya-bhattacharya: The error message in this case is a generic message "Expected 200 OK, got 400 Bad Request. Failed to deploy app".Not sure if asserting this is correct.
Changed test case to directly check the exception received from configurePipeline()

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
ananya-bhattacharya: Since the mapreduce completes the job execution; using Assert.fail(), results in assertion error.
Removed try catch and used Assert.assertEquals() to check the "FAILED" status from program history.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
+# XML Parser Transform
+
+Description
+-----------
+The XML Parser uses XPath to extract field from a complex XML Event. This should generally be used in conjunction with
+the XML Reader Batch Source. The XML Source Reader will provide individual events to the XML Parser and the XML Parser
+will be responsible for extracting fields from the events and mapping them to output schema.
```
awholegunch: The XML Parser Transform uses XPath to extract fields from a complex XML event. This plugin should generally be
used in conjunction with the XML Reader Batch Source. The XML Reader will provide individual events to the
XML Parser, which will be responsible for extracting fields from the events and mapping them to the output schema.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+Use Case
+--------
+The transform takes input record that contain xml events/records, parses it using the xpath specified and returns a
+structured record according to the schema specified by user.
+For example, this plugin can be used in conjuction with XML Reader Batch Source to extract values from XMLNews documents
+and to create structured records which will be easier to query.
```
awholegunch: The transform takes an input record that contain XML events or records, parses it using the specified XPaths and
returns a structured record according to the specified schema. For example, this plugin can be used in conjunction with
the XML Reader Batch Source to extract values from XMLNews documents and create structured records which are easier to query.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+Properties
+----------
+
+**input:** Specifies the field in input that should be considered as source of XML event or record.
```
awholegunch: The field in the input that is the source of the XML event or record.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+
+**encoding:** Specifies XML encoding type(default is UTF-8).
```
awholegunch: The source XML character set encoding (default UTF-8).

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+       {
+            "name": "XMLParser",
+            "plugin": {
+                "name": "XMLParser",
+                "type": "transform",
+                "label": "XMLParser",
+                "properties": {
+                    "encoding": "UTF-8",
+                     "processOnError": "Ignore error and continue",
+                      "xpathMappings": "category://book/@category,
```
awholegunch: This does not line up correctly with the line above. The double-quote marks should line up, for "encoding", "processOnError", "xpathMappings", "fieldTypeMapping", and  "input".
Also, the order of these should match the order given in the properties section.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+    |          | <title lang=\"en\">Harry Potter</title><author>J K. Rowling</author><year>2005</year><price> |
+    |          | 49.99</price></book></bookstore>                                                             |
+    +=========================================================================================================+
+
+Output record will contain all the output fields specified by user:
+
+    +=========================================================================================================+
+    | category  | title              | year   |  price  | subcategory                                         |
+    +=========================================================================================================+
+    | cooking   | Everyday Italian   | null   |  null   | <subcategory><type>Continental</type></subcategory> |
```
awholegunch: year here should be 2005, based on the above example.

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+                "label": "XMLParser",
+                "properties": {
+                    "encoding": "UTF-8",
+                     "processOnError": "Ignore error and continue",
+                      "xpathMappings": "category://book/@category,
+                                        title://book/title,
+                                        year:/bookstore/book[price>35.00]/year,
+                                        price:/bookstore/book[price>35.00]/price,
+                                        subcategory://book/subcategory",
+                      "fieldTypeMapping": "category:string,title:string,year:int,price:double,subcategory:string",
```
awholegunch: Why not combine the field type and the xpath mapping into one, rather than re-type all the field names, potentially causing an error?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+Properties
+----------
+
+**input:** Specifies the field in input that should be considered as source of XML event or record.
+
+**encoding:** Specifies XML encoding type(default is UTF-8).
+
+**xpathMappings:** Specifies a mapping from XPath of XML record to field name.
```
awholegunch: Mapping of the field names to the XPaths of the XML record. A comma-separated list, each element of which is a field name, followed by a colon, followed by an XPath location path. Location paths can include predicates and supports XPath 1.0.

transform-plugins/docs/XMLParser-transform.md
awholegunch: Which version of XPath is supported, and what controls this?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }
+       }
+
+For example, suppose the transform recieves the input record:
+
+    +=========================================================================================================+
+    | offset   | body                                                                                         |
+    +=========================================================================================================+
+    | 1        | <bookstore><book category="cooking"><subcategory><type>Continental</type></subcategory>      |
```
awholegunch: Should this not be:

```
<book category=\"cooking\">
```

or should line 68 be

```
<title lang="en">
```

?

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+ * Parses XML Event using XPath.
+ * This should generally be used in conjunction with the XML Reader Batch Source.
+ */
+@Plugin(type = Transform.PLUGIN_TYPE)
+@Name("XMLParser")
+@Description("Parse XML events based on XPath")
+public class XMLParser extends Transform<StructuredRecord, StructuredRecord> {
+
+  private static final String EXIT_ON_ERROR = "Exit on error";
+  private static final String WRITE_ERROR_DATASET = "Write to error dataset";
```
awholegunch, line 64: Maybe these should have names that indicate that they are similar? How does one specify the error dataset?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+
+**xpathMappings:** Specifies a mapping from XPath of XML record to field name.
+
+**fieldTypeMapping:** Specifies the field name as specified in xpathMappings and its corresponding data type.
+The data type can be of following types- boolean, int, long, float, double, bytes, string
+
+**processOnError:** Specifies what happens in case of error.
+                     1. Ignore the error record
+                     2. Stop processing upon encoutering error
+                     3. Write error record to different dataset
```
awholegunch: This needs to match the values in the code, otherwise I don't think this will work:

```
**processOnError:** The action to take in case of an error. One of these strings:

                      - "Ignore error and continue"
                      - "Exit on error":  Stops processing upon encountering an error
                      - "Write to error dataset":  Writes the error record to an error dataset and continues
```

transform-plugins/docs/XMLParser-transform.md
awholegunch: Should this property be **xPathMappings**, since "Path" is capitalized in "XPath"?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+----------
+
+**input:** Specifies the field in input that should be considered as source of XML event or record.
+
+**encoding:** Specifies XML encoding type(default is UTF-8).
+
+**xpathMappings:** Specifies a mapping from XPath of XML record to field name.
+
+**fieldTypeMapping:** Specifies the field name as specified in xpathMappings and its corresponding data type.
+The data type can be of following types- boolean, int, long, float, double, bytes, string
```
awholegunch: Mapping of field names in the output schema to data types. Consists of a comma-separated list, each element of which is a field name followed by a colon and a type, where the field names are the same as used in the **xpathMappings**, and the type is one of boolean, int, long, float, double, bytes, or string.

transform-plugins/docs/XMLParser-transform.md
awholegunch: What does this support? Which expressions and operators?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+                     2. Stop processing upon encoutering error
+                     3. Write error record to different dataset
+
+Example
+-------
+
+This example parses the xml record received in "body" field of the input record, according to the
+xpathMappings specified, for each field name.
+The type output schema will be created, using the type specified for each field in "fieldTypeMapping".
+
```
awholegunch, line 45:
```
This example parses an XML record received in the "body" field of the input record following the
*xpathMappings* for each field name. The output structured record will be created  using the type
specified for each field in the "fieldTypeMapping". Only years and prices will be passed on for books with a price
over 35.00:
```

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+                                        year:/bookstore/book[price>35.00]/year,
+                                        price:/bookstore/book[price>35.00]/price,
+                                        subcategory://book/subcategory",
+                      "fieldTypeMapping": "category:string,title:string,year:int,price:double,subcategory:string",
+                      "input": "body"
+                }
+            }
+       }
+
+For example, suppose the transform recieves the input record:
```
awholegunch:
```
For example, suppose the transform receives these input records:
```

awholegunch: These are two records, right?

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+    +=========================================================================================================+
+    | 1        | <bookstore><book category="cooking"><subcategory><type>Continental</type></subcategory>      |
+    |          | <title lang=\"en\">Everyday Italian</title><author>Giada De Laurentiis</author><year>2005    |
+    |          | </year><price>30.00</price></book></bookstore>                                               |
+    | 2        | <bookstore><book category="children"><subcategory><type>Series</type></subcategory>          |
+    |          | <title lang=\"en\">Harry Potter</title><author>J K. Rowling</author><year>2005</year><price> |
+    |          | 49.99</price></book></bookstore>                                                             |
+    +=========================================================================================================+
+
+Output record will contain all the output fields specified by user:
```
awholegunch:
```
The output records will contain:
```

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+For example, suppose the transform recieves the input record:
+
+    +=========================================================================================================+
+    | offset   | body                                                                                         |
+    +=========================================================================================================+
+    | 1        | <bookstore><book category="cooking"><subcategory><type>Continental</type></subcategory>      |
+    |          | <title lang=\"en\">Everyday Italian</title><author>Giada De Laurentiis</author><year>2005    |
+    |          | </year><price>30.00</price></book></bookstore>                                               |
+    | 2        | <bookstore><book category="children"><subcategory><type>Series</type></subcategory>          |
+    |          | <title lang=\"en\">Harry Potter</title><author>J K. Rowling</author><year>2005</year><price> |
```
awholegunch:
```
<author>J. K. Rowling</author>
```

missing a period after "J"

transform-plugins/docs/XMLParser-transform.md
```diff
@@ -0,0 +1,82 @@
...
+    |          | 49.99</price></book></bookstore>                                                             |
+    +=========================================================================================================+
+
+Output record will contain all the output fields specified by user:
+
+    +=========================================================================================================+
+    | category  | title              | year   |  price  | subcategory                                         |
+    +=========================================================================================================+
+    | cooking   | Everyday Italian   | null   |  null   | <subcategory><type>Continental</type></subcategory> |
+    | children  | Harry Potter       | 2005   | 49.99   | <subcategory><type>Series</type></subcategory>      |
```
awholegunch: Should the output of the `subcategory`field not be:

```
<type>Continental</type>
```

to match the others???

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+        notOutput.append(fieldName + ";");
+      }
+    }
+    if (notOutput.length() > 0) {
+      throw new IllegalArgumentException("Following fields are not present in output schema :" +
+                                           notOutput.toString());
+    }
+  }
+
+  private Map<String, String> getxPathMapping() {
```
awholegunch: should this function be `getXPathMapping`?

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+    Map<String, String> map = new HashMap<>();
+    String[] xpaths = config.xpathFieldMapping.split(",");
+    for (String xpath : xpaths) {
+      String[] xpathmap = xpath.split(":"); //name:xpath[,name:xpath]*
+      String fieldName = xpathmap[0].trim();
+      if (Strings.isNullOrEmpty(fieldName)) {
+        throw new IllegalArgumentException("Field name cannot be null or empty.");
+      } else if (xpathmap.length < 2 || Strings.isNullOrEmpty(xpathmap[1])) {
+        throw new IllegalArgumentException(String.format("Xpath for field name, %s cannot be null or empty.",
```
awholegunch:
```
"XPath for field name %s cannot be null or empty."
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+          builder.set(fieldName, null);
+        } else {
+          builder.set(fieldName, TypeConvertor.get(value, type));
+        }
+      }
+      emitter.emit(builder.build());
+    } catch (Exception e) {
+      switch (config.processOnError) {
+        case EXIT_ON_ERROR:
+          throw new IllegalStateException("Terminating process in error : " + e.getMessage());
```
awholegunch:
```
"Terminating process on error: " + e.getMessage());
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+      XPath xpath = xpathFactory.newXPath();
+      StructuredRecord.Builder builder = StructuredRecord.builder(outSchema);
+      for (Schema.Field field : outSchema.getFields()) {
+        String fieldName = field.getName();
+        //To evaluate a node, the type(Nodelist or Node) should be known before hand.
+        //Since, the type is not specified from user inputs, taking everything as NodeList and then evaluating.
+        NodeList nodeList = (NodeList) xpath.compile(xPathMapping.get(fieldName)).evaluate(document,
+                                                                                           XPathConstants.NODESET);
+        if (nodeList.getLength() > 1) {
+          throw new IllegalArgumentException("Cannot specify XPath that are arrays");
```
awholegunch:
```
"Cannot specify an XPath that is an array")
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
```diff
@@ -0,0 +1,282 @@
...
+        String fieldName = params[0].trim();
+        if (Strings.isNullOrEmpty(fieldName)) {
+          throw new IllegalArgumentException("Field name cannot be null or empty.");
+        } else if (params.length < 2 || Strings.isNullOrEmpty(params[1])) {
+          throw new IllegalArgumentException("Type cannot be null. Please specify type for " + fieldName);
+        }
+        Schema.Field field = Schema.Field.of(fieldName, Schema.nullableOf(Schema.of(Schema.Type.valueOf(
+          params[1].trim().toUpperCase()))));
+        if (fields.contains(field)) {
+          throw new IllegalArgumentException("Field " + fieldName + " already has type specified. Check the mapping.");
```
awholegunch: Which mapping should be checked?

transform-plugins/widgets/XMLParser-transform.json
```diff
@@ -0,0 +1,74 @@
...
+  "metadata": {
+    "spec-version": "1.0"
+  },
+  "configuration-groups": [
+    {
+      "label": "XML Parser",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Input field to Parse as XML Record",
```
awholegunch:
```
Input field to parse as an XML record
```

awholegunch:
I can look at this again...

transform-plugins/docs/XMLParser-transform.md
ananya-bhattacharya: The XPath version is 1.0 as available with Java 7

transform-plugins/docs/XMLParser-transform.md
ananya-bhattacharya: Implementing 2 different widgets (design approach #2) as mentioned in wiki
https://wiki.cask.co/display/CE/XML+Parser+Plugin

The field names in the 2 widgets are validated while publishing the pipeline using configurePipeline()

transform-plugins/docs/XMLParser-transform.md
ananya-bhattacharya: The xPathMappings will accept all the expressions and operators supported by XPath 1.0

transform-plugins/src/main/java/co/cask/hydrator/plugin/XMLParser.java
ananya-bhattacharya, line 64: These are actually 2 different options available in "Error handling"(processOnError) drop-down, and therefore having separate names for them.Please let me know if this should be changed.

transform-plugins/docs/XMLParser-transform.md
ananya-bhattacharya: In cases where the xpath evalutes to a node containing child nodes extracting only child nodes can result in multiple XML events being emitted for a single record. Therefore, including the parent: <subcategory><type>Continental</type>
to keep this as a single valid XML event that can be parsed again.

gokulavasan: Would it be possible to include the multiple XML events case in the above example?

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,315 @@
...
+  @Test
+  public void testInvalidConfig() throws Exception {
+    XMLParser.Config config = new XMLParser.Config("body", "UTF-8", "category://book/@category,title://book/title," +
+      "year:/bookstore/book[price>35.00]/year,price:/bookstore/book[price>35.00]/price,subcategory://book/subcategory",
+                                                   "category:,title:string,price:double,year:int,subcategory:string",
+                                                   "Exit on error");
+
+    MockPipelineConfigurer configurer = new MockPipelineConfigurer(INPUT);
+    try {
+      new XMLParser(config).configurePipeline(configurer);
```
gokulavasan, line 65: Add `Assert.fail()` here if we expect configurePipeline to fail

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,315 @@
...
+        .set("body", "<book category=\"COOKING\"><title lang=\"en\">Everyday Italian</title>" +
+          "<author>Giada De Laurentiis</author><year>2005</year><price>30.00</price></book>").build()
+    );
+    MockSource.writeInput(inputManager, input);
+    MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+    mrManager.start();
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
+
+    DataSetManager<Table> outputManager = getDataset(sinkTable);
+    List<StructuredRecord> outputRecords = MockSink.readOutput(outputManager);
```
gokulavasan, line 115: Add a check to make sure there is only one record from the call `MockSink.readOutput`

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
```diff
@@ -0,0 +1,315 @@
...
+    Assert.assertEquals(31, invalidEntry.getErrorCode());
+    Assert.assertEquals("offset", 1, invalidEntry.getInvalidRecord().get("offset"));
+    Assert.assertEquals("body", "<book category=\"COOKING\"><title lang=\"en\">Everyday Italian</title><author>Giada " +
+      "De Laurentiis</author><year>2005</year><price>30.00</price>" +
+      "</book>", invalidEntry.getInvalidRecord().get("body"));
+  }
+
+  @Test
+  public void testIgnoreError() throws Exception {
+    XMLParser.Config config = new XMLParser.Config("body", "UTF-8", "title:/book/title,author:/book/author," +
```
gokulavasan: Can we use a different encoding (other than UTF-8 - say UTF-16) for one of the test cases? Want to make sure the other encoding values work.

gokulavasan:
Few comments. Please address them. Overall code looks good.

transform-plugins/src/test/java/co/cask/hydrator/plugin/XMLParserTest.java
ananya-bhattacharya, line 115: We are asserting the outputRecords.size() in line #116.
Please let me know if this is okay or we need to check any extra parameters.

gokulavasan, line 115: ah sorry. missed that. that is fine.

gokulavasan:
Code LGTM 👍 . Please squash commits before merging. @awholegunch @ajainarayanan Do you have any comments wrt docs or UI JSON?

transform-plugins/widgets/XMLParser-transform.json
```diff
@@ -0,0 +1,74 @@
...
+  "configuration-groups": [
+    {
+      "label": "XML Parser",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Input field to parse as an XML record",
+          "name": "input",
+          "widget-attributes": {
+            "width": "large"
```
ajainarayanan: We don't actually support width(sizing of widget) today. Not sure if this was in the docs.

ajainarayanan:
One minor comment. Otherwise 👍 LGTM on widget json

nitinmotgi:
@ananya-bhattacharya can you squash the merges

ananya-bhattacharya:
Squashing and Merging. Included the changed for the width property in widget type input

anwar6953:
@sreevatsanraman @ananya-bhattacharya 
Since https://github.com/caskdata/hydrator-plugins/pull/309 was merged, this PR can be closed?
Just a note - https://github.com/caskdata/hydrator-plugins/pull/309 has much fewer lines than this PR, so not sure if some changes are missed.

ananya-bhattacharya:
@anwar6953 
#309 has all the review comments from this PR(#293) incorporated, and the number of lines for all the files for XMLParser, are same in the two PRs.
The extra lines in PR #293 are due to the error that had occurred during squashing.

prinam:
#309 has been merged. Closing this one.

PR #295, opened 2016-06-23
CDAP-6255 Handle the case where a field is null in HDFSSink by writing an empty string in its place

gokulavasan:
JIRA : https://issues.cask.co/browse/CDAP-6255

Modified the test case to check for the case where a field is null.

### DIFF
--- a/hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
+++ b/hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
@@ -50,6 +50,7 @@ import javax.annotation.Nullable;
 @Name("HDFS")
 @Description("Batch HDFS Sink")
 public class HDFSSink extends ReferenceBatchSink<StructuredRecord, Text, NullWritable> {
+  public static final String NULL_STRING = "\0";
   private HDFSSinkConfig config;
 
   public HDFSSink(HDFSSinkConfig config) {
@@ -76,7 +77,9 @@ public class HDFSSink extends ReferenceBatchSink<StructuredRecord, Text, NullWri
   public void transform(StructuredRecord input, Emitter<KeyValue<Text, NullWritable>> emitter) throws Exception {
     List<String> dataArray = new ArrayList<>();
     for (Schema.Field field : input.getSchema().getFields()) {
-      dataArray.add(input.get(field.getName()).toString());
+      Object fieldValue = input.get(field.getName());
+      String data = (fieldValue != null) ? fieldValue.toString() : NULL_STRING;
+      dataArray.add(data);
     }
     emitter.emit(new KeyValue<>(new Text(Joiner.on(",").join(dataArray)), NullWritable.get()));
   }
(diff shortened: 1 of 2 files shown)


hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
```diff
@@ -76,7 +76,8 @@ public void prepareRun(BatchSinkContext context) throws Exception {
   public void transform(StructuredRecord input, Emitter<KeyValue<Text, NullWritable>> emitter) throws Exception {
     List<String> dataArray = new ArrayList<>();
     for (Schema.Field field : input.getSchema().getFields()) {
-      dataArray.add(input.get(field.getName()).toString());
+      String data = (input.get(field.getName()) != null) ? input.get(field.getName()).toString() : "";
+      dataArray.add(data);
```
anwar6953, line 82: should we add null instead of ""?
What's the end difference?
What about not adding it at all?

gokulavasan, line 82: @anwar6953 I am fine with either "null" or "" (empty string). I don't think not adding is right because then the user has no idea which field went missing (say we have fname,lname,city and the lname is null, now if we don't put in anything then there is no way to know if the lname was null or city was null).

anwar6953, line 82: Not adding it at all is definitely not a good idea, as you say.
I can't think of any pro/con of using empty string vs "null".
So, lgtm

anwar6953:
LGTM

hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
```diff
@@ -76,7 +76,8 @@ public void prepareRun(BatchSinkContext context) throws Exception {
   public void transform(StructuredRecord input, Emitter<KeyValue<Text, NullWritable>> emitter) throws Exception {
     List<String> dataArray = new ArrayList<>();
     for (Schema.Field field : input.getSchema().getFields()) {
-      dataArray.add(input.get(field.getName()).toString());
+      String data = (input.get(field.getName()) != null) ? input.get(field.getName()).toString() : "";
```
anwar6953: no need to get the field from the input twice.
Would be better to do the lookup just once, and reuse it from a local var

anwar6953:
minor comment; remainder LGTM

hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
anwar6953, line 82: It just comes down to:
If we use "null", we won't be able to distinguish between string value of "null" vs actual null value of that string.
If we use "", we won't be able to distinguish between string value of "" vs actual null value of that string.

gokulavasan:
@anwar6953 Can you take another look? Thanks for the review!

hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
```diff
@@ -50,6 +50,7 @@
 @Name("HDFS")
 @Description("Batch HDFS Sink")
 public class HDFSSink extends ReferenceBatchSink<StructuredRecord, Text, NullWritable> {
+  public static final String NULL_STRING = new String(new char[]{'\0'});
```
anwar6953: nit - wouldn't this be simpler:

``` java
String NULL_STRING = "\0";
```

anwar6953:
1 minor comment.
Looks good to merge, besides that.

gokulavasan:
@anwar6953 Thanks for the review. Addressed comment. Will merge after build passes

PR #296, opened 2016-06-23
Remove aliasing of batch sinks

anwar6953:
We don't need to alias batch sinks with various things (i.e. path, column family, etc).
We can simply use the configured reference name as its name.

https://issues.cask.co/browse/CDAP-6258

### DIFF
--- a/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCassandraSink.java
+++ b/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCassandraSink.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 Cask Data, Inc.
+ * Copyright © 2015-2016 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -65,8 +65,7 @@ public class BatchCassandraSink
 
   @Override
   public void prepareRun(BatchSinkContext context) {
-    context.addOutput(Output.of(config.referenceName, new CassandraOutputFormatProvider(config))
-                        .alias(config.columnFamily));
+    context.addOutput(Output.of(config.referenceName, new CassandraOutputFormatProvider(config)));
   }
 
   @Override
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3AvroBatchSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3AvroBatchSink.java
@@ -24,6 +24,7 @@ import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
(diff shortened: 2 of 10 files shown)


bdmogal:
Code itself LGTM. why was aliasing done before, though?

anwar6953:
Not sure. It was done as a part of this:
https://github.com/caskdata/hydrator-plugins/pull/235

@gokulavasan ?

gokulavasan:
@anwar6953 @bdmogal Is reference name assured to be unique in a pipeline? Say if we have two HBase Sinks and both of them are 'reference named' as 'HBase'. Will that pipeline fail? If it won't then we could possibly use the alias to differentiate between the two sinks. Same for sources to actually.

anwar6953:
I don't think so. It will be a runtime failure if user specifies non-unique reference names.
However, that's an issue even with the aliasing we are doing now.

gokulavasan:
@anwar6953 With aliasing the assumption is that you won't be writing to the same hdfs path in two different HDFS sinks.

anwar6953:
Right, but someone using Cassandra sink to column family A and to HBase
sink column family A will have the same issue.

gokulavasan:
@anwar6953 Well in that case the reference names are expected to different. HBase, Cassandra. I just explained the reasoning behind using alias. Nothing much to add otherwise. @poornachandra any comments?

anwar6953:
Their reference names would be different, but their alias being the same would cause an issue during runtime.

So basically, with the current aliasing, there's a requirement on the user that reference names be unique, except in cases where the alias is unique.

This PR has the requirement that user defines unique reference names for their sinks. Probably, we should file a JIRA to make this a check during deploy time. Then we can simply reject a pipeline deploy that has non-unique reference names in sinks.

gokulavasan:
@anwar6953 Agreed. Have to confirm if it is ok to have reference names to be unique. I think this issue is even more important for batch sources especially when we have joins.

poornachandra:
Reference names need not be unique in a pipeline. For instance, the user may choose to use the DB name as a reference name for both source and sink. 

As far as I know the only thing that needs to be unique in a pipeline is the label for a stage. Can we use the label as the alias?

anwar6953:
I don't think label is propagated to the plugin APIs. But otherwise, that would work.
Alternatively, we could make a backend change to require reference names be unique.

CuriousVini:
@anwar6953 As BatchSinkContext is a TransformContext, we do expose getStageName() in the context. Can we use that?

anwar6953:
Ooh, good point. We do have that.
We can alias the outputs with their label / stage name, unless anyone sees issues with that.

gokulavasan:
@anwar6953  +1 for aliasing the output with their stageName! It will also make the multi input implementation easier since we can figure out from which stage the data is coming from based on the input name (since they will be one and the same)

bdmogal:
What name does a sink get in a multiple outputs scenario when no alias is specified? Can it default to `stageName`? Doesn't it seem like an overkill to have every sink do this?

anwar6953:
If alias is not specified, an output will get what is currently passed as a
reference name. I think these are not guaranteed to be unique, which is why
gokul suggested we not use it, but instead stage name.

bdmogal:
right, that's what I meant. If an alias is not specified, instead of getting the reference name, can it get the stage name? It seems bad to enforce that every plugin must alias their output if they wanted to use it in a multi-output scenario.

gokulavasan:
@bdmogal @anwar6953 @poornachandra We are planning to override/suffix (in ETLMapReduce code) the alias set by the user to uniquely identify the source/sink.

anwar6953:
@gokulavasan 
Updated the PR to also remove aliasing for batch sources.

gokulavasan:
LGTM 👍

PR #297, opened 2016-06-29
CDAP-6238 Use the configs for default file system and its implementation while configuring the job in plugin.

sagarkapare:
Backport of PR #292


chtyim:
LGTM

PR #300, opened 2016-06-30
fixing deps for SQS plugin

russorat:
fix for https://issues.cask.co/browse/CDAP-6318

PR #301, opened 2016-07-01
removing dup exports

russorat:
fix for @awholegunch 's comment from https://github.com/caskdata/hydrator-plugins/pull/300


wolf31o2:
👍 LGTM

PR #304, opened 2016-07-06
CDAP-6176, Excel Input Reader implementation

romy-khetan:


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+      Sheet sheet = null; // sheet can be used as common for XSSF and HSSF workbook
+      try {
+        Workbook workbook = WorkbookFactory.create(fileIn);
+
+        if (sheetIndx < 0 && sheetName != null && !sheetName.isEmpty()) {
+          sheet = workbook.getSheet(sheetName);
+        } else {
+          sheet = workbook.getSheetAt(sheetIndx);
+        }
+
```
nitinmotgi: remove line

core-plugins/docs/ExcelInputReader-batchsource.md
```diff
@@ -0,0 +1,128 @@
...
+
+**ifErrorRecord:** Specifies the action to be take in case of an error.
+
+**errorDatasetName:** Table name to keep the error record encountered while processing the excel file(s).
+
+
+Example
+-------
+
+This example reads data from a hdfs excel file "hdfs://10.222.73.37:9000/tmp/inventory.xlsx"  and parses it
```
nitinmotgi: What is this ip address for ? can we not just have `<namenode-hostname>:9000`

nitinmotgi: Also it seems like based on sentence that it can only read that file. So, I would recommend removing it completely.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * Reads excel spread sheet, where the keys are the offset in the excel file and the text is the complete record.
+   */
+  public static class ExcelRecordReader extends RecordReader<LongWritable, Text> {
+
+    public static final String END = "END";
+    public static final String MID = "MID";
+    public static final String CELL_SEPERATION = "\t";
```
nitinmotgi: Should we pick something that's not printable, but still on ASCII table ? The cell data might have a tab no ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+      Iterator<Cell> cellIterator = row.cellIterator();
+
+      sb.append(row.getRowNum()).append(CELL_SEPERATION);
+      sb.append(file).append(CELL_SEPERATION);
+      sb.append(row.getSheet().getSheetName()).append(CELL_SEPERATION);
+
+      int count = fileRowLimitMap.get(file.getName());
+      if (fileRowLimitMap.get(file.getName()) - 1 == 0 || !rows.hasNext()) {
+        sb.append(END).append(CELL_SEPERATION);
+
```
nitinmotgi: Remove line

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+          case Cell.CELL_TYPE_BOOLEAN:
+            sb.append(colName)
+              .append(COLUMN_SEPERATION).append(cell.getBooleanCellValue()).append(CELL_SEPERATION);
+            break;
+
+          case Cell.CELL_TYPE_NUMERIC:
+            sb.append(colName)
+              .append(COLUMN_SEPERATION).append(cell.getNumericCellValue()).append(CELL_SEPERATION);
+            break;
+
```
nitinmotgi: Remove line

core-plugins/widgets/ExcelInputReader-batchsource.json
```diff
@@ -17,7 +17,7 @@
         },
         {
           "widget-type": "textbox",
-          "label": "Memory Table",
+          "label": "File Tracking Table",
```
nitinmotgi, line 20: Track Files

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+      // Reads the excel file, selects the sheet to be read.
+
+      String sheetName = job.get(SHEET_NAME);
+      int sheetIndx = job.getInt(SHEET_NO, -1);
+
+      Sheet sheet = null; // sheet can be used as common for XSSF and HSSF workbook
+      try {
+        Workbook workbook = WorkbookFactory.create(fileIn);
+
+        if (sheetIndx < 0 && sheetName != null && !sheetName.isEmpty()) {
```
nitinmotgi: So if we are providing both sheet name and sheet index, then sheet name overrides sheet index. I could be wrong, but I thought we wanted to override it the other way.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+    private Path file;
+
+    // Specifies the row index.
+    private long rowIdx;
+
+    // Specifies last row num.
+    private int lastRowNum;
+
+    // Keeps row limits for each excel file(s)
+    private Map<String, Integer> fileRowLimitMap = new HashMap<>();
```
nitinmotgi: Why do we need this ? Doesn't each mapper get only one file, we don't need to keep a map.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -0,0 +1,251 @@
...
+        if (rows.hasNext()) {
+          rowIdx = 1;
+        }
+        rows.next();
+      }
+    }
+
+    @Override
+    public boolean nextKeyValue() throws IOException, InterruptedException {
+      if (!rows.hasNext() || (fileRowLimitMap.get(file.getName()) == 0)) {
```
nitinmotgi: Isn't check for fileRowLimitMap redudant as rows.hasNext() wouldn't have anything

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  @Override
+  public void transform(KeyValue<LongWritable, Object> input, Emitter<StructuredRecord> emitter) throws Exception {
+
+    getOutputSchema();
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    String inputValue = input.getValue().toString();
+
+    String[] excelRecords = inputValue.split("\t");
```
nitinmotgi: You want to use constant to represent \t

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+                                       " . Terminating processing", new Throwable());
+      }
+    }
+
+    filePrevRowNumMap.put(fileName, prevRowNum);
+
+    Map<String, String> excelColumnValueMap = new HashMap<>();
+
+    for (String columns : excelRecords) {
+      String[] columnValueArray = columns.split("\r");
```
nitinmotgi: Use constant to represent \r

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+
+    String[] excelRecords = inputValue.split("\t");
+
+    String fileName = excelRecords[1];
+    String sheetName = excelRecords[2];
+    String ifEndRow = excelRecords[3];
+
+    int prevRowNum = Integer.parseInt(excelRecords[0]);
+
+    if (filePrevRowNumMap.containsKey(fileName)) {
```
nitinmotgi: Not sure why you need the map here as the mapper is only processing one file right ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+      if (prevRowNum - filePrevRowNumMap.get(fileName) > 1 &&
+        excelInputreaderConfig.terminateIfEmptyRow.equalsIgnoreCase("true")) {
+        throw new ExecutionException("Encountered empty row while reading Excel file :" + fileName +
```
nitinmotgi: Will this terminate processing and fail the job or terminate processing and is successful. Looks like this will fail the job.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+      }
+    }
+
+    filePrevRowNumMap.put(fileName, prevRowNum);
+
+    Map<String, String> excelColumnValueMap = new HashMap<>();
+
+    for (String columns : excelRecords) {
+      String[] columnValueArray = columns.split("\r");
+      if (columnValueArray.length > 1) {
```
nitinmotgi: if there is only one column, this would be a problem right as you are comparing it be greater than 1.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+        throw new ExecutionException("Encountered empty row while reading Excel file :" + fileName +
+                                       " . Terminating processing", new Throwable());
+      }
+    }
+
+    filePrevRowNumMap.put(fileName, prevRowNum);
+
+    Map<String, String> excelColumnValueMap = new HashMap<>();
+
+    for (String columns : excelRecords) {
```
nitinmotgi: If excelRecords has more rows, then it would override excelColumnMap. Down below you use this to write to structured record and it seems wrong. This logic needs a re-write assuming that excelRecords can have more than 1 row in it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+        if (columnMapping.containsKey(columnName)) {
+          excelColumnValueMap.put(columnMapping.get(columnName), columnValue);
+        } else {
+          excelColumnValueMap.put(columnName, columnValue);
+        }
+      }
+    }
+
+    try {
+      for (Schema.Field field : outputSchema.getFields()) {
```
nitinmotgi, line 189: Ok! this logic looks totally wrong to me. Here is why : You are iterating over excelRecords. I am assuming excelRecords has more than 1 record (row) and the excelColumnValueMap gets overwritten everytime when it iterates through the excelRecords. Now, you take that and write to StructureRecord down here. So, it will write only the last record value to the output. This section needs to be within excelRecords loop. If all this makes sense, please re-write this section.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+          builder.convertAndSet(fieldName, excelColumnValueMap.get(fieldName));
+        } else {
+          builder.set(fieldName, NULL);
+        }
+      }
+
+      builder.set(FILE, new Path(fileName).getName());
+      builder.set(SHEET, sheetName);
+
+      if (ifEndRow.equalsIgnoreCase(END)) {
```
nitinmotgi, line 203: If emitter fails then you should not record that the file has been processed right ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+          errorRecordBuilder.set(FILE, fileName);
+          errorRecordBuilder.set(SHEET, sheetName);
+          errorRecordBuilder.set(RECORD, inputValue);
+          writeErrorRecordToDataset(errorRecordBuilder.build());
+          break;
+        default:
+          //ignore on error
+          break;
+      }
+
```
nitinmotgi: Remove line

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+    }
+
```
nitinmotgi: Remove line

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  /**
+   * Writes proceesed file name to memory table.
+   * @param fileName
+   */
+  private void trackProcessedFiles(String fileName) {
```
nitinmotgi: Why do we need this two line function and it's used in only one place. Makes it more readable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+    KeyValueTable processedFileMemoryTable = batchRuntimeContext.getDataset(excelInputreaderConfig.memoryTableName);
+    processedFileMemoryTable.write(fileName, String.valueOf(System.currentTimeMillis()));
+  }
+
+  /**
+   * Writes error record to error dataset.
+   * @param errorRecord in the form of {@link StructuredRecord}
+   * throws {@link IOException}
+   */
+  private void writeErrorRecordToDataset(StructuredRecord errorRecord) throws IOException {
```
nitinmotgi: It's same with this.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -0,0 +1,457 @@
...
+  /**
+   * Returns list of all the processed file names which are kept in memory table.
+   * @param batchSourceContext
+   * @return processedFiles
+   */
+  private List<String> getAllProcessedFiles(BatchSourceContext batchSourceContext) {
+    KeyValueTable table = batchSourceContext.getDataset(excelInputreaderConfig.memoryTableName);
+    List<String> processedFiles = new ArrayList<>();
+    try (CloseableIterator<KeyValue<byte[], byte[]>> filesIterable = table.scan(null, null)) {
+      while (filesIterable.hasNext()) {
```
nitinmotgi: Will the key value table keep growing or is there a TTL set on this. After a while this table will become large and it would take a long time to return. We need to be able to set TTL and we should ask for this configuration from user right ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
romy-khetan: Replaced input widget for accepting sheet name or sheet number. User can either specify sheet name or sheet index. By default, sheet number will be selected as input.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
romy-khetan: fileRowLimitMap check has been provided to check the user specified row limits on the excel sheet.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
romy-khetan: This will fail the job as we are considering "termination in case of empty row" input as a negative case, so that user should be able to know that processing has not been done completely and it is terminated.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
romy-khetan: This array contains a single column-value pair.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
romy-khetan, line 189: excelRecords contains just one row. Renamed this variable for better readability.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
romy-khetan: Added an input widget to take TTL from user.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
romy-khetan, line 203: Yes. Moved the file tracking code after emitter.

romy-khetan:
Updated PR with code review comments

nitinmotgi:
LGTM 👍

PR #305, opened 2016-07-08
Initial Draft for SolrSearch sink plugin.[CDAP 6178]

rashig:
SolrSearch Sink Plugin to write data to a Sinlge Node Solr or to SolrCloud. For design and implementation, please refer:
https://wiki.cask.co/display/CE/Solr+Search+or+Cloudera+Search


solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,119 @@
...
+  public String getIdField() {
+    return idField;
+  }
+
+  /**
+   * Creates a map for input field and its mapping for output field, if present.
+   *
+   * @return Map of input fields and its mapping
+   */
+  public Map<String, String> getOutputFieldMappings() {
```
anwar6953: Is this logic copied from somewhere? If so, please refactor it to some common place.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,199 @@
...
+        String.format("Idfield '%s' does not exist in the input schema %s", config.getIdField(), inputSchema));
+    }
+  }
+
+  /**
+   * Validates whether the CDAP data types coming from the input schema are compatible with the Solr data types or not.
+   *
+   * @param inputSchema
+   */
+  private void validateInputFieldsDataType(Schema inputSchema) {
```
anwar6953: seems like we can refactor some of the methods from this class and the batch sink. The implementation of some methods seem to be duplicate.
Consider a Solr utility class, which both sinks can use.

solrsearch-plugins/src/test/java/co.cask.hydrator.plugin/SolrSearchSinkTest.java
```diff
@@ -0,0 +1,291 @@
...
+      StructuredRecord.builder(inputSchema).set("rowid", "1").set("firstname", "Brett").set("lastname", "Lee").set
+        ("address", "NE lake side").set("pincode", 480001).build(),
+      StructuredRecord.builder(inputSchema).set("rowid", "2").set("firstname", "John").set("lastname", "Ray").set
+        ("address", "SE lake side").set("pincode", 480002).build()
+    );
+    MockSource.writeInput(inputManager, input);
+
+    MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+    mrManager.start();
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
```
anwar6953, line 164: I think you should validate the data in solr, similar to how you do in the realtime test case

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,216 @@
...
+  private File tempDir;
+
+  public SolrSearchSink(SolrSearchSinkConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void prepareRun(BatchSinkContext batchSinkContext) throws Exception {
+    // Setting FileOutputFormat.setOutputPath in order to resolve the 'Output directory not set' error.
+    tempDir = Files.createTempDir();
```
anwar6953: its performing all IO to this temp dir?
That seems quite inefficiency.
Instead, can you use NullOutputFormat?
https://hadoop.apache.org/docs/r2.4.1/api/org/apache/hadoop/mapreduce/lib/output/NullOutputFormat.html

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  @Override
+  public void destroy() {
+    solrClient.shutdown();
+  }
+
+  @Override
+  public void onRunFinish(boolean succeeded, BatchSinkContext context) {
+    File directory = new File(tempDir.getAbsolutePath() + "/temp");
```
anwar6953: this deletion can be removed if you use:
https://hadoop.apache.org/docs/r2.4.1/api/org/apache/hadoop/mapreduce/lib/output/NullOutputFormat.html

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  /**
+   * Verifies whether the Solr server is running or not. Also, validates whether the collection provided by user is
+   * registered with the Solr server or not.
+   */
+  private void verifySolrConfiguration() {
+    SolrClient client = getSolrConnection();
+    try {
+      client.ping();
```
anwar6953: is it possible to ping without specifying a collection?
That way, we can know whether the issue is the host/port or the collection not existing?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
rashig: Kept this method in config class, as it is related with SolrSearch sink config and utilized by both batch and realtime SolrSearch plugins.
Should I move this to SolrUtility class ?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: While creating SolrClient object, need to pass the name of a collection when its a Single Node Solr mode. Also, during SolrCloud mode, to make sure that indexing requests should go to the specific collection only, I am adding the collection name while establishing the connection.
Also, I was getting different exceptions (during single node and SolrCloud mode) when Solr server is not running or collection doesn't exists. That's why I kept the message generic.

I will try to separate out both: the server not running and collection not exists cases, and I will update it accordingly.

solrsearch-plugins/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.solr</groupId>
+      <artifactId>solr-solrj</artifactId>
+      <version>5.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-logging</groupId>
+      <artifactId>commons-logging</artifactId>
```
albertshau, line 45: why is commons logging needed?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+
+import co.cask.cdap.api.data.schema.Schema;
+import org.apache.solr.client.solrj.SolrClient;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
+import org.apache.solr.client.solrj.impl.HttpSolrClient;
+
+/**
+ * Utility class for SolrSearchSink and RealtimeSolrSearchSink plugins.
+ */
+public class SolrSearchSinkUtility {
```
albertshau: you can put these methods in the SolrSearchSinkConfig class instead of having a utility class where all methods take the config as an argument.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final String SINGLE_NODE_SOLR_MODE = "Single Node Solr";
```
albertshau: lets make these 'singlenode' and 'cloud' instead of multi word values.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+      solrClient = solrCloudClient;
+    }
+    return solrClient;
+  }
+
+  /**
+   * Validates whether the host entered for Single Node Solr instance is proper or not.
+   */
+  public static void validateSolrConnectionString(SolrSearchSinkConfig config) {
+    if (config.getSolrMode().equals(SINGLE_NODE_SOLR_MODE) && config.getSolrHost().contains(",")) {
```
albertshau: do we actually need a mode setting then? Can we just assume multiple hosts means cloud and one host means singlenode?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+        switch (field.getSchema().getType()) {
+          case BOOLEAN:
+          case INT:
+          case LONG:
+          case FLOAT:
+          case DOUBLE:
+          case STRING:
+            break;
+          default:
+            throw new IllegalArgumentException(
```
albertshau: this means nullable fields are not supported? I would think the id field can't be nullable but other fields can be

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Validates whether the Id field provided by user is present in the input schema or not.
+   *
+   * @param inputSchema
+   */
+  public static void validateIdField(Schema inputSchema, SolrSearchSinkConfig config) {
+    if (inputSchema != null && inputSchema.getField(config.getIdField()) == null) {
+      throw new IllegalArgumentException(
```
albertshau: also need to check that the id field is not nullable.  I would imagine there are also type restrictions on what can be an id right?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,113 @@
...
+    // Setting outputformat class in order to resolve the 'Output directory not set' error.
+    Job job = batchSinkContext.getHadoopJob();
+    job.setOutputFormatClass(NullOutputFormat.class);
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    SolrSearchSinkUtility.validateSolrConnectionString(config);
+    SolrSearchSinkUtility.verifySolrConfiguration(config);
```
albertshau: at configuration time, we should only be checking static things, which means things that will never change.

For example, checking whether the host is a valid url is something you can check, because host will never change. But checking if you can ping the solr server isn't, because the server may happen to be down during pipeline deployment, or the index may not be created yet. None of these scenarios mean the pipeline is invalid, just that the pipeline can't run right at this moment.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,113 @@
...
+    for (Schema.Field field : structuredRecord.getSchema().getFields()) {
+      solrFieldName = field.getName();
+      if (outputFieldMap.containsKey(solrFieldName)) {
+        document.addField(outputFieldMap.get(solrFieldName), structuredRecord.get(solrFieldName));
+      } else {
+        document.addField(solrFieldName, structuredRecord.get(solrFieldName));
+      }
+    }
+    solrClient.add(document);
+    solrClient.commit();
```
albertshau, line 104: this type of pattern is ok for the realtime sink, but not for a batch sink.  Batch sink should be using a Hadoop output format like the SolrOutputFormat included in the solr project.

The purpose of the transform method is to convert the input into key-values that the OutputFormat understands how to write.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public RealtimeSolrSearchSink(SolrSearchSinkConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    SolrSearchSinkUtility.validateSolrConnectionString(config);
+    SolrSearchSinkUtility.verifySolrConfiguration(config);
```
albertshau: similar comment in the batch sink about what to check and what not to check.

Also, to avoid duplicate code, you can move the validation methods into the SolrSearchSinkConfig class.

albertshau:
the realtime sink looks good, but the batch sink needs to be reworked

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,71 @@
...
+
+**solrMode:** Solr mode to connect to. For example, Single Node Solr or SolrCloud.
+
+**solrHost:** The hostname and port for the Solr server. For example, localhost:8983 if Single Node Solr or
+zkHost1:2181,zkHost2:2181,zkHost3:2181 for SolrCloud.
+
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**idField:** Field that will determine the unique id for the document to be indexed. It should match a fieldname
+in the structured record of the input.
```
awholegunch, line 29: "It should match a field name": should that be "must match"?

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,71 @@
...
+**referenceName:** This will be used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**solrMode:** Solr mode to connect to. For example, Single Node Solr or SolrCloud.
+
+**solrHost:** The hostname and port for the Solr server. For example, localhost:8983 if Single Node Solr or
+zkHost1:2181,zkHost2:2181,zkHost3:2181 for SolrCloud.
+
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**idField:** Field that will determine the unique id for the document to be indexed. It should match a fieldname
```
awholegunch: fieldname -> field name

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,71 @@
...
+The Solr server should be running prior to creating the application.
+
+All the fields that user wants to index into the Solr server, should be properly declared and defined in Solr's
+schema.xml file. The Solr server schema should be properly defined prior to creating the application.
+
+If idField('id') in the input record is NULL, then that particular record will be filtered out.
+
+Example
+-------
+This example connects to 'Sinlge Node Solr' server, which is running locally( at default port 8983), and writes the
```
awholegunch:
```
This example connects to a 'Single Node Solr' server, running locally at the default port of 8983, and writes the
```

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
```diff
@@ -0,0 +1,116 @@
...
+    String urlString;
+    SolrClient solrClient = null;
+    if (config.getSolrMode().equals(SINGLE_NODE_SOLR_MODE)) {
+      urlString = "http://" + config.getSolrHost() + "/solr/" + config.getCollectionName();
+      solrClient = new HttpSolrClient(urlString);
+    } else if (config.getSolrMode().equals(SOLR_CLOUD_MODE)) {
+      CloudSolrClient solrCloudClient = new CloudSolrClient(config.getSolrHost());
+      solrCloudClient.setDefaultCollection(config.getCollectionName());
+      solrClient = solrCloudClient;
+    }
```
awholegunch: Shouldn't an error get thrown if one of the allowed modes isn't passed?

solrsearch-plugins/src/test/java/co.cask.hydrator.plugin/SolrSearchSinkTest.java
```diff
@@ -0,0 +1,376 @@
...
+    NamespaceId.DEFAULT.artifact("etlrealtime", CURRENT_VERSION.getVersion());
+  private static final ArtifactSummary REALTIME_APP_ARTIFACT =
+    new ArtifactSummary(REALTIME_APP_ARTIFACT_ID.getArtifact(), REALTIME_APP_ARTIFACT_ID.getVersion());
+  private static final ArtifactRange REALTIME_ARTIFACT_RANGE = new ArtifactRange(Id.Namespace.DEFAULT, "etlrealtime",
+                                                                                 CURRENT_VERSION, true,
+                                                                                 CURRENT_VERSION, true);
+  private static final ArtifactRange BATCH_ARTIFACT_RANGE = new ArtifactRange(Id.Namespace.DEFAULT, "etlbatch",
+                                                                              CURRENT_VERSION, true,
+                                                                              CURRENT_VERSION, true);
+  private static final String SINGLE_NODE_SOLR_MODE = "Single Node Solr";
```
awholegunch: Shouldn't this string have no spaces?

```
"SingleNodeSolr";
```

awholegunch: Also, this is now defined twice, and the other is above. One place only?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
awholegunch: This is also defined in `SolrSearchSinkTest`

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,119 @@
...
+ * Config class for SolrSearchSink and RealtimeSolrSearchSink plugins.
+ */
+public class SolrSearchSinkConfig extends ReferencePluginConfig {
+
+  @Description("Solr mode to connect to. For example, Single Node Solr or SolrCloud.")
+  private final String solrMode;
+
+  @Description("The hostname and port for the Solr server seperated by colon. For example, localhost:8983 for " +
+    "Single Node Solr or comma-seperated list of hostname and port, zkHost1:2181,zkHost2:2181,zkHost3:2181 " +
+    "for SolrCloud.")
```
awholegunch, line 43: seperated -> separated

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Config class for SolrSearchSink and RealtimeSolrSearchSink plugins.
+ */
+public class SolrSearchSinkConfig extends ReferencePluginConfig {
+
+  @Description("Solr mode to connect to. For example, Single Node Solr or SolrCloud.")
+  private final String solrMode;
+
+  @Description("The hostname and port for the Solr server seperated by colon. For example, localhost:8983 for " +
```
awholegunch: separated by a colon.

solrsearch-plugins/pom.xml
rashig, line 45: Code works fine when run through JUnit test but while deploying jar on standalone server it gives following error:
Unexpected error: request=PUT /v3/namespaces/default/apps/Test user=<null>: java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory and doesn’t allow pipeline to be published.
That’s why I have to explicitly include commons-logging jar in SolrSearch module.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
rashig: There is a possibility that user can run SolrCloud on one zookeeper (one host) only, may be for testing purposes. In this case, then it will be difficult to identify whether the single host is for SolrCloud or Single Node. That’s why I kept the mode setting.

Assuming that if we will always get multiple hosts for SorCloud, then we can remove the mode setting.  Let me know your thoughts on this.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
rashig: Id field could be null in the input data, and that we are handling in transform/write section. If Id field value is null, then that row will be filtered out and will not be indexed to Solr.
Also, as such there are no restrictions on unique field type from Solr, it could be any Solr supported data types (primitive data types). 
And we are already validating schema for the input fields in validateInputFieldsDataType().

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkUtility.java
rashig: User is allowed to select the mode from two of the provided options, SingleNode and SolrCloud (dropdown on UI), So there is no possibility for options other than allowed modes.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig, line 104: Started working on SolrSearch batch sink using SolrOutputFormat class, and I will update the PR once done.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -116,4 +92,95 @@ public String getIdField() {
...
+  /**
+   * Returns the SolrClient for establishing the connection with Solr server, using the hostname and port provided by
+   * user.
+   *
+   * @return Solr client
+   */
+  public SolrClient getSolrConnection() {
+    String urlString;
+    SolrClient solrClient = null;
+    if (solrMode.equals(SINGLE_NODE_MODE)) {
```
albertshau, line 105: lets ignore case when checking mode

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -116,4 +92,95 @@ public String getIdField() {
...
+  }
+
+  /**
+   * Validates whether the Id field provided by user is present in the input schema or not.
+   *
+   * @param inputSchema
+   */
+  public void validateIdField(Schema inputSchema) {
+    if (inputSchema != null && inputSchema.getField(idField) == null) {
+      throw new IllegalArgumentException(
```
albertshau, line 149: you need to check the id field's schema as well right? I'm assuming ids can't be arbitrary objects, and can't be nullable.

solrsearch-plugins/src/test/java/co.cask.hydrator.plugin/SolrSearchSinkTest.java
```diff
@@ -103,11 +103,10 @@
   private static final ArtifactRange BATCH_ARTIFACT_RANGE = new ArtifactRange(Id.Namespace.DEFAULT, "etlbatch",
                                                                               CURRENT_VERSION, true,
                                                                               CURRENT_VERSION, true);
-  private static final String SINGLE_NODE_SOLR_MODE = "Single Node Solr";
+  private static final String SINGLE_NODE_MODE = "SingleNode";
```
albertshau, line 106: you can make these fields public in the config class so that there's a single source of truth.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
rashig, line 105: These values are coming from the ‘select widget’ on UI, which will always have the same case for its options.
Please let me know, if I still need to check values by ignoring case.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
rashig, line 149: We are already verifying the data type for each field (that will include Id field column also) coming in the input structured record, in validateInputFieldsDataType(Schema inputSchema) method.
Please let me know, if we need to repeat the check for ID field here.

rashig, line 149: Id field was kept nullable, assuming the scenario where the sink could receive a nullable schema from the previous stage. Restricting Id field schema to be non-nullable, would force the user to always remove the nulls before passing it to the Solr search sink. Therefore, we were handling the null values in transform.
Please let me know, if we still need to change this and have non nullable schema for Id field.

albertshau, line 149: so what happens if the id is null? Is that record dropped? Please be sure to document the behavior.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
albertshau, line 105: the UI is not the only way to create a pipeline. Please ignore case when checking.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
ameyk, line 149: Yes Albert. Thats correct. Its documented in wiki and plug-in documentation.

sreevatsanraman:
This PR should be reopened against release/1.4

PR #308, opened 2016-07-11
Adding Joiner plugin for hydrator

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-6195
Build: http://builds.cask.co/browse/HYP-DBT23-10

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
@@ -1,0 +1,332 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.joiner;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.etl.api.JoinConfig;
+import co.cask.cdap.etl.api.JoinElement;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.StageConfigurer;
+import co.cask.cdap.etl.api.batch.BatchJoiner;
+import co.cask.cdap.etl.api.batch.BatchJoinerContext;
+import co.cask.cdap.etl.api.batch.BatchJoinerRuntimeContext;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Table;
+import com.google.gson.Gson;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.Path;
+
+/**
+ * Batch joiner to join records from multiple inputs
+ */
+@Plugin(type = BatchJoiner.PLUGIN_TYPE)
+@Name("Joiner")
+@Description("Performs join operation on records from each input based on required inputs. If all the inputs are " +
+  "required inputs, inner join will be performed. Otherwise inner join will be performed on required inputs and " +
+  "records from non-required inputs will only be present if they match join criteria. If there are no required " +
+  "inputs, outer join will be performed")
+public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, StructuredRecord> {
+  private static final Gson GSON = new Gson();
+  private final JoinerConfig conf;
+  private Map<String, Schema> inputSchemas;
+  private Schema outputSchema;
+  private Map<String, List<String>> perStageJoinKeys;
+  private Table<String, String, String> perStageSelectedFields;
+  private Set<String> requiredInputs;
+  private Multimap<String, String> duplicateFields = ArrayListMultimap.create();
+
+  public Joiner(JoinerConfig conf) {
+    this.conf = conf;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Map<String, Schema> inputSchemas = stageConfigurer.getInputSchemas();
+    init();
+    //validate the input schema and get the output schema for it
+    stageConfigurer.setOutputSchema(getOutputSchema(inputSchemas));
+  }
+
+  @Override
+  public void prepareRun(BatchJoinerContext context) throws Exception {
+    if (conf.getNumPartitions() != null) {
+      context.setNumPartitions(conf.getNumPartitions());
+    }
+  }
+
+  @Override
+  public void initialize(BatchJoinerRuntimeContext context) throws Exception {
+    init();
+    inputSchemas = context.getInputSchemas();
+    outputSchema = context.getOutputSchema();
+  }
+
+  @Override
+  public StructuredRecord joinOn(String stageName, StructuredRecord record) throws Exception {
+    List<Schema.Field> fields = new ArrayList<>();
+    Schema schema = record.getSchema();
+
+    List<String> joinKeys = perStageJoinKeys.get(stageName);
+    int i = 1;
+    for (String joinKey : joinKeys) {
+      Schema.Field joinField = Schema.Field.of(String.valueOf(i++), schema.getField(joinKey).getSchema());
+      fields.add(joinField);
+    }
+    Schema keySchema = Schema.recordOf("join.key", fields);
+    StructuredRecord.Builder keyRecordBuilder = StructuredRecord.builder(keySchema);
+    i = 1;
+    for (String joinKey : joinKeys) {
+      keyRecordBuilder.set(String.valueOf(i++), record.get(joinKey));
+    }
+
+    return keyRecordBuilder.build();
+  }
+
+  @Override
+  public JoinConfig getJoinConfig() {
+    return new JoinConfig(requiredInputs);
+  }
+
+  @Override
+  public StructuredRecord merge(StructuredRecord joinKey, Iterable<JoinElement<StructuredRecord>> joinRow) {
+    StructuredRecord.Builder outRecordBuilder = StructuredRecord.builder(outputSchema);
+
+    for (JoinElement<StructuredRecord> joinElement : joinRow) {
+      String stageName = joinElement.getStageName();
+      StructuredRecord record = joinElement.getInputRecord();
+
+      Map<String, String> selectedFields = perStageSelectedFields.row(stageName);
+
+      for (Schema.Field field : record.getSchema().getFields()) {
+        String inputFieldName = field.getName();
+
+        // drop the field if not part of fieldsToRename
+        if (!selectedFields.containsKey(inputFieldName)) {
+          continue;
+        }
+
+        String outputFieldName = selectedFields.get(inputFieldName);
+        outRecordBuilder.set(outputFieldName, record.get(inputFieldName));
+      }
+    }
+    return outRecordBuilder.build();
+  }
+
+  private void init() {
+    perStageJoinKeys = conf.getPerStageJoinKeys();
+    requiredInputs = conf.getInputs();
+    perStageSelectedFields = conf.getPerStageSelectedFields();
+  }
+
+  @Path("outputSchema")
+  public Schema getOutputSchema(GetSchemaRequest request) {
+    try {
+      return getOutputSchema(request.inputSchemas);
+    } catch (IllegalArgumentException e) {
+      if (!duplicateFields.isEmpty()) {
+        throw new BadRequestException(GSON.toJson(duplicateFields.asMap()));
+      } else {
+        throw new BadRequestException(e.getMessage());
+      }
+    }
+  }
+
+  /**
+   * Endpoint request for output schema.
+   */
+  public static class GetSchemaRequest extends JoinerConfig {
+    public Map<String, Schema> inputSchemas;
+  }
+
+  private Schema getOutputSchema(Map<String, Schema> inputSchemas) {
+    validateRequiredInputs(inputSchemas);
+
+    // stage name to input schema
+    Map<String, Schema> inputs = new HashMap<>(inputSchemas);
+    // Selected Field name to output field info
+    Map<String, OutputFieldInfo> outputFieldInfo = new LinkedHashMap<>();
+    List<String> duplicateAliases = new ArrayList<>();
+
+    // order of fields in output schema will be same as order of selectedFields
+    Set<Table.Cell<String, String, String>> rows = perStageSelectedFields.cellSet();
+    for (Table.Cell<String, String, String> row : rows) {
+      String stageName = row.getRowKey();
+      String inputFieldName = row.getColumnKey();
+      String alias = row.getValue();
+      Schema inputSchema = inputs.get(stageName);
+
+      // TODO As per CDAP-6402 handle cases where input schema is null at configure time.
+      if (inputSchema == null) {
+        throw new IllegalArgumentException(String.format("Input schema for input stage %s can not be null", stageName));
+      }
+
+      if (outputFieldInfo.containsKey(alias)) {
+        OutputFieldInfo outInfo = outputFieldInfo.get(alias);
+        if (duplicateAliases.add(alias)) {
+          duplicateFields.put(outInfo.getStageName(), outInfo.getInputFieldName());
+        }
+        duplicateFields.put(stageName, inputFieldName);
+        continue;
+      }
+
+      Schema.Field inputField = inputSchema.getField(inputFieldName);
+      if (inputField == null) {
+        throw new IllegalArgumentException(String.format(
+          "Invalid field: %s of stage '%s' does not exist in input schema %s.",
+          inputFieldName, stageName, inputSchema));
+      }
+      // set nullable fields for non-required inputs
+      if (requiredInputs.contains(stageName)) {
+        outputFieldInfo.put(alias, new OutputFieldInfo(alias, stageName, inputFieldName,
+                                                       Schema.Field.of(alias, inputField.getSchema())));
+      } else {
+        outputFieldInfo.put(alias, new OutputFieldInfo(alias, stageName, inputFieldName,
+                                                       Schema.Field.of(alias,
+                                                                       Schema.nullableOf(inputField.getSchema()))));
+      }
+    }
+
+    if (!duplicateFields.isEmpty()) {
+      throw new IllegalArgumentException(String.format("Output schema must not have any duplicate field names, but " +
+                                                         "found: %s for aliases: %s", duplicateFields,
+                                                       duplicateAliases));
+    }
+    return Schema.recordOf("join.output", getOutputFields(outputFieldInfo.values()));
+  }
+
+  private List<Schema.Field> getOutputFields(Collection<OutputFieldInfo> fieldsInfo) {
+    List<Schema.Field> outputFields = new ArrayList<>();
+    for (OutputFieldInfo fieldInfo : fieldsInfo) {
+      outputFields.add(fieldInfo.getField());
+    }
+    return outputFields;
+  }
+
+  /**
+   * Class to hold information about output fields
+   */
+  private class OutputFieldInfo {
+    private String name;
+    private String stageName;
+    private String inputFieldName;
+    private Schema.Field field;
+
+    public OutputFieldInfo(String name, String stageName, String inputFieldName, Schema.Field field) {
+      this.name = name;
+      this.stageName = stageName;
+      this.inputFieldName = inputFieldName;
+      this.field = field;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public void setName(String name) {
+      this.name = name;
+    }
+
+    public String getStageName() {
+      return stageName;
+    }
+
+    public void setStageName(String stageName) {
+      this.stageName = stageName;
+    }
+
+    public String getInputFieldName() {
+      return inputFieldName;
+    }
+
+    public void setInputFieldName(String inputFieldName) {
+      this.inputFieldName = inputFieldName;
+    }
+
+    public Schema.Field getField() {
+      return field;
+    }
+
+    public void setField(Schema.Field field) {
+      this.field = field;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      OutputFieldInfo that = (OutputFieldInfo) o;
+
+      if (!name.equals(that.name)) {
+        return false;
+      }
+      if (!stageName.equals(that.stageName)) {
+        return false;
+      }
+      if (!inputFieldName.equals(that.inputFieldName)) {
+        return false;
+      }
+      return field.equals(that.field);
+    }
+
+    @Override
+    public int hashCode() {
+      int result = name.hashCode();
+      result = 31 * result + stageName.hashCode();
+      result = 31 * result + inputFieldName.hashCode();
+      result = 31 * result + field.hashCode();
+      return result;
+    }
+
+    @Override
+    public String toString() {
+      return "OutputFieldInfo{" +
+        "name='" + name + '\'' +
+        ", stageName='" + stageName + '\'' +
+        ", inputFieldName='" + inputFieldName + '\'' +
+        ", field=" + field +
+        '}';
+    }
+  }
+
+  private void validateRequiredInputs(Map<String, Schema> inputSchemas) {
+    for (String requiredInput : requiredInputs) {
+      if (!inputSchemas.containsKey(requiredInput)) {
+        throw new IllegalArgumentException(String.format("Provided required input %s is not an input stage name.",
+                                                         requiredInput));
+      }
+    }
+  }
+}
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
@@ -1,0 +1,177 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.joiner;
+
+import co.cask.cdap.api.annotation.Description;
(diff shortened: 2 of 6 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -0,0 +1,264 @@
...
+    put.add("actor_name", "samuel");
+    filmActorTable.put(put);
+    filmActorManager.flush();
+
+    // 1: 1, matrix, action
+    // 2: 1, matrix, thriller
+    // 3: 2, equilibrium, action
+    DataSetManager<Table> filmCategoryManager = getDataset(filmCategoryDatasetName);
+    Table filmCategoryTable = filmCategoryManager.get();
+    put = new Put(Bytes.toBytes(1));
```
anwar6953: look into using MockTableSource

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -0,0 +1,92 @@
...
+  @Nullable
+  public String getSelectedFields() {
+    return selectedFields;
+  }
+
+  public String getJoinKeys() {
+    return joinKeys;
+  }
+
+  @Nullable
```
anwar6953, line 102: this doesn't seem like it will ever return null. empty string is possibility

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
CuriousVini: Why to use mock?

anwar6953: look at other usages of it. It helps set up the source data, I think.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+    return Collections.<String>emptyList();
+  }
+
+  @Path("outputSchema")
+  public Schema getOutputSchema(GetSchemaRequest request) {
+    try {
+      return getOutputSchema(request.inputSchemas);
+    } catch (IllegalArgumentException e) {
+      throw new BadRequestException(e.getMessage());
```
anwar6953: I think you should also send the `e` into the BadRequestException. can you do that? That way, the root stacktrace isn't lost

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private void init() {
+    perStageJoinKeys = getJoinKeys();
+    requiredInputs = getRequiredInputs();
+    perStageFieldsToRename = getPerStageFieldsToRename();
+  }
+
+  private Map<String, List<String>> getJoinKeys() {
+    Map<String, List<String>> stageToKey = new HashMap<>();
+    if (!Strings.isNullOrEmpty(conf.getJoinKeys())) {
```
anwar6953: I think it would be cleaner if you invert this condition. throw the exception on the next line. that way, the remainder of the function can be un-indented.

anwar6953: same in the `getPerStageFieldsToRename` method

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  @Override
+  public JoinConfig getJoinConfig() {
+    return new JoinConfig(requiredInputs);
+  }
+
+  @Override
+  public StructuredRecord merge(StructuredRecord joinKey, Iterable<JoinElement<StructuredRecord>> joinRow) {
+    StructuredRecord.Builder outRecordBuilder;
```
anwar6953: can combine lines 131 and 132.
why separate the declaration and initialization?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+  private Map<String, Schema> inputSchemas;
+  private Schema outputSchema;
+  private Map<String, List<String>> perStageJoinKeys;
+  private Table<String, String, String> perStageFieldsToRename;
+  private Iterable<String> requiredInputs;
+  private static final Gson GSON = new Gson();
+  private static final Type MAP_STRING_MAP = new TypeToken<Map<String, Map<String, String>>>() {
+  }.getType();
+
+  private static final Gson GSON1 = new GsonBuilder()
```
anwar6953: unused?
Does the used GSON object not need custom type adapters? For Schema? Structured Record?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Batch joiner to join records from multiple inputs
+ */
+@Plugin(type = BatchJoiner.PLUGIN_TYPE)
+@Name("Joiner")
+@Description("Performs join operation on records from each input based on required inputs. If all the inputs are " +
+  "required inputs, inner join will be performed. Otherwise inner join will be performed on required inputs and non " +
+  "records from non-required inputs will only be present if they match join criteria.")
+public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, StructuredRecord> {
+  private final JoinerConfig conf;
```
anwar6953, line 57: put non-static fields below the static ones

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+        }
+      }
+    } else {
+      throw new IllegalArgumentException("selectedFields can not be empty. Please provide at least 1 selectedFields");
+    }
+    return perStageFieldsToRename;
+  }
+
+  Iterable<String> getRequiredInputs() {
+    if (conf.getRequiredInputs() != null) {
```
anwar6953: conf.getRequiredInputs() method seems like it can/will return empty string, instead of null. can you check that

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+
+        // check if there are any duplicate fields
+        if (outputFields.containsKey(outputFieldName)) {
+          duplicateFieldsMap.put(stageName, com.google.common.base.Joiner.on(",")
+            .join(duplicateFieldsMap.get(stageName), outputFieldName));
+          continue;
+        }
+
+        // set nullable fields for non-required inputs
+        if (!Iterables.isEmpty(requiredInputs) && Iterables.contains(requiredInputs, stageName)) {
```
anwar6953: is the `isEmpty` method call necessary? Wouldn't the `contains` method check that?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,289 @@
...
+          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, inputField.getSchema()));
+        } else {
+          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, Schema.nullableOf(inputField.getSchema()
+          )));
```
anwar6953: just putting the close-parens on the next line seems strange. can you split the statement somewhere else?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -0,0 +1,92 @@
...
+    this.selectedFields = selectedFields;
+    this.requiredInputs = requiredInputs;
+  }
+
+  @Nullable
+  public Integer getNumPartitions() {
+    return numPartitions;
+  }
+
+  @Nullable
```
anwar6953: this doesn't seem like it will ever return null. empty string is possibility

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfigTest.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+package co.cask.hydrator.plugin.batch.joiner;
+
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.hydrator.plugin.transform.MockPipelineConfigurer;
+import com.google.gson.Gson;
+import org.junit.Assert;
+import org.junit.Test;
+import org.python.google.common.collect.ImmutableMap;
```
anwar6953: why org.python ... ?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfigTest.java
```diff
@@ -0,0 +1,173 @@
...
+                                                                         "film_name"));
+    selectedFields.put("filmActor", com.google.common.collect.ImmutableMap.of("actor_name", "renamed_actor"));
+    selectedFields.put("filmCategory", com.google.common.collect.ImmutableMap.of("category_name", "renamed_category"));
+
+    JoinerConfig config = new JoinerConfig("film.film_id=filmActor.film_id=filmCategory.film_id&" +
+                                             "film.film_name=filmActor.film_name=filmCategory.film_name",
+                                           GSON.toJson(selectedFields), "film");
+
+    MockPipelineConfigurer pipelineConfigurer = new MockPipelineConfigurer(
+      ImmutableMap.<String, Schema>of("film", filmSchema, "filmActor", filmActorSchema,
```
anwar6953: isn't `String, Schema` implicit? (Diamond syntax)

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -0,0 +1,264 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Tests for Joiner.
+ */
+public class JoinerTestRun extends ETLBatchTestBase {
+  private static final Gson GSON = new Gson();
```
anwar6953: insert space after line 53

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -0,0 +1,264 @@
...
+    filmTable.put(put);
+    filmManager.flush();
+
+    // 1: 1, matrix, alex
+    // 2: 1, matrix, bob
+    // 3: 2, equilibrium, cathie
+    // 4: 3, avatar, samuel
+    DataSetManager<Table> filmActorManager = getDataset(filmActorDatasetName);
+    Table filmActorTable = filmActorManager.get();
+    put = new Put(Bytes.toBytes(1));
```
anwar6953: private methods for creating these `Puts` would make the code more readable/clean

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class JoinerConfig extends PluginConfig {
+  private static final String NUM_PARTITIONS_DESC = "Number of partitions to use when joining. " +
+    "If not specified, the execution framework will decide how many to use.";
+  private static final String JOIN_KEY_DESC = "List of join keys to perform join operation. The list is " +
+    "separated by join condition. Join key from each input stage will be prefixed with '<stageName>.' And the " +
+    "relation among join keys from different inputs is represented by '='. For example: customers.customer_id=items" +
+    ".c_id&customers.customer_name=items.c_name will mean the join key is a composite key of customer id and customer" +
+    " name from customers and items input stages and join will be performed on equality of the join keys.";
+  private static final String SELECTED_FIELDS = "Map which represents fields to be selected and " +
```
albertshau: this seems like a pretty difficult format to understand and get correct.  Can we make this simpler for users by supporting a sql-like syntax where its a list, but people can say `as xyz` to rename the field to `xyz`? So something like `A.id as user_id, B.id as product_id`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+    return outRecordBuilder.build();
+  }
+
+  private void init() {
+    perStageJoinKeys = getJoinKeys();
+    requiredInputs = getRequiredInputs();
+    perStageFieldsToRename = getPerStageFieldsToRename();
+  }
+
+  private Map<String, List<String>> getJoinKeys() {
```
albertshau: I think it would be cleaner to have this method in the JoinConfig class. Then you could test validation of config settings by just creating a config and calling the method instead of going through the plugin itself. It would also make the logic re-usable in case we add more joins.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+          throw new IllegalArgumentException("There should be one join key from each of the stages. Please add join " +
+                                               "keys for each stage.");
+        }
+        for (KeyValue<String, String> keyValue : keyValues) {
+          String stageName = keyValue.getKey();
+          String joinKey = keyValue.getValue();
+          if (!stageToKey.containsKey(stageName)) {
+            stageToKey.put(stageName, new ArrayList<String>());
+          }
+          stageToKey.get(stageName).add(joinKey);
```
albertshau: don't you need to strip the prefix? Or is that done elsewhere?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+        }
+      }
+    } else {
+      throw new IllegalArgumentException("Join keys can not be empty");
+    }
+    return stageToKey;
+  }
+
+  private Table<String, String, String> getPerStageFieldsToRename() {
```
albertshau: lets move this into JoinConfig too

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+   * Endpoint request for output schema.
+   */
+  public static class GetSchemaRequest extends JoinerConfig {
+    public Map<String, Schema> inputSchemas;
+  }
+
+
+  private Schema getOutputSchema(Map<String, Schema> inputSchemas) {
+    validateRequiredInputs(inputSchemas);
+    // sort the input schemas by input names to get the deterministic order of fields for output schema
```
albertshau: if I were a user, I would expect the output schema to have fields ordered by what I put into the selectedFields setting,

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+          perStageFieldsToRename.put(stageToRenameField.getKey(), entry.getKey(), entry.getValue());
+        }
+      }
+    } else {
+      throw new IllegalArgumentException("selectedFields can not be empty. Please provide at least 1 selectedFields");
+    }
+    return perStageFieldsToRename;
+  }
+
+  Iterable<String> getRequiredInputs() {
```
albertshau: we should return a set in case they put in duplicates.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
albertshau: a Set seems like a more natural data structure here than an Iterable

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, inputField.getSchema()));
+        } else {
+          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, Schema.nullableOf(inputField.getSchema()
+          )));
+        }
+      }
+    }
+
+    if (!duplicateFieldsMap.isEmpty()) {
+      throw new IllegalArgumentException(GSON.toJson(duplicateFieldsMap));
```
albertshau: there needs to be a human-readable error message here. Something about field name conflicts in the output and which fields caused the problem.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -0,0 +1,282 @@
...
+  private Schema getOutputSchema(Map<String, Schema> inputSchemas) {
+    validateRequiredInputs(inputSchemas);
+    // sort the input schemas by input names to get the deterministic order of fields for output schema
+    Map<String, Schema> inputs = new TreeMap<>(inputSchemas);
+    Map<String, Schema.Field> outputFields = new LinkedHashMap<>();
+    Map<String, String> duplicateFieldsMap = new HashMap<>();
+
+    for (Map.Entry<String, Schema> input : inputs.entrySet()) {
+      String stageName = input.getKey();
+      Schema inputSchema = input.getValue();
```
albertshau: You need to handle the case where this schema is null. For now you could throw an exception about all inputs need to have a schema set, then add a TODO and jira to handle cases where schema is not known at configure time.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
albertshau: I don't think we're adding the Mock plugin in ETLBatchTestBase

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -0,0 +1,264 @@
...
+    workflowManager.start();
+    workflowManager.waitForFinish(5, TimeUnit.MINUTES);
+
+    DataSetManager<TimePartitionedFileSet> outputManager = getDataset(joinedDatasetName);
+    TimePartitionedFileSet fileSet = outputManager.get();
+    List<GenericRecord> outputRecords = readOutput(fileSet, outputSchema);
+
+    Assert.assertEquals(5, outputRecords.size());
+
+    Assert.assertEquals("1", outputRecords.get(0).get("film_id").toString());
```
albertshau: Is there anything guaranteeing order of output records?  I don't think there is, which means you should be comparing sets

albertshau:
there needs to be a test case for more than 2 inputs, and a test case where not all inputs are required. We also need to add a widgets and doc file, though that can be done in a separate PR.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
CuriousVini: `KeyValueListParser kvParser = new KeyValueListParser("\\s*=\\s*", "\\.");` strips the prefix.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
CuriousVini, line 102: It is `Nullable` as requiredInputs are optional.

CuriousVini:
@albertshau Thanks for the comments! As outer join is not supported yet, I will create another PR with added test case where not all the inputs are required. Also, will create another PR for doc and widgets.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
albertshau, line 102: right but the no-args constructor sets it to an empty string, which means this can never return null. I think you want to set selectedFields and requiredInputs to null in the no-args constructor.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -29,14 +42,18 @@
...
-    "being map of old field name to alias. There should be at least one selected field. There must not be any " +
-    "duplicate fields in output.";
+    "separated by '&'. Join key from each input stage will be prefixed with '<stageName>.' And the " +
+    "relation among join keys from different inputs is represented by '='. For example: " +
+    "customers.customer_id=items.c_id&customers.customer_name=items.c_name means the join key is a composite key" +
+    " of customer id and customer name from customers and items input stages and join will be performed on equality " +
+    "of the join keys.";
+  private static final String SELECTED_FIELDS = "Comma-separated list of fields to be selected and renamed " +
+    "in join output from each input stage. Each selected input field name needs to be present in the output must be " +
+    "prefixed with '<input_stage_name>. The syntax for specifying alias for each selected field is similar to sql. " +
```
albertshau: `'<input_stage_name>` -> `'<input_stage_name>'`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -89,4 +104,79 @@ public String getJoinKeys() {
...
+    ImmutableTable.Builder<String, String, String> tableBuilder = new ImmutableTable.Builder<>();
+
+    if (Strings.isNullOrEmpty(selectedFields)) {
+      throw new IllegalArgumentException("selectedFields can not be empty. Please provide at least 1 selectedFields");
+    }
+
+    for (String selectedField : Splitter.on(',').trimResults().omitEmptyStrings().split(selectedFields)) {
+      Iterable<String> stageOldNameAliasPair = Splitter.on(" as ").trimResults().omitEmptyStrings()
+        .split(selectedField);
+      if (Iterables.size(stageOldNameAliasPair) != 2) {
```
albertshau: we shouldn't require people to use ' as '. Instead, if ' as ' is not used, you can just have the alias be the same as the field name.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -227,55 +166,153 @@ public Schema getOutputSchema(GetSchemaRequest request) {
...
 
-        // set nullable fields for non-required inputs
-        if (!Iterables.isEmpty(requiredInputs) && Iterables.contains(requiredInputs, stageName)) {
-          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, inputField.getSchema()));
-        } else {
-          outputFields.put(outputFieldName, Schema.Field.of(outputFieldName, Schema.nullableOf(inputField.getSchema()
-          )));
+      if (outputFieldInfo.containsKey(alias)) {
+        OutputFieldInfo outInfo = outputFieldInfo.get(alias);
+        if (!duplicateAliases.contains(alias)) {
```
albertshau: nit: slightly more efficient if you do `if (duplicateAliases.add(alias)) {`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -227,55 +166,153 @@ public Schema getOutputSchema(GetSchemaRequest request) {
...
+          duplicateFields.put(outInfo.getStageName(), outInfo.getInputFieldName());
+          duplicateAliases.add(alias);
         }
+        duplicateFields.put(stageName, inputFieldName);
+        continue;
       }
+
+      Schema.Field inputField = inputSchema.getField(inputFieldName);
```
albertshau, line 200: need to handle the case where this is null, which happens if the user specifies a field that is not in the input schema.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -89,4 +104,79 @@ public String getJoinKeys() {
...
+        throw new IllegalArgumentException(String.format("Invalid syntax. Selected Fields must be of syntax " +
+                                                           "<stageName>.<oldFieldName> as <alias>, but found %s",
+                                                         selectedField));
+      }
+
+      Iterable<String> stageOldNamePair = Splitter.on('.').trimResults().omitEmptyStrings().
+        split(Iterables.get(stageOldNameAliasPair, 0));
+
+      if (Iterables.size(stageOldNameAliasPair) != 2) {
```
albertshau, line 157: this is slightly too restrictive, as it is legal for field names to contain a '.'.  Should instead be splitting on the first '.'

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfigTest.java
```diff
@@ -59,38 +56,25 @@ public void testJoinerConfig() {
...
-    selectedFields.put("filmActor", com.google.common.collect.ImmutableMap.of("actor_name", "renamed_actor"));
-    selectedFields.put("filmCategory", com.google.common.collect.ImmutableMap.of("category_name", "renamed_category"));
-
-
     JoinerConfig config = new JoinerConfig("film.film_id=filmActor.film_id=filmCategory.film_id&" +
                                              "film.film_name=filmActor.film_name=filmCategory.film_name",
-                                           GSON.toJson(selectedFields), "film,filmActor,filmCategory");
+                                           selectedFields, "film,filmActor,filmCategory");
 
     MockPipelineConfigurer pipelineConfigurer = new MockPipelineConfigurer(
```
albertshau, line 66: shouldn't need to use MockPipelineConfigurer anywhere, just call the relevant method on the config object.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -227,55 +166,153 @@ public Schema getOutputSchema(GetSchemaRequest request) {
...
+    for (OutputFieldInfo fieldInfo : fieldsInfo) {
+      outputFields.add(fieldInfo.getField());
+    }
+    return  outputFields;
+  }
+
+  /**
+   * Class to hold information about outputfields
+   */
+  class OutputFieldInfo {
```
albertshau: nit: this can be private

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -227,55 +166,153 @@ public Schema getOutputSchema(GetSchemaRequest request) {
...
+    }
+    return Schema.recordOf("join.output", getOutputFields(outputFieldInfo.values()));
+  }
+
+  private List<Schema.Field> getOutputFields(Collection<OutputFieldInfo> fieldsInfo) {
+    List<Schema.Field> outputFields = new ArrayList<>();
+    for (OutputFieldInfo fieldInfo : fieldsInfo) {
+      outputFields.add(fieldInfo.getField());
+    }
+    return  outputFields;
```
albertshau: remove extra space

albertshau:
took another pass, a couple comments

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfig.java
CuriousVini, line 102: done

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfigTest.java
CuriousVini, line 66: Right. I will remove `MockPipelineConfigurer` for methods present in `JoinConfig.java`. As `getOutputSchema()` is only present in Joiner plugin as private method, I am keeping `MockPipelineConfigurer` for some test methods which require output schema for assertion.

CuriousVini:
@albertshau Addressed comments and added 3 more test cases in JoinConfig. Please take another look when you get a chance. Thanks!

albertshau:
lgtm

CuriousVini:
Build is green. Squashing and merging.

PR #315, opened 2016-07-16
Fixing bug where not specifying fields to renamed throws IllegalArgum…

CuriousVini:
…ent exception

Issue: https://issues.cask.co/browse/CDAP-6503
Build: http://builds.cask.co/browse/HYP-DBT29-1


albertshau:
lgtm

PR #316, opened 2016-07-18
[CDAP-6527]: SSHAction Hydrator Plugin

thegreenfrog:
Implement generic SSHAction class for Hydrator. Runs command on remote machine and gives option to collect output to store in macro afterwards.

### DIFF
--- a/core-plugins/docs/SSHAction-action.md
+++ b/core-plugins/docs/SSHAction-action.md
@@ -1,0 +1,58 @@
+# SSH Action
+
+
+Description
+-----------
+Establishes an SSH connection with remote machine to execute command on that machine.
+
+
+Use Case
+--------
+This action can be used when data needs to be copied to HDFS from a remote server before the pipeline starts.
+It can also be used to move data for archiving before the pipeline starts.
+
+
+Properties
+----------
+
+**host:** The host name of the remote machine where the command needs to be executed.
+
+**port:** The port used to make SSH connection. Defaults to 22.
+
+**user:** The username used to connect to host.
+
+**privateKeyFile:** The file path to Private key.
+
+**password:** The password associated with private key.
+
+**command:** The command to be executed on the remote host. This includes the filepath of the script and any arguments
+needing to be passed.
+
+**output:** Script output variable used as the key in the instance the output from a script needs to be saved
+and passed through the Action Context.
+
+Example
+-------
+This example runs a script on demo@example.com:
+
+    {
+        "name": "SSHAction",
+          "plugin": {
+            "name": "SSHAction",
+            "type": "action",
+            "label": "SSHAction",
+            "artifact": {
+              "name": "core-plugins",
+              "version": "1.4.0-SNAPSHOT",
+              "scope": "SYSTEM"
+          },
+          "properties": {
+              "host": "example.com",
+              "password": "pass",
+              "port": "22",
+              "user": "demo",
+              "privateKeyFile": "/Users/Jose/.ssh/id_rsa",
+              "command": "~/scripts/remoteScript.sh"
+            }
+          }
+    }
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -30,6 +30,12 @@
   <artifactId>core-plugins</artifactId>
 
   <dependencies>
+    <!-- https://mvnrepository.com/artifact/ch.ethz.ganymed/ganymed-ssh2 -->
+    <dependency>
+      <groupId>ch.ethz.ganymed</groupId>
+      <artifactId>ganymed-ssh2</artifactId>
+      <version>262</version>
+    </dependency>
     <dependency>
       <groupId>co.cask.cdap</groupId>
       <artifactId>hydrator-test</artifactId>
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.action;
+
+import ch.ethz.ssh2.Connection;
+import ch.ethz.ssh2.Session;
+import ch.ethz.ssh2.StreamGobbler;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+
+import co.cask.cdap.etl.api.action.Action;
+import co.cask.cdap.etl.api.action.ActionContext;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
+import com.google.common.io.CharStreams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import javax.annotation.Nullable;
+
+/**
+ * SSH into a remote machine and execute a script on that machine.
+ * A user must specify username and keypair authentication credentials.
+ * Optionals include port and machine URL
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name("SSHAction")
+@Description("Action to run a script on remote machine")
+public class SSHAction extends Action {
+  private final SSHActionConfig config;
+
+  private static final Logger LOG = LoggerFactory.getLogger(SSHAction.class);
+
+  public SSHAction(SSHActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(final ActionContext context) throws Exception {
+    Connection connection = new Connection(config.host);
+    Session session = connection.openSession();
+    try {
+      connection.connect();
+
+      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
+      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+        throw new IOException(String.format("SSH authentication error when connecting to %s@%s on port %d",
+                                            config.user, config.host, config.port));
+      }
+
+      LOG.debug("Connection established with the host {}", config.host);
+      session.execCommand(config.command);
+
+      InputStream stdout = new StreamGobbler(session.getStdout());
+      BufferedReader outBuffer = new BufferedReader(new InputStreamReader(stdout, Charsets.UTF_8));
+      InputStream stderr = new StreamGobbler(session.getStderr());
+      BufferedReader errBuffer = new BufferedReader(new InputStreamReader(stderr, Charsets.UTF_8));
+
+      String out = CharStreams.toString(outBuffer);
+      String err = CharStreams.toString(errBuffer);
+
+      LOG.debug("Output: {}", out);
+      LOG.error("Errors: {}", err);
+
+      if (config.output != null) {
+        context.getArguments().set(config.output, out);
+      }
+    } finally {
+      connection.close();
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    this.config.validate();
+  }
+
+  /**
+   * Config class that contains all the properties needed to SSH into the remote machine and run the script.
+   */
+  public static class SSHActionConfig extends PluginConfig {
+
+    @Description("Host name of the remote machine where the command needs to be executed.")
+    private String host;
+
+    @Nullable
+    @Description("Port to connect to. Defaults to 22")
+    private Integer port;
+
+    @Description("User name used to connect to host")
+    private String user;
+
+    @Description("File path to Private key")
+    private String privateKeyFile;
+
+    @Nullable
+    @Description("Password associated with private key")
+    private String password;
+
+    @Description("Command to be executed on the remote host. Should include filepath of script and any arguments")
+    private String command;
+
+    @Nullable
+    @Description("Script output variable used as the key to store output from the script " +
+      "to pass through Action Context")
+    private String output;
+
+    public SSHActionConfig() {
+      port = 22;
+    }
+
+    @VisibleForTesting
+    public SSHActionConfig(String host, String user, String privateKeyFile, int port, String password,
+                           String command, String output) {
+      this.host = host;
+      this.port = port;
+      this.user = user;
+      this.privateKeyFile = privateKeyFile;
+      this.password = password;
+      this.command = command;
+      this.output = output;
+    }
+
+    public void validate() {
+      //check that port is not negative
+      if (port < 0) {
+        throw new IllegalArgumentException("Port cannot be negative");
+      }
+    }
+  }
+}
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
@@ -21,6 +21,7 @@ import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
 import co.cask.cdap.datapipeline.DataPipelineApp;
 import co.cask.cdap.etl.api.PipelineConfigurable;
+import co.cask.cdap.etl.api.action.Action;
 import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.batch.ETLBatchApplication;
 import co.cask.cdap.proto.Id;
@@ -31,6 +32,7 @@ import co.cask.cdap.proto.id.NamespaceId;
 import co.cask.cdap.test.TestBase;
 import co.cask.cdap.test.TestConfiguration;
 import co.cask.hydrator.plugin.batch.action.EmailAction;
+import co.cask.hydrator.plugin.batch.action.SSHAction;
 import co.cask.hydrator.plugin.batch.aggregator.DedupAggregator;
 import co.cask.hydrator.plugin.batch.aggregator.GroupByAggregator;
 import co.cask.hydrator.plugin.batch.joiner.Joiner;
@@ -114,6 +116,7 @@ public class ETLBatchTestBase extends TestBase {
     // add the artifact for etl batch app
     addAppArtifact(DATAPIPELINE_ARTIFACT_ID, DataPipelineApp.class,
                    BatchSource.class.getPackage().getName(),
+                   Action.class.getPackage().getName(),
                    PipelineConfigurable.class.getPackage().getName(),
                    "org.apache.avro.mapred", "org.apache.avro", "org.apache.avro.generic", "org.apache.avro.io",
                    // these are not real exports for the application, but are required for unit tests.
@@ -151,7 +154,8 @@ public class ETLBatchTestBase extends TestBase {
                       GroupByAggregator.class,
                       DedupAggregator.class,
                       Joiner.class,
-                      EmailAction.class);
+                      EmailAction.class,
+                      SSHAction.class);
   }
 
   protected List<GenericRecord> readOutput(TimePartitionedFileSet fileSet, Schema schema) throws IOException {
--- a/core-plugins/widgets/SSHAction-action.json
+++ b/core-plugins/widgets/SSHAction-action.json
@@ -1,0 +1,48 @@
+{
+  "metadata": {
+    "spec-version": "1.0"
+  },
+  "configuration-groups": [
+    {
+      "label": "SSH Properties",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Hostname",
+          "name": "host"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Port",
+          "name": "port"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Username",
+          "name": "user"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Private Key File",
+          "name": "privateKeyFile"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Private Key Passphrase",
+          "name": "password"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Command",
+          "name": "command"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Script Output Key Name",
+          "name": "output"
+        }
+      ]
+    }
+  ],
+  "outputs": []
+}


sagarkapare:
Please file a JIRA for this issue and also include JIRA id in the commit message while merging.

core-plugins/pom.xml
```diff
@@ -216,6 +222,7 @@
               co.cask.hydrator.plugin.validator.*;
               org.apache.avro.mapreduce;
               parquet.avro.*;
+              ch.ethz.ssh2.*;
```
chtyim: Why need to be exported? I don't think we should.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+    try {
+      connection.connect();
+
+      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
+      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+        throw new IOException(String.format("Unable to establish SSH connection for %s@%s on port %d",
+                                            config.user, config.host, config.port));
+      }
+
+      LOG.info(String.format("Connection established with the host %s", config.host));
```
chtyim: Don't use `String.format` when using logger. Use the `{}` place holder instead.

`LOG.info("Connection established with host {}", config.host);`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+        line = errBuffer.readLine();
+      }
+
+      LOG.info("Output: {}", outBuilder.toString());
+      LOG.info("Errors: {}", errBuilder.toString());
+
+      if (config.output != null) {
+        context.getArguments().set(config.output, outBuilder.toString());
+      }
+      session.close();
```
chtyim: Do this need to be in finally? Or the connection.close() will close this as well?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+                                            config.user, config.host, config.port));
+      }
+
+      LOG.info(String.format("Connection established with the host %s", config.host));
+
+      Session session = connection.openSession();
+      session.execCommand(config.cmd);
+
+      InputStream stdout = new StreamGobbler(session.getStdout());
+      BufferedReader outBuffer = new BufferedReader(new InputStreamReader(stdout));
```
chtyim: Always provide the charset when creating `InputStreamReader`.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+      InputStream stderr = new StreamGobbler(session.getStderr());
+      BufferedReader errBuffer = new BufferedReader(new InputStreamReader(stderr));
+
+      StringBuilder outBuilder = new StringBuilder();
+      StringBuilder errBuilder = new StringBuilder();
+
+      String line = outBuffer.readLine();
+      while (line != null) {
```
chtyim: You can use `CharStreams.toString(outBuffer)`. No need to do it yourself.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+    @Description("Port to connect to. Defaults to 22")
+    private int port;
+
+    @Description("User name used for conn")
+    private String user;
+
+    @Description("Private key file.")
+    private String privateKeyFile;
+
+    @Description("Password")
```
chtyim: What password? The password for the user or the password for the private key file?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Nullable
+    @Description("Port to connect to. Defaults to 22")
+    private int port;
+
+    @Description("User name used for conn")
+    private String user;
+
+    @Description("Private key file.")
+    private String privateKeyFile;
```
chtyim, line 118: Where does the file come from? Does it need to be on the cluster? When secure store is available, will it be using the secure store instead?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: Private Key File. ill update the description!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Description("Password")
+    private String password;
+
+    @Description("Command to be executed on the remote host.")
+    private String cmd;
+
+    @Nullable
+    @Description("Script output variable")
```
chtyim: What is variable? I don't think we have this concept in hydrator?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -114,12 +116,13 @@ public static void setupTest() throws Exception {
...
     addAppArtifact(DATAPIPELINE_ARTIFACT_ID, DataPipelineApp.class,
                    BatchSource.class.getPackage().getName(),
+                   Action.class.getPackage().getName(),
                    PipelineConfigurable.class.getPackage().getName(),
                    "org.apache.avro.mapred", "org.apache.avro", "org.apache.avro.generic", "org.apache.avro.io",
                    // these are not real exports for the application, but are required for unit tests.
                    // the stupid hive-exec jar pulled in by cdap-unit-test contains ParquetInputSplit...
                    // without this, different classloaders will be used for ParquetInputSplit and we'll see errors
-                   "parquet.hadoop.api", "parquet.hadoop", "parquet.schema", "parquet.io.api");
+                   "parquet.hadoop.api", "parquet.hadoop", "parquet.schema", "parquet.io.api", "ch.ethz.ssh2");
```
chtyim: Shouldn't need to expose the ssh2 package.

core-plugins/pom.xml
thegreenfrog: it wasn't building successfully without it when I was working with it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: ok! What's the reasoning behind this? Just convention?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog, line 118: yes it will be using secure store.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: this will be passed by the user to specify what the key will be to store values in secure store, such as a table name for a dbplugin.

core-plugins/pom.xml
albertshau: this should not be exported, and it shouldn't have any effect on the build. what error were you seeing?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    this.config.validate();
+  }
+
+  /**
+   * Config class that contains all the properties needed to SSH into the remote machine and run the script.
+   */
+  public static class SSHActionConfig extends PluginConfig {
+
+    @Description("Host name of the remote machine where the command need to be executed.")
```
albertshau: need -> needs

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+    private String host;
+
+    @Nullable
+    @Description("Port to connect to. Defaults to 22")
+    private int port;
```
albertshau: nullable properties need  to be objects. Integer instead of int.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @Description("User name used for conn")
```
albertshau: used to connect to the host.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: yes it should.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
albertshau: we're storing the output in the secure store? That seems wrong, should we just log whatever comes back in stdout as info and whatever comes back in stderr as error?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
albertshau: in order to set the default value, you should create a no-argument constructor and set the default value there.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
thegreenfrog: ok. will remove.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,159 @@
...
+    private String user;
+
+    @Description("Private key file.")
```
albertshau: please change this description to reflect the intended behavior.

core-plugins/pom.xml
thegreenfrog: Forget honestly. It was back when building a CDAP app in a testcase wasn't recognizing the package.

albertshau:
please add a corresponding widgets .json file and docs .md file for the plugin.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: we do. But if the user specifies a name for this config, meaning they know the output will have the needed key name in secure store.

albertshau: what use case are we trying to solve with this?

thegreenfrog:
@albertshau @chtyim please review changes.

core-plugins/pom.xml
chtyim: Do you have the stacktrace?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
chtyim: One less unnecessary object creation.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
chtyim: BTW, I think we use set the command execution status (succeeded, failed) as part of the `SettableArguments` of the action. That is something that downstream stages can use.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,152 @@
...
+
+      LOG.info("Output: {}", out);
+      LOG.info("Errors: {}", err);
+
+      if (config.output != null) {
+        context.getArguments().set(config.output, out);
+      }
+    } finally {
+      session.close();
+      connection.close();
```
chtyim, line 93: Is this the right order? You open the session first, followed by the connection. Shouldn't the free up in reverse order? Also, will `session.close` or `connection.close` throw exception? If either of them will, you need to use nested try-finally blocks to do the close correctly. E.g.

``` java
connection.connect();
try {
  Session session = connection.openSession();
  try {
    // actual work.
  } finally {
    session.close();
  }
} finally {
  connection.close();
}
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,152 @@
...
+
+      InputStream stdout = new StreamGobbler(session.getStdout());
+      BufferedReader outBuffer = new BufferedReader(new InputStreamReader(stdout, Charsets.UTF_8));
+      InputStream stderr = new StreamGobbler(session.getStderr());
+      BufferedReader errBuffer = new BufferedReader(new InputStreamReader(stderr, Charsets.UTF_8));
+
+      String out = CharStreams.toString(outBuffer);
+      String err = CharStreams.toString(errBuffer);
+
+      LOG.info("Output: {}", out);
```
chtyim: debug.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,152 @@
...
+    Session session = connection.openSession();
+    try {
+      connection.connect();
+
+      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
+      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+        throw new IOException(String.format("Unable to establish SSH connection for %s@%s on port %d",
+                                            config.user, config.host, config.port));
+      }
+      LOG.info("Connection established with the host {}", config.host);
```
chtyim: debug?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: if the user runs a script that outputs the tablename that they want their dbplugin to execute a MYSQL query on, then they can set "output" to that in the config, and then the code will save the tablename in the context variable under the key set in "output"

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog, line 93: oh true it should be in reverse order. Neither throw an exception though.

thegreenfrog:
@chtyim updated files. please scan again! thanks.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
albertshau: ok so its not writing to the secure store, its setting a pipeline argument. And the output of the script means stdout right? Please add some more detail to the description, otherwise users will not know what this is for.

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,56 @@
+# SSH Action
+
+
+Description
+-----------
+Establishes an SSH conncetion with remote machine to execute command on that machine.
```
anwar6953: connection

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+Use Case
+--------
+This action can be used when data needs to be copied to HDFS from a remote server before the pipleline starts.
```
anwar6953: pipeline

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+**port:** The port used to make SSH connection. Defaults to 22.
+
+**user:** The username used to connect to host.
+
+**privateKeyFile:** The file path to Private key.
+
+**password:** The password associated with private key
+
+**cmd:** The command to be executed on the remote host.
```
anwar6953: This has to be a file path?

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,56 @@
...
+                        "plugin": {
+                            "name": "SSHAction",
+                            "type": "action",
+                            "label": "SSHAction",
+                            "artifact": {
+                                "name": "core-plugins",
+                                "version": "1.4.0-SNAPSHOT",
+                                "scope": "SYSTEM"
+                            },
+                            "properties": {
```
anwar6953: indentation of these properties looks off

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,150 @@
...
+
+      InputStream stdout = new StreamGobbler(session.getStdout());
+      BufferedReader outBuffer = new BufferedReader(new InputStreamReader(stdout, Charsets.UTF_8));
+      InputStream stderr = new StreamGobbler(session.getStderr());
+      BufferedReader errBuffer = new BufferedReader(new InputStreamReader(stderr, Charsets.UTF_8));
+
+      String out = CharStreams.toString(outBuffer);
+      String err = CharStreams.toString(errBuffer);
+
+      LOG.info("Errors: {}", err);
```
anwar6953: Should this be LOG.error

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,150 @@
...
+    @Description("Port to connect to. Defaults to 22")
+    private Integer port;
+
+    @Description("User name used to connect to host")
+    private String user;
+
+    @Description("File path to Private key")
+    private String privateKeyFile;
+
+    @Description("Password associated with private key")
```
anwar6953, line 121: can this be nullable? It's always required?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Nullable
+    @Description("Script output variable")
+    private String output;
+
+    public SSHActionConfig() {
+      port = 22;
+    }
+
+    public SSHActionConfig(String host, String user, String privateKeyFile, int port, String password,
```
anwar6953, line 137: Is this constructor used only for testing?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,150 @@
...
+  @Override
+  public void run(final ActionContext context) throws Exception {
+    Connection connection = new Connection(config.host);
+    Session session = connection.openSession();
+    try {
+      connection.connect();
+
+      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
+      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+        throw new IOException(String.format("Unable to establish SSH connection for %s@%s on port %d",
```
anwar6953: Isn't this specifically an authentication error? The exception indicates that it could be some generic ssh connection error. Make the message as specific as possible, in my opinion.

core-plugins/docs/SSHAction-action.md
thegreenfrog: I think we ultimately are going to use secure storage, so this is only temporary.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog, line 121: Oh I guess it can be passed as null to the Authentication method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog, line 137: yes. Currently I didn't include unit tests, because we're trying to figure out if we can even do unit tests. LMK if I should remove this.

core-plugins/docs/SSHAction-action.md
anwar6953: I'm referring to the `cmd` property.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
anwar6953, line 137: just mark it
@VisibleForTesting (or indicate that its just for testing with a comment)

core-plugins/docs/SSHAction-action.md
thegreenfrog: oh, no it should include the file path and any arguments. Changing.

core-plugins/docs/SSHAction-action.md
thegreenfrog: seems fine to me

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,58 @@
...
+
+**cmd:** The command to be executed on the remote host. This includes the filepath of the script and any arguments
+needing to be passed.
+
+**output:** Script output variable used as the key in the instance the output from a script needs to be saved
+and passed through the Action Context.
+
+Example
+-------
+This example runs a script on Jose@23.345.1.117:
```
sagarkapare: don't put ip address. use demo user and host such as `demo@example.com`

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,58 @@
...
+                            "name": "SSHAction",
+                            "type": "action",
+                            "label": "SSHAction",
+                            "artifact": {
+                                "name": "core-plugins",
+                                "version": "1.4.0-SNAPSHOT",
+                                "scope": "SYSTEM"
+                            },
+                            "properties": {
+                                "host": "23.345.1.117",
```
sagarkapare: don't put IP address. use `example.com`

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,58 @@
...
+                                "password": "pass",
+        			                  "port": "22",
+        			                  "user": "Jose",
+        			                  "privateKeyFile": "/Users/Jose/.ssh/id_rsa",
```
sagarkapare: should be updated once you changed the user name

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,58 @@
...
+                            "label": "SSHAction",
+                            "artifact": {
+                                "name": "core-plugins",
+                                "version": "1.4.0-SNAPSHOT",
+                                "scope": "SYSTEM"
+                            },
+                            "properties": {
+                                "host": "23.345.1.117",
+                                "password": "pass",
+        			                  "port": "22",
```
sagarkapare: nit: indentation is off

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,154 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import javax.annotation.Nullable;
+
+/**
+ * SSH into a remote machine and execute a script on that machine.
+ * A user must specify username and keypair authentication credentials.
```
sagarkapare, line 47: This will change once we get a private key from secure store. Can you file a JIRA for this?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,154 @@
...
+
+      InputStream stdout = new StreamGobbler(session.getStdout());
+      BufferedReader outBuffer = new BufferedReader(new InputStreamReader(stdout, Charsets.UTF_8));
+      InputStream stderr = new StreamGobbler(session.getStderr());
+      BufferedReader errBuffer = new BufferedReader(new InputStreamReader(stderr, Charsets.UTF_8));
+
+      String out = CharStreams.toString(outBuffer);
+      String err = CharStreams.toString(errBuffer);
+
+      LOG.error("Errors: {}", err);
```
sagarkapare, line 87: should also `LOG` the stdout of the script.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,154 @@
...
+
+      if (config.output != null) {
+        context.getArguments().set(config.output, out);
+      }
+    } finally {
+      connection.close();
+      session.close();
```
sagarkapare: should we first close the session and then connection? what does the library doc say?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Description("File path to Private key")
+    private String privateKeyFile;
+
+    @Nullable
+    @Description("Password associated with private key")
+    private String password;
+
+    @Description("Command to be executed on the remote host. Should include filepath of script and any arguments")
+    private String cmd;
```
sagarkapare: better call it `command`

core-plugins/widgets/SSHAction-action.json
```diff
@@ -0,0 +1,48 @@
...
+          "name": "privateKeyFile"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Private Key Passphrase",
+          "name": "password"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Remote Command",
```
sagarkapare: should just be `Command`

core-plugins/widgets/SSHAction-action.json
```diff
@@ -0,0 +1,48 @@
...
+          "name": "cmd"
```
sagarkapare: `command`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog, line 87: I think @chtyim suggested we remove that above?

sagarkapare:
Minor nits else LGTM. Squash the commits before merging and also make sure to run hydrator build.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
albertshau, line 87: "debug?" meant log at debug level, not remove.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: oh actually the docs say "Close the connection to the SSH-2 server. All assigned sessions will be closed, too." Missed that. So then we don't even need to specify `session.close()`

core-plugins/docs/SSHAction-action.md
```diff
@@ -0,0 +1,58 @@
...
+
+**port:** The port used to make SSH connection. Defaults to 22.
+
+**user:** The username used to connect to host.
+
+**privateKeyFile:** The file path to Private key.
+
+**password:** The password associated with private key.
+
+**cmd:** The command to be executed on the remote host. This includes the filepath of the script and any arguments
```
sagarkapare: command

core-plugins/docs/SSHAction-action.md
albertshau: everything seems too far right. "name" and "plugin" should be at the same level

thegreenfrog: oh i see. 👍

ajainarayanan:
Widget json LGTM 👍

PR #317, opened 2016-07-18
[CDAP-6371] Add Joiner widget json

elfenheart:


core-plugins/widgets/Joiner-batchjoiner.json
```diff
@@ -0,0 +1,46 @@
+{
+  "metadata": {
+    "spec-version": "1.3"
+  },
+  "inputs": {
+    "multipleInputs": true
```
CuriousVini, line 6: Does this property mean we only allow more than 2 inputs for joiner plugin and not any other plugin?

CuriousVini:
Just one question. Otherwise LGTM

core-plugins/widgets/Joiner-batchjoiner.json
elfenheart, line 6: This property means that the plugin allow multiple inputs with different schema. If the config does not have this property, it means that the schema for all inputs have to be the same

ajainarayanan:
👍 LGTM

PR #318, opened 2016-07-18
make table name a macro in batch writable sinks (table, kvtable, cube)

gsps1:
https://issues.cask.co/browse/CDAP-6360

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
@@ -16,8 +16,9 @@
 
 package co.cask.hydrator.plugin.batch.sink;
 
+import co.cask.cdap.api.dataset.DatasetManagementException;
 import co.cask.cdap.api.dataset.DatasetProperties;
-import co.cask.cdap.api.plugin.PluginProperties;
+import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
@@ -37,7 +38,11 @@ import java.util.Map;
  * @param <VAL_OUT> the type of value the sink outputs
  */
 public abstract class BatchWritableSink<IN, KEY_OUT, VAL_OUT> extends BatchSink<IN, KEY_OUT, VAL_OUT> {
+  private final PluginConfig pluginConfig;
 
+  protected BatchWritableSink(PluginConfig pluginConfig) {
+    this.pluginConfig = pluginConfig;
+  }
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     String datasetName = getProperties().get(Properties.BatchReadableWritable.NAME);
@@ -49,7 +54,11 @@ public abstract class BatchWritableSink<IN, KEY_OUT, VAL_OUT> extends BatchSink<
     properties.remove(Properties.BatchReadableWritable.NAME);
     properties.remove(Properties.BatchReadableWritable.TYPE);
 
-    pipelineConfigurer.createDataset(datasetName, datasetType, DatasetProperties.builder().addAll(properties).build());
+    // null check for tests, as macro fields map and properties are null in test cases
+    if (pluginConfig.getProperties() != null && !pluginConfig.containsMacro(Properties.BatchReadableWritable.NAME)) {
+      pipelineConfigurer.createDataset(datasetName, datasetType,
+                                       DatasetProperties.builder().addAll(properties).build());
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
```diff
@@ -37,7 +38,11 @@
  * @param <VAL_OUT> the type of value the sink outputs
  */
 public abstract class BatchWritableSink<IN, KEY_OUT, VAL_OUT> extends BatchSink<IN, KEY_OUT, VAL_OUT> {
+  private final PluginConfig pluginConfig;
 
+  protected BatchWritableSink(PluginConfig pluginConfig) {
+    this.pluginConfig = pluginConfig;
+  }
```
albertshau, line 46: add newline between methods

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
```diff
@@ -49,7 +54,11 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     properties.remove(Properties.BatchReadableWritable.NAME);
     properties.remove(Properties.BatchReadableWritable.TYPE);
 
-    pipelineConfigurer.createDataset(datasetName, datasetType, DatasetProperties.builder().addAll(properties).build());
+    // null check for tests, as macro fields map and properties are null in test cases
+    if (pluginConfig.getProperties() != null && !pluginConfig.containsMacro(Properties.BatchReadableWritable.NAME)) {
```
albertshau: rather than doing this, I think it would be better to have a BatchWritableConfig with a single macro enabled name field that all the other configs extend.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
```diff
@@ -58,8 +67,16 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   protected abstract Map<String, String> getProperties();
 
   @Override
-  public void prepareRun(BatchSinkContext context) {
-    PluginProperties pluginProperties = context.getPluginProperties();
-    context.addOutput(pluginProperties.getProperties().get(Properties.BatchReadableWritable.NAME));
+  public void prepareRun(BatchSinkContext context) throws DatasetManagementException {
+    Map<String, String> properties = getProperties();
+    if (pluginConfig.containsMacro(Properties.BatchReadableWritable.NAME)) {
+      if (!context.datasetExists(properties.get(Properties.BatchReadableWritable.NAME))) {
```
albertshau, line 76: rather than doing this, I think it would be better to have a BatchReadableWritableConfig with a single macro enabled name field that all the other configs extend.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
```diff
@@ -49,7 +54,11 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     properties.remove(Properties.BatchReadableWritable.NAME);
     properties.remove(Properties.BatchReadableWritable.TYPE);
 
-    pipelineConfigurer.createDataset(datasetName, datasetType, DatasetProperties.builder().addAll(properties).build());
+    // null check for tests, as macro fields map and properties are null in test cases
```
albertshau, line 59: why is it null for tests? not sure I understand. Are we calling configurePipeline on some plugin that has been manually created in the test? Why would we do that?

albertshau:
what about the sources? Is that coming later?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
gsps1, line 59: in tests, https://github.com/caskdata/hydrator-plugins/blob/develop/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/TableSinkTest.java#L45

we create a TableSinkConfig manually, but since this is constructed using the constructor, the parent class fields `macroFields` and `properties` would be null. 

without this check, it will throw NPE,  we do a similar check here,
https://github.com/caskdata/hydrator-plugins/blob/develop/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java#L85

gsps1:
@albertshau addressed the comments. will send sources in a separate PR.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
albertshau, line 59: ok I see. That seems like just a bad test though. It's testing schema validation through the plugin instead of just testing the SchemaValidator.  Can we change these bad tests?  We should avoid adding logic strictly for tests.

albertshau, line 59: if its just that one test that is failing, I think its enough just to move the test to SchemaValidatorTest.

albertshau:
just one comment. Based on how widespread it is, may make sense to do it in a separate PR.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
gsps1, line 59: i don't see a `SchemaValidatorTest` all the test cases in TableSinkTest, CubeSinkTest and KVTableSinkTest do the similar behavior.

gsps1:
Build green : http://builds.cask.co/browse/HYP-DBT33-4/test

gsps1:
@albertshau I'd do that in separate PR, when i send the PR for batch source changes, will include the changes for tests and do the clean up, merging this now.

PR #319, opened 2016-07-18
Fixing renaming field when alias is not present in config

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-6529
Build: http://builds.cask.co/browse/HYP-DBT32-3

Summary:
- Addressing issue when selected field does not get populated correctly when alias is not specified.
- Fixing Rest end point for getOutputSchema to use request instead of config
- Adding test case for selecting fields without specifying alias.


albertshau:
lgtm

CuriousVini:
Build is green. Merging.

PR #320, opened 2016-07-18
make stream batch source stream name, duration and delay macro enabled

gsps1:
https://issues.cask.co/browse/CDAP-6360

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
@@ -17,6 +17,7 @@
 package co.cask.hydrator.plugin.batch.source;
 
 import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.batch.Input;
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
```diff
@@ -94,7 +96,9 @@
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
     streamBatchConfig.validate();
-    pipelineConfigurer.addStream(new Stream(streamBatchConfig.name));
+    if (!streamBatchConfig.containsMacro(Properties.Stream.NAME)) {
+      pipelineConfigurer.addStream(new Stream(streamBatchConfig.name));
```
gsps1, line 100: @albertshau @chtyim should we allow this in configure or not ? since we are not allowing stream creation with stream names from macro at runtime. would it make sense to disallow this too ?

gsps1:
Build Running: http://builds.cask.co/browse/HYP-DBT34

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java
albertshau, line 100: you mean because cdap doesn't allow stream creation at runtime?  I think its still a valid use case to have a macro as the stream name, it just means that the stream will have to be created by somebody else. We should mention that in the plugin docs though.

albertshau:
add a note in the docs about using a macro in stream name, then lgtm

sreevatsanraman:
Added JIRA https://github.com/caskdata/hydrator-plugins/pull/320 to address comments

PR #321, opened 2016-07-18
[CDAP-6527]:

thegreenfrog:
Fix Connection error caused by attempting to open session before establishing connection


sagarkapare:
LGTM. Please file a JIRA for adding a test case for this issue.

PR #324, opened 2016-07-20
[CDAP-6566]: Handle more exceptions in SSHAction and trim carriage return

thegreenfrog:
Trimmed trailing carriage return from end of output of the SSH command. Handled more exceptions during the setup process. If remote command has exit code greater than 0, then also throws an exception

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
@@ -85,8 +85,16 @@ public class SSHAction extends Action {
 
       LOG.debug("Output: {}", out);
       LOG.error("Errors: {}", err);
+      Integer exitCode = session.getExitStatus();
+      if (exitCode != null && exitCode != 0) {
+        throw new IOException(String.format("Error: command %s running on hostname %s finished with exit code: %d",
+                                            config.command, config.host, exitCode));
+      }
 
       if (config.output != null) {
+        //removes the carriage return at the end of the line
+        out = out.endsWith("\n") ? out.substring(0, out.length() - 1) : out;
+
         context.getArguments().set(config.output, out);
       }
     } finally {


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -63,7 +63,11 @@ public SSHAction(SSHActionConfig config) {
   public void run(final ActionContext context) throws Exception {
     Connection connection = new Connection(config.host);
     try {
-      connection.connect();
+      try {
+        connection.connect();
+      } catch (IOException e) {
+        throw new IOException(String.format("Attempt to connect to host %s failed", config.host));
```
wolf31o2: Will this surface _why_ the connection failed?

albertshau: why do we need to catch and then throw an exception without the original cause? Is there something wrong with the original exception?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -73,7 +77,11 @@ public void run(final ActionContext context) throws Exception {
 
       LOG.debug("Connection established with the host {}", config.host);
       Session session = connection.openSession();
-      session.execCommand(config.command);
+      try {
+        session.execCommand(config.command);
+      } catch (IOException e) {
+        throw new IOException(String.format("Exception thrown while executing command: %s", e.getMessage()));
```
albertshau: same question here

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -85,9 +93,14 @@ public void run(final ActionContext context) throws Exception {
...
       LOG.error("Errors: {}", err);
+      Integer exitCode = session.getExitStatus();
+      if (exitCode != null && exitCode > 0) {
+        throw new IOException(String.format("Error: command finished with exit code: %d", exitCode));
+      }
 
       if (config.output != null) {
-        context.getArguments().set(config.output, out);
+        //regex removes the carriage return at the end of the line
+        context.getArguments().set(config.output, out.replaceAll("\\s+$", ""));
```
wolf31o2: Removes carriage return and any other whitespace characters at the end of the output. Expected?

albertshau: the ssh library always adds a newline at the end of stdout? This will actually do more than just remove a newline, it will remove all whitespace from the end stdout, including any actual whitespace returned by the command. 

If we know the library always appends a newline, you can just check that out ends with a newline and do a substring.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: oh yeah it doesn't show the stack trace. will update it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: ok that works too.

thegreenfrog:
Ok changes made. Take another look @albertshau @wolf31o2 !

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -85,8 +85,16 @@ public void run(final ActionContext context) throws Exception {
...
       LOG.error("Errors: {}", err);
+      Integer exitCode = session.getExitStatus();
+      if (exitCode != null && exitCode > 0) {
+        throw new IOException(String.format("Error: command finished with exit code: %d", exitCode));
+      }
 
       if (config.output != null) {
+        //removes the carriage return at the end of the line
+        String isThereCarriage = out.length() > 1 ? out.substring(out.length() - 1) : out;
+        out = isThereCarriage.equals("\n") ? out.substring(0, out.lastIndexOf("\n")) : out;
```
albertshau: just do `out.endsWith("\n") ? out.substring(0, out.length() - 1) : out`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -85,8 +85,16 @@ public void run(final ActionContext context) throws Exception {
 
       LOG.debug("Output: {}", out);
       LOG.error("Errors: {}", err);
+      Integer exitCode = session.getExitStatus();
+      if (exitCode != null && exitCode > 0) {
+        throw new IOException(String.format("Error: command finished with exit code: %d", exitCode));
```
albertshau: lets put the actual command and hostname in the error message. Otherwise somebody sees the log, but then has to figure out which pipeline and action its from in order to find the actual command.

thegreenfrog:
@albertshau updated the code based on your comments!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -85,8 +85,16 @@ public void run(final ActionContext context) throws Exception {
 
       LOG.debug("Output: {}", out);
       LOG.error("Errors: {}", err);
+      Integer exitCode = session.getExitStatus();
+      if (exitCode != null && exitCode > 0) {
```
albertshau: most systems have non-negative exit codes, but its still safer just to check the exitCode != 0 in case negative exit codes are allowed.

albertshau:
just one more minor comment, otherwise lgtm. fix then merge.

PR #325, opened 2016-07-20
Fixes widget json for plugins that uses non-editable-schema-editor to conform avro schema spec

ajainarayanan:


sreevatsanraman:
LGTM

PR #326, opened 2016-07-20
Do not add partition to the underlying PFS through SnapshotFileSet

sagarkapare:
`SnapshotFileSet` embeds the `PartitionFileSet`.
1. When MapReduce job finishes `MapReduceRuntimeService.destroy` gets called.
2. `destroy` method in turn calls the `commitOutput` which adds a partition to the embedded PFS through `PartitionFileSet.onSuccess` method. 
3. `destroy` method then calls the `MapReduce.destroy` method of the user's MR program.
4. When pipeline is using `SnapshotFileSet` (through `SnapshotFileBatchSink`), `destroy` method from step 3 results into `SnapshotFileSet.onSuccess` method which again tries to create the partition already created in the step 2 above.

This is causing `ETLSnapshotTestRun.testMultiSnapshotOutput` (unit test case) and `BatchAggregatorTest` (ITN) failures.

Fix is to not to create partition in the `SnapshotFileSet`.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/dataset/SnapshotFileSet.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/dataset/SnapshotFileSet.java
@@ -96,10 +96,6 @@ public class SnapshotFileSet {
   public void onSuccess(long snapshotTime) throws IOException, InterruptedException {
     Location lock = lock();
     try {
-      // add partition
-      PartitionKey partitionKey = PartitionKey.builder().addLongField("snapshot", snapshotTime).build();
-      files.addPartition(partitionKey, String.valueOf(snapshotTime));
-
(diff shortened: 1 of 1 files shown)


anwar6953:
We have a check against this in PartitionedFileSetDataset:
https://github.com/caskdata/cdap/blob/develop/cdap-data-fabric/src/main/java/co/cask/cdap/data2/dataset2/lib/partitioned/PartitionedFileSetDataset.java#L248-L253

You didn't see that warning? Perhaps the path is not the same, which is why that equality didn't match?

anwar6953:
LGTM. Perhaps file a JIRA to track why the code in PartitionedFileSetDataset.java intended to handle this isn't working.

sagarkapare:
Filed JIRA https://issues.cask.co/browse/CDAP-6578. Merging this PR.

PR #328, opened 2016-07-21
[CDAP-6597]: Fix Error logs in SSHAction

thegreenfrog:
Only show 'Errors: ' message after SSHAction execution if there are errors to be shown

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
@@ -83,8 +83,12 @@ public class SSHAction extends Action {
       String out = CharStreams.toString(outBuffer);
       String err = CharStreams.toString(errBuffer);
 
-      LOG.debug("Output: {}", out);
-      LOG.error("Errors: {}", err);
+      if (out.length() > 0) {
+        LOG.debug("Stdout: {}", out);
+      }
+      if (err.length() > 0) {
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -84,7 +84,9 @@ public void run(final ActionContext context) throws Exception {
       String err = CharStreams.toString(errBuffer);
 
       LOG.debug("Output: {}", out);
```
albertshau: same thing for stdout?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -84,7 +84,9 @@ public void run(final ActionContext context) throws Exception {
...
-      LOG.error("Errors: {}", err);
+      if (err.length() > 0) {
+        LOG.error("Errors: {}", err);
```
albertshau: its also more accurate to call this stderr and output stdout. Lots of times people aren't very judicious about what goes to stdout vs stderr

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
thegreenfrog: fair point.

albertshau:
lgtm

PR #329, opened 2016-07-21
CDAP-5144 Remove Script Transform plugin

awholegunch:
Deprecated back in CDAP 3.3.0, to be removed as part of CDAP 3.5.0.

Fix for https://issues.cask.co/browse/CDAP-5144


anwar6953:
I wonder if there's any implementation difference between this and the JavaScriptTransform, that we'd be missing by removing this.
That's the only thing I think we should check.

anwar6953:
Just checked - there's no functionality loss, with just the JavaScriptTransform.
LGTM

PR #330, opened 2016-07-21
[CDAP-6360] Removed old macros and updated plugins that used them

knazir:
Several Hydrator plugins used to extend MacroConfig instead of PluginConfig in order to support macros. This made all of their properties macro-enabled. MacroConfig and all outdated macro-related classes have been removed from Hydrator and the plugins that used them have been updated to defer certain checks/validations to runtime if macro syntax is detected.


albertshau:
lgtm 💯

knazir:
Guild is breen: http://builds.cask.co/browse/HYP-DBT40-5 going to merge.

PR #334, opened 2016-07-24
Feature/bump hydrator version in develop to 1.5.0-SNAPSHOT

sreevatsanraman:


anwar6953:
Know what the deal here is?
https://github.com/caskdata/hydrator-plugins/blob/1b7686d6d3d50773af69022090f9efa261d301cd/copybookreader-plugins/pom.xml#L34
Can it be removed?

anwar6953:
Seems like that module should use whatever's defined in parent pom.
Instead of redefining the version of cdap to use.

anwar6953:
lgtm

PR #336, opened 2016-07-24
Remove project-version property from copybook

sreevatsanraman:
- Specify dependency in parent pom for cdap project. Then project-version property can be removed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,18 @@
       </dependency>
       <dependency>
         <groupId>co.cask.cdap</groupId>
+        <artifactId>cdap-api</artifactId>
+        <version>${cdap.version}</version>
+        <scope>provided</scope>
+      </dependency>
+      <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -162,6 +162,18 @@
...
         <groupId>co.cask.cdap</groupId>
+        <artifactId>cdap-api</artifactId>
+        <version>${cdap.version}</version>
+        <scope>provided</scope>
+      </dependency>
+      <dependency>
+        <groupId>co.cask.cdap</groupId>
+        <artifactId>cdap-etl-proto</artifactId>
+        <version>${cdap.version}</version>
+        <scope>provided</scope>
```
anwar6953, line 173: is this provided?

anwar6953, line 173: same for develop, if necessary (in case its not provided by program classloader on distributed cdap).

PR #338, opened 2016-07-25
HYDRA-327 : do macro check before null check in BatchWritableSink

gsps1:
After the change in default value assignment in CDAP. we get null values for fields with macros during configure time. so its good to skip those if they are macros.

Issue: https://issues.cask.co/browse/HYDRA-327

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
@@ -47,18 +47,18 @@ public abstract class BatchWritableSink<IN, KEY_OUT, VAL_OUT> extends BatchSink<
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
-    String datasetName = getProperties().get(Properties.BatchReadableWritable.NAME);
-    Preconditions.checkArgument(datasetName != null && !datasetName.isEmpty(), "Dataset name must be given.");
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchWritableSink.java
```diff
@@ -47,18 +47,18 @@ protected BatchWritableSink(BatchReadableWritableConfig batchReadableWritableCon
...
-    String datasetType = getProperties().get(Properties.BatchReadableWritable.TYPE);
-    Preconditions.checkArgument(datasetType != null && !datasetType.isEmpty(), "Dataset type must be given.");
-
-    Map<String, String> properties = Maps.newHashMap(getProperties());
-    properties.remove(Properties.BatchReadableWritable.NAME);
-    properties.remove(Properties.BatchReadableWritable.TYPE);
-
     // null check for tests, as macro fields map and properties are null in test cases
     if (batchReadableWritableConfig.getProperties() != null &&
       !batchReadableWritableConfig.containsMacro(Properties.BatchReadableWritable.NAME)) {
```
knazir: Would just doing the ```!containsMacro(...)``` check be good enough? Name is not nullable so it would only be null in the case macro syntax was provided?

albertshau:
lgtm

gsps1:
build failing on xm reader test and excel tests. looking into it.

gsps1:
https://issues.cask.co/browse/HYDRA-357
https://issues.cask.co/browse/HYDRA-356

PR #339, opened 2016-07-26
Refactoring Joiner testcases

CuriousVini:
Issue: https://issues.cask.co/browse/HYDRA-308
Build: http://builds.cask.co/browse/HYP-DBT47-2

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
@@ -51,9 +51,28 @@ import java.util.concurrent.TimeUnit;
  * Tests for Joiner.
  */
 public class JoinerTestRun extends ETLBatchTestBase {
+  private static final Schema FILM_SCHEMA = Schema.recordOf(
+    "film",
+    Schema.Field.of("film_id", Schema.of(Schema.Type.STRING)),
+    Schema.Field.of("film_name", Schema.of(Schema.Type.STRING)));
+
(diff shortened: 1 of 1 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -612,64 +497,81 @@ private void verifyOuterJoinWithoutRequiredInputs(Schema outputSchema, TimeParti
...
+                                                  getBobRecord2(avroOutputSchema),
+                                                  getAlexRecord1(avroOutputSchema),
+                                                  getAlexRecord2(avroOutputSchema),
+                                                  getCathieRecord1(avroOutputSchema),
+                                                  getAvatarRecord1(avroOutputSchema),
+                                                  humtumRecord, sultanRecord);
+    Assert.assertEquals(expected, actual);
+  }
+
+  GenericRecord getBobRecord1(org.apache.avro.Schema avroOutputSchema) {
```
albertshau: these can be constants right? or do we need new instances of them for some reason?

albertshau:
just one question, otherwise lgtm

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerTestRun.java
CuriousVini: As we use different `avroOutputSchema` depending on type of the join (For outer joins, we use nullable fields), we need new instances of records.

CuriousVini:
Build is green. Merging.

PR #340, opened 2016-07-26
Adding validations for schema type for join keys

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-6551
Build: http://builds.cask.co/browse/HYP-DBT48-4

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
@@ -70,7 +70,7 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
     Map<String, Schema> inputSchemas = stageConfigurer.getInputSchemas();
-    init();
+    init(inputSchemas);
     //validate the input schema and get the output schema for it
     stageConfigurer.setOutputSchema(getOutputSchema(inputSchemas));
   }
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/joiner/Joiner.java
```diff
@@ -140,16 +140,38 @@ public StructuredRecord merge(StructuredRecord joinKey, Iterable<JoinElement<Str
...
+        Schema.Field field = schema.getField(joinKey);
+        schemaList.add(field.getSchema());
+      }
+      if (prevSchemaList != null && !prevSchemaList.equals(schemaList)) {
+        throw new IllegalArgumentException(String.format("For stage %s, Schemas of joinKeys %s are expected to be: " +
+                                                           "%s, but found: %s",
+                                                         stageName, entry.getValue(), prevSchemaList.toString(),
+                                                         schemaList.toString()));
+      }
+      prevSchemaList = new ArrayList<>(schemaList);
```
albertshau: don't think you need to copy, since schemaList is newly created in each loop.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/joiner/JoinerConfigTest.java
```diff
@@ -184,4 +184,25 @@ public void testJoinerConfigWithDuplicateOutputFields() {
...
+    Schema filmCategorySchema = Schema.recordOf(
+      "filmCategory",
+      Schema.Field.of("film_id", Schema.of(Schema.Type.LONG)),
+      Schema.Field.of("film_name", Schema.of(Schema.Type.STRING)),
+      Schema.Field.of("category_name", Schema.of(Schema.Type.STRING)));
+
+    JoinerConfig config = new JoinerConfig("film.film_id=filmActor.film_id=filmCategory.film_id&" +
+                                             "film.film_name=filmActor.film_name=filmCategory.film_name",
+                                           selectedFields, "film,filmActor,filmCategory");
+    MockPipelineConfigurer pipelineConfigurer = new MockPipelineConfigurer(
```
albertshau: we should avoid adding new usages of this class and eventually remove it, as it guarantees that whenever the Hydrator PipelineConfigurer API changes, this repo will break.

I think we can just make validateJoinKeySchemas package private and call it directly from this class.

albertshau:
a couple of small comments, otherwise lgtm

CuriousVini:
Addressed comments. Build is green. Squashing and merging.

PR #342, opened 2016-07-26
[CDAP-6661] Order by aggregate plugin

ananya-bhattacharya:
The Order By a aggregator plugin that sorts structured records based on fields and order specified by user. 
For design and use case please refer - https://wiki.cask.co/display/CE/Order+By

### DIFF
--- a/orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
+++ b/orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.aggregator;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.format.StructuredRecordStringConverter;
+import com.google.common.base.Joiner;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.reflect.TypeToken;
+import org.apache.commons.lang.ArrayUtils;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.Writable;
+import org.apache.hadoop.io.WritableComparable;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * CustomWritable for the composite key.
+ */
+public class CompositeKey implements Writable, WritableComparable<CompositeKey> {
+
+  private static final String FIELD_DELIMITER = "->";
+  private final Gson gson = new GsonBuilder().create();
+  private final Type mapType = new TypeToken<LinkedHashMap>() {  }.getType();
+  private final Type schemaType = new TypeToken<Schema>() {  }.getType();
+  private String structureRecordJSON;
+  private String sortFieldsJSON;
+  private String schemaJSON;
+
+  public CompositeKey() {
+  }
+
+  public CompositeKey(Text structureRecordJSON, String sortFieldsJSON, String schemaJSON) {
+    this.structureRecordJSON = structureRecordJSON.toString();
+    this.sortFieldsJSON = sortFieldsJSON;
+    this.schemaJSON = schemaJSON;
+  }
+
+  public static int getSortingOrder(String order) {
+    if (order.equalsIgnoreCase("asc")) {
+      return 1;
+    } else {
+      return -1;
+    }
+  }
+
+  /**
+   * Get sorting order for structured records depending on the type of value contained in key
+   *
+   * @param key               contains the value to sort the data
+   * @param value             determines whether the objects are to be sorted in ascending or descending order for key
+   * @param structuredRecord1 Structured record
+   * @param structuredRecord2 Structured record to compare against
+   * @return sorting order of the complete structured record
+   */
+  public static int compareRecords(String key, String value, StructuredRecord structuredRecord1,
+                                   StructuredRecord structuredRecord2) {
+    int res;
+    Schema.Type type = structuredRecord1.getSchema().getField(key).getSchema().getType();
+    if (type.equals(Schema.Type.STRING)) {
+      res = getSortingOrder(value) * ((String) structuredRecord1.get(key))
+        .compareTo((String) structuredRecord2.get(key));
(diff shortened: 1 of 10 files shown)


orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+import java.lang.reflect.Type;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * CustomWritable for the composite key.
+ */
+public class CompositeKey implements Writable, WritableComparable<CompositeKey> {
+
+  private static final String FIELD_DELIMITER = "->";
```
nitinmotgi: Why such a delimiter ? Can't we use something standard ?

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+  private final Type mapType = new TypeToken<LinkedHashMap>() {  }.getType();
+  private final Type schemaType = new TypeToken<Schema>() {  }.getType();
+  private String structureRecordJSON;
+  private String sortFieldsJSON;
+  private String schemaJSON;
+
+  public CompositeKey() {
+  }
+
+  public CompositeKey(Text structureRecordJSON, String sortFieldsJSON, String schemaJSON) {
```
nitinmotgi: structuredRecord not structureRecord

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Get sorting order for structured records depending on the type of value contained in key
+   *
+   * @param key               contains the value to sort the data
+   * @param value             determines whether the objects are to be sorted in ascending or descending order for key
+   * @param structuredRecord1 Structured record
+   * @param structuredRecord2 Structured record to compare against
+   * @return sorting order of the complete structured record
+   */
+  public static int compareRecords(String key, String value, StructuredRecord structuredRecord1,
```
nitinmotgi, line 80: Comparison has to support recursive complex records. So you might want to implement a recursive function for doing that. StructuredRecord can have basic types as well as Record, Union, etc.

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+      res = getSortingOrder(value) * Byte.compare((Byte) structuredRecord1.get(key),
+                                                  (Byte) structuredRecord2.get(key));
+    } else {
+      throw new IllegalArgumentException("Sorting can be performed only on the following CDAP data types:  " +
+                                           "String, Int, Long, Float, Double, Byte");
+    }
+    return res;
+  }
+
+  @Override
```
nitinmotgi, line 141: Doc

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+  public void write(DataOutput dataOutput) throws IOException {
```
nitinmotgi, line 145: Doc

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
nitinmotgi, line 80: Take it back. You have implemented it below.

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+   * @return structured records sorted on the keys
+   */
+  private int compareNestedRecords(String key, String value, StructuredRecord structuredRecord1,
+                                   StructuredRecord structuredRecord2) {
+
+    if (key.contains(FIELD_DELIMITER)) {
+      String[] nestedRecord = key.split(FIELD_DELIMITER);
+      Schema.Field field = structuredRecord1.getSchema().getField(nestedRecord[0].trim());
+      Schema.Type type = field.getSchema().getType();
+      if (type.equals(Schema.Type.RECORD)) {
```
nitinmotgi, line 213: Don't you want to handle UNION and such ?

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
```diff
@@ -0,0 +1,198 @@
...
+                                    structuredRecord2);
+      } else {
+        throw new IllegalArgumentException("Nested structured defined by -> should be of type RECORD");
+      }
+    } else {
+      return compareRecords(key, value, structuredRecord1, structuredRecord2);
+    }
+  }
+
+  public String getStructureRecordJSON() {
```
nitinmotgi: StructuredRecord not StructureRecord

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKeyDriver.java
```diff
@@ -0,0 +1,46 @@
...
+public class CompositeKeyDriver {
+  public static void main(String[] args) throws Exception {
+    Job job = Job.getInstance(new Configuration(), "Order By");
+    job.setJarByClass(CompositeKeyDriver.class);
+
+    job.setMapperClass(CompositeKeyMapper.class);
+    job.setMapOutputKeyClass(CompositeKey.class);
+    job.setMapOutputValueClass(Text.class);
+
+    job.setPartitionerClass(CompositeKeyPartitioner.class);
```
nitinmotgi, line 35: @albertshau how can we get this into planner or app -- not sure which one. What do you think is the best approach to be able set the different comparator / partitioner when OrderBy is included. Please advice.

albertshau, line 35: what are we doing here? This is just a mapreduce job and not a plugin at all.

Anything we do needs to work for both Spark and for MapReduce. We should not be relying on such mapreduce specific features.

Can't order by be implemented by simply grouping with one partition (reducer)?

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKey.java
ananya-bhattacharya, line 213: The if(#172) statement checks, if the path specified by user for selecting the nested field is valid, by checking the parent is of type RECORD.
For the current implementation we are only supporting simple types with nullable (as UNION): String, Int, Long, Short, Float, Double, Byte as mentioned in the use-case.

orderby-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/CompositeKeyDriver.java
ananya-bhattacharya, line 35: To sort the structured records using sort and group comparator we had used the MR job as a first part of the implementation of the Order-by plugin.
https://wiki.cask.co/display/CE/Order+By -  Implementation Tips #1 and #2
Please let me know, if this approach has to be changed and if we need to explore the plugin classes for implementing the order-by plugin.

abhinavbansal246:
Closing PR against release, please open against develop

PR #344, opened 2016-07-26
HYDRA-338 fix parent lower bound versions

albertshau:
These plugins require 3.5.x versions of the Hydrator apps, as they
contain various implementations of the new plugin types, like join.


CuriousVini:
LGTM.

PR #346, opened 2016-07-26
[CDAP-6744][HYDRA-341]: Email Action Test Case uses deprecated macros function. Hydrator Maven Build doesn't run test.

thegreenfrog:
Email Action Test uses deprecated `runtime` macro function. Changed it to `logicalStartTime`


sagarkapare:
Why is it not failing on bamboo builds?

thegreenfrog:
That is being taken care of in another JIRA. Apparently it was not included in the test build.

sagarkapare:
Better to address https://issues.cask.co/browse/HYDRA-341 in the same PR.

thegreenfrog:
fixed both issues @sagarkapare

sagarkapare:
LGTM 👍

PR #348, opened 2016-07-27
Update SSHAction

knazir:
SSHAction reads the private key file from secure storage instead of a file path. Also updated docs to reflect property name changes. "passphrase" may be more appropriate than password. 

JIRA:  https://issues.cask.co/browse/HYDRA-169

### DIFF
--- a/core-plugins/docs/SSHAction-action.md
+++ b/core-plugins/docs/SSHAction-action.md
@@ -21,15 +21,17 @@ Properties
 
 **user:** The username used to connect to host.
 
-**privateKeyFile:** The file path to Private key.
+**privateKey:** The private key to be used to perform the secure shell action. Users can also specify a macro that will
+pull the private key from the secure key management store in CDAP such as ${secure(myPrivateKey)}.
 
-**password:** The password associated with private key.
+**passphrase:** Passphrase used to decrypt the provided private key in "privateKey".
 
 **command:** The command to be executed on the remote host. This includes the filepath of the script and any arguments
 needing to be passed.
 
-**output:** Script output variable used as the key in the instance the output from a script needs to be saved
-and passed through the Action Context.
+**outputKey:** The key used to store the output of the command run by the action.
+Plugins that run at later stages in the pipeline can retrieve the command's output using this key through
+macro substitution (e.g. ${outputKey} where "outputKey" is the key specified). Defaults to "sshOutput".
 
 Example
 -------
@@ -48,11 +50,12 @@ This example runs a script on demo@example.com:
           },
           "properties": {
               "host": "example.com",
-              "password": "pass",
               "port": "22",
               "user": "demo",
-              "privateKeyFile": "/Users/Jose/.ssh/id_rsa",
-              "command": "~/scripts/remoteScript.sh"
+              "privateKey": "${secure(myPrivateKey)}",
+              "passphrase": "pass",
+              "command": "~/scripts/remoteScript.sh",
+              "outputKey": "myActionOutput"
             }
           }
     }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
@@ -21,6 +21,7 @@ import ch.ethz.ssh2.Session;
 import ch.ethz.ssh2.StreamGobbler;
 
 import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.plugin.PluginConfig;
@@ -30,13 +31,11 @@ import co.cask.cdap.etl.api.action.Action;
 import co.cask.cdap.etl.api.action.ActionContext;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
-import com.google.common.base.Strings;
 import com.google.common.io.CharStreams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -61,12 +60,14 @@ public class SSHAction extends Action {
 
   @Override
   public void run(final ActionContext context) throws Exception {
+    // now that macros have been substituted, try validation again
+    config.validate();
+
     Connection connection = new Connection(config.host);
     try {
       connection.connect();
 
-      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
-      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+      if (!connection.authenticateWithPublicKey(config.user, config.privateKey.toCharArray(), config.passphrase)) {
         throw new IOException(String.format("SSH authentication error when connecting to %s@%s on port %d",
                                             config.user, config.host, config.port));
       }
@@ -95,12 +96,9 @@ public class SSHAction extends Action {
                                             config.command, config.host, exitCode));
       }
 
-      if (config.output != null) {
-        //removes the carriage return at the end of the line
-        out = out.endsWith("\n") ? out.substring(0, out.length() - 1) : out;
-
-        context.getArguments().set(config.output, out);
-      }
+      //removes the carriage return at the end of the line
+      out = out.endsWith("\n") ? out.substring(0, out.length() - 1) : out;
+      context.getArguments().set(config.outputKey, out);
     } finally {
       connection.close();
     }
@@ -108,7 +106,9 @@ public class SSHAction extends Action {
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
-    this.config.validate();
+    if (!config.containsMacro("port")) {
+      config.validate();
+    }
   }
 
   /**
@@ -116,44 +116,54 @@ public class SSHAction extends Action {
    */
   public static class SSHActionConfig extends PluginConfig {
     @Description("Command to be executed on the remote host. Should include filepath of script and any arguments")
+    @Macro
     private String command;
 
     @Description("Host name of the remote machine where the command needs to be executed.")
+    @Macro
     private String host;
 
     @Nullable
     @Description("Port to connect to. Defaults to 22")
+    @Macro
     private Integer port;
 
     @Description("User name used to connect to host")
+    @Macro
     private String user;
 
-    @Description("File path to Private key")
-    private String privateKeyFile;
+    @Description("The private key to be used to perform the secure shell action. Users can also specify a macro that " +
+      "will pull the private key from the secure key management store in CDAP such as ${secure(myPrivateKey)}.")
+    @Macro
+    private String privateKey;
 
     @Nullable
-    @Description("Password associated with private key")
-    private String password;
+    @Description("Passphrase used to decrypt the provided private key associated with \"privateKeyLookupKey\"")
+    @Macro
+    private String passphrase;
 
     @Nullable
-    @Description("Variable used to hold desired key name to be associated with script output in the instance " +
-      "the script output contains information needed to be stored in the ActionContext for later use in the pipeline.")
-    private String output;
+    @Description("The key used to store the output of the command run by the action. Plugins that run at later " +
+      "stages in the pipeline can retrieve the command's output using this key through macro substitution " +
+      "${outputKey} where \"outputKey\" is the key specified. Defaults to \"sshOutput\".")
+    @Macro
+    private String outputKey;
 
     public SSHActionConfig() {
-      port = 22;
+      this.port = 22;
+      this.outputKey = "sshOutput";
     }
 
     @VisibleForTesting
-    public SSHActionConfig(String command, String host, String user, String privateKeyFile, int port, String password,
-                           String output) {
+    public SSHActionConfig(String command, String host, String user, String privateKeyFileLookupKey, Integer port,
+                           String passphrase, String outputKey) {
       this.command = command;
       this.host = host;
-      this.port = port;
+      this.port = (port == null) ? 22 : port;
       this.user = user;
-      this.privateKeyFile = privateKeyFile;
-      this.password = password;
-      this.output = output;
+      this.privateKey = privateKeyFileLookupKey;
+      this.passphrase = passphrase;
+      this.outputKey = outputKey;
     }
 
     public void validate() {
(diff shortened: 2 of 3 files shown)


knazir:
Manually testing now and build running: http://builds.cask.co/browse/HYP-DBT50-2

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -116,49 +124,52 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    private String passphrase;
 
     @Nullable
     @Description("Variable used to hold desired key name to be associated with script output in the instance " +
       "the script output contains information needed to be stored in the ActionContext for later use in the pipeline.")
+    @Macro
     private String output;
 
-    public SSHActionConfig() {
-      port = 22;
```
anwar6953, line 144: this isn't the value used by default?

core-plugins/docs/SSHAction-action.md
```diff
@@ -21,9 +21,9 @@ Properties
 
 **user:** The username used to connect to host.
 
-**privateKeyFile:** The file path to Private key.
+**privateKeyFileLookupKey:** The file path to Private key.
```
bdmogal: nit: lowercase p

bdmogal: Also, can this parameter be named better? Seems very verbose for something that's exposed to users. What does it do? Also seems like it could use some more detailed docs.

core-plugins/docs/SSHAction-action.md
```diff
@@ -21,9 +21,9 @@ Properties
...
 
-**password:** The password associated with private key.
+**passphrase:** The password associated with private key.
```
bdmogal: Is this related to the above property? If so, should mention the relationship.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
knazir, line 144: would this constructor ever be called? o-o

anwar6953, line 144: I think so.
But you're saying that the default value of 22 is not being set... so... not sure.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -61,12 +62,19 @@ public SSHAction(SSHActionConfig config) {
...
+
     Connection connection = new Connection(config.host);
     try {
       connection.connect();
 
-      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
-      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+      // get private key from secure store using user-specified key
+      char[] privateKey =
+        new String(context.getSecureData(context.getNamespace(), config.privateKeyFileLookupKey).get()).toCharArray();
```
bdmogal: Should specify charset while converting from `byte[]` to `String`. Use `Bytes.toString(...)`.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
anwar6953, line 144: Seems like other code depends on it...
https://github.com/caskdata/hydrator-plugins/blob/develop/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java#L103-L108

Note: regex to find all default constructors (intellij has regex search):
`public\ \w+\(\)\ \{`

anwar6953, line 144: What was the symptom observed in this JIRA?
https://issues.cask.co/browse/HYDRA-354
NPE? Or some other error?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -116,49 +124,52 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
 
     @Nullable
-    @Description("Password associated with private key")
-    private String password;
+    @Description("Passphrase associated with private key")
+    @Macro
+    private String passphrase;
 
     @Nullable
     @Description("Variable used to hold desired key name to be associated with script output in the instance " +
```
bdmogal: Unrelated to your PR, but there are a couple of problems with this description:
1. Its hard to understand because of the long sentence.
2. It uses terms like `ActionContext`, which are very code specific. 

What does it do? Also, is it too internal to be exposed to users?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
knazir, line 144: Not sure what I did to be honest, can't reproduce the behavior. I may have just used " " a few times. Will update the PR, thanks.

core-plugins/docs/SSHAction-action.md
knazir: Yeah fair point, we could try `privateKeyLookupKey`

core-plugins/docs/SSHAction-action.md
knazir: Yes it is, will update that.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
knazir: From my understanding it's meant to be exposed to users to (almost exclusively) substitute plugin properties with a macro `${output}` where "output" is the key they specified. I agree it could use some work. Will try to update it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -116,49 +125,57 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
 
     @Nullable
-    @Description("Variable used to hold desired key name to be associated with script output in the instance " +
-      "the script output contains information needed to be stored in the ActionContext for later use in the pipeline.")
-    private String output;
+    @Description("The key used to store the output of the command run by the action. Plugins that run at later " +
+      "stages in the pipeline can retrieve the command's output using this key through macro substitution " +
+      "${outputKey} where \"outputKey\" is the key specified.")
+    @Macro
+    private String outputKey;
```
knazir, line 150: Since it's nullable, should a default key be given if none is provided like "sshOutput" or something?

core-plugins/docs/SSHAction-action.md
```diff
@@ -21,15 +21,16 @@ Properties
 
 **user:** The username used to connect to host.
 
-**privateKeyFile:** The file path to Private key.
+**privateKeyLookupKey:** Key used to lookup private key from secure storage.
```
bdmogal: Now it has 2 keys :) Actually, is it mandatory that this will always be a macro? If someone doesn't care about security, couldn't they just dump their private key in a textarea? I think we can just call it `privateKey`. The description can be something like - "The private key to be used to perform the secure shell action. Users can also specify a key that will be looked up from the secure key management store in CDAP". However, in that case you'll need a way to distinguish between whether its a secure key or just text.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -61,12 +62,20 @@ public SSHAction(SSHActionConfig config) {
...
 
-      String password = Strings.isNullOrEmpty(config.password) ? null : config.password;
-      if (!connection.authenticateWithPublicKey(config.user, new File(config.privateKeyFile), password)) {
+      // get private key from secure store using user-specified key
+      char[] privateKey =
+        Bytes.toString(context.getSecureData(context.getNamespace(), config.privateKeyLookupKey).get())
+          .toCharArray();
+      String passphrase = Strings.isNullOrEmpty(config.passphrase) ? "" : config.passphrase;
+
+      if (!connection.authenticateWithPublicKey(config.user, privateKey, passphrase)) {
```
bdmogal: so the method says authenticate with public key, but we ask for a private key?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
bdmogal, line 150: Yes, and it should be specified in the description.

core-plugins/docs/SSHAction-action.md
knazir: That's a good point, I think I got too caught up in secure storage/macros.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
knazir: Right, it's a part of the ganymed2 SSH library. It's very poorly named...

bdmogal: but does it work with public keys? If so we should accept a public key.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -116,49 +114,59 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-    private String output;
+    @Description("The key used to store the output of the command run by the action. Plugins that run at later " +
+      "stages in the pipeline can retrieve the command's output using this key through macro substitution " +
+      "${outputKey} where \"outputKey\" is the key specified. Defaults to \"sshOutput\".")
+    @Macro
+    private String outputKey;
 
     public SSHActionConfig() {
       port = 22;
+      outputKey = "sshOutput";
```
gsps1: use `this.outputKey`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -116,49 +114,59 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-      this.output = output;
+      this.privateKey = privateKeyFileLookupKey;
+      this.passphrase = passphrase;
+      this.outputKey = outputKey;
     }
 
     public void validate() {
       //check that port is not negative
-      if (port < 0) {
+      if (!containsMacro("port") && port < 0) {
```
gsps1: when you are calling  validate during runtime. `containsMacro("port")` will still be true even and you wouldn't check port if its negative or not.

knazir: Oh, I thought that we went with an empty set at runtime. Just checked PluginInstantiator and that's not the case.. will fix. Should the same check `if (port < 0)` just be copied to the run method @gsps1?

core-plugins/docs/SSHAction-action.md
```diff
@@ -21,15 +21,17 @@ Properties
 
 **user:** The username used to connect to host.
 
-**privateKeyFile:** The file path to Private key.
+**privateKey:** The private key to be used to perform the secure shell action. Users can also specify a macro that will
```
gsps1, line 24: it might be better to say `Users can store the private key in CDAP's secure store and use macro to read the private key from the secure store using ...` ask @awholegunch for review

gsps1:
btw though unrelated to your changes, i noticed we are not closing the session, it's important to do so.  can you make that change as well?

Edit : seems like connection close closes the session too.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
gsps1, line 144: @knazir so we use reflection to instantiate and set values. but internally it will call default constructor before trying to set values for the variables through reflection. so say the port was not provided in the config, due to this default constructor being called, this config will have a value of `22` for port and whatever you set inside.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
knazir: Nope, it's just poorly named:

@param pemPrivateKey A <code>char[]</code> containing a DSA or RSA private key of the user in OpenSSH key format (PEM, you can't miss the "-----BEGIN DSA PRIVATE KEY-----" or "-----BEGIN RSA PRIVATE KEY-----" tag). The char array may contain linebreaks/linefeeds.

```
public synchronized boolean authenticateWithPublicKey(String user, char[] pemPrivateKey, String password);
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -95,68 +96,82 @@ public void run(final ActionContext context) throws Exception {
...
-      this.password = password;
-      this.output = output;
+      this.privateKey = privateKeyFileLookupKey;
+      this.passphrase = passphrase;
+      this.outputKey = outputKey;
+    }
+
+    public void validateConfigureTime() {
+      //check that port is not negative
+      if (!containsMacro("port") && port < 0) {
```
gsps1: i'd say just have validate only do `if (port < 0)` and in configurePipeline call validate if `config.containsMacro("port")`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
```diff
@@ -95,65 +96,74 @@ public void run(final ActionContext context) throws Exception {
...
       }
 
-      if (config.output != null) {
-        //removes the carriage return at the end of the line
-        out = out.endsWith("\n") ? out.substring(0, out.length() - 1) : out;
-
-        context.getArguments().set(config.output, out);
-      }
+      //removes the carriage return at the end of the line
+      out = out.endsWith("\n") ? out.substring(0, out.length() - 1) : out;
```
gsps1, line 100: @sagarkapare should we handle multiple carriage return at the end ?

knazir:
Manual testing looked good. Hydrator plugins tests are still broken so can't get a green build...

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/SSHAction.java
gsps1, line 100: spoke with sagar, we decided that's not needed

gsps1:
code changes LGTM, please squash before merging.

PR #350, opened 2016-07-29
Added errorDataset property to XMLParser json

ananya-bhattacharya:
Fix for - https://issues.cask.co/browse/CDAP-6813
Added errorDataset name to allow user to set the error data-set that is to be used.


sreevatsanraman:
LGTM

PR #356, opened 2016-08-03
Add support for nullable input fields in CSVParser

knazir:
Added support for the CSVParser transform to accept nullable fields. Null input fields allows all other fields to be propagated and otherwise-parsed fields to be set to null.

### DIFF
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /**
  * Transformation that parses a field as CSV Record into {@link StructuredRecord}.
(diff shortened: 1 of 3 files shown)


transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -108,7 +108,8 @@ public void configurePipeline(PipelineConfigurer configurer) throws IllegalArgum
         throw new IllegalArgumentException(
           "Field " + config.field + " is not present in the input schema");
       } else {
-        if (!inputSchemaField.getSchema().getType().equals(Schema.Type.STRING)) {
+        if (!inputSchemaField.getSchema().getType().equals(Schema.Type.STRING) &&
```
albertshau: you can use Schema's isNullable and getNonNullable methods to help out here.

albertshau: lets move the schema validation logic into a package private method and add a unit test that calls it. `validateSchema(Schema inputSchema)`

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -189,6 +212,11 @@ public void transform(StructuredRecord in, Emitter<StructuredRecord> emitter) th
     // Field has to string to be parsed correctly. For others throw an exception.
     String body = in.get(config.field);
 
+    // Handles case where nullable input field was null
+    if (body == null) {
```
albertshau: lets add a short blurb in the field description and in the .md doc file describing that records with a null field get filtered out.

albertshau:
few small comments

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
knazir: whoops, kind of re-invented the wheel here, thanks

albertshau: It's more readable to do:

```
Schema fieldSchema = inputSchemaField.getSchema();
Schema.Type fieldType = fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
if (fieldType != Schema.Type.STRING) {  
```

albertshau:
one minor comment, otherwise lgtm. fix and merge

russorat:
Spoke to @knazir about this yesterday. Since this plugin allows non-csv fields to be passed through, we can't simply drop the record. We decided it would be better to pass the non-csv records, and add null for any records expected from the csv field. If those fields are not nullable in the output schema, it will fail, but that is expected.

knazir:
Thanks for the look through @albertshau, I'll update the changes for the point @russorat brought up.

knazir:
Rebased with release/1.4's CSVParser changes. Would appreciate another look through. Now, if the input field to parse is null, all other fields are propagated and those that would have otherwise been parsed are set to null.

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -247,7 +266,8 @@ private StructuredRecord createStructuredRecord(CSVRecord record, StructuredReco
     private final String format;
 
     @Name("field")
-    @Description("Specify the field that should be used for parsing into CSV.")
+    @Description("Specify the field that should be used for parsing into CSV. Input records with a null field are " +
```
albertshau: the bit about being filtered is no longer true

albertshau:
one comment about a description. otherwise lgtm after conflicts are fixed.

russorat:
lgtm

knazir:
Guild is breen: http://builds.cask.co/browse/HYP-BAD3-6 and manual tests worked. Going to merge.

PR #357, opened 2016-08-03
HDFS Action Plugin

thegreenfrog:
Implements Plugin that allows to moving files in HDFS.

Testing ongoing.

### DIFF
--- a/core-plugins/docs/HDFSAction-action.md
+++ b/core-plugins/docs/HDFSAction-action.md
@@ -1,0 +1,52 @@
+# HDFS Action
+
+
+Description
+-----------
+Moves a file or files within an HDFS cluster.
+
+
+Use Case
+--------
+This action can be used when a file or files need to be moved to a new location in an HDFS cluster, 
+often required when archiving files.
+
+
+Properties
+----------
+**sourcePath:** The full HDFS path of the file or directory that is to be moved. In the case of a directory, if
+fileRegex is set, then only files in the source directory matching the wildcard regex will be moved.
+Otherwise, all files in the directory will be moved. For example: `hdfs://hostname/tmp`.
+
+**destPath:** The valid, full HDFS destination path in the same cluster where the file or files are to be moved.
+If a directory is specified with a file sourcePath, the file will be put into that directory. If sourcePath is
+a directory, it is assumed that destPath is also a directory. HDFSAction will not catch this inconsistency.
+
+**fileRegex:** Wildcard regular expression to filter the files in the source directory that will be moved.
+
+**continueOnError:** Indicates if the pipeline should continue if the move process fails. If all files are not 
+successfully moved, the action will not return the files already moved to their original locations.
+
+
+Example
+-------
+This example moves a file from `/source/path` to `/dest/path`:
+
+    {
+        "name": "HDFSAction",
+        "plugin": {
+            "name": "HDFSAction",
+            "type": "action",
+            "artifact": {
+                "name": "core-plugins",
+                "version": "1.4.0-SNAPSHOT",
+                "scope": "SYSTEM"
+            },
+            "properties": {
+                "sourcePath": "hdfs://example.com:8020/source/path",
+                "destPath": "hdfs://example.com:8020/dest/path",
+                "fileRegex": ".*\.txt",
+                "continueOnError": "false"
+            }
+        }
+    }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.action;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.action.Action;
+import co.cask.cdap.etl.api.action.ActionContext;
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * Action that moves file(s) within HDFS in the same cluster.
+ * A user must specify file/directory path and destination file/directory path
+ * Optionals include fileRegex
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name("HDFSFileMoveAction")
+@Description("Action to move files within HDFS")
+public class HDFSFileMoveAction extends Action {
+  private static final Logger LOG = LoggerFactory.getLogger(HDFSFileMoveAction.class);
+
+  private HDFSActionConfig config;
+
+  public HDFSFileMoveAction(HDFSActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path source = new Path(config.sourcePath);
+
+    Path dest = new Path(config.destPath);
+
+    FileSystem fileSystem = source.getFileSystem(new Configuration());
+    fileSystem.mkdirs(dest.getParent());
+
+    if (fileSystem.getFileStatus(source).isFile()) { //moving single file
+
+      try {
+        if (!fileSystem.rename(source, dest)) {
+          if (!config.continueOnError) {
+            throw new IOException(String.format("Failed to rename file {} to {}", source.toString(), dest.toString()));
+          }
+          LOG.error("Failed to move file {} to {}", source.toString(), dest.toString());
+        }
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          throw e;
+        }
+        LOG.error("Failed to move file {} to {}", source.toString(), dest.toString(), e);
+      }
+      return;
+    }
+
+    // Moving contents of directory
+    FileStatus[] listFiles;
+    if (config.fileRegex != null) {
+      PathFilter filter = new PathFilter() {
+        private final Pattern pattern = Pattern.compile(config.fileRegex);
+
+        @Override
+        public boolean accept(Path path) {
+          return pattern.matcher(path.getName()).matches();
+        }
+      };
+
+      listFiles = fileSystem.listStatus(source, filter);
+    } else {
+      listFiles = fileSystem.listStatus(source);
+    }
+
+    if (listFiles.length == 0) {
+      if (config.fileRegex != null) {
+        LOG.warn("Not moving any files of type {} from source {}", config.fileRegex, source.toString());
+      } else {
+        LOG.warn("Not moving any files from source {}", source.toString());
+      }
+    }
+
+    if (fileSystem.isFile(dest)) {
+      throw new IllegalArgumentException(String.format("destPath %s needs to be a directory since sourcePath is a " +
+                                                         "directory", config.destPath));
+    }
+    fileSystem.mkdirs(dest); //create destination directory if necessary
+
+    for (FileStatus file: listFiles) {
+      source = file.getPath();
+
+      try {
+        if (!fileSystem.rename(source, dest)) {
+          if (!config.continueOnError) {
+            throw new IOException(String.format("Failed to rename file {} to {}", source.toString(), dest.toString()));
+          }
+          LOG.error("Failed to move file {} to {}", source.toString(), dest.toString());
+        }
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          throw e;
+        }
+        LOG.error("Failed to move file {} to {}", source.toString(), dest.toString(), e);
+      }
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or directory that is to be moved. In the case of a directory, if " +
+      "fileRegex is set, then only files in the source directory matching the wildcard regex will be moved. " +
+      "Otherwise, all files in the directory will be moved. For example: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The valid, full HDFS destination path in the same cluster where the file or files are to be moved. " +
+      "If a directory is specified with a file sourcePath, the file will be put into that directory. If sourcePath " +
+      "is a directory, it is assumed that destPath is also a directory. HDFSAction will not catch this inconsistency.")
+    private String destPath;
+
+    @Description("Wildcard regular expression to filter the files in the source directory that will be moved")
+    @Nullable
+    private String fileRegex;
+
+    @Description("Indicates if the pipeline should continue if the move process fails")
+    private boolean continueOnError;
+
+    @VisibleForTesting
+    HDFSActionConfig(String sourcePath, String destPath, String fileRegex, boolean continueOnError) {
+      this.sourcePath = sourcePath;
+      this.destPath = destPath;
+      this.fileRegex = fileRegex;
+      this.continueOnError = continueOnError;
+    }
+  }
+}
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
@@ -1,0 +1,231 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.action;
+
+import co.cask.cdap.api.workflow.NodeStatus;
+import co.cask.cdap.datapipeline.SmartWorkflow;
+import co.cask.cdap.etl.api.action.Action;
+import co.cask.cdap.etl.api.batch.BatchSink;
+import co.cask.cdap.etl.api.batch.BatchSource;
+import co.cask.cdap.etl.proto.v2.ETLBatchConfig;
+import co.cask.cdap.etl.proto.v2.ETLPlugin;
+import co.cask.cdap.etl.proto.v2.ETLStage;
+import co.cask.cdap.proto.Id;
+import co.cask.cdap.proto.ProgramRunStatus;
+import co.cask.cdap.proto.ProgramType;
+import co.cask.cdap.proto.RunRecord;
+import co.cask.cdap.proto.WorkflowNodeStateDetail;
+import co.cask.cdap.proto.artifact.AppRequest;
+import co.cask.cdap.test.ApplicationManager;
+import co.cask.cdap.test.WorkflowManager;
+import co.cask.hydrator.plugin.batch.ETLBatchTestBase;
+import com.google.common.collect.ImmutableMap;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hdfs.MiniDFSCluster;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test for {@link HDFSFileMoveAction}
+ */
+public class HDFSFileMoveActionTestRun extends ETLBatchTestBase {
+  @ClassRule
+  public static TemporaryFolder folder = new TemporaryFolder();
+
+  private static MiniDFSCluster dfsCluster;
+  private static FileSystem fileSystem;
+
+  @BeforeClass
+  public static void buildMiniDFS() throws Exception {
+    // Setup MiniDFSCluster
+    File baseDir = folder.newFolder();
+    org.apache.hadoop.conf.Configuration conf = new org.apache.hadoop.conf.Configuration();
+    conf.set(MiniDFSCluster.HDFS_MINIDFS_BASEDIR, baseDir.getAbsolutePath());
+    MiniDFSCluster.Builder builder = new MiniDFSCluster.Builder(conf);
+    dfsCluster = builder.build();
+    dfsCluster.waitActive();
+    fileSystem = FileSystem.get(conf);
+  }
+
+  @AfterClass
+  public static void cleanupMiniDFS() throws Exception {
+    // Shutdown MiniDFSCluster
+    dfsCluster.shutdown();
+    folder.delete();
+  }
+
+  @Test
+  public void testFailedHDFSAction() throws Exception {
+    //expect action phase and therefore pipeline to fail because of invalid source path
+
+    Path outputDir = dfsCluster.getFileSystem().getHomeDirectory();
+
+    fileSystem.mkdirs(new Path("source"));
+    fileSystem.createNewFile(new Path("source/test.txt"));
+    fileSystem.createNewFile(new Path("source/test.json"));
+
+    ETLStage action = new ETLStage(
+      "HDFSFileMoveAction",
+      new ETLPlugin("HDFSFileMoveAction", Action.PLUGIN_TYPE,
+                    ImmutableMap.of("sourcePath", outputDir.toUri().toString() + "/test",
+                                    "destPath", outputDir.toUri().toString() + "/",
+                                    "fileRegex", ".*\\.txt",
+                                    "continueOnError", "false"),
+                    null));
+    ETLStage source = new ETLStage("source",
+                                   new ETLPlugin("KVTable", BatchSource.PLUGIN_TYPE,
+                                                 ImmutableMap.of("name", "hdfsTestSource"), null));
+    ETLStage sink = new ETLStage("sink", new ETLPlugin("KVTable", BatchSink.PLUGIN_TYPE,
+                                                       ImmutableMap.of("name", "hdfsTestSink"), null));
+    ETLBatchConfig etlConfig = ETLBatchConfig.builder("* * * * *")
+      .addStage(source)
+      .addStage(sink)
+      .addStage(action)
+      .addConnection(action.getName(), source.getName())
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start();
+    manager.waitForFinish(3, TimeUnit.MINUTES);
+
+//    List<RunRecord> history = appManager.getHistory(new Id.Program(appId, ProgramType.WORKFLOW, SmartWorkflow.NAME),
+//                                                    ProgramRunStatus.FAILED);
+//    Assert.assertTrue(history.size() == 1);
+//
+//    Map<String, WorkflowNodeStateDetail> nodesInFailedProgram =
+//      manager.getWorkflowNodeStates(history.get(0).getPid());
+//    Assert.assertTrue(nodesInFailedProgram.size() == 1);
+//
+//    //check that HDFSFileMoveAction node failed
+//    Assert.assertTrue(nodesInFailedProgram.values().iterator().next().getNodeStatus().equals(NodeStatus.FAILED));
+
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test.txt")));
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test.json")));
+  }
+
+  @Test
+  public void testHDFSAction() throws Exception {
+    //moves only test.txt from `/source` to `/dest`, which is also created during the Action process
+
+    Path outputDir = dfsCluster.getFileSystem().getHomeDirectory();
+
+    fileSystem.mkdirs(new Path("source"));
+    fileSystem.createNewFile(new Path("source/test.txt"));
+    fileSystem.createNewFile(new Path("source/test.json"));
+
+    ETLStage action = new ETLStage(
+      "HDFSFileMoveAction",
+      new ETLPlugin("HDFSFileMoveAction", Action.PLUGIN_TYPE,
+                    ImmutableMap.of("sourcePath", outputDir.toUri().toString() + "/source",
+                                    "destPath", outputDir.toUri().toString() + "/dest",
+                                    "fileRegex", ".*\\.txt",
+                                    "continueOnError", "false"),
+                    null));
+    ETLStage source = new ETLStage("source",
+                                   new ETLPlugin("KVTable", BatchSource.PLUGIN_TYPE,
+                                                 ImmutableMap.of("name", "hdfsTestSource"), null));
+    ETLStage sink = new ETLStage("sink", new ETLPlugin("KVTable", BatchSink.PLUGIN_TYPE,
+                                                       ImmutableMap.of("name", "hdfsTestSink"), null));
+    ETLBatchConfig etlConfig = ETLBatchConfig.builder("* * * * *")
+      .addStage(source)
+      .addStage(sink)
+      .addStage(action)
+      .addConnection(action.getName(), source.getName())
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start();
+    manager.waitForFinish(3, TimeUnit.MINUTES);
+
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dest/test.txt")));
+    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dest/test.json")));
+  }
+
+  @Test
+  public void testInvalidDestHDFSAction() throws Exception {
+    //destPath is an existing text file, not a directory, so the pipeline should fail
+
+    Path outputDir = dfsCluster.getFileSystem().getHomeDirectory();
+
+    fileSystem.mkdirs(new Path("source"));
+    fileSystem.createNewFile(new Path("source/test.txt"));
+    fileSystem.createNewFile(new Path("source/test2.txt"));
+    fileSystem.createNewFile(new Path("source/test.json"));
+    fileSystem.createNewFile(new Path("dest/dest.txt"));
+
+    ETLStage action = new ETLStage(
+      "HDFSFileMoveAction",
+      new ETLPlugin("HDFSFileMoveAction", Action.PLUGIN_TYPE,
+                    ImmutableMap.of("sourcePath", outputDir.toUri().toString() + "/source",
+                                    "destPath", outputDir.toUri().toString() + "/dest/dest.txt",
+                                    "fileRegex", ".*\\.txt",
+                                    "continueOnError", "false"),
+                    null));
+    ETLStage source = new ETLStage("source",
+                                   new ETLPlugin("KVTable", BatchSource.PLUGIN_TYPE,
+                                                 ImmutableMap.of("name", "hdfsTestSource"), null));
+    ETLStage sink = new ETLStage("sink", new ETLPlugin("KVTable", BatchSink.PLUGIN_TYPE,
+                                                       ImmutableMap.of("name", "hdfsTestSink"), null));
+    ETLBatchConfig etlConfig = ETLBatchConfig.builder("* * * * *")
+      .addStage(source)
+      .addStage(sink)
+      .addStage(action)
+      .addConnection(action.getName(), source.getName())
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start();
+    manager.waitForFinish(3, TimeUnit.MINUTES);
+
+//    List<RunRecord> history = appManager.getHistory(new Id.Program(appId, ProgramType.WORKFLOW, SmartWorkflow.NAME),
+//                                                    ProgramRunStatus.FAILED);
+//    Assert.assertTrue(history.size() == 1);
+//
+//    Map<String, WorkflowNodeStateDetail> nodesInFailedProgram =
+//      manager.getWorkflowNodeStates(history.get(0).getPid());
+//    Assert.assertTrue(nodesInFailedProgram.size() == 1);
+//
+//    //check that HDFSFileMoveAction node failed
+//    Assert.assertTrue(nodesInFailedProgram.values().iterator().next().getNodeStatus().equals(NodeStatus.FAILED));
+
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test.txt")));
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test2.txt")));
+    Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test.json")));
+  }
+}
--- a/core-plugins/widgets/HDFSAction-action.json
+++ b/core-plugins/widgets/HDFSAction-action.json
@@ -1,0 +1,40 @@
+{
+  "metadata": {
+    "spec-version": "1.0"
+  },
+  "configuration-groups": [
+    {
+      "label": "Properties",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Source Path",
+          "name": "sourcePath"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Destination Path",
+          "name": "destPath"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Wildcard",
+          "name": "fileRegex"
+        },
+        {
+          "widget-type": "select",
+          "label": "Continue Pipeline Upon Error",
+          "name": "continueOnError",
+          "widget-attributes": {
+            "values": [
+              "true",
+              "false"
+            ],
+            "default": "false"
+          }
+        }
+      ]
+    }
+  ],
+  "outputs": []
+}
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -70,6 +70,34 @@
       <artifactId>cdap-etl-realtime</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-minicluster</artifactId>
+      <version>2.3.0</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.servlet</groupId>
+          <artifactId>servlet-api</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-logging</groupId>
+          <artifactId>commons-logging</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.sun.jersey</groupId>
+          <artifactId>jersey-core</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>org.twitter4j</groupId>
       <artifactId>twitter4j-core</artifactId>
     </dependency>
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/BatchPluginsTestSuite.java
@@ -18,6 +18,7 @@ package co.cask.hydrator.plugin.batch;
 
 import co.cask.cdap.common.test.TestSuite;
 import co.cask.hydrator.plugin.batch.action.EmailActionTestRun;
+import co.cask.hydrator.plugin.batch.action.HDFSFileMoveActionTestRun;
 import co.cask.hydrator.plugin.batch.aggregator.DedupTestRun;
 import co.cask.hydrator.plugin.batch.aggregator.GroupByTestRun;
 import org.junit.runner.RunWith;
@@ -36,7 +37,8 @@ import org.junit.runners.Suite;
   GroupByTestRun.class,
   DedupTestRun.class,
   ETLFTPTestRun.class,
-  EmailActionTestRun.class
+  EmailActionTestRun.class,
+  HDFSFileMoveActionTestRun.class
 })
 public class BatchPluginsTestSuite extends ETLBatchTestBase {
 }


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,106 @@
...
+        //Moving directory, so destination filePath doesn't have filename
+        dest = new Path(config.destPath + listFiles[i].getName());
+      } else {
+        dest = new Path(config.destPath);
+      }
+
+      try {
+        hdfs.rename(source, dest);
+      } catch (IOException e) {
+        LOG.error("Failed to move file {} to {} for reason: {}", source.toString(), dest.toString(), e.getMessage());
```
anwar6953: we're losing the stacktrace of the IOE, like this.
I think we should log the entire e, not just its message.

albertshau:
this isn't even a plugin (missing annotations). Please finish the entire thing (coding, testing, etc.) before opening a review.

anwar6953:
need rst file? json file for widget?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,106 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Move file(s) within HDFS.
+ * A user must specify file/directory path and destination file/directory path
+ */
+public class HDFSAction extends Action {
+  private HDFSActionConfig config;
+
+  private static final Logger LOG = LoggerFactory.getLogger(HDFSAction.class);
```
chtyim: Put static fields before non-static fields

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.io.File;
+import java.io.FileFilter;
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+/**
+ * Move file(s) within HDFS.
+ * A user must specify file/directory path and destination file/directory path
+ */
```
anwar6953: do we need @Plugin annotation?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(HDFSAction.class);
+
+  public HDFSAction(HDFSActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    //determine how many times we have to run this
+    File folder = new File(config.sourcePath);
```
chtyim: Why look at the local file system?? Isn't this plugin to operate on files on HDFS?

anwar6953: Can the `config.sourcePath` point to a single file, instead of a folder? Variable name seems to imply its always a directory. Also, what code changes would be required to support a single file as the source (i.e. we can't call `listFiles` below).

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: oh yes I do.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: @chtyim it was intended that the user provides the full hdfs path. Does that answer your question?

thegreenfrog:
Made changes based on comments. 
@albertshau is it possible to create a unit test for an action moving stuff in hdfs?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public HDFSAction(HDFSActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    FileSystem hdfs = new Path(config.sourcePath).getFileSystem(new Configuration());
+
+    if (new File(config.sourcePath).isFile()) {//moving single file
```
chtyim: It won't work since the file is on HDFS. Shouldn't you use `hdfs.getFileStatus`?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,118 @@
...
+      Path source = new Path(config.sourcePath);
```
chtyim: Don't keep recreating the path. Just create it once when you use that to get the `FileSystem` object.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,118 @@
...
+      Path dest = new Path(config.destPath);
+      try {
+        hdfs.rename(source, dest);
+      } catch (IOException e) {
+        LOG.error("Failed to move file {} to {} for reason: {}", source.toString(), dest.toString(), e);
```
chtyim: Why just log? So failure to move is not a failure?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,118 @@
...
+      }
+      return;
+    }
+
+    //moving contents of directory
+    File folder = new File(config.sourcePath);
```
chtyim: Again, shouldn't use `File`.

albertshau:
yes you can write a unit test. It won't be working on hdfs, since its a unit test. It will work on local files.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,128 @@
...
+
+  public HDFSAction(HDFSActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path source = new Path(config.sourcePath);
+    Path dest = new Path(config.destPath);
+    FileSystem hdfs = source.getFileSystem(new Configuration());
```
anwar6953: name it `fileSystem`, since its not always hdfs (such as when running in Standalone)

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,128 @@
...
+          return pattern.matcher(path.getName()).matches();
+        }
+      };
+
+      listFiles = hdfs.listStatus(source, filter);
+    } else {
+      listFiles = hdfs.listStatus(source);
+    }
+
+    for (int i = 0; i < listFiles.length; i++) {
```
anwar6953: nit - I think for-each syntax will look cleaner.

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+Description
+-----------
+Moves file(s) within HDFS.
+
+
+Use Case
+--------
+This action can be used when files needs to be moved to a new location in HDFS for the purpose of archiving,
```
anwar6953: Would it make sense to also support the use case of `copy` in this plugin also?
(copy vs move)

thegreenfrog: perhaps? I think we currently just need to be able to move files in HDFS, so that's why i just went with the `rename()` call`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,126 @@
...
+
+      listFiles = fileSystem.listStatus(source, filter);
+    } else {
+      listFiles = fileSystem.listStatus(source);
+    }
+
+    for (FileStatus file: listFiles) {
+      source = file.getPath();
+      //Moving directory, so destination filePath doesn't have filename
+      dest = new Path(config.destPath);
```
chtyim: So all source files move to the same dest?

thegreenfrog:
@albertshau testing on cluster in hdfs works.

moved only text files from `hdfs://hdfsaction-cluster11967-1000.dev.continuuity.net/tmp/actionTestFolder` to `hdfs://hdfsaction-cluster11967-1000.dev.continuuity.net/tmp/actionTestFolder/destDir` with wildcard regex `.*\.txt`

One thing I am not entirely sure about that Terence brought up was whether or not the plugin should fail if the move returns some error. Currently, it just logs the error. Thoughts?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: yes

chtyim: What's the point of doing so?

thegreenfrog: archiving data

chtyim: hum? How can renaming to the same file achieve that? It fact, it will fail for the second file.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,106 @@
...
+    file.getParentFile().mkdirs();
+    file.createNewFile();
+    File jsonFile = new File("testFiles/source/test.json");
+    jsonFile.createNewFile();
+
+    ETLStage action = new ETLStage(
+      "HDFSAction",
+      new ETLPlugin("HDFSAction", Action.PLUGIN_TYPE,
+                    ImmutableMap.of("sourcePath", "testFiles/source",
+                                    "destPath", "testFiles/",
```
chtyim: Why need a `/` suffix?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Test for {@link HDFSAction}
+ */
+public class HDFSActionTestRun extends ETLBatchTestBase {
+  private final String sourcePath = "/some/Source/Path";
+  private final String destPath = "/some/dest/Path";
+
+  @Test
+  public void testHDFSAction() throws Exception {
+    File file = new File("testFiles/source/test.txt");
```
chtyim: This test is not really testing HDFS. Shouldn't you start a mini DFSCluster and test with it?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: We're moving files to different locations in HDFS.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
thegreenfrog: Yes it is moving local files. But I also tested it by running this in a pipeline on a Coopr cluster. Are you saying I can write a unit test that creates a mini distributed cluster?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
anwar6953: @chtyim I think if the destination is a parameter like '/tmp/', then it will move all the files into the tmp folder.

anwar6953:
I think the action should throw an exception if a file move operation fails.

Edit: or maybe have an option to suppress failures (by simply logging them) and still move other files.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
chtyim: So it assumes the dest is a directory that already exists? Shouldn't we check? Shall we create the directory if it is missing? Also, I don't think it's good to require the user input "/tmp/". What if the "/" suffix is missing?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test for {@link HDFSAction}
+ */
+public class HDFSActionTestRun extends ETLBatchTestBase {
+
+  @Test
+  public void testHDFSAction() throws Exception {
+    File file = new File("testFiles/source/test.txt");
```
chtyim: Should use `TemporaryFolder` `ClassRule` instead. It handles resource cleanup automatically, hence you don't need the `deleteDirectory` method. Also unit-test shouldn't create file in non-target directory.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: Well the path would be something like `hdfs://hostname/tmp/`, so it would be hard to miss the suffix, in my opinion.

chtyim: If the user miss it what would happen? He will lose all the file except the last one?

chtyim: I think it is better if you construct the target path using the dest as the parent and the source name as the child

anwar6953: What about the case of moving just a single file?

thegreenfrog: Could we just throw an exception if it is incorrect?

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,48 @@
...
+              "version": "1.4.0-SNAPSHOT",
+              "scope": "SYSTEM"
+          },
+          "properties": {
+              "sourcePath": "hdfs://123.23.12.4344:10000/source/path",
+              "destPath": "hdfs://123.23.12.4344:10000/dest/path",
+              "fileRegex": ".*\.txt",
+              "continueOnError": "false"
+            }
+          }
```
albertshau: why are these italicized?

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,48 @@
...
+**continueOnError:** Indicates whether or not the pipeline should stop if the file move process encounters an error.
+
+
+Example
+-------
+This example moves a file from /source/path to /dest/path:
+
+    {
+        "name": "HDFSAction",
+          "plugin": {
```
albertshau: fix indentation

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,48 @@
...
+            "name": "HDFSAction",
+            "type": "action",
+            "label": "HDFSAction",
+            "artifact": {
+              "name": "core-plugins",
+              "version": "1.4.0-SNAPSHOT",
+              "scope": "SYSTEM"
+          },
```
albertshau: fix indentation

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,48 @@
...
+          "properties": {
```
albertshau: fix indentation

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("Full HDFS path of file/directory")
```
albertshau: include an example in the description

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+    private String sourcePath;
+
+    @Description("Full HDFS path of desired destination path")
+    private String destPath;
+
+    @Nullable
```
albertshau: this needs a description

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+    private String fileRegex;
+
+    private boolean continueOnError;
```
albertshau: this needs a description.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -63,13 +64,15 @@ public void run(ActionContext context) throws Exception {
       try {
         fileSystem.rename(source, dest);
       } catch (IOException e) {
+        if (!config.continueOnError) {
```
anwar6953: so if its false, what's the expected behavior?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
albertshau: the class javadoc says this is optional, but it is not optional here. If you want it to be optional, it needs to be a Boolean and annotated as nullable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+    @Description("Full HDFS path of desired destination path")
+    private String destPath;
+
+    @Nullable
+    private String fileRegex;
+
+    private boolean continueOnError;
+
+    HDFSActionConfig() {
+      continueOnError = false;
```
albertshau: there is no point in this if continueOnError is not optional

anwar6953: constructor doesn't just hold default values?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("Full HDFS path of file/directory")
+    private String sourcePath;
+
+    @Description("Full HDFS path of desired destination path")
```
albertshau: what happens if the source is a file? Is this path then a file? Or is it always a directory? Need to include these details so people know what it does.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/**
+ * Move file(s) within HDFS.
+ * A user must specify file/directory path and destination file/directory path
+ * Optionals include fileRegex
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name("HDFSAction")
```
albertshau: this is not a good name if it is just moving files. FileMove would be more descriptive.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
albertshau: this is the full path, so something like hdfs:/...? If so, it is possible for somebody to have a source from one cluster and a destination in another cluster. You need to document that this is only moving files in one cluster.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+    Path dest = new Path(config.destPath);
+    FileSystem fileSystem = source.getFileSystem(new Configuration());
+
+    if (fileSystem.getFileStatus(source).isFile()) { //moving single file
+
+      try {
+        fileSystem.rename(source, dest);
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          e.printStackTrace();
```
albertshau: shouldn't this throw the exception?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path source = new Path(config.sourcePath);
+    Path dest = new Path(config.destPath);
+    FileSystem fileSystem = source.getFileSystem(new Configuration());
+
+    if (fileSystem.getFileStatus(source).isFile()) { //moving single file
+
+      try {
+        fileSystem.rename(source, dest);
```
albertshau: according to javadocs, this returns whether it was successful or not. Not sure when it returns false versus throwing an exception, but you should be checking the return value and failing if the rename failed.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          e.printStackTrace();
+        }
+        LOG.error("Failed to move file {} to {} for reason: {}", source.toString(), dest.toString(), e);
```
albertshau: the number of '{}' don't match the number of arguments you gave. The exception is always the last thing you pass and shouldn't be something that is substituted in. You probably don't need 'for reason: {}'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
```diff
@@ -0,0 +1,140 @@
...
+    for (FileStatus file: listFiles) {
+      source = file.getPath();
+      //Moving directory, so destination filePath doesn't have filename
+      dest = new Path(config.destPath);
+
+      try {
+        fileSystem.rename(source, dest);
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          e.printStackTrace();
```
albertshau: shouldn't this throw the exception?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
albertshau: I think you just need to ensure that the destination exists as a directory.  You should clearly document what the destination needs to be if the source is a single file, a single directory, files in a directory.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,144 @@
...
+      .addStage(action)
+      .addConnection(action.getName(), source.getName())
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start(ImmutableMap.of("logical.start.time", "0"));
```
albertshau: there is no need to set logical start time.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,144 @@
...
+    manager.waitForFinish(3, TimeUnit.MINUTES);
+
+    List<RunRecord> history = appManager.getHistory(new Id.Program(appId, ProgramType.WORKFLOW, SmartWorkflow.NAME),
+                                                    ProgramRunStatus.FAILED);
+    Assert.assertTrue(history.size() == 1); //make sure pipeline didn't fail
```
albertshau: there is history regardless of whether it failed or not. In fact, aren't you making sure it does fail?

core-plugins/widgets/HDFSAction-action.json
```diff
@@ -0,0 +1,38 @@
...
+          "widget-type": "textbox",
+          "label": "Destination Path",
+          "name": "destPath"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Wildcard",
+          "name": "fileRegex"
+        },
+        "widget-type": "select",
```
albertshau: why are these colons weird? How are you creating all these text files? They have so many strange characters

albertshau: this isn't correct, all of these need to be in a json object. Did you even test this?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
albertshau: its not optional, so the user is forced to set it, so there is no point in a default. If it made optional, then this is the correct way to set the default.

core-plugins/widgets/HDFSAction-action.json
thegreenfrog: This was a recent addition based on what Ali recommended. I'll be testing it on a cluster after I make all the changes.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
thegreenfrog: oh yes. But doesn't this only return something if the program failed by setting the `ProgramRunStatus.FAILED`?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
```diff
@@ -0,0 +1,144 @@
...
+ * Test for {@link HDFSAction}
+ */
+public class HDFSActionTestRun extends ETLBatchTestBase {
+  @ClassRule
+  public static TemporaryFolder folder = new TemporaryFolder();
+
+  @Test
+  public void testFailedHDFSAction() throws Exception {
+
+    folder.newFolder("source");
```
chtyim: I think I had a previous comment saying to use the mini DFSCluster to test instead of using local file.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
thegreenfrog: We're going to need to create an action to move files between remote machines too though. Would HDFSFileMove be ok? Or is that too verbose?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSActionTestRun.java
albertshau: o yeah you're right, didn't see the ProgramRunStatus.FAILED. I guess its just a misleading comment then.

core-plugins/widgets/HDFSAction-action.json
albertshau: again, you need to test everything before asking people to review it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSAction.java
albertshau: thats fine

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
+# HDFS Action
+
+
+Description
+-----------
+Moves file(s) within HDFS.
```
awholegunch: Moves a file or files within an HDFS cluster.

core-plugins/docs/HDFSAction-action.md
awholegunch:
```
This action can be used when a file or files need to be moved to a new location in an HDFS cluster, 
often required when archiving files.
```

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+          "artifact": {
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+          },
+        },
+        "properties": {
+          "sourcePath": "hdfs://123.23.12.4344:10000/source/path",
+          "destPath": "hdfs://123.23.12.4344:10000/dest/path",
+          "fileRegex": ".*\.txt",
```
awholegunch: Is this a backslash or should this be a pipe?

```
".*\.txt"
```

or

```
".*|.txt"
```

or something else?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("Full HDFS path of file/directory. In the case of a directory, if fileRegex is set, " +
+      "then all files in the directory matching the wildcard regex will be moved. Otherwise, all files in the " +
```
awholegunch: `then only files in the source directory matching the wildcard regex will be moved.`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+      "directory will be moved. Ex: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The full HDFS destination path in the same cluster where user wants to move the file(s) in " +
+      "sourcePath. It is assumed that the path is valid. For moving a file, this means that the parent directories " +
+      "exist. For moving multiple files in a directory or an entire directory, the path to the desired directory " +
+      "location should be passed and that directory and all parent directories should already exist.")
+    private String destPath;
+
+    @Description("Wildcard regex to filter type of files in the directory to move")
```
awholegunch: "Wildcard regular expression to filter the source files in the directory that will be moved"

core-plugins/docs/HDFSAction-action.md
thegreenfrog: backslash

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+    @Nullable
+    private String fileRegex;
+
+    @Description("Whether or not the Pipeline should stop if the file move process fails")
```
awholegunch:
```
Indicates if the pipeline should continue if the move process fails.
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+      PathFilter filter = new PathFilter() {
+        private final Pattern pattern = Pattern.compile(config.fileRegex);
+
+        @Override
+        public boolean accept(Path path) {
+          return pattern.matcher(path.getName()).matches();
+        }
+      };
+
+      listFiles = fileSystem.listStatus(source, filter);
```
awholegunch, line 95: Do you need to test if filter is null, or listFiles is null? What if there is no match? Should we not get a message such as:

```
LOG.error("Failed to move file {} to {} using regex {}", source.toString(), dest.toString(), config.fileRegex);
```

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+**sourcePath:** The full HDFS path of the file/directory that the user wants to move. In the case of a directory,
+if fileRegex is set, then all files in the directory matching the wildcard regex will be moved. Otherwise,
+all files in the directory will be moved. Ex: hdfs://hostname/tmp
+
+**destPath:** The full HDFS destination path in the same cluster where the user wants to move the file(s) specified
+in the sourcePath. It is assumed that the path is valid. For moving a file, this means that the parent directories
+exist. For moving multiple files in a directory or an entire directory, the path to the desired directory location
+should be passed and that directory and all parent directories should already exist.
+
+**fileRegex:** The wildcard regex to filter what kind of files in the directory to move.
```
awholegunch: **fileRegex:** Wildcard regular expression to filter the files in the source directory that will be moved.

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+--------
+This action can be used when files needs to be moved to a new location in HDFS for the purpose of archiving,
+for example.
+
+
+Properties
+----------
+**sourcePath:** The full HDFS path of the file/directory that the user wants to move. In the case of a directory,
+if fileRegex is set, then all files in the directory matching the wildcard regex will be moved. Otherwise,
+all files in the directory will be moved. Ex: hdfs://hostname/tmp
```
awholegunch:
```
**sourcePath:** The full HDFS path of the file or directory that is to be moved. In the case of a directory,
 if fileRegex is set, then only files in the source directory matching the wildcard regex will be moved. 
Otherwise, all files in the directory will be moved. For example: `hdfs://hostname/tmp`.
```

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+
+**destPath:** The full HDFS destination path in the same cluster where the user wants to move the file(s) specified
+in the sourcePath. It is assumed that the path is valid. For moving a file, this means that the parent directories
+exist. For moving multiple files in a directory or an entire directory, the path to the desired directory location
+should be passed and that directory and all parent directories should already exist.
```
awholegunch:
```
**destPath:** The valid, full HDFS destination path in the same cluster where the file or files 
are to be moved. For moving a single file, this means that all parent directories must already exist.
For moving multiple files in a directory or an entire directory, the path to the desired directory 
is passed and that directory and all parent directories must already exist.
```

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+
+**fileRegex:** The wildcard regex to filter what kind of files in the directory to move.
+
+**continueOnError:** Indicates whether or not the pipeline should stop if the file move process fails.
```
awholegunch: Indicates if the pipeline should continue if the move process fails.

awholegunch:
Can you edit the documentation markdown file, and then edit the source annotations to match?

core-plugins/docs/HDFSAction-action.md
albertshau: I commented on this already, but these are italicized when they shouldn't be. Thats why the backslash looks weird.

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+          },
+        },
+        "properties": {
+          "sourcePath": "hdfs://123.23.12.4344:10000/source/path",
+          "destPath": "hdfs://123.23.12.4344:10000/dest/path",
+          "fileRegex": ".*\.txt",
+          "continueOnError": "false"
```
albertshau: remove italics (same with line above and lines below)

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+          "name": "HDFSAction",
+          "type": "action",
+          "label": "HDFSAction",
+          "artifact": {
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+          },
+        },
+        "properties": {
```
albertshau: this is wrong, properties belongs in the plugin section

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+      "name": "HDFSAction",
+        "plugin": {
+          "name": "HDFSAction",
+          "type": "action",
+          "label": "HDFSAction",
+          "artifact": {
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+          },
```
albertshau: is the comma in italics or is that my eyes playing tricks? Please make sure you don't have italics anywhere.
Also, this is not valid json because of the hanging comma.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
albertshau, line 95: filter is never null, and I'm guessing listStatus returns an empty list instead of null? Is that correct Chris?

albertshau, line 95: though it may be worth warn logging if there is nothing to move, as that is probably not the intent of the plugin.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+      } catch (IOException e) {
+        if (!config.continueOnError) {
+          throw e;
+        }
+        LOG.error("Failed to move file {} to {}", source.toString(), dest.toString(), e);
+      }
+      return;
+    }
+
+    //moving contents of directory
```
awholegunch: BTW: Comments should have initial cpaitalization and a space after the "//":

```
// Moving contents of directory
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+
+      listFiles = fileSystem.listStatus(source, filter);
+    } else {
+      listFiles = fileSystem.listStatus(source);
+    }
+
+    for (FileStatus file: listFiles) {
+      source = file.getPath();
+      //Moving directory, so destination filePath doesn't have filename
+      dest = new Path(config.destPath);
```
albertshau: none of the earlier comments about this were addressed. You need to make sure the destination exists and is a directory. If this is a file, you will end up overwriting all source files except the last one.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,153 @@
...
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("Full HDFS path of file/directory. In the case of a directory, if fileRegex is set, " +
+      "then only files in the directory matching the wildcard regex will be moved. Otherwise, all files in the " +
+      "directory will be moved. Ex: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The full HDFS destination path in the same cluster where user wants to move the file(s) in " +
+      "sourcePath. It is assumed that the path is valid. For moving a file, this means that the parent directories " +
+      "exist. For moving multiple files in a directory or an entire directory, the path to the desired directory " +
+      "location should be passed and that directory and all parent directories should already exist.")
```
albertshau: We should create the directory and its parents if they don't exist.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,172 @@
...
+ * Test for {@link HDFSFileMoveAction}
+ */
+public class HDFSFileMoveActionTestRun extends ETLBatchTestBase {
+  @ClassRule
+  public static TemporaryFolder folder = new TemporaryFolder();
+
+  private MiniDFSCluster dfsCluster;
+  private FileSystem fileSystem;
+
+  @Before
```
albertshau: this should be BeforeClass, as you just want to do this once for this entire class instead of once per test method.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,172 @@
...
+    File baseDir = folder.newFolder();
+    org.apache.hadoop.conf.Configuration conf = new org.apache.hadoop.conf.Configuration();
+    conf.set(MiniDFSCluster.HDFS_MINIDFS_BASEDIR, baseDir.getAbsolutePath());
+    MiniDFSCluster.Builder builder = new MiniDFSCluster.Builder(conf);
+    dfsCluster = builder.build();
+    dfsCluster.waitActive();
+    fileSystem = FileSystem.get(conf);
+  }
+
+  @After
```
albertshau: similarly, should be AfterClass

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,172 @@
...
+  public void afterTest() throws Exception {
+    // Shutdown MiniDFSCluster
+    if (dfsCluster != null) {
```
albertshau: how can this be null?

core-plugins/widgets/HDFSAction-action.json
```diff
@@ -0,0 +1,40 @@
...
+          "name": "destPath"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Wildcard",
+          "name": "fileRegex"
+        },
+        {
+          "widget-type": "select",
+          "label": "Stop Pipeline Upon Error",
```
albertshau: you've labeled it as stop on error, but a true value continues on error, it doesn't stop.

core-plugins/docs/HDFSAction-action.md
thegreenfrog: I think Github is italicizing it... because on IntelliJ it is definitely not italicized.

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+**continueOnError:** Indicates whether or not the pipeline should stop if the file move process fails.
+
+
+Example
+-------
+This example moves a file from /source/path to /dest/path:
+
+    {
+      "name": "HDFSAction",
+        "plugin": {
```
albertshau, line 37: Seems like I've commented on indentation multiple times, but please fix indentation. name and plugin should not be at different levels.

core-plugins/docs/HDFSAction-action.md
albertshau: hm ok then ignore my comments. I guess it's interpreting the '*' specially?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
thegreenfrog, line 95: `if (config.fileRegex != null) {` checks if the filter has been set? Yeah I'll add a log.warn(?) statement if there is nothing to move.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
thegreenfrog: creating a new `Path` object will thrown an `IllegalArgumentException` if the path does not exist. I'll add a check to ensure it is a directory though.

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+Properties
+----------
+**sourcePath:** The full HDFS path of the file or directory that is to be moved. In the case of a directory, if
+fileRegex is set, then only files in the source directory matching the wildcard regex will be moved.
+Otherwise, all files in the directory will be moved. For example: `hdfs://hostname/tmp`.
+
+**destPath:** The valid, full HDFS destination path in the same cluster where the file or files are to be moved.
+For moving a single file, this means that all parent directories must already exist.For moving multiple files in a
```
awholegunch: Space required: `must already exist. For moving`

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+directories must already exist.
+
+**fileRegex:** Wildcard regular expression to filter the files in the source directory that will be moved.
+
+**continueOnError:** Indicates if the pipeline should continue if the move process fails.
+
+
+Example
+-------
+This example moves a file from /source/path to /dest/path:
```
awholegunch:
```
This example moves a file from `/source/path` to `/dest/path`:
```

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+        "type": "action",
+        "label": "HDFSAction",
+        "artifact": {
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+        },
+        "properties": {
+            "sourcePath": "hdfs://123.23.12.4344:10000/source/path",
+            "destPath": "hdfs://123.23.12.4344:10000/dest/path",
```
awholegunch: A couple of questions:
- Can this be a domain name rather than an IP address? If so, `example.com` is reserved for documentation, and you could then use:

`"hdfs://example.com:10000/source/path"`
- What part does the port play? Which ports can and cannot be used? Does it have to be `10000`? Does it have to be the `router.bind.port`? If so, should it actually be specified?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
anwar6953: Is that true? How would you create a file on the file system via the Path
API then?

core-plugins/docs/HDFSAction-action.md
thegreenfrog: Oh actually I think the proper port is 8020 (http://www.cloudera.com/documentation/enterprise/latest/topics/cdh_ig_ports_cdh5.html). And yes it can be a domain name. I can use `example.com`

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+Example
+-------
+This example moves a file from `/source/path` to `/dest/path`:
+
+{
+    "name": "HDFSAction",
+    "plugin": {
+        "name": "HDFSAction",
+        "type": "action",
+        "label": "HDFSAction",
```
albertshau: should remove label, as it is something the UI injects, it is not used by the backend anywhere (UI shouldn't be putting it there).

core-plugins/docs/HDFSAction-action.md
```diff
@@ -0,0 +1,53 @@
...
+        "artifact": {
+            "name": "core-plugins",
+            "version": "1.4.0-SNAPSHOT",
+            "scope": "SYSTEM"
+        },
+        "properties": {
+            "sourcePath": "hdfs://example.com:8020/source/path",
+            "destPath": "hdfs://example.com:8020/dest/path",
+            "fileRegex": ".*\.txt",
```
albertshau: does this render as italics in our UI as well? Wondering if it is a markdown thing or if its just a github thing.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,166 @@
...
+    if (listFiles.length == 0) {
+      if (config.fileRegex != null) {
+        LOG.warn("Not moving any files of type {} from source {}", config.fileRegex, source.toString());
+      } else {
+        LOG.warn("Not moving any files from source {}", source.toString());
+      }
+    }
+
+    //Moving directory, so destination filePath doesn't have filename
+    dest = new Path(config.destPath);
```
albertshau: this isn't needed, you set it already at the start of the method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,166 @@
...
+    if (!fileSystem.getFileStatus(dest).isDirectory()) { //directory path was not passed
```
albertshau: mentioned this in the last round of reviews, but we should be trying to create the destination if it doesn't exist.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,166 @@
...
+                                                         "is a directory", dest.toString(), source.toString()));
+    }
+
+    for (FileStatus file: listFiles) {
+      source = file.getPath();
+
+      try {
+        if (!fileSystem.rename(source, dest)) {
+          if (!config.continueOnError) {
+            throw new IOException(String.format("Failed to rename file {} to {}", source.toString(), dest.toString()));
```
albertshau, line 120: if a file fails to move, we don't try and move files we've already moved back to their original location. Should mention that in the doc.

albertshau:
couple comments

core-plugins/docs/HDFSAction-action.md
awholegunch: It's a bug in the GitHub dif display. Look at the final markdown (https://github.com/caskdata/hydrator-plugins/blob/8fd63d8b62077e26f1c66780f660b76ea6c47c7b/core-plugins/docs/HDFSAction-action.md) to see that it doesn't matter.

awholegunch: But what is an issue is that four (4) leading spaces needs to be added to each line in this block so that it is rendered as a literal.

awholegunch:
I've added a comment about the Markdown.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
thegreenfrog, line 120: ok. I added an `mkdirs` call before creating the destination path variable to create all necessary parent directories.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,170 @@
...
+      .addStage(action)
+      .addConnection(action.getName(), source.getName())
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start(ImmutableMap.of("logical.start.time", "0"));
```
anwar6953: can remove `ImmutableMap.of("logical.start.time", "0")`, right?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,170 @@
...
```
anwar6953: can remove `ImmutableMap.of("logical.start.time", "0")`, right?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -0,0 +1,170 @@
...
+  @AfterClass
+  public static void cleanupMiniDFS() throws Exception {
+    // Shutdown MiniDFSCluster
+    dfsCluster.shutdown();
+    folder.delete();
+  }
+
+  @Test
+  public void testFailedHDFSAction() throws Exception {
+
```
anwar6953, line 84: can you add a comment as to why we expect it to fail

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,168 @@
...
+  private HDFSActionConfig config;
+
+  public HDFSFileMoveAction(HDFSActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path source = new Path(config.sourcePath);
+    new File(config.destPath).mkdirs();
```
albertshau: this doesn't work. File is java's api for local files. Its also incorrect logic in case the the destination is a file and not a directory. Please add or modify a test case for this. Don't just add code and assume it works. You need to test it out.

thegreenfrog: ok. I changed it to make an `mkdir` call to the filesystem. If the destination is a file, it would make sense to create any necessary directories in the destination file path though, right?

albertshau: yes but I believe it would create the destination as a directory instead of a file. In any case, you should be able to see the behavior in your test case.

thegreenfrog: I edited the test case to use a non-existent destination path to test this element of the Action.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path source = new Path(config.sourcePath);
+
+    Path dest = new Path(config.destPath);
+
+    FileSystem fileSystem = source.getFileSystem(new Configuration());
+    fileSystem.mkdirs(dest);
```
albertshau: did you try when dest is supposed to be a file and not a directory?  I think this creates a directory in that scenario, which will result in behavior different than what is documented. You can use dest.getParent() to get the parent to mkdirs, then if its a single file move you don't do anything else, but if its multiple files, you then later create the dest directory.

albertshau: add a test for this case too

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,169 @@
...
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or directory that is to be moved. In the case of a directory, if" +
+      "fileRegex is set, then only files in the source directory matching the wildcard regex will be moved." +
+      "Otherwise, all files in the directory will be moved. For example: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The valid, full HDFS destination path in the same cluster where the file or files are to be moved." +
+      "For moving a single file, this means that all parent directories must already exist.For moving multiple files" +
```
albertshau: Space after 'exist.'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,169 @@
...
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or directory that is to be moved. In the case of a directory, if" +
```
albertshau: space after last 'if'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,169 @@
...
+      "fileRegex is set, then only files in the source directory matching the wildcard regex will be moved." +
```
albertshau: space after 'moved.'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,169 @@
...
+      "Otherwise, all files in the directory will be moved. For example: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The valid, full HDFS destination path in the same cluster where the file or files are to be moved." +
```
albertshau: space after 'moved.'

albertshau:
has this been tested on a secure cluster?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
thegreenfrog: yeah you're right. It tags the dest path as a directory always.... thanks for catching that

thegreenfrog:
@albertshau So seems like it's impossible to check if the destination is actually a directory or file using any of `fileSystem` API, so I just updated the docs and code to make it clear that a source directory must also have a destination directory. Tested it on a cluster.

albertshau:
isFile(Path f). isDirectory(Path f).

thegreenfrog:
@albertshau made changes based on conversation.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveAction.java
```diff
@@ -0,0 +1,167 @@
...
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class HDFSActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or directory that is to be moved. In the case of a directory, if " +
+      "fileRegex is set, then only files in the source directory matching the wildcard regex will be moved. " +
+      "Otherwise, all files in the directory will be moved. For example: hdfs://hostname/tmp")
+    private String sourcePath;
+
+    @Description("The valid, full HDFS destination path in the same cluster where the file or files are to be moved. " +
+      "If a directory is specified with a file sourcePath, the file will be put into that directory. If sourcePath is" +
```
albertshau: need a space after 'is'

albertshau:
one minor comment about a description, otherwise logic lgtm. Did you test on a _secure_ cluster? I'm wondering if the tokens and everything are used correctly.

thegreenfrog:
yeah i tested it on a hadoop-secure-autobuild cluster.

albertshau:
lgtm

thegreenfrog:
the guild is breen 
http://builds.cask.co/browse/HYP-BAD38-3

PR #368, opened 2016-08-08
Merge 1 3

albertshau:

PR #371, opened 2016-08-08
Uncomment RedundantModifier since this is the current active release

wolf31o2:
This was accidentally commented in #368 for this branch. The `release/1.4` branch is under active development, so we should perform this test. We only disabled it on patch-only branches.


albertshau:
lgtm

PR #374, opened 2016-08-09
fix pass through for csv parser

albertshau:
I resolved the conflict incorrectly for this class, causing pass
through to work incorrectly. This fixes the behavior.


sreevatsanraman:
lgtm

PR #375, opened 2016-08-09
[CDAP-6535] Snapshot sink cleans up older snapshots

knazir:
Snapshot sinks should now clean up snapshots older than a time specified by the user.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
@@ -22,6 +22,7 @@ import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.hydrator.common.TimeParser;
 import co.cask.hydrator.plugin.common.SnapshotFileSetConfig;
 import co.cask.hydrator.plugin.dataset.SnapshotFileSet;
 import com.google.gson.Gson;
@@ -29,6 +30,7 @@ import com.google.gson.reflect.TypeToken;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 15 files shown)


anwar6953:
what happens if used doesn't configure/define it? Doesn't seem like you're handling that case (nor documenting what happens in that case)

knazir:
Whoops. Thanks @anwar6953, I added a null check. It's a no-op like TPFS sinks.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
```diff
@@ -76,6 +77,12 @@ public void onRunFinish(boolean succeeded, BatchSinkContext context) {
     super.onRunFinish(succeeded, context);
     if (succeeded) {
       try {
+        if (config.getCleanPartitionsOlderThan() != null) {
+          long cutoffTime =
+            context.getLogicalStartTime() - TimeParser.parseDuration(config.getCleanPartitionsOlderThan());
+          snapshotFileSet.deleteMatchingPartitionsByTime(cutoffTime);
+          LOG.info("Cleaning up old partition for timestamp {}", cutoffTime);
```
anwar6953: this log statement is inaccurate

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
```diff
@@ -76,6 +77,12 @@ public void onRunFinish(boolean succeeded, BatchSinkContext context) {
     super.onRunFinish(succeeded, context);
     if (succeeded) {
       try {
+        if (config.getCleanPartitionsOlderThan() != null) {
+          long cutoffTime =
+            context.getLogicalStartTime() - TimeParser.parseDuration(config.getCleanPartitionsOlderThan());
```
albertshau, line 94: should also parse it in configurePipeline() and check its a valid duration

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
```diff
@@ -76,6 +77,12 @@ public void onRunFinish(boolean succeeded, BatchSinkContext context) {
...
+          snapshotFileSet.deleteMatchingPartitionsByTime(cutoffTime);
+          LOG.info("Cleaning up snapshots older than {}", cutoffTime);
+        }
         snapshotFileSet.onSuccess(context.getLogicalStartTime());
       } catch (Exception e) {
         LOG.error("Exception updating state file with value of latest snapshot, ", e);
```
albertshau: this may no longer be accurate if the error happened when dropping partitions.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchSink.java
```diff
@@ -76,6 +77,12 @@ public void onRunFinish(boolean succeeded, BatchSinkContext context) {
     super.onRunFinish(succeeded, context);
     if (succeeded) {
       try {
+        if (config.getCleanPartitionsOlderThan() != null) {
```
albertshau, line 92: I think this should happen after the onSuccess() call. If an exception occurs when cleaning up old partitions, it would leave things in a weird state.

albertshau:
a few comments

albertshau:
lgtm

knazir:
Guild is breen: http://builds.cask.co/browse/HYP-BAD21-9 going to squash and merge.

PR #377, opened 2016-08-10
nested schema issue fix

yug095:
fix the lira:
https://issues.cask.co/browse/HYDRA-91

http://builds.cask.co/browse/HYP-BAD26-2

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToAvroTransformer.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToAvroTransformer.java
@@ -33,28 +33,31 @@ import java.util.Map;
 public class StructuredToAvroTransformer extends RecordConverter<StructuredRecord, GenericRecord> {
 
   private final Map<Integer, Schema> schemaCache;
+  private final co.cask.cdap.api.data.schema.Schema outputCDAPSchema;
   private final Schema outputAvroSchema;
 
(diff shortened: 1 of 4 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToAvroTransformer.java
```diff
@@ -63,7 +63,12 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws IOExcep
       if (schemaField == null) {
         throw new IllegalArgumentException("Input record does not contain the " + fieldName + " field.");
       }
-      recordBuilder.set(fieldName, convertField(structuredRecord.get(fieldName), schemaField.getSchema()));
+      if (schemaField.getSchema().getType() == co.cask.cdap.api.data.schema.Schema.Type.RECORD) {
```
albertshau: this is not a good way to fix it. convertField already has a switch statement on the field type, you shouldn't duplicate that logic here, and shouldn't be creating a new transformer for each recursive call.

Instead, just remove the outputAvroSchema variable completely. The avro schema should always be calculated from the  cdap schema.

albertshau:
lgtm assuming tests pass. Have you verified that the pipeline in the jira works now?

yug095:
@albertshau Yes, it works now.

yug095:
Bamboo Build Green, will merge it.

PR #379, opened 2016-08-10
Kafka and HDFS Plugins Macro Update

knazir:
Enabled macros for the following DB plugin properties:

`HDFSSink`
- path
- timeSuffix

`KafkaSource`
- partitions
- topic
- zkConnect
- kafkaBrokers
- defaultOffset
- format

### DIFF
--- a/hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
+++ b/hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
@@ -17,6 +17,7 @@
 package co.cask.hydrator.plugin;
 
 import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
 import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.batch.Output;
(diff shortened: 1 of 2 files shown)


hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
```diff
@@ -122,18 +122,26 @@ private static String outputFormatClassName(String option) {
...
     private String timeSufix;
 
     public HDFSSinkConfig(String referenceName, String path, String suffix, String outputFormat) {
       super(referenceName);
       this.path = path;
       this.timeSufix = suffix;
     }
+
+    private void validate() {
+      if (!containsMacro("timeSuffix") && !Strings.isNullOrEmpty(timeSufix)) {
```
anwar6953: why do we need to check `!containsMacro("timeSuffix")`?
Why isn't it sufficient to check `!Strings.isNullOrEmpty(timeSufix)`

knazir: ah you're right... strings are null at configure time by default

albertshau:
lgtm

knazir:
Guild is breen: http://builds.cask.co/browse/HYP-BAD25 going to squash and merge.

PR #380, opened 2016-08-11
Fixing getOutputSchema for joins

CuriousVini:
Issue: https://issues.cask.co/browse/HYDRA-406
Build: http://builds.cask.co/browse/HYP-BAD27-6
- This error occurred because we do not use config object in rest calls for output schema, my bad!
- Adding test to avoid this error in future. Also tested it on cluster
- Adding validation for join keys if the join key fields are not present in input schema, otherwise we will get NPE.


albertshau:
lgtm

CuriousVini:
Build is green. Merging

PR #381, opened 2016-08-11
HYDRA-665 move runtime check for $CONDITIONS

albertshau:
Moving the runtime check for $CONDITIONS in the import query so
that it is only checked if numSplits is not set to 1.


albertshau:
http://builds.cask.co/browse/HYP-BAD29

sreevatsanraman:
LGTM

PR #383, opened 2016-08-12
add TINYINT and SMALLINT transform

yug095:
When getting JDBC types of small int and tiny int, instead of transforming them to short by default, transform them to Integer. 
JIRA:
https://issues.cask.co/browse/HYDRA-47
Build and Test:
http://builds.cask.co/browse/HYP-BAD31-6
Here are the data type mapping. 
https://msdn.microsoft.com/en-us/library/ms378878(v=sql.110).aspx
http://docs.oracle.com/javase/6/docs/technotes/guides/jdbc/getstart/mapping.html
Found that only tinyint and smallint mapping are different. add all necessary type in test.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
@@ -152,11 +152,7 @@ public class DBRecord implements Writable, DBWritable {
       switch (sqlColumnType) {
         case Types.SMALLINT:
         case Types.TINYINT:
-          if (original.getClass() == Short.class) {
-            return ((Short) original).intValue();
-          } else {
-            break;
-          }
+          return ((Number) original).intValue();
         case Types.NUMERIC:
         case Types.DECIMAL:
           return ((BigDecimal) original).doubleValue();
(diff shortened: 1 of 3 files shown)


albertshau:
Can we add test cases? What about the other types mentioned in the jira?

database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
```diff
@@ -150,6 +150,13 @@ public void write(PreparedStatement stmt) throws SQLException {
   private Object transformValue(int sqlColumnType, Object original) throws SQLException {
     if (original != null) {
       switch (sqlColumnType) {
+        case Types.SMALLINT:
+        case Types.TINYINT:
+          if (original.getClass() == Short.class) {
```
albertshau: you should be able to do ((Number) original).intValue() instead of checking the class. Should be a little more robust in case there's a silly jdbc driver that makes all numbers Doubles or something weird like that.

albertshau:
one comment, otherwise lgtm

yug095:
Build is green, merged.

PR #355, opened 2016-08-02
fix HYDRA-125 cannot read from TPFS Parquet source

yaojiefeng:
JIRA: https://issues.cask.co/browse/HYDRA-125
Build: http://builds.cask.co/browse/HYP-DBT56-1
I cannot reproduce the error described in this JIRA, but get a runtime exception when reading from TPFS parquet source on the cluster:
java.lang.RuntimeException: readObject can't find class
    at co.cask.cdap.internal.app.runtime.batch.dataset.input.TaggedInputSplit.readClass(TaggedInputSplit.java:162)
    at co.cask.cdap.internal.app.runtime.batch.dataset.input.TaggedInputSplit.readFields(TaggedInputSplit.java:146)
    at org.apache.hadoop.io.serializer.WritableSerialization$WritableDeserializer.deserialize(WritableSerialization.java:71)
    at org.apache.hadoop.io.serializer.WritableSerialization$WritableDeserializer.deserialize(WritableSerialization.java:42)
    at org.apache.hadoop.mapred.MapTask.getSplitDetails(MapTask.java:372)
    at org.apache.hadoop.mapred.MapTask.runNewMapper(MapTask.java:754)
    at org.apache.hadoop.mapred.MapTask.run(MapTask.java:341)
    at org.apache.hadoop.mapred.YarnChild$2.run(YarnChild.java:168)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:415)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1657)
    at org.apache.hadoop.mapred.YarnChild.main(YarnChild.java:162)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at co.cask.cdap.internal.app.runtime.batch.distributed.MapReduceContainerLauncher.launch(MapReduceContainerLauncher.java:98)
    at org.apache.hadoop.mapred.YarnChild.main(Unknown Source)
Caused by: java.lang.ClassNotFoundException: Class parquet.hadoop.ParquetInputSplit not found
    at org.apache.hadoop.conf.Configuration.getClassByName(Configuration.java:2101)
[...]

PR #384, opened 2016-08-12
Revert "fix HYDRA-125 cannot read from TPFS Parquet source"

albertshau:
Reverts caskdata/hydrator-plugins#355


sreevatsanraman:
LGTM

albertshau:
the change breaks backwards compatibility, and it breaks spark on some versions of some distros.

PR #385, opened 2016-08-12
Fixing DBSource to setOutputSchema()

CuriousVini:
Issue: https://issues.cask.co/browse/HYDRA-663, https://issues.cask.co/browse/HYDRA-606
Build: http://builds.cask.co/browse/HYP-BAD35-5

### DIFF
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -47,6 +47,10 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
+**schema:** The schema of records output by the source. This will be used in place of whatever schema comes 
+back from the query. However, it must match the schema that comes back from the query, 
+except it can mark fields as nullable and can contain a subset of the fields. 
+
 Example
 -------
 This example connects to a database using the specified 'connectionString', which means
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
@@ -18,11 +18,12 @@ package co.cask.hydrator.plugin;
 
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.data.schema.UnsupportedTypeException;
-import com.google.common.base.Throwables;
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
@@ -48,6 +49,7 @@ import javax.management.ReflectionException;
  */
 public final class DBUtils {
   private static final Logger LOG = LoggerFactory.getLogger(DBUtils.class);
+  public static final String OVERRIDE_SCHEMA = "co.cask.hydrator.db.override.schema";
 
   /**
    * Performs any Database related cleanup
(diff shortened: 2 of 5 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -262,6 +276,11 @@ private String getJDBCPluginId() {
     @Macro
     Integer numSplits;
 
+    @Nullable
+    @Name(OUTPUT_SCHEMA)
+    @Description("Output schema for schema propogation")
```
elfenheart: nit: typo `propagation`

sreevatsanraman:
Will this change make output schema for db source mandatory?

albertshau:
need to update the markdown doc too.

CuriousVini:
@sreevatsanraman It is still similar behaviour for other plugins, lets say Database -> transform -> sink. But if Database plugin is just before Joiner in pipeline, i.e Database -> Joiner, then we will need output schema. Otherwise Joiner will throw `java.lang.IllegalArgumentException: Joiner cannot have any null input schemas`.

sreevatsanraman:
What will happen if the user doesn't enter the output schema for a plugin. Use-case is - I would like to dump all the DB Entries into a TPFS, for this I don't need to enter an output schema either via getSchema or manually entering the schema. Will this assumption break with this change?

CuriousVini:
@sreevatsanraman outputSchema is optional for DBSource plugin. I have tried with DBSource -> HdfsSink without getSchema(). It works fine. But, since Joiner requires inputs to be non-null, this change will propagate outputSchema only if it is set. Otherwise, behavior would be same as before.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -294,6 +310,16 @@ private void validate() {
...
       }
+
+    }
+
+    private Schema getSchema() {
+      try {
+        return Schema.parseJson(schema);
+      } catch (IOException e) {
+        throw new IllegalArgumentException(
+          String.format("Unable to parse schema '%s'. Reason: %s", schema, e.getMessage()));
```
anwar6953: instead of just putting `e.getMessage()` as part of the new exception message, I think you should pass the entire `e` as the second parameter to the constructor of `IllegalArgumentException`.

database-plugins/docs/Database-batchsink.md
```diff
@@ -47,6 +47,10 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
+**schema:** The schema of records output by the source. This will be used in place of whatever schema comes 
+back from the query. This should only be used if there is a bug in your jdbc driver.
```
albertshau: We don't need to say only use if there is a bug, since we want people to use it so that we can propagate schema.

We should instead say it must match the schema that comes back from the query, except it can mark fields as nullable and can contain a subset of the fields.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -99,6 +100,29 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
    * list of {@link co.cask.cdap.api.data.schema.Schema.Field},
    * where name of the field is same as column name and type of the field is obtained using {@link DBUtils#getType(int)}
    * @param resultSet result set of executed query
+   * @param schemaStr schema string to override resultant schema
+   *
+   * @return list of schema fields
+   * @throws SQLException
+   * @throws IOException
+   */
+  public static List<Schema.Field> getSchemaFields(ResultSet resultSet, String schemaStr) throws SQLException {
```
albertshau: annotate schemaStr as nullable.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -99,6 +100,29 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
...
+    if (!Strings.isNullOrEmpty(schemaStr)) {
+      try {
+        Schema schema = Schema.parseJson(schemaStr);
+        return schema.getFields();
```
albertshau: we should still get the fields from the result set. We should then check that the user provided schema is compatible with the result schema. Each field in the user provided schema exists in the result schema and is the same type, or compatible non-nullable type.

If we don't do this, the user will get a very confusing error message if they give a wrong schema.

CuriousVini:
@albertshau Since jdbc drivers can return wrong null codes, I have not added validation for nullable fields. Apart from that addressed all other comments.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -99,29 +100,50 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
...
+            throw new IllegalArgumentException(String.format("Schema field %s has type %s but in input record found " +
+                                                               "type %s ",
+                                                             field.getName(), simpleSchema.getType(),
+                                                             resultsetFieldSchema.getType()));
+          }
+        }
         return schema.getFields();
       } catch (IOException e) {
-        LOG.error("Can not covert schema string {} to Schema", schemaStr);
+        throw new IllegalArgumentException(String.format("Can not covert schema string %s to Schema", schemaStr), e);
```
albertshau: for messaging consistency:
Can not convert schema string -> Unable to parse schema string

albertshau: its a little cleaner to just have the parseJson in this try-catch. Though nothing else in the try can throw an IOException, if somebody adds some logic later that does throw IOException, we'll end up with a misleading error message.

albertshau:
minor comments, otherwise lgtm. fix and merge

CuriousVini:
Addressed comments. Build is green. squashing and merging.

PR #389, opened 2016-08-12
Fix macro validation for DB plugins

knazir:
At runtime, properties that were provided macro syntax when configured will still have `containsMacro(property)` return true. This ensures that validation always happens at runtime and only conditionally at configure time.


knazir:
Making a change to PluginConfig: https://issues.cask.co/browse/CDAP-7046 that invalidates this PR.

PR #390, opened 2016-08-12
Fix HDFSSink property validation

knazir:
We are changing the general approach to validate properties conditionally with macros. This should update HDFSSink to align with updates to other plugins.


hdfs-plugins/src/main/java/co/cask/hydrator/plugin/HDFSSink.java
```diff
@@ -139,10 +140,17 @@ public HDFSSinkConfig(String referenceName, String path, String suffix, String o
...
-      // if macro provided, timeSuffix will be null at configure time
-      if (!Strings.isNullOrEmpty(timeSufix)) {
-        new SimpleDateFormat(timeSufix);
+    private void validateProperties() {
+      validateProperty("timeSuffix");
+    }
+
+    private void validateProperty(String property) {
+      switch (property) {
+        case "timeSuffix":
```
gsps1, line 149: for cases like this when there is just a single property, we don't need a switch case.

knazir:
Making a change to PluginConfig: https://issues.cask.co/browse/CDAP-7046 that invalidates this PR.

PR #392, opened 2016-08-13
Fix MockActionContext Compilation Errors

knazir:
Unfortunately, I can't seem to convince Hydrator Plugins to build with the latest from the release/3.5 CDAP branch locally. This should should fix the compilation errors caused by https://github.com/caskdata/cdap/pull/6399 from the ActionContext changes and allow coopr clusters to build...


sagarkapare:
LGTM

PR #397, opened 2016-08-15
initial implementation for dynamic partitioning sink plugin

russorat:
The purpose of this plugin is to allow users to use values in each record for partitioning. For example, if a user has records with "name" and "original_create_date" and they specify they want to partition on original_create_date, the resulting File Dataset will have a partition for each value in original_create_date. The use case for this is a user loading historical data from a DB or set of files and would like to keep the data partitioned on the original transaction or load date instead of the run date of the load.


russorat:
closing this and will reopen against develop for a later release.

PR #401, opened 2016-08-16
[HYDRA-668] - Decision Tree Regression

ananya-bhattacharya:
Spark plugins that trains and predicts label data based on Decision Tree Regression.
For design please refer : https://wiki.cask.co/display/CE/Decision+Tree+Regression


spark-plugins/docs/DecisionTreeRegressor-sparkcompute.md
```diff
@@ -0,0 +1,63 @@
...
+     | 2     | 4     | DL      | N523US  | 1756      | ATL    | MLB         | 442      | 2058      | 2229    |
+     | 20    | 1     | AA      | N3GEAA  | 1162      | DEN    | DFW         | 641      | 905       | 1205    |
+     | 4     | 6     | AA      | N3CBAA  | 1314      | MIA    | IAH         | 964      | 1225      | 1420    |
+     +=======================================================================================================+
+
+
+Output records will contain all the fields along with the predicted field:
+
+     +=================================================================================================================+
+     | DayOf | DayOf | Carrier | TailNum | FlightNum | Origin | Destination | Distance | Departure | Arrival | Dealyed |
```
russorat: Spelling for Dealyed

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,47 @@
+# DecisionTreeTrainer
+
+
+Description
+-----------
+Trains a regression model based upon a particular label and features of a record. Saves this model to a FileSet.
+
+Use Case
+--------
+This sink can be used when you have a sample data and you want to use it, to build a Decision Tree Regression model.
```
russorat: remove comma

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,47 @@
...
+This model can be used further for predicting the label.
```
russorat: this is a trainer so it can't do any predicting. you will need to use the regressor for that.

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,47 @@
...
+
+Properties
+----------
+**fileSetName:** The name of the FileSet to save the model to.
+
+**path:** Path of the FileSet to save the model to.
+
+**features:** A comma-separated sequence of fields to use for training. Features to be used, must be of simple type :
+String, int, double, float, long, bytes, boolean.
```
russorat: no need to list them, just say simple type.

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,47 @@
...
+    {
+      "name": "DecisionTreeRegression",
+      "type": "sparkcompute",
+      "properties": {
+            "fileSetName": "decision-tree-model",
+            "path": "decisionTree",
+            "features": "dofM,dofW,scheduleDepTime,scheduledArrTime,carrier,elapsedTime,origin,dest",
+            "predictionField": "delayed",
+            "maxDept":9,
+            "maxBins"100
```
russorat: json not properly formatted

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+      return;
+    }
+    validateSchema(inputSchema);
+    // otherwise, we have a constant input schema. Get the input schema and
+    // add a field to it, on which the prediction will be set
+    outputSchema = getOutputSchema(inputSchema);
+    stageConfigurer.setOutputSchema(outputSchema);
+  }
+
+  private void validateSchema(Schema inputSchema) {
```
russorat: can you move this to the Config class and use it as part of config.validate?

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+    String[] fields = config.features.split(",");
+    for (String field : fields) {
+      Schema.Field inputField = inputSchema.getField(field);
+      if (inputField == null) {
+        throw new IllegalArgumentException(String.format("Field %s does not exists in the input schema", field));
+      }
+      Schema.Type features = inputField.getSchema().getType();
+      Preconditions.checkArgument(features.isSimpleType(), "Field to classify must be of simple type : String, int, " +
```
russorat: Can this be a nullable type? if so, this should handle that as well.

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+          if (!(type.equals(Schema.Type.INT) || type.equals(Schema.Type.DOUBLE) || type.equals(Schema.Type.FLOAT) ||
+            type.equals(Schema.Type.LONG))) {
+            return structuredRecord.get(field);
+          } else {
+            return null;
+          }
+        }
+      }).distinct().zipWithIndex().collectAsMap();
+
+      if (!(map == null ||
```
russorat: no need to split lines here.

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+    JavaSparkContext javaSparkContext = context.getSparkContext();
+    SparkContext sparkContext = JavaSparkContext.toSparkContext(javaSparkContext);
+    final DecisionTreeModel loadedModel = DecisionTreeModel.load(sparkContext, modelLocation.toURI().getPath());
+
+    JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
+      @Override
+      public StructuredRecord call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<Double>();
+        for (String field : fields) {
+          if (categoricalFeaturesMap.keySet().contains(field)) {
```
russorat: does it make sense to store this instead of getting the keyset every loop? does the keyset change?

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+      }
+    });
+    return output;
+  }
+
+  /**
+  *Creates a builder based off the given record.
+  */
+  private StructuredRecord.Builder cloneRecord(StructuredRecord record) {
+    Schema schemaToUse = outputSchema != null ? outputSchema : getOutputSchema(record.getSchema());
```
russorat: could you add ( ) around outputSchema != null for readability?

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  private Schema getOutputSchema(Schema inputSchema) {
+    return getOutputSchema(inputSchema, config.predictionField);
+  }
+
+  private Schema getOutputSchema(Schema inputSchema, String predictionField) {
+    List<Schema.Field> fields = new ArrayList<>(inputSchema.getFields());
+    fields.add(Schema.Field.of(predictionField, Schema.of(Schema.Type.DOUBLE)));
+    return Schema.recordOf(inputSchema.getRecordName() + ".predicted", fields);
```
russorat: no issues with using periods in schema names?

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
ananya-bhattacharya: This was mentioned to indicate that the model created using the trainer will be used later by the regressor during prediction.
However,  removing this line, to avoid confusion.

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
ananya-bhattacharya: Not getting any error for the period in schema name while testing.
Please let me know if this still needs to be changed.

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,225 @@
...
+
+    final HashMap<String, Map<Object, Long>> categoricalFeaturesMap = new HashMap<>();
+    final String[] fields = config.features.split(",");
+
+    for (final String field : fields) {
+      Map<Object, Long> map = input.map(new Function<StructuredRecord, Object>() {
+        @Override
+        public Object call(StructuredRecord structuredRecord) throws Exception {
+          Schema schema = structuredRecord.getSchema().getField(field).getSchema();
+          if (structuredRecord.get(field) == null && !schema.isNullable()) {
```
russorat: if structuredRecord.get(field) == null, wouldn't line 95 throw a NPE?

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,225 @@
...
+            throw new IllegalArgumentException(String.format("null value found for non-nullable field %s", field));
+          }
+          Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+          if (!(type.equals(Schema.Type.INT) || type.equals(Schema.Type.DOUBLE) || type.equals(Schema.Type.FLOAT) ||
```
russorat: Is it easier to just check if it equals STRING or BYTES? What about BOOLEAN

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,197 @@
...
+      Map<Object, Long> map = input.map(new Function<StructuredRecord, Object>() {
+        @Override
+        public Object call(StructuredRecord structuredRecord) throws Exception {
+          Schema schema = structuredRecord.getSchema().getField(field).getSchema();
+          if (structuredRecord.get(field) == null && !schema.isNullable()) {
+            throw new IllegalArgumentException(String.format("null value found for non-nullable field %s", field));
+          }
+          Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+          if (!(type.equals(Schema.Type.INT) || type.equals(Schema.Type.DOUBLE) || type.equals(Schema.Type.FLOAT) ||
+            type.equals(Schema.Type.LONG))) {
```
russorat: same comment as above

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,197 @@
...
+    });
+
+    trainingData.cache();
+    Set<String> keys = categoricalFeaturesMap.keySet();
+    for (String key : keys) {
+      categoricalFeaturesInfo.put(categoricalFeaturesList.indexOf(key), categoricalFeaturesMap.get(key).size());
+    }
+
+    config.maxDepth = config.maxDepth == null ? 10 : config.maxDepth;
+    config.maxBins = config.maxBins == null ? 100 : config.maxBins;
```
russorat: Should the values 10 and 100 be put into static vars? DEFAULT_MAX_DEPTH and DEFAULT_MAX_BINS?

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,197 @@
...
+    JavaSparkContext sparkContext = context.getSparkContext();
+    FileSet outputFS = context.getDataset(config.fileSetName);
+    model.save(JavaSparkContext.toSparkContext(sparkContext),
+               outputFS.getBaseLocation().append(config.path).toURI().getPath());
+
+  }
+
+  @Override
+  public void prepareRun(SparkPluginContext sparkPluginContext) throws Exception {
+    // no-op; no need to do anything
```
russorat: Do you need to override here if you aren't doing anything?

spark-plugins/src/test/java/co/cask/hydrator/plugin/batch/spark/test/DecisionTreeRegressionTest.java
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Unit tests for {@Link DecisionTreeTrainer} and {@link DecisionTreeRegressor} classes.
+ */
+public class DecisionTreeRegressionTest extends HydratorTestBase {
+
+  @ClassRule
+  public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", false);
+
+  protected static final ArtifactId DATAPIPELINE_ARTIFACT_ID = NamespaceId.DEFAULT.artifact("data-pipeline", "3.2.0");
+  protected static final ArtifactSummary DATAPIPELINE_ARTIFACT = new ArtifactSummary("data-pipeline", "3.2.0");
```
russorat: Should these versions be 3.5.0?

russorat:
Small comments.

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeRegressor.java
ananya-bhattacharya: Updated code, to check STRING and BOOLEAN using the type.equals().
BYTES will be directly converted to double when creating the dense vector list. Since there can be too many values for this, and creating index may not be the best approach, therefore, not handling BYTES using type.equals().

spark-plugins/src/main/java/co/cask/hydrator/plugin/batch/spark/DecisionTreeTrainer.java
ananya-bhattacharya: SparkSink does not implement prepareRun and it's abstract in BatchConfigurable, so it has to be implemented here.
Since, Decision Tree does not need any job config to be set in prepare run, hence keeping this empty.

spark-plugins/pom.xml
```diff
@@ -193,6 +193,47 @@
...
+      <artifactId>apache-rat-plugin</artifactId>
+      <version>0.10</version>
+        <executions>
+          <execution>
+            <id>rat-check</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+            <configuration>
```
anwar6953: Isn't it sufficient to define this configuration at the root pom.xml file?

ananya-bhattacharya: This is done to exclude rat check for "test resources" folder which contains "testData.csv" used for training the data.
Please let me know if the exclusion can be mentioned at the root pom level to exclude all resources, or is it okay to keep it specific to this case.

spark-plugins/pom.xml
```diff
@@ -193,6 +193,47 @@
...
+                <exclude>plugins/**</exclude>
+                <exlcude>**/*.patch</exlcude>
+                <exclude>**/logrotate.d/**</exclude>
+                <exclude>**/limits.d/**</exclude>
+                <exclude>**/*.json</exclude>
+                <exclude>**/*.json.template</exclude>
+                <exclude>**/MANIFEST.MF</exclude>
+                <!-- exclude any hack classes -->
+                <exclude>**/org/apache/hadoop/**</exclude>
+                <exclude>**/resources/**</exclude>
```
russorat: It looks like you should just be able to add this line to the root pom.xml in the rat-check section and you would be good.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,229 @@
...
+
+    // load the model from a file in the model fileset
+    JavaSparkContext javaSparkContext = context.getSparkContext();
+    SparkContext sparkContext = JavaSparkContext.toSparkContext(javaSparkContext);
+    final DecisionTreeModel loadedModel = DecisionTreeModel.load(sparkContext, modelLocation.toURI().getPath());
+
+    JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
+      @Override
+      public StructuredRecord call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<Double>();
```
russorat: this can just be ArrayList<>() i believe.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/DecisionTreeRegressionTest.java
```diff
@@ -0,0 +1,299 @@
...
+  private List<StructuredRecord> getInputData() throws IOException {
+    List<StructuredRecord> messagesToWrite = new ArrayList<>();
+    File file = new File(this.getClass().getResource("/trainData.csv").getFile());
+    BufferedReader bufferedInputStream = new BufferedReader(new FileReader(file));
+    String line;
+    while ((line = bufferedInputStream.readLine()) != null) {
+      String[] flightData = line.split(",");
+      Double depdelaymins = Double.parseDouble(flightData[11]);
+      //For binary classification create delayed field containing values 1.0 and 0.0 depending on the delay time.
+      double delayed = depdelaymins > 40 ? 1.0 : 0.0;
```
russorat: camelCase variable name.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/Flight.java
```diff
@@ -0,0 +1,247 @@
...
+import co.cask.cdap.api.data.schema.Schema;
+
+public class Flight {
+  private final Schema schema =
+    Schema.recordOf("flightData", Schema.Field.of("dofM", Schema.nullableOf(Schema.of(Schema.Type.INT))),
+                    Schema.Field.of("dofW", Schema.nullableOf(Schema.of(Schema.Type.INT))),
+                    Schema.Field.of("carrier", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                    Schema.Field.of("tailNum", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                    Schema.Field.of("flightNum", Schema.nullableOf(Schema.of(Schema.Type.INT))),
+                    Schema.Field.of("originID", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
```
russorat: minor: originId

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/Flight.java
```diff
@@ -0,0 +1,247 @@
...
+                    Schema.Field.of("arrDelay", Schema.nullableOf(Schema.of(Schema.Type.DOUBLE))),
+                    Schema.Field.of("elapsedTime", Schema.nullableOf(Schema.of(Schema.Type.DOUBLE))),
+                    Schema.Field.of("distance", Schema.nullableOf(Schema.of(Schema.Type.INT))),
+                    Schema.Field.of("delayed", Schema.nullableOf(Schema.of(Schema.Type.DOUBLE))));
+  public int dofM;
+  public String carrier;
+  private int dofW;
+  private String tailNum;
+  private int flightNum;
+  private String originID;
```
russorat: same originId

russorat:
Minor updates and a change in the pom.xml. fix those and it LGTM

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/Flight.java
```diff
@@ -0,0 +1,247 @@
...
+    temp = Double.doubleToLongBits(scheduleDepTime);
+    result = 31 * result + (int) (temp ^ (temp >>> 32));
+    result = 31 * result + (deptime != null ? deptime.hashCode() : 0);
+    result = 31 * result + depDelayMins.hashCode();
+    result = 31 * result + (scheduledArrTime != null ? scheduledArrTime.hashCode() : 0);
+    result = 31 * result + (arrTime != null ? arrTime.hashCode() : 0);
+    result = 31 * result + (arrDelay != null ? arrDelay.hashCode() : 0);
+    result = 31 * result + (elapsedTime != null ? elapsedTime.hashCode() : 0);
+    result = 31 * result + distance;
+    result = 31 * result + delayed.hashCode();
```
anwar6953: since delayed is nullable, this could result in a NPE

spark-plugins/pom.xml
anwar6953: This is fine

spark-plugins/docs/DecisionTreeRegressor-sparkcompute.md
```diff
@@ -0,0 +1,68 @@
...
+
+Properties
+----------
+**fileSetName:** The name of the FileSet to load the model from.
+
+**path:** Path of the FileSet to load the model from.
+
+**featuresToInclude:** A comma-separated sequence of fields for Decision Tree Regression. Features to be used, must be
+from one of the following type: int, long, float or double. Both *featuresToInclude* and *featuresToExclude* fields
+annot be specified.
```
albertshau: cannot

spark-plugins/docs/DecisionTreeRegressor-sparkcompute.md
```diff
@@ -0,0 +1,68 @@
...
+
+**featuresToExclude:** A comma-separated sequence of fields to be excluded when calculating prediction. If empty, all
+the fields will be considered for calculating  prediction. Both *featuresToInclude* and *featuresToExclude* fields
```
albertshau: calculating prediction -> prediction

spark-plugins/docs/DecisionTreeRegressor-sparkcompute.md
```diff
@@ -0,0 +1,68 @@
...
+**predictionField:** The field on which to set the prediction. It will be of type double.
+
+
+Example
+-------
+This example uses ``dofM, dofW, scheduleDepTime, scheduledArrTime, carrier, elapsedTime, origin, dest`` fields from the
+input record as featuresToInclude to predict ``delayed`` field.
+
+    {
+      "name": "DecisionTreeRegression",
```
albertshau: use consistent indentation (4 spaces) instead of 2 here and 4 elsewhere

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,50 @@
...
+considered for training. Features to be used, must be from one of the following type: int, long, float or double. Both
+*featuresToInclude* and *featuresToExclude* fields cannot be specified.
+
+**featuresToExclude:** A comma-separated sequence of fields to be excluded when training. If empty, all the fields will
+be considered for training.  Both *featuresToInclude* and *featuresToExclude* fields cannot be specified.
+
+**labelField:** The field from which to get the prediction. It must be of type double.
+
+**maxDepth:** Maximum depth of the tree.
+              For example, depth 0 means 1 leaf node; depth 1 means 1 internal node + 2 leaf nodes. Default is 10.
```
albertshau: Should use consistent indentation. Other fields have this all the way to the left.

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,50 @@
...
+requires maxBins to be at least as large as the number of values in each categorical feature. Default is 100.
+
+
+Example
+-------
+This example uses the fields ``dofM, dofW, scheduleDepTime, scheduledArrTime, carrier, elapsedTime, origin, dest`` from
+the input record as featuresToInclude and ``delayed`` field as the label to train the model.
+
+    {
+        	"name": "DecisionTreeRegression",
```
albertshau: use consistent indentation (4 spaces)

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,50 @@
...
+        	"type": "sparkcompute",
+        	"properties": {
+        		"fileSetName": "decision-tree-model",
+        		"path": "decisionTree",
+        		"featuresToInclude": "dofM,dofW,scheduleDepTime,scheduledArrTime,carrier,elapsedTime,origin,dest",
+        		"labelField": "delayed",
+        		"maxDept": "9",
```
albertshau: maxDepth

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+Example
+-------
+This example uses the fields ``dofM, dofW, scheduleDepTime, scheduledArrTime, carrier, elapsedTime, origin, dest`` from
+the input record as featuresToInclude and ``delayed`` field as the label to train the model.
+
+    {
+        	"name": "DecisionTreeRegression",
+        	"type": "sparkcompute",
```
albertshau: sparksink

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+import javax.ws.rs.Path;
+
+/**
+ * SparkCompute that uses a trained model to tag input records using Decision Tree regression.
+ */
+@Plugin(type = SparkCompute.PLUGIN_TYPE)
+@Name(DecisionTreeRegressor.PLUGIN_NAME)
+@Description("Uses a trained Decision Tree Regression model and regress records.")
+public class DecisionTreeRegressor extends SparkCompute<StructuredRecord, StructuredRecord> {
+  public static final String PLUGIN_NAME = "DecisionTreeRegressor";
```
albertshau: I think it is more clear to call this DecisionTreePredictor.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+import java.util.Arrays;
+import java.util.List;
+import javax.ws.rs.Path;
+
+/**
+ * SparkCompute that uses a trained model to tag input records using Decision Tree regression.
+ */
+@Plugin(type = SparkCompute.PLUGIN_TYPE)
+@Name(DecisionTreeRegressor.PLUGIN_NAME)
+@Description("Uses a trained Decision Tree Regression model and regress records.")
```
albertshau: Uses a Decision Tree Regression model to make predictions.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+  private DecisionTreeRegressorConfig config;
+  private Schema outputSchema;
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    // if null, the input schema is unknown, or has multiple schemas.
+    if (inputSchema == null) {
+      outputSchema = null;
```
albertshau: can remove this line since you're just returning 2 lines down.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+      stageConfigurer.setOutputSchema(null);
+      return;
+    }
+    config.validate(inputSchema);
+    // otherwise, we have a constant input schema. Get the input schema and
+    // add a field to it, on which the prediction will be set
+    outputSchema = getOutputSchema(inputSchema);
```
albertshau: there is no need to set any class variable in configurePipeline, since it is only called once when the pipeline is published. In other words, this can be a variable local to this method.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+    stageConfigurer.setOutputSchema(outputSchema);
+  }
+
+
+  @Override
+  public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext context,
+                                             JavaRDD<StructuredRecord> input) throws Exception {
+    FileSet fileSet = context.getDataset(config.fileSetName);
+    Location modelLocation = fileSet.getBaseLocation().append(config.path);
+    if (!modelLocation.exists()) {
```
albertshau: should load the model in the initialize() method. This will make sure that the model is only loaded once in realtime pipelines.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+    JavaSparkContext javaSparkContext = context.getSparkContext();
+    SparkContext sparkContext = JavaSparkContext.toSparkContext(javaSparkContext);
+    final DecisionTreeModel loadedModel = DecisionTreeModel.load(sparkContext, modelLocation.toURI().getPath());
+
+    JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
+      @Override
+      public StructuredRecord call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<>();
+        for (String field : fields) {
+          featureList.add(Double.valueOf(record.get(field).toString()));
```
albertshau: use `((Number) record.get(field)).doubleValue()` instead. It is more efficient than parsing a string.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
albertshau: actually, we should throw an exception if the input schema is null.

The reason is that we require a constant input schema, otherwise the record fields might get mapped to different features if featuresToInclude is not specified.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        Vector vector = Vectors.dense(Doubles.toArray(featureList));
```
albertshau: in most ML use cases, the vector is sparse, not dense. Better to use Vectors.sparse().

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
albertshau: in a lot of use cases, the vector will be sparse, meaning record.get(field) will be null a lot of the time.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+  @Override
+  public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext context,
+                                             JavaRDD<StructuredRecord> input) throws Exception {
+    FileSet fileSet = context.getDataset(config.fileSetName);
+    Location modelLocation = fileSet.getBaseLocation().append(config.path);
+    if (!modelLocation.exists()) {
+      throw new IllegalArgumentException(String.format("Failed to find model to use for Regression. Location does " +
+                                                         "not exist: {}.", modelLocation));
+    }
+    final List<String> fields = config.getFeatureList(input.first().getSchema());
```
albertshau: if the input is empty, this will throw a NPE.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+      }
+    });
+    return output;
+  }
+
+  /**
+   * Creates a builder based off the given record.
+   */
+  private StructuredRecord.Builder cloneRecord(StructuredRecord record) {
+    Schema schemaToUse = (outputSchema != null) ? outputSchema : getOutputSchema(record.getSchema());
```
albertshau: you can set outputSchema when you get input.first() in the transform method. That way you don't need this check here.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+    List<Schema.Field> fields = new ArrayList<>();
+    fields.addAll(schemaToUse.getFields());
+    fields.addAll(Arrays.asList(Schema.Field.of(config.predictionField, Schema.of(Schema.Type.DOUBLE))));
```
albertshau: fields.add(Schema.Field.of(config.predictionField, Schema.of(Schema.Type.DOUBLE)))

albertshau: actually, schemaToUse isn't needed at all. You can just use outputSchema instead of building another schema for each record.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+        throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
+                                             "Please specify fields for one");
+      }
+      List<String> fields = getFeatureList(inputSchema);
+      for (String field : fields) {
+        Schema schema = inputSchema.getField(field).getSchema();
+        Schema.Type features = schema.isNullableSimple() ? schema.getNonNullable().getType() : schema.getType();
+        Preconditions.checkArgument(features.isSimpleType(), "Field to be used must be of type number: int, long, " +
+          "float, double but was of type %s for field %s ", features, field);
+        if (features.equals(Schema.Type.BYTES) || features.equals(Schema.Type.STRING) ||
```
albertshau: a little easier to read if you just do one check here instead of the isSimpleType() check and then this check and then a null check below

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+        "exist in the input schema.");
+    }
+
+    private List<String> getFeatureList(Schema inputSchema) {
+      List<String> fields = new ArrayList<>();
+      List<Schema.Field> inputSchemaFields = inputSchema.getFields();
+      List<String> excludeFeatures = new ArrayList<>();
+      if (!Strings.isNullOrEmpty(featuresToInclude)) {
+        fields = Arrays.asList(featuresToInclude.split(","));
+        for (String field : fields) {
```
albertshau: it's a little more robust to do `for (String field : Splitter.on(',').trimResults().split(featuresToInclude))`, since it will handle the scenario if a user includes whitespace before or after the comma.

albertshau: o nevermind, i didn't see the trim() call below.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+            throw new IllegalArgumentException(String.format("Field %s does not exists in the input schema", field));
+          }
+        }
+      } else {
+        if (!Strings.isNullOrEmpty(featuresToExclude)) {
+          excludeFeatures = Arrays.asList(featuresToExclude.split(","));
+        }
+        for (Schema.Field field : inputSchemaFields) {
+          String fieldName = field.getName();
+          if (!fieldName.equals(predictionField)) {
```
albertshau: its better to combine the conditions instead of having layers of nesting.

```
if (!fieldName.equals(predictionField) && (excludeFeatures.isEmpty() || !excludeFeature.contains(fieldName))) {
  fields.add(fieldName);
}
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
```diff
@@ -0,0 +1,243 @@
...
+        for (String field : fields) {
+          field = field.trim();
+          Schema.Field inputField = inputSchema.getField(field);
+          if (inputField == null) {
+            throw new IllegalArgumentException(String.format("Field %s does not exists in the input schema", field));
+          }
+        }
+      } else {
+        if (!Strings.isNullOrEmpty(featuresToExclude)) {
+          excludeFeatures = Arrays.asList(featuresToExclude.split(","));
```
albertshau: more robust to do `excludeFeatures.addAll(Splitter.on(',').trimResults().split(featuresToExclude))` since it will handle the scenario if a user includes whitespace before or after the comma.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+import co.cask.cdap.api.dataset.lib.FileSet;
+import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.batch.SparkExecutionPluginContext;
+import co.cask.cdap.etl.api.batch.SparkPluginContext;
+import co.cask.cdap.etl.api.batch.SparkSink;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import com.google.common.primitives.Doubles;
+import joptsimple.internal.Strings;
```
albertshau: think this got mistakenly used instead of guava's Strings

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+ */
+@Plugin(type = SparkSink.PLUGIN_TYPE)
+@Name(DecisionTreeTrainer.PLUGIN_NAME)
+@Description("Trains a regression model based upon a particular label and featuresToInclude of a record.")
+public class DecisionTreeTrainer extends SparkSink<StructuredRecord> {
+  public static final String PLUGIN_NAME = "DecisionTreeTrainer";
+  //Impurity measure of the homogeneity of the labels at the node. Expected value for regression is "variance".
+  private static final String IMPURITY = "variance";
+  private static final int DEFAULT_MAX_DEPTH = 10;
+  private static final int DEFAULT_MAX_BINS = 100;
```
albertshau: its a little cleaner for these defaults to be in the config. In the no-args constructor for the config, you can set these values.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+      "+ 2 leaf nodes. Default is 10.")
+    private Integer maxDepth;
+
+    @Nullable
+    @Description("Maximum number of bins used for splitting when discretizing continuous featuresToInclude. " +
+      "DecisionTree requires maxBins to be at least as large as the number of values in each categorical feature. " +
+      "Default is 100.")
+    private Integer maxBins;
+
+    public DecisionTreeTrainerConfig(String fileSetName, String path, String featuresToInclude,
```
albertshau: create a no-args constructor where the defaults are set

```
DecisionTreeTrainerConfig() {
  maxBins = 100;
  maxDepth = 10;
  ...
}
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+  private static final String IMPURITY = "variance";
+  private static final int DEFAULT_MAX_DEPTH = 10;
+  private static final int DEFAULT_MAX_BINS = 100;
+  private DecisionTreeTrainerConfig config;
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    pipelineConfigurer.createDataset(config.fileSetName, FileSet.class);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
```
albertshau: for a similar reason as the predictor, we should error if the input schema is null

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+    private final String path;
+
+    @Nullable
+    @Description("A comma-separated sequence of fields to use for training. If empty, all the fields will be " +
+      "considered for training. Features to be used, must be from one of the following type: int, long, float or " +
+      "double. Both featuresToInclude and featuresToExclude fields cannot be specified.")
+    private final String featuresToInclude;
+
+    @Nullable
+    @Description("A comma-separated sequence of fields to exclude when training. If empty, all the fields will be " +
```
albertshau: all fields except the label will be used for training

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+  public static class DecisionTreeTrainerConfig extends PluginConfig {
+
+    @Description("The name of the FileSet to save the model to.")
+    private final String fileSetName;
+
+    @Description("Path of the FileSet to save the model to.")
+    private final String path;
+
+    @Nullable
+    @Description("A comma-separated sequence of fields to use for training. If empty, all the fields will be " +
```
albertshau: all fields except the label will be used for training

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+      this.path = path;
+      this.featuresToInclude = featuresToInclude;
+      this.featuresToExclude = featuresToExclude;
+      this.cardinalityMapping = cardinalityMapping;
+      this.labelField = labelField;
+      this.maxDepth = maxDepth;
+      this.maxBins = maxBins;
+    }
+
+    private void validate(Schema inputSchema) {
```
albertshau: a lot of this logic is actually the same for the trainer and the predictor. It is better to create a DecisionTreeConfig, or perhaps a FeatureConfig, that all the configs extend. That way the same logic can be used for all the ML trainer and predictor plugins.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+      List<String> featureList = getFeatureList(inputSchema);
+      Map<Integer, Integer> outputFieldMappings = new HashMap<>();
+      if (!Strings.isNullOrEmpty(cardinalityMapping)) {
+        for (String field : Splitter.on(',').trimResults().split(cardinalityMapping)) {
+          String[] value = field.split(":");
+          if (value.length == 2) {
+            outputFieldMappings.put(featureList.indexOf(value[0]), Integer.parseInt(value[1]));
+          } else {
+            throw new IllegalArgumentException(
+              String.format("Either key or value is missing for 'Categorical Feature  Cardinality Mapping'. Please " +
```
albertshau: this can also happen if there are more than 2 values. Probably better to just have some generic message like:

```
"Invalid categorical feature mapping '%s'. Please make sure it is in the format 'feature':'cardinality'.", field
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    private Map<Integer, Integer> getCategoricalFeatureInfo(Schema inputSchema) {
+      List<String> featureList = getFeatureList(inputSchema);
+      Map<Integer, Integer> outputFieldMappings = new HashMap<>();
+      if (!Strings.isNullOrEmpty(cardinalityMapping)) {
+        for (String field : Splitter.on(',').trimResults().split(cardinalityMapping)) {
+          String[] value = field.split(":");
+          if (value.length == 2) {
+            outputFieldMappings.put(featureList.indexOf(value[0]), Integer.parseInt(value[1]));
```
albertshau: should also catch NumberFormatException in case the number is invalid, and throw an exception with a nice message.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+          throw new IllegalArgumentException(String.format("Value of Prediction field %s value must not be null",
+                                                           config.labelField));
+        }
+        return new LabeledPoint(prediction, Vectors.dense(Doubles.toArray(featureList)));
+      }
+    });
+
+    trainingData.cache();
+
+    config.maxDepth = config.maxDepth == null ? DEFAULT_MAX_DEPTH : config.maxDepth;
```
albertshau: won't be needed if you set defaults in the no-args constructor for the config.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+    JavaRDD<LabeledPoint> trainingData = input.map(new Function<StructuredRecord, LabeledPoint>() {
+      @Override
+      public LabeledPoint call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<>();
+        for (String field : fields) {
+          featureList.add(Double.valueOf(record.get(field).toString()));
+        }
+        Double prediction = record.get(config.labelField);
+        if (prediction == null) {
+          throw new IllegalArgumentException(String.format("Value of Prediction field %s value must not be null",
```
albertshau: Label field

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+                                                           config.labelField));
+        }
+        return new LabeledPoint(prediction, Vectors.dense(Doubles.toArray(featureList)));
```
albertshau: it's common for the feature vector to be sparse, so we should use Vectors.sparse() instead.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
```diff
@@ -0,0 +1,259 @@
...
+    //Empty categoricalFeaturesInfo indicates all features are continuous.
+    Map<Integer, Integer> categoricalFeaturesInfo = config.getCategoricalFeatureInfo(schema);
+    final List<String> fields = config.getFeatureList(schema);
+
+    JavaRDD<LabeledPoint> trainingData = input.map(new Function<StructuredRecord, LabeledPoint>() {
+      @Override
+      public LabeledPoint call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<>();
+        for (String field : fields) {
+          featureList.add(Double.valueOf(record.get(field).toString()));
```
albertshau: need to handle the case when record.get(field) is null. This will be common, as many vectors are usually sparse.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/DecisionTreeRegressionTest.java
```diff
@@ -0,0 +1,338 @@
...
+    // manually trigger the pipeline
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForFinish(5, TimeUnit.MINUTES);
+
+    Assert.assertEquals("FAILED", workflowManager.getHistory().get(0).getStatus().name());
+  }
+
+  @Test
+  public void testInvalidFeatures() throws Exception {
```
albertshau: is this just testing the validate methods in the config? If so, it's better to just 'new' a Config object and call the validate method(s).  That is much faster than this, which requires creating a cdap application and running a workflow.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/DecisionTreeRegressionTest.java
```diff
@@ -0,0 +1,338 @@
...
+    // write records to source
+    DataSetManager<Table> inputManager = getDataset(Id.Namespace.DEFAULT, "flightRecords");
+    MockSource.writeInput(inputManager, messagesToWrite);
+
+    // manually trigger the pipeline
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForFinish(5, TimeUnit.MINUTES);
+    //featurs can only have type: int, long, float, double
+    Assert.assertEquals("FAILED", workflowManager.getHistory().get(0).getStatus().name());
```
albertshau: similar comment here. The config class should have its own test class for all the validation that needs to be done.

albertshau:
overall approach looks good. Comments are mostly about cleanup

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
ananya-bhattacharya: Agreed. But the input schema (fetched using pipelineConfigurer.getStageConfigurer.getInputSchema()) from MockSource to sink stage is null in the configurePipeline() method. The null check is also kept to avoid deploy failure during unit testing.
Will change the comment for if block, to reflect only Junit test scenario. 

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java
ananya-bhattacharya: Raised PR #459 for common logic.
Will update this PR, once that is reviewed and merged, since it is common for the other spark ml plugins.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeRegressor.java
albertshau: Ah ok I see. I have opened https://issues.cask.co/browse/HYDRATOR-947 to add a way to set output schema for the MockSource. Will try to merge it today so that it will be available for you to use.

albertshau: Merged it. You should be able to use it once the build runs and the jar is pushed to maven central. Instead of doing `MockSource.getPlugin(String)`, use the `MockSource.getPlugin(String, Schema)`, and it will set the output schema to what you specified.

spark-plugins/docs/DecisionTreeTrainer-sparksink.md
```diff
@@ -0,0 +1,54 @@
...
+-------
+This example uses the fields ``dofM, dofW, scheduleDepTime, scheduledArrTime, carrier, elapsedTime, originId, destId``
+from the input record as features and ``delayed`` field as the label to train the model.
+
+    {
+        "name": "DecisionTreeRegression",
+        "type": "sparksink",
+        "properties": {
+            "fileSetName": "decision-tree-model",
+        		"path": "decisionTree",
```
albertshau: use consistent indentation

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreePredictor.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @Override
+  public void initialize(SparkExecutionPluginContext context) throws Exception {
+    super.initialize(context);
+    FileSet fileSet = context.getDataset(config.fileSetName);
+    Location modelLocation = fileSet.getBaseLocation().append(config.path);
+    if (!modelLocation.exists()) {
+      throw new IllegalArgumentException(String.format("Failed to find model to use for Regression. Location does " +
+                                                         "not exist: {}.", modelLocation));
```
albertshau: '{}' only works for the Logger. You need to use '%s' for String.format()

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/Flight.java
```diff
@@ -0,0 +1,271 @@
...
+    if (!elapsedTime.equals(flight.elapsedTime)) {
+      return false;
+    }
+    return delayed != null ? delayed.equals(flight.delayed) : flight.delayed == null;
+
+  }
+
+  @Override
+  public int hashCode() {
+    int result;
```
albertshau: nit: we generally use java.util.Objects.hash(dofM, dofW, ....) for this.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/Flight.java
```diff
@@ -0,0 +1,271 @@
...
+    if (destId != flight.destId) {
+      return false;
+    }
+    if (Double.compare(flight.scheduleDepTime, scheduleDepTime) != 0) {
+      return false;
+    }
+    if (distance != flight.distance) {
+      return false;
+    }
+    if (!carrier.equals(flight.carrier)) {
```
albertshau: this can throw an exception if carrier is null. You can use java.util.Objects.equals() and it will handle nulls for you. For consistent style with other classes, you can do something like:

```
return dofM == flight.dofM &&
  dofW == flight.dofW && 
  Objects.equals(carrier, flight.carrier) && 
  ...
```

albertshau (changes requested):
only a few minor comments left. lgtm after that.

albertshau:
lgtm 👍

PR #404, opened 2016-08-17
[HYDRA-681] Remove redundant null checks for plugin config properties

knazir:
After https://github.com/caskdata/cdap/pull/6434, PluginConfig will never have null properties, even in unit tests. We should remove the now-redundant checks and unit tests should still work.

Issue: https://issues.cask.co/browse/HYDRA-681
Build: http://builds.cask.co/browse/HYP-BAD63-1


knazir:
Guild is breen: http://builds.cask.co/browse/HYP-BAD63-1

albertshau:
lgtm

PR #407, opened 2016-08-17
Update to CDAP 3.4.4-SNAPSHOT

wolf31o2:
Use 3.4.4-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
     <!-- properties for script build step that creates the config files for the artifacts -->
     <widgets.dir>widgets</widgets.dir>
     <docs.dir>docs</docs.dir>
-    <app.parents>system:cdap-etl-batch[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT),system:cdap-etl-realtime[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT),system:cdap-data-pipeline[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT)</app.parents>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -122,7 +122,7 @@
     <!-- properties for script build step that creates the config files for the artifacts -->
     <widgets.dir>widgets</widgets.dir>
     <docs.dir>docs</docs.dir>
-    <app.parents>system:cdap-etl-batch[3.4.0,3.5.0-SNAPSHOT),system:cdap-etl-realtime[3.4.0,3.5.0-SNAPSHOT),system:cdap-data-pipeline[3.4.0,3.5.0-SNAPSHOT)</app.parents>
+    <app.parents>system:cdap-etl-batch[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT),system:cdap-etl-realtime[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT),system:cdap-data-pipeline[3.4.4-SNAPSHOT,3.5.0-SNAPSHOT)</app.parents>
```
albertshau: we shouldn't change the lower bound here. In case somebody upgrades from hydrator 3.4.0 to 3.4.4, we still want plugins from this branch to be usable by older 3.4.x versions.

albertshau:
one comment, otherwise lgtm

pom.xml
wolf31o2: Ahh, gotcha. Will update.

PR #410, opened 2016-08-17
HYDRA-692 export package windower is in to make it visible

albertshau:


knazir:
lgtm.. 😅

core-plugins/pom.xml
```diff
@@ -266,6 +266,7 @@
               co.cask.hydrator.plugin.realtime.*;
               co.cask.hydrator.plugin.transform.*;
               co.cask.hydrator.plugin.validator.*;
+              co.cask.hydrator.plugin.streaming.*;
```
awholegunch: Nit: should this moved two lines up, to be alphabetical?

PR #413, opened 2016-08-18
Check for bounding and splitby fields only if split size is greater than 1

sreevatsanraman:
JIRA - https://issues.cask.co/browse/CDAP-7110
Without this fix a pipeline with DBSource cannot be published.
Build: http://builds.cask.co/browse/HYP-BAD72-1


knazir:
lgtm 👍

PR #416, opened 2016-08-19
hdfs delete plugin

thegreenfrog:


thegreenfrog:
Build: http://builds.cask.co/browse/HYP-BAD75

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,117 @@
...
+      listFiles = fileSystem.listStatus(path);
+    }
+
+    for (FileStatus file: listFiles) {
+      path = file.getPath();
+      if (!fileSystem.delete(path, true)) {
+        if (!config.continueOnError) {
+          throw new IOException(String.format("Removal of %s was unsuccessful.", path.toString()));
+        }
+        LOG.error("Removal of {} was unsuccessful.", path.toString());
```
nitinmotgi: It's a warning not a error.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS delete command.
+   */
+  public class HDFSDeleteActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or files that need to be deleted." +
+      " For example: hdfs://hostname/tmp")
+    private String path;
+
+    @Description("Wildcard regular expression to filter the files in the source directory that will be moved")
```
nitinmotgi: Just Regular expression. What is Wildcard regular expression ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Action that deletes file(s) within HDFS in the same cluster.
+ * A user must specify file/directory path.
+ * Optionals include fileRegex
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name("HDFSFileDeleteAction")
+@Description("Action to delete files within HDFS")
```
nitinmotgi: within = on

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,117 @@
...
+    if (config.fileRegex != null) {
+      PathFilter filter = new PathFilter() {
+        private final Pattern pattern = Pattern.compile(config.fileRegex);
+
+        @Override
+        public boolean accept(Path path) {
+          return pattern.matcher(path.getName()).matches();
+        }
+      };
+
```
nitinmotgi: remove space

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,116 @@
...
+  public class HDFSDeleteActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or files that need to be deleted." +
+      " For example: hdfs://hostname/tmp")
+    private String path;
+
+    @Description("Wildcard regular expression to filter the files in the source directory that will be moved")
+    @Nullable
+    private String fileRegex;
+
+    @Description("Indicates if the pipeline should continue if the delete process fails")
```
nitinmotgi: delete fails and not delete process fails.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
thegreenfrog: But we would throw an exception if the user wants the pipeline to fail, so isn't it an error?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private HDFSDeleteActionConfig config;
+
+  public HDFSFileDeleteAction(HDFSDeleteActionConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public void run(ActionContext context) throws Exception {
+    Path path = new Path(config.path);
```
nitinmotgi, line 59: How will this work in the secured cluster ? Doesn't this require you to pass tokens ?

anwar6953, line 59: I believe the tokens just have to be set on the current `UserGroupInformation` and they will be extracted in the `getFileSystem` method call below.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
anwar6953: If we reach this log statement, we've decided not to throw an exception.
So +1 to just having it a `WARN`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
anwar6953, line 59: So, @thegreenfrog doesn't have to worry about passing any tokens.

thegreenfrog:
Build: http://builds.cask.co/browse/HYP-BAD75-3

Works when testing on Secure Cluster

core-plugins/docs/HDFSFileDeleteAction-action.md
```diff
@@ -0,0 +1,44 @@
...
+Use Case
+--------
+This action can be used to remove a file or files in an HDFS cluster.
+
+
+Properties
+----------
+**path:** The valid, full HDFS destination path the file or files to be removed at located.
+
+**fileRegex:** Wildcard regular expression to filter the files in the source directory that will be moved.
```
albertshau: looks like the description for move. update it for delete.

core-plugins/docs/HDFSFileDeleteAction-action.md
```diff
@@ -0,0 +1,44 @@
...
+
+**continueOnError:** Indicates if the pipeline should continue if the delete process fails. If all files are not
```
albertshau, line 23: same here

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS delete command.
+   */
+  public class HDFSDeleteActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or files that need to be deleted." +
+      " For example: hdfs://hostname/tmp")
+    private String path;
+
+    @Description("Regular expression to filter the files in the source directory that will be moved")
```
albertshau: update the description

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,115 @@
...
+          throw new IOException(String.format("Removal of %s was unsuccessful.", path.toString()));
+        }
+        LOG.warn("Removal of {} was unsuccessful.", path.toString());
+      }
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
```
albertshau: if the fileRegex exists, try compiling it here so that an invalid one fails early at pipeline creation rather than failing at runtime. If we're not doing this in the file move action, should do it there too.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,115 @@
...
+        private final Pattern pattern = Pattern.compile(config.fileRegex);
+
+        @Override
+        public boolean accept(Path path) {
+          return pattern.matcher(path.getName()).matches();
+        }
+      };
+      listFiles = fileSystem.listStatus(path, filter);
+    } else {
+      listFiles = fileSystem.listStatus(path);
```
albertshau, line 75: what does this do if the path is a file?

albertshau, line 75: I think you need to handle the case when the path is a file separately from when its a directory.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,115 @@
...
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS delete command.
+   */
+  public class HDFSDeleteActionConfig extends PluginConfig {
+    @Description("The full HDFS path of the file or files that need to be deleted." +
+      " For example: hdfs://hostname/tmp")
```
albertshau: description should include detail about what happens when its a file versus a directory. I would expect this to behave like `hadoop fs -rm -r path`. If the path is a file, its deleted. If the path is a directory and there is no regex, the directory and its contents are deleted. If the path is a directory and there is a regex, only the files and directories matching that regex are deleted. (This is not the current behavior based on the code).

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+      };
+      listFiles = fileSystem.listStatus(path, filter);
+    } else {
+      listFiles = fileSystem.listStatus(path);
+    }
+
+    for (FileStatus file: listFiles) {
+      path = file.getPath();
+      if (!fileSystem.delete(path, true)) {
```
albertshau: this can throw an exception. You need to catch and handle appropriately based on continueOnError

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
albertshau: please add test cases for these scenarios.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -87,7 +80,17 @@ public boolean accept(Path path) {
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    if (config.fileRegex != null) {
+      filter = new PathFilter() {
+        private final Pattern pattern = Pattern.compile(config.fileRegex);
```
albertshau: just do Pattern.compile(config.fileRegex). You don't need everything else.  If it throws an exception, catch it and throw an IllegalArgumentException about the file regex being invalid with the exception message.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -60,15 +61,7 @@ public void run(ActionContext context) throws Exception {
...
-    if (config.fileRegex != null) {
-      PathFilter filter = new PathFilter() {
-        private final Pattern pattern = Pattern.compile(config.fileRegex);
-
-        @Override
-        public boolean accept(Path path) {
-          return pattern.matcher(path.getName()).matches();
-        }
-      };
+    if (filter != null) {
```
albertshau: configurePipeline() and run() are called in different contexts. configure is when the pipeline is deployed. run is per run. So setting filter in configure will not make it available here.

albertshau:
comment about removing a file was not addressed

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
thegreenfrog, line 75: ListFiles is just of size one if the path is a file.

thegreenfrog:
Ok. Made changes based on all comments.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -88,14 +96,11 @@ public void run(ActionContext context) throws Exception {
...
-
-        @Override
-        public boolean accept(Path path) {
-          return pattern.matcher(path.getName()).matches();
-        }
-      };
+      try {
+        Pattern filter = Pattern.compile(config.fileRegex);
+      } catch (Exception e) {
+        throw new IllegalArgumentException(String.format("File regex is invalid", e));
```
albertshau: The exception should be the 2nd argument passed to the constructor, not an argument of String.format(). Should also include more information for it to be useful.

`throw new IllegalArgumentException(String.format("File regex '%s' is invalid: %s", config.fileRegex, e.getMessage()), e);`

core-plugins/docs/HDFSFileDeleteAction-action.md
```diff
@@ -0,0 +1,44 @@
...
+        "plugin": {
+            "name": "HDFSFileDeleteAction",
+            "type": "action",
+            "artifact": {
+                "name": "core-plugins",
+                "version": "1.4.0-SNAPSHOT",
+                "scope": "SYSTEM"
+            },
+            "properties": {
+                "sourcePath": "hdfs://example.com:8020/source/path",
```
albertshau: sourcePath -> path

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -213,19 +195,166 @@ public void testInvalidDestHDFSAction() throws Exception {
...
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(DATAPIPELINE_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "hdfsDeleteDirectoryActionTest");
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+    WorkflowManager manager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    manager.start();
+    manager.waitForFinish(3, TimeUnit.MINUTES);
+
+    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir/source/test.txt")));
```
albertshau: should also assert /dir/source/test2.txt does not exist

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -213,19 +195,166 @@ public void testInvalidDestHDFSAction() throws Exception {
...
+  }
+
+  @Test
+  public void testHDFSDeleteDirectoryAction() throws Exception {
+    Path outputDir = dfsCluster.getFileSystem().getHomeDirectory();
+
+    fileSystem.mkdirs(new Path("dir/source"));
+    fileSystem.createNewFile(new Path("dir/source/test.txt"));
+    fileSystem.createNewFile(new Path("dir/source/test2.txt"));
+    fileSystem.createNewFile(new Path("dir/source/dir2/test2.txt"));
```
albertshau: does this actually create the file? You don't have to create the directory first?

thegreenfrog: I guess my test never checks that. Ill create the dir beforehand.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -213,19 +195,166 @@ public void testInvalidDestHDFSAction() throws Exception {
...
     Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test2.txt")));
     Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/source/test.json")));
   }
+
+  @Test
+  public void testHDFSDeleteAction() throws Exception {
+
+    Path outputDir = dfsCluster.getFileSystem().getHomeDirectory();
+
+    fileSystem.mkdirs(new Path("dir"));
```
albertshau: is the same directory used in all tests?  It doesn't have any affect right now, but it's better to use a different directory per test, or make sure you clean up after the test finishes to make sure tests are isolated from each other.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
albertshau: actually, should be good enough just to check that /dir/source does not exist

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
albertshau: should be calling assertTrue() on these statements

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
thegreenfrog: ok i can change the directories used.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -354,8 +364,8 @@ public void testHDFSDeleteDirectoryAction() throws Exception {
     manager.start();
     manager.waitForFinish(3, TimeUnit.MINUTES);
 
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir/source/test.txt")));
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir/source/test2.txt")));
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir/source/dir2/test2.txt")));
+    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test.txt")));
+    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test2.txt")));
+    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/dir2/test2.txt")));
```
albertshau: should assert the directory itself is deleted. /dir1/source does not exist.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -97,9 +97,10 @@ public boolean accept(Path path) {
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     if (config.fileRegex != null) {
       try {
-        Pattern filter = Pattern.compile(config.fileRegex);
+        Pattern.compile(config.fileRegex);
       } catch (Exception e) {
-        throw new IllegalArgumentException(String.format("File regex is invalid", e));
+        throw new IllegalArgumentException(String.format("File regex %s is invalid: %s",
+                                                         config.fileRegex, e.getMessage(), e));
```
albertshau: again... the exception is not an argument to String.format. It is an argument to the constructor. Your IDE should be highlighting this as a warning.

albertshau:
2 comments left

albertshau:
there is a keytab checked in...

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -91,16 +110,17 @@ public boolean accept(Path path) {
...
     }
+
   }
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     if (config.fileRegex != null) {
       try {
-        Pattern.compile(config.fileRegex);
+        Pattern filter = Pattern.compile(config.fileRegex);
```
albertshau: why add this back? this should generate a warning for an unused variable. Just do Pattern.compile

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -364,8 +366,8 @@ public void testHDFSDeleteDirectoryAction() throws Exception {
     manager.start();
     manager.waitForFinish(3, TimeUnit.MINUTES);
 
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test.txt")));
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test2.txt")));
-    Assert.assertFalse(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/dir2/test2.txt")));
+    Assert.assertFalse(fileSystem.isDirectory(new Path(outputDir.toUri().toString() + "/dir1/source")));
```
albertshau, line 368: isn't it supposed to delete this directory?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -75,8 +75,27 @@ public boolean accept(Path path) {
       listFiles = fileSystem.listStatus(path);
     }
 
+
+
```
albertshau: remove extra lines

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/HDFSFileDeleteAction.java
```diff
@@ -75,8 +75,27 @@ public boolean accept(Path path) {
...
     for (FileStatus file: listFiles) {
-      path = file.getPath();
+      Path currPath = file.getPath();
+      try {
```
albertshau: move this common logic into a method instead of having 2 copies in the code.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
albertshau, line 368: o it's asserting false. How come its asserting false for isDirectory() instead of exists()?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
```diff
@@ -337,6 +338,7 @@ public void testHDFSDeleteDirectoryAction() throws Exception {
     Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test.txt")));
     Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/test2.txt")));
     Assert.assertTrue(fileSystem.exists(new Path(outputDir.toUri().toString() + "/dir1/source/dir2/test2.txt")));
+    Assert.assertTrue(fileSystem.isDirectory(new Path(outputDir.toUri().toString() + "/dir1/source")));
```
albertshau: don't need this, it's redundant because of the line above.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/action/HDFSFileMoveActionTestRun.java
thegreenfrog, line 368: I just figured it is more specific for this Assert case.

core-plugins/docs/HDFSFileDeleteAction-action.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+Use Case
+--------
+This action can be used to remove a file or files in an HDFS cluster.
+
+
+Properties
+----------
+**path:** The valid, full HDFS destination path the file or files to be removed at located.
```
albertshau: this should have all the info in your plugin config's description

albertshau:
one doc comment, otherwise lgtm. fix and merge.

PR #409, opened 2016-08-17
Added macros to core plugin batch sources

knazir:
Enabled macros on the following properties:

`AzureBatchSource(s)`:
- account
- container
- storageKey

`BatchReadableSource(s)`:
- name

`ExcelInputReader`:
- filePath
- filePattern
- memoryTableName
- tableExpiryPeriod
- reprocess
- sheet
- sheetValue
- skipFirstRow
- terminateIfEmptyRow
- rowsLimit
- ifErrorRecord
- errorDatasetName

`FTPBatchSource`:
- path
- fileSystemProperties
- fileRegex
- inputFormatClassName

`FileBatchSource`:
- path
- fileSystemProperties
- fileRegex
- timeTable
- inputFormatClass
- maxSplitSize

`KVTableSource`
- name

`S3BatchSource(s)`:
- accessID
- accessKey

`SnapshotFileBatchSource(s)`:
- name
- basePath
- fileProperties

`TimePartitionedFileSetSource(s)`:
- name
- basePath
- duration
- delay

`XMLReaderBatchSource`:
- path
- nodePath
- pattern
- actionAfterProcess
- targetFolder
- reprocessingRequired
- tableName
- tableExpiryPeriod
- temporaryFolder

`TableSource`:
- rowField

PR #419, opened 2016-08-20
Create dataset during runtime only if it is not created during configure.

sagarkapare:


knazir:
Lgtm

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -309,7 +309,7 @@ private void createDatasets(@Nullable PipelineConfigurer pipelineConfigurer, @Nu
 
         if (pipelineConfigurer != null) {
           pipelineConfigurer.createDataset(excelInputreaderConfig.errorDatasetName, Table.class, datasetProperties);
-        } else if (context != null) {
+        } else if (context != null && !context.datasetExists(excelInputreaderConfig.errorDatasetName)) {
           context.createDataset(excelInputreaderConfig.errorDatasetName, Table.class.getName(), datasetProperties);
```
anwar6953, line 313: Shouldn't this internally bail out if it already exists?

sagarkapare, line 313: We create datasets either during configure time or during runtime depending on the field is macro or not. In this case there is a logical error in which if field is not macro then we were creating dataset in configure time as well as in runtime which causes MapReduce to fail. So fix is to do not attempt the creation at runtime if dataset already created during configure time. 
Similar checks are done at other places https://github.com/caskdata/hydrator-plugins/pull/409/files#diff-a2609c590a55b09cdb5f0132ea8f14a4R119 but was missing here. 

I think this logic can be further simplified but can be done as separate PR (post 3.5).

anwar6953, line 313: Lgtm.

Please file a jira, since I don't think user should have to protect against
already exists.
Like in app configure, they simply call create dataset.

sagarkapare:
ITN is green - http://builds.cask.co/browse/CDAP-ITCC19-JOB1-23. Merging.

PR #423, opened 2016-08-22
bump version

sagarkapare:


bdmogal:
wonder why app.parents is only in a few sub-modules. other than that, lgtm

sreevatsanraman:
@bdmogal good question. I am not sure why it has to be in the sub-modules. It should only be in the parent. I verified that building from this branch generates right plugin.json. Opened this JIRA to track https://issues.cask.co/browse/HYDRATOR-894

bdmogal:
lgtm

PR #341, opened 2016-07-26
SolrSearch Sink Plugin Implementation.

rashig:
SolrSearch Sink Plugin to write data to a SinlgeNode Solr or to SolrCloud. For design and implementation, please refer:
https://wiki.cask.co/display/CE/Solr+Search+or+Cloudera+Search

PR #425, opened 2016-08-22
SolrSearch Sink Implementation

rashig:
SolrSearch Sink Plugin write data to a SinlgeNode Solr or to SolrCloud. For design and implementation, please refer:
https://wiki.cask.co/display/CE/Solr+Search+or+Cloudera+Search

### DIFF
--- a/solrsearch-plugins/docs/SolrSearch-batchsink.md
+++ b/solrsearch-plugins/docs/SolrSearch-batchsink.md
@@ -1,0 +1,77 @@
+# SolrSearch Batch Sink
+
+
+Description
+-----------
+The Solr search batch sink takes the structured record from the input source and indexes it into the Solr server or
+SolrCloud using the collection and key field specified by the user.
+
+The incoming fields from the previous stage in pipelines are mapped to Solr fields. Also, user is able to specify the
+mode of the Solr to connect to. For example, SingleNode Solr or SolrCloud.
+
+Use Case
+--------
+Solr search batch sink is used to write data to the Solr server or SolrCloud. For example, the plugin can be used in
+conjuction with a stream batch source to parse a file and read its contents in Solr.
+
+Properties
+----------
+**referenceName:** This will be used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**solrMode:** Solr mode to connect to. For example, SingleNode Solr or SolrCloud.
+
+**solrHost:** The hostname and port for the Solr server. For example, localhost:8983 if SingleNode Solr or
+zkHost1:2181,zkHost2:2181,zkHost3:2181 for SolrCloud.
+
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**keyField:** Field that will determine the unique key for the document to be indexed. It must match a field name
+in the structured record of the input.
+
+**batchSize:** Number of documents to create a batch and send it to Solr for indexing. After each batch, commit will
+be triggered. Default batch size is 10000. (Macro-enabled)
+
+**outputFieldMappings:** List of the input fields to map to the output Solr fields. This is a comma-separated list of
+key-value pairs, where each pair is separated by a colon ':' and specifies the input and output names. For example,
+'firstname:fname,lastname:lname' specifies that the 'firstname' should be renamed to 'fname' and the 'lastname'
+should be renamed to 'lname'.
+
+Conditions
+----------
+The Solr server should be running prior to creating the application.
+
+All the fields that user wants to index into the Solr server, should be properly declared and defined in Solr's
+schema.xml file. The Solr server schema should be properly defined prior to creating the application.
+
+If keyField('id') in the input record is NULL, then that particular record will be filtered out.
+
+Example
+-------
+This example connects to a 'SinlgeNode Solr' server, running locally at the default port of 8983, and writes the
+data to the specified collection (test_collection). The data is indexed using the id field coming in the input record
+. And also the fieldname 'office address' is mapped to the 'address' field in Solr's index.
+
+    {
+      "name": "SolrSearch",
+      "type": "batchsink",
+        "properties": {
+          "solrMode": "SingleNode",
+          "solrHost": "localhost:8983",
+          "collectionName": "test_collection",
+          "keyField": "id",
+          "batchSize": "10000",
+          "outputFieldMappings": "office address:address"
+        }
+    }
+
+For example, suppose the Solr search sink receives the input record:
+
+    +===================================================================================================+
+    | id : STRING | firstname : STRING  | lastname : STRING |  office address : STRING  | pincode : INT |
+    +===================================================================================================+
+    | 100A        | John                | Wagh              |  NE Lakeside              | 480001        |
+    | 100B        | Brett               | Lee               |  SE Lakeside              | 480001        |
+    +===================================================================================================+
+
+ Once Solr search sink plugin execution is completed, all the rows from input data will be indexed in the
+ test_collection with the fields id, firstname, lastname, address and pincode.
--- a/solrsearch-plugins/docs/SolrSearch-realtimesink.md
+++ b/solrsearch-plugins/docs/SolrSearch-realtimesink.md
@@ -1,0 +1,73 @@
+# SolrSearch Realtime Sink
+
+
+Description
+-----------
+The Solr search realtime sink takes the structured record from the input source and indexes it into the Solr server
+or SolrCloud using the collection and key field specified by the user.
+
+The incoming fields from the previous stage in pipelines are mapped to Solr fields. Also, user is able to specify the
+mode of the Solr to connect to. For example, SingleNode Solr or SolrCloud.
+
+Use Case
+--------
+Solr search realtime sink is used to write data to the Solr server or SolrCloud. For example, the plugin can be used in
+conjuction with a realtime source to parse a file and read its contents in Solr.
+
+Properties
+----------
+**referenceName:** This will be used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**solrMode:** Solr mode to connect to. For example, SingleNode Solr or SolrCloud.
+
+**solrHost:** The hostname and port for the Solr server. For example, localhost:8983 if SingleNode Solr or
+zkHost1:2181,zkHost2:2181,zkHost3:2181 for SolrCloud.
+
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**keyField:** Field that will determine the unique key for the document to be indexed. It must match a field name
+in the structured record of the input.
+
+**outputFieldMappings:** List of the input fields to map to the output Solr fields. This is a comma-separated list of
+key-value pairs, where each pair is separated by a colon ':' and specifies the input and output names. For example,
+'firstname:fname,lastname:lname' specifies that the 'firstname' should be renamed to 'fname' and the 'lastname'
+should be renamed to 'lname'.
+
+Conditions
+----------
+The Solr server should be running prior to creating the application.
+
+All the fields that user wants to index into the Solr server, should be properly declared and defined in Solr's
+schema.xml file. The Solr server schema should be properly defined prior to creating the application.
+
+If idField('id') in the input record is NULL, then that particular record will be filtered out.
+
+Example
+-------
+This example connects to a 'SinlgeNode Solr' server, running locally at the default port of 8983, and writes the
+data to the specified collection (test_collection). The data is indexed using the id field coming in the input record
+. And also the fieldname 'office address' is mapped to the 'address' field in Solr's index.
+
+    {
+      "name": "SolrSearch",
+      "type": "realtimesink",
+        "properties": {
+          "solrMode": "SingleNode",
+          "solrHost": "localhost:8983",
+          "collectionName": "test_collection",
+          "keyField": "id",
+			    "outputFieldMappings": "office address:address"
+        }
+    }
+
+For example, suppose the Solr search sink receives the input record:
+
+    +===================================================================================================+
+    | id : STRING | firstname : STRING  | lastname : STRING |  office address : STRING  | pincode : INT |
+    +===================================================================================================+
+    | 100A        | John                | Wagh              |  NE Lakeside              | 480001        |
+    | 100B        | Brett               | Lee               |  SE Lakeside              | 480001        |
+    +===================================================================================================+
+
+ Once Solr search sink plugin execution is completed, all the rows from input data will be indexed in the
+ test_collection with the fields id, firstname, lastname, address and pincode.
--- a/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
+++ b/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.hydrator.plugin.batch;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.batch.Output;
+import co.cask.cdap.api.data.batch.OutputFormatProvider;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.StageMetrics;
+import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.batch.BatchSink;
+import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.cdap.format.StructuredRecordStringConverter;
+import co.cask.hydrator.plugin.common.SolrOutputFormat;
+import co.cask.hydrator.plugin.common.SolrRecordWriter;
+import co.cask.hydrator.plugin.common.SolrSearchSinkConfig;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import org.apache.hadoop.io.Text;
+
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Batch SolrSearch Sink Plugin - Writes data to a SingleNode Solr or to SolrCloud.
+ */
+@Plugin(type = BatchSink.PLUGIN_TYPE)
+@Name("SolrSearch")
+@Description("This plugin allows users to build the pipelines to write data to Solr. The input fields coming from " +
+  "the previous stage of the pipeline are mapped to Solr fields. User can also specify the mode of the Solr to " +
+  "connect to. For example, SingleNode Solr or SolrCloud.")
+public class SolrSearchSink extends BatchSink<StructuredRecord, Text, Text> {
+  private static final Gson GSON = new Gson();
+  private static final Type SCHEMA_TYPE = new TypeToken<Schema>() { }.getType();
+  private final BatchSolrSearchConfig batchConfig;
+  private StageMetrics metrics;
+
+  public SolrSearchSink(BatchSolrSearchConfig batchConfig) {
+    this.batchConfig = batchConfig;
+  }
+
+  @Override
+  public void prepareRun(BatchSinkContext context) throws Exception {
+    context.addOutput(Output.of(batchConfig.referenceName, new SolrSearchSink.SolrOutputFormatProvider(batchConfig)));
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    batchConfig.validateSolrConnectionString();
+    if (inputSchema != null) {
+      batchConfig.validateKeyField(inputSchema);
+      batchConfig.validateInputFieldsDataType(inputSchema);
+    }
+    batchConfig.validateOutputFieldMappings();
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    metrics = context.getMetrics();
+    //Calling testSolrConnection() before each mapper, to ensure that the connection is alive and available for
+    //indexing.
+    batchConfig.testSolrConnection();
+  }
+
+  @Override
+  public void transform(StructuredRecord structuredRecord, Emitter<KeyValue<Text, Text>> emitter) throws Exception {
+    Schema recordSchema = structuredRecord.getSchema();
+    batchConfig.validateKeyField(recordSchema);
+    batchConfig.validateInputFieldsDataType(recordSchema);
+
+    if (structuredRecord.get(batchConfig.getKeyField()) == null) {
+      metrics.count("invalid", 1);
+      return;
+    }
+    emitter.emit(new KeyValue<Text, Text>((new Text(GSON.toJson(recordSchema, SCHEMA_TYPE))),
+                                          new Text(StructuredRecordStringConverter.toJsonString(structuredRecord))));
+  }
+
+  /**
+   * Output format provider for Batch SolrSearch Sink.
+   */
+  private static class SolrOutputFormatProvider implements OutputFormatProvider {
+    private static final String DEAFULT_BATCHSIZE = "10000";
+    private Map<String, String> conf;
+
+    private SolrOutputFormatProvider(BatchSolrSearchConfig batchConfig) {
+      this.conf = new HashMap<>();
+      conf.put(SolrRecordWriter.SERVER_URL, batchConfig.getSolrHost());
+      conf.put(SolrRecordWriter.SERVER_MODE, batchConfig.getSolrMode());
+      conf.put(SolrRecordWriter.COLLECTION_NAME, batchConfig.getCollectionName());
+      conf.put(SolrRecordWriter.KEY_FIELD, batchConfig.getKeyField());
+      conf.put(SolrRecordWriter.BATCH_SIZE, batchConfig.getBatchSize() == null ? DEAFULT_BATCHSIZE :
+        batchConfig.getBatchSize());
+      if (batchConfig.getOutputFieldMappings() == null) {
+        conf.put(SolrRecordWriter.FIELD_MAPPINGS, "");
+      } else {
+        conf.put(SolrRecordWriter.FIELD_MAPPINGS, batchConfig.getOutputFieldMappings());
+      }
+    }
+
+    @Override
+    public String getOutputFormatClassName() {
+      return SolrOutputFormat.class.getName();
+    }
+
+    @Override
+    public Map<String, String> getOutputFormatConfiguration() {
+      return conf;
+    }
+  }
+
+  /**
+   * Config class for Batch SolrSearch sink.
+   */
+  public static class BatchSolrSearchConfig extends SolrSearchSinkConfig {
+    @Description("Number of documents to create a batch and send it to Solr for indexing. After each batch, " +
+      "commit will be triggered. Default batch size is 10000. (Macro-enabled)")
+    @Nullable
+    @Macro
+    private final String batchSize;
+
+    public BatchSolrSearchConfig(String referenceName, String solrMode, String solrHost, String collectionName,
+                                 String keyField, @Nullable String outputFieldMappings, @Nullable String batchSize) {
+      super(referenceName, solrMode, solrHost, collectionName, keyField, outputFieldMappings);
+      this.batchSize = batchSize;
+    }
+
+    /**
+     * Returns the batch size given as input by user.
+     *
+     * @return batch size
+     */
+    @Nullable
+    public String getBatchSize() {
+      return batchSize;
+    }
+  }
+}
--- a/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrOutputFormat.java
+++ b/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrOutputFormat.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.hydrator.plugin.common;
+
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.NullOutputFormat;
+
+/**
+ * SolrOutputFormat - Output format class for Solr.
+ */
+public class SolrOutputFormat extends NullOutputFormat {
+  @Override
+  public RecordWriter getRecordWriter(TaskAttemptContext context) {
+    return new SolrRecordWriter(context);
+  }
+}
--- a/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
+++ b/solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.hydrator.plugin.common;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.format.StructuredRecordStringConverter;
+import com.google.gson.Gson;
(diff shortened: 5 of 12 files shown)


gsps1:
@rashig have you tried using the EmbeddedSolrServer ? 
Reference : https://wiki.searchtechnologies.com/index.php/Unit_Testing_with_Embedded_Solr

rashig:
@gsps1 
I have already explored the option of EmbeddedSolrServer for testing the complete SolrSearch plugin functionality. The EmbeddedSolrServer works fine independently, however to use it with CDAP test cases, the actual SolrClient object(both Cloud and HttpClient) will need to be mocked with the object of EmbeddedSolrServer. 
However, mocking was not working as expected since SolrClient object is created by private method and called more than once during the program workflow and also extending the HydartorTestBase with different mocking API's, results into errors.

Still exploring the options for unit testing and will raise a separate PR for it.

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,75 @@
...
+    {
+      "name": "SolrSearch",
+      "type": "batchsink",
+        "properties": {
+          "solrMode": "SingleNode",
+          "solrHost": "localhost:8983",
+          "collectionName": "test_collection",
+          "keyField": "id",
+          "batchSize": "10000",
+			    "outputFieldMappings": "office address:address"
```
gsps1: formatting is off

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,75 @@
...
+
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**keyField:** Field that will determine the unique key for the document to be indexed. It must match a field name
+in the structured record of the input.
+
+**batchSize:** Number of documents to create a batch and send it to Solr for indexing. After each batch, commit will
+be triggered. Default batch size is 10000.
+
+**outputFieldMappings:** List of the input fields to map to the output Solr fields. The key specifies the name of the
```
gsps1: is this a list or comma separated string.

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,75 @@
...
+          "keyField": "id",
+          "batchSize": "10000",
+			    "outputFieldMappings": "office address:address"
+        }
+    }
+
+For example, suppose the Solr search sink receives the input record:
+
+    +===================================================================================================+
+    | id : STRING | firstname : STRING  | lastname : STRING |  Office Address : STRING  | pincode : INT |
```
gsps1: config uses "office address" here its "Office Address" do we ignore case ?

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,75 @@
...
+Use Case
+--------
+Solr search batch sink is used to write data to the Solr server or SolrCloud. For example, the plugin can be used in
+conjuction with a stream batch source to parse a file and read its contents in Solr.
+
+Properties
+----------
+**referenceName:** This will be used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**solrMode:** Solr mode to connect to. For example, Single Node Solr or SolrCloud.
```
gsps1: use exact name "SingleNode" as that's how it looks in the example config. doesn't need a space.

solrsearch-plugins/docs/SolrSearch-realtimesink.md
```diff
@@ -0,0 +1,71 @@
+# SolrSearch Realtime Sink
+
+
+Description
+-----------
```
gsps1, line 5: similar comments to batch on this doc too.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  /**
+   * Config class for Batch SolrSearch sink.
+   */
+  public static class BatchSolrSearchConfig extends SolrSearchSinkConfig {
+    @Description("Number of documents to create a batch and send it to Solr for indexing. After each batch, " +
+      "commit will be triggered. Default batch size is 10000.")
+    private final String batchSize;
```
gsps1, line 143: you should enable macros for these fields. maybe in the next PR. but we should consider that when writing plugins.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+    batchConfig.validateSolrConnectionString();
+    if (inputSchema != null) {
+      batchConfig.validateKeyField(inputSchema);
+      batchConfig.validateInputFieldsDataType(inputSchema);
+    }
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    batchConfig.verifySolrConfiguration();
```
gsps1: this check should be done in configurePipeline stage. so it will fail early during deploy time than at runtime.

If you are using macros, then it makes sense to do this check at prepareRun, but currently you are not using macros either.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  @Override
+  public void transform(StructuredRecord structuredRecord, Emitter<KeyValue<Text, Text>> emitter) throws Exception {
+    batchConfig.validateKeyField(structuredRecord.getSchema());
```
gsps1: store structuredRecord once and reuse than calling it multiple times.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+      return conf;
+    }
+  }
+
+  /**
+   * Config class for Batch SolrSearch sink.
+   */
+  public static class BatchSolrSearchConfig extends SolrSearchSinkConfig {
+    @Description("Number of documents to create a batch and send it to Solr for indexing. After each batch, " +
+      "commit will be triggered. Default batch size is 10000.")
```
gsps1: you talk about default batch size, but its not nullable ? don't think that default size is applied. update the description.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+                                          new Text(StructuredRecordStringConverter.toJsonString(structuredRecord))));
+  }
+
+  /**
+   * Output format provider for Batch SolrSearch Sink.
+   */
+  private static class SolrOutputFormatProvider implements OutputFormatProvider {
+    private Map<String, String> conf;
+
+    SolrOutputFormatProvider(BatchSolrSearchConfig batchConfig) {
```
gsps1: private

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,146 @@
...
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    batchConfig.verifySolrConfiguration();
+  }
+
+  @Override
+  public void transform(StructuredRecord structuredRecord, Emitter<KeyValue<Text, Text>> emitter) throws Exception {
+    batchConfig.validateKeyField(structuredRecord.getSchema());
+    batchConfig.validateInputFieldsDataType(structuredRecord.getSchema());
+
+    if (structuredRecord.get(batchConfig.getKeyField()) == null) {
```
gsps1, line 94: do you need this ? keyField is non-nullable, when do you think it will be null?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+    return outputFieldMappings;
+  }
+
+  /**
+   * Creates a map for input field and its mapping for output field, if present.
+   *
+   * @return Map of input fields and its mapping
+   */
+  public Map<String, String> getOutputFieldMap() {
+    Map<String, String> outputFieldMap = new HashMap<String, String>();
```
gsps1, line 119: this method is called often, so every time this is called, you create a  map, parse, add to map and return. very inefficient. parse and store the map in a local variable and return that variable if non-null for subsequent calls.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+    if (StringUtils.isNotEmpty(outputFieldMappings)) {
+      for (String field : Splitter.on(',').trimResults().split(outputFieldMappings)) {
+        String[] value = field.split(":");
+        if (value.length == 2) {
+          outputFieldMap.put(value[0], value[1]);
+        } else {
+          throw new IllegalArgumentException(
+            String.format("Either key or value is missing for 'Fields to rename'. Please make sure that " +
+                            "both key and value are present."));
```
gsps1: `String.format("expected field and its rename field to be separated by ':' example oldFieldName:newFieldName, but received this field %s, field)` including the problematic field in error message is more useful for users to debug.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+   *
+   * @return
+   */
+  @Nullable
+  public String getOutputFieldMappings() {
+    return outputFieldMappings;
+  }
+
+  /**
+   * Creates a map for input field and its mapping for output field, if present.
```
gsps1, line 113: you return an empty map if output field is not preset. include that in javadoc.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+      throw new IllegalArgumentException(String.format("Multiple hosts '%s' found for Single Node Solr.",
+                                                       solrHost));
+    }
+  }
+
+  /**
+   * Verifies whether the Solr server is running or not. Also, validates whether the collection provided by user is
+   * registered with the Solr server or not.
+   */
+  public void verifySolrConfiguration() {
```
gsps1: maybe call it "testSolrConnection"

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+  private static final Gson GSON = new Gson();
+  private static final Type SCHEMA_TYPE = new TypeToken<Schema>() { }.getType();
+  private final SolrSearchSinkConfig config;
+  List<SolrInputDocument> documentList = new ArrayList<SolrInputDocument>();
+  private SolrClient solrClient;
+  private Configuration conf;
+
+  public SolrRecordWriter(TaskAttemptContext context) {
+    conf = context.getConfiguration();
+    config = new SolrSearchSinkConfig(null, conf.get(SERVER_MODE), conf.get(SERVER_URL), conf.get(COLLECTION_NAME),
```
gsps1, line 58: why referenceName is null ? would this cause any issue @albertshau ? also is there any issue with using the same reference name from the config ?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+  public static final String SERVER_URL = "solr.server.url";
+  public static final String SERVER_MODE = "solr.server.mode";
+  public static final String COLLECTION_NAME = "solr.server.collection";
+  public static final String KEY_FIELD = "solr.server.keyfield";
+  public static final String FIELD_MAPPINGS = "solr.output.field.mappings";
+  public static final String BATCH_SIZE = "solr.batch.size";
+  private static final Gson GSON = new Gson();
+  private static final Type SCHEMA_TYPE = new TypeToken<Schema>() { }.getType();
+  private final SolrSearchSinkConfig config;
+  List<SolrInputDocument> documentList = new ArrayList<SolrInputDocument>();
```
gsps1: this should be private final too. same with the following fields.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrOutputFormat.java
```diff
@@ -0,0 +1,30 @@
...
+package co.cask.hydrator.plugin.common;
+
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.output.NullOutputFormat;
+
+/**
+ * SolrOutputFormat - Output format class for Solr.
+ */
+public class SolrOutputFormat extends NullOutputFormat {
```
gsps1, line 25: implement `OutputFormatProvider` instead of extending `NullOutputFormat`, look here for example,
https://github.com/caskdata/hydrator-plugins/blob/develop/cassandra-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/BatchCassandraSink.java#L183

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+      solrFieldName = field.getName();
+      if (config.getOutputFieldMap().containsKey(solrFieldName)) {
+        document.addField(config.getOutputFieldMap().get(solrFieldName), structuredRecord.get(solrFieldName));
+      } else {
+        document.addField(solrFieldName, structuredRecord.get(solrFieldName));
+      }
+    }
+    documentList.add(document);
+    try {
+      if (documentList.size() == Integer.parseInt(conf.get(BATCH_SIZE))) {
```
gsps1: get this batch size in constructor and store it in a variable. not efficient to do this for every write. also use this batch size as initial size for the array list you are creating.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
```diff
@@ -0,0 +1,103 @@
...
+                                      conf.get(KEY_FIELD), conf.get(FIELD_MAPPINGS));
+    solrClient = config.getSolrConnection();
+  }
+
+  @Override
+  public void write(Text key, Text value) throws IOException {
+    String solrFieldName;
+    SolrInputDocument document = new SolrInputDocument();
+
+    Schema inputSchema = GSON.fromJson(key.toString(), SCHEMA_TYPE);
```
gsps1, line 73: should we handle invalid/error record in a better way than to throw exception? @albertshau ?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+  @Description("Name of the collection where data will be indexed and stored in Solr.")
+  private final String collectionName;
+  @Description("Field that will determine the unique key for the document to be indexed. It must match a " +
+    "field name in the structured record of the input.")
+  private final String keyField;
+  @Description("List of the input fields to map to the output Solr fields. The key specifies the name of the field to" +
+    " rename, with its corresponding value specifying the new name for that field.")
+  @Nullable
+  private final String outputFieldMappings;
+
```
gsps1, line 55: enable macros for these fields, in the next PR.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,221 @@
...
+  public SolrSearchSinkConfig(String referenceName, String solrMode, String solrHost, String collectionName,
+                              String keyField, String outputFieldMappings) {
```
gsps1: add Nullable annotation to `outputFieldMappings`

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,106 @@
...
+    config.validateSolrConnectionString();
+    if (inputSchema != null) {
+      config.validateKeyField(inputSchema);
+      config.validateInputFieldsDataType(inputSchema);
+    }
+  }
+
+  @Override
+  public void initialize(RealtimeContext context) throws Exception {
+    uniqueKey = config.getKeyField();
```
gsps1: why uniqueKey, be consistent call it keyField

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,106 @@
...
+    int numRecordsWritten = 0;
+    String solrFieldName;
+    List<SolrInputDocument> documentList = new ArrayList<SolrInputDocument>();
+    SolrInputDocument document;
+
+    for (StructuredRecord structuredRecord : structuredRecords) {
+      config.validateKeyField(structuredRecord.getSchema());
+      config.validateInputFieldsDataType(structuredRecord.getSchema());
+
+      if (structuredRecord.get(uniqueKey) == null) {
```
gsps1: so we will skip records that doesn't have keyField ? then update the description for keyField to be more specific.

you don't seem to be doing this skip for SolrBatch, did you miss adding it ?

gsps1:
took a pass, address the comments. also about your test case attempt, can you checkout a branch with your work in progress, so we can look at it and see what's missing.

gsps1:
not related, but i noticed your source branch is develop, its better to create and work on feature branch, so its easier/cleaner when you want to sync your repo with remote to get develop changes, etc.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: Sure. Will enable the macros for batchSize parameter. Is there any other parameter that we need to enable with macros?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: As discussed and suggested by @albertshau , at configuration time, we should only be checking static things, which means things that will never change.
https://github.com/caskdata/hydrator-plugins/pull/305/files/d9ea06c15a570d2e3ca7a2d2742a392084e274fd#r70322603
Also, as suggested by @anwar6953 , verifying this in initalize() method. 
https://github.com/caskdata/hydrator-plugins/pull/341/files/affafe1a0fbc9c0fedeaec6147e5ddee85a0b933#r74375529
@albertshau  @gsps1  @anwar6953 , can you please suggest the correct method to check this validation?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: KeyField can be of nullable type.  There in no restriction on user. And if the value for keyfield is null, then the row will be filtered out. For more details and discussion:
https://github.com/caskdata/hydrator-plugins/pull/305/files/d9ea06c15a570d2e3ca7a2d2742a392084e274fd..656f6da70d375a0342b0e19c885a974693a6a165#r70747197

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
rashig: 'referenceName' is not required in SolrRecordWriter class. Therefore, keeping this null.
Tested this using Junit and pipeline  and it runs without any issue.
Please let me know, if I am missing anything.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrOutputFormat.java
rashig: As discussed and suggested, changed SolrOutputFormat class that will extend NullOutputFormat.
https://github.com/caskdata/hydrator-plugins/pull/341/files/affafe1a0fbc9c0fedeaec6147e5ddee85a0b933#r74376124

Also, there is SolrOutputFormatProvider(inside SolrSearchSink) class which is following the same structure like the BatchCassandraSink and calls SolrOutputFormat class which is custom OutputFormat class for Solr.
Please let me know, if I am missing anything.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
rashig: 'batchSize' is coming as part of configuration specified by the user  and not available at the time of creating document list.
Therefore, not using this at the time of list creation.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
rashig: Enabling macros for each parameter, may result into the new pipeline which user could clone and create. For example, if user changes the 'collectionName' , it  might change the schema which in turn changes the 'keyField' and 'outputFieldMappings'. Also, for every 'solrMode', the 'solrHost' and other parameters might also change.
Only 'batchSize' is a parameter that can be changed for a pipeline, for every run without affecting the final output and Solr schema.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
rashig: Yes, we are skipping the records where the keyField is null. Details are mentioned in documentations under the "Conditions" section and in design on WIKI as well.
Also, we are doing the same in SolrSearch batch mode.
Please let me know, if anything is missing.

rashig:
Regarding the test cases, had discussion with Sree, will raise a separate PR for the test cases. So that we can debug it properly.

rashig:
Sure. Will create feature branches for future plugins and we have already started that.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,106 @@
...
+    keyField = config.getKeyField();
+    outputFieldMap = config.getOutputFieldMap();
+    solrClient = config.getSolrConnection();
+    config.testSolrConnection();
+  }
+
+  @Override
+  public int write(Iterable<StructuredRecord> structuredRecords, DataWriter dataWriter) throws Exception {
+    int numRecordsWritten = 0;
+    String solrFieldName;
```
anwar6953: no need to have this declaration on separate line than the initialization

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,106 @@
...
+    List<SolrInputDocument> documentList = new ArrayList<SolrInputDocument>();
+    SolrInputDocument document;
```
anwar6953: no need to have this declaration on separate line than the initialization

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Batch SolrSearch Sink Plugin - Writes data to a Sinlge Node Solr or to SolrCloud.
+ */
+@Plugin(type = BatchSink.PLUGIN_TYPE)
+@Name("SolrSearch")
+@Description("This plugin allows users to build the pipelines to write data to Solr. " +
```
abhinavbansal246: nit: break the lines right before 120 character limits

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public void transform(StructuredRecord structuredRecord, Emitter<KeyValue<Text, Text>> emitter) throws Exception {
+    StructuredRecord record = structuredRecord;
+    batchConfig.validateKeyField(record.getSchema());
+    batchConfig.validateInputFieldsDataType(record.getSchema());
+
+    if (record.get(batchConfig.getKeyField()) == null) {
+      return;
```
abhinavbansal246, line 96: Is it possible to log number of records skipped because of keys not present?

anwar6953: or emit it as a metric?

rashig: Tried writing dropped records into the error dataset using 

emitter.emitError(new InvalidEntry<>(31, "Invalid Record", new KeyValue<>
(new Text(GSON.toJson(record.getSchema(), SCHEMA_TYPE)), 
new Text (StructuredRecordStringConverter.toJsonString(record)))));

However, while running the pipeline, getting this error:
Transform : SolrSearch has error records, but does not have a error dataset configured.
Please let me know, if I am missing anything.

rashig: Working on the error dataset implementation, will update the PR once done.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
gsps1: if some field is used to determine schema and validate, then it makes sense to not make them macro enabled.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrRecordWriter.java
gsps1: not sure what you mean.. i asked you to initialize batchSize in the constructor `public SolrRecordWriter(TaskAttemptContext context) {` which is where you initialize `conf`. `batchSize` is not something that's going to change for every write, as its part of config.

rashig: Already changed this in the latest commit 0701f13.
Please let me know, if I am missing anything.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
rashig: Only batchSize seems to be the right parameter in this case, for enabling macro.
Added macro to batchSize, in the latest commit.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: As discussed with @abhinavbansal246 and team, will use inbuilt capabilities to capture the error/dropped records.
Hence, currently not including the error dataset in this PR. 
Will create separate JIRA as enhancement task for including this feature, once the framework supports the error dataset in sink type of plugins.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,154 @@
...
+    batchConfig.validateSolrConnectionString();
+    if (inputSchema != null) {
+      batchConfig.validateKeyField(inputSchema);
+      batchConfig.validateInputFieldsDataType(inputSchema);
+    }
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    batchConfig.testSolrConnection();
```
gsps1, line 85: can you add a comment about why you want to testConnection in initialize ? do you want each map task to test connection ? or can this be skipped to be done in configure or initialize stage?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public void transform(StructuredRecord structuredRecord, Emitter<KeyValue<Text, Text>> emitter) throws Exception {
+    StructuredRecord record = structuredRecord;
```
gsps1: why store duplicate ? i believe the suggestion was to store schema, something like 
`Schema schema = structuredRecord.getSchema()` and use this schema for repeated use.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
gsps1: for now emit a metric as ali suggested, `metrics.count("invalid", 1);`

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: As discussed and suggested by @albertshau , at configuration time, we should only be checking static things, which means things that will never change.
However, calling testConnection() inside transform method will not be feasible at all, as it will be called for each record.
So the last option that seems feasible (Also @anwar6953 has suggested the same), was to call testConnection() in initialize() method. And also, it would be good to check connection before each mapper, so that we can catch the failures atleast before transform method.
For discussion details, please refer:
https://github.com/caskdata/hydrator-plugins/pull/425/files/b1c95907c73cf27a2d0bf514a704ca8ed9d8d2dd#r77234560 
Added comments in the code also.

Please let me know, if this is OK?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,160 @@
...
+      batchConfig.validateKeyField(inputSchema);
+      batchConfig.validateInputFieldsDataType(inputSchema);
+    }
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    metrics = context.getMetrics();
+    /* Calling testSolrConnection() before each mapper, to ensure that the connection is alive and available for
+    indexing.*/
```
gsps1: nit : use single line comments

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
```diff
@@ -0,0 +1,160 @@
...
+    Schema recordSchema = structuredRecord.getSchema();
+    batchConfig.validateKeyField(recordSchema);
+    batchConfig.validateInputFieldsDataType(recordSchema);
+
+    if (structuredRecord.get(batchConfig.getKeyField()) == null) {
+      metrics.count("invalid", 1);
+      return;
+    }
+    emitter.emit(new KeyValue<Text, Text>(new Text(GSON.toJson(recordSchema, SCHEMA_TYPE)),
+                                          new Text(StructuredRecordStringConverter.toJsonString(structuredRecord))));
```
gsps1, line 99: any reason for emitting `keyvalue<text,text>` or can it be Strings ? if there's a reason  add comments.

solrsearch-plugins/docs/SolrSearch-batchsink.md
```diff
@@ -0,0 +1,75 @@
...
+**collectionName:** Name of the collection where data will be indexed and stored in Solr.
+
+**keyField:** Field that will determine the unique key for the document to be indexed. It must match a field name
+in the structured record of the input.
+
+**batchSize:** Number of documents to create a batch and send it to Solr for indexing. After each batch, commit will
+be triggered. Default batch size is 10000. (Macro-enabled)
+
+**outputFieldMappings:** A comma-separated list of the input fields to map to the output Solr fields. The key specifies
+the name of the field to rename, with its corresponding value specifying the new name for that field.
```
gsps1: this doesn't say anything about using ":" for separating field from its field mapping, as that's the example you are using. what if there's a colon in field name, how do you escape that ?

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/realtime/RealtimeSolrSearchSink.java
```diff
@@ -0,0 +1,111 @@
...
+    if (inputSchema != null) {
+      config.validateKeyField(inputSchema);
+      config.validateInputFieldsDataType(inputSchema);
+    }
+  }
+
+  @Override
+  public void initialize(RealtimeContext context) throws Exception {
+    keyField = config.getKeyField();
+    outputFieldMap = config.getOutputFieldMap();
```
gsps1: if field map is not in the expected format, then it would throw exception, but that should be checked and thrown at config time than during initialize.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,223 @@
...
+    super(referenceName);
+    this.solrMode = solrMode;
+    this.solrHost = solrHost;
+    this.collectionName = collectionName;
+    this.keyField = keyField;
+    this.outputFieldMappings = outputFieldMappings;
+  }
+
+  /**
+   * Returns the Solr mode selected as input by user.
```
gsps1: remove `selected as input by user` from java docs, same for other methods.

gsps1 (commented):
few more comments.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/common/SolrSearchSinkConfig.java
```diff
@@ -0,0 +1,223 @@
...
+    return outputFieldMappings;
+  }
+
+  /**
+   * Creates a map for input field and its mapping for output field, if present.
+   * If mapping is not present, then empty map will be returned.
+   *
+   * @return Map of input fields and its mapping or empty
+   */
+  public Map<String, String> getOutputFieldMap() {
```
gsps1: its better to make use of `KeyValueListParser` for doing the parsing look at how ProjectionTransform does it for parsing rename field. its similar.

solrsearch-plugins/src/main/java/co.cask.hydrator.plugin/batch/SolrSearchSink.java
rashig: Kept type to org.apache.hadoop.io.Text as the SolrOutputFormat and SolrRecordWriter classes are specifically related to Solr indexing with mapreduce functionality. And since SolrRecordWriter extends RecordWriter<Text, Text>, hence kept emitter.emit(new KeyValue<Text, Text> type to Text also. 

However, tested the plugin by changing the type to java.lang.String (in both SolrRecordWriter and SolrSearchSink classes), it worked.
Please let me know, should we change it to String type or the Text is OK.

gsps1:
LGTM, squash commits.

PR #426, opened 2016-08-22
[CDAP-6661] Order by aggregate plugin

ananya-bhattacharya:
The Order By a aggregator plugin that sorts structured records based on fields and order specified by user. 
For design and use case please refer - https://wiki.cask.co/display/CE/Order+By


ananya-bhattacharya:
This PR is reopened against https://github.com/caskdata/hydrator-plugins/pull/342.

sreevatsanraman:
@ananya-bhattacharya this PR should not be merged. This is just investigation work. The changes for order by will need an update to data pipeline app. As mentioned in the Implementation Tips in the design wiki https://wiki.cask.co/display/CE/Order+By

Implementation Tips
- Investigate how ‘Group Comparator’ and ‘Sort Comparator’ work together and be used to achieve the functionality for this plugin.
- Build a simple map-reduce program to show understand how the above functionality work — Implement Sort Comparator using StructuredRecord
- If the above works, then Data Pipeline Application Template need to be modified to the set the sort class comparator and this shouldn’t affect the other plugins.

ananya-bhattacharya:
Yes, agreed. Discussed this with @albertshau; the requirements will need to be revisited.
@prinam is going to share the updates on this.

russorat:
Closing for now

PR #347, opened 2016-07-27
[CDAP-6662] Logistic Regression Trainer implementation

romy-khetan:

PR #427, opened 2016-08-22
[HYDRATOR-517] Logistic Regression Trainer and Classifier Plugins implementation

romy-khetan:


abhinavbansal246:
https://github.com/caskdata/hydrator-plugins/pull/347

spark-plugins/docs/LogisticRegressionClassifier-sparkcompute.md
```diff
@@ -0,0 +1,70 @@
...
+
+Use Case
+--------
+This transform can be used when you have a saved Logistic Regression model and want to classify data.
+
+Properties
+----------
+**fileSetName:** The name of the FileSet to load the model from.
+
+**path:** Path of the FileSet to load the model from.
```
abhinavbansal246, line 16: is path always required? if not please mention here

spark-plugins/docs/LogisticRegressionClassifier-sparkcompute.md
```diff
@@ -0,0 +1,70 @@
...
+
+**fieldsToClassify:** A space-separated sequence of fields to use for classification.
```
albertshau: should make it comma separated so that it is consistent with other plugins

spark-plugins/docs/LogisticRegressionClassifier-sparkcompute.md
```diff
@@ -0,0 +1,77 @@
...
+**featureFieldsToInclude:** A comma-separated sequence of fields that needs to be used for training.
+
+**featureFieldsToExclude:** A comma-separated sequence of fields that needs to be excluded from being used in training.
+
+**predictionField:** The field on which prediction needs to be set. It will be of type double.
+
+**numFeatures:** The number of features to use when classifying with the trained model. This should be the same as
+the number of features used to train the model in LogisticRegressionTrainer. Default is 100.
+
+Condition
```
albertshau: you don't need this extra section. Just write the sentences without numbering.

spark-plugins/docs/LogisticRegressionTrainer-sparksink.md
```diff
@@ -0,0 +1,53 @@
...
+**featureFieldsToExclude:** A comma-separated sequence of fields that needs to be excluded from being used in training.
+
+**labelField:** The field from which to get the prediction. It must be of type double.
+
+**numFeatures:** The number of features to train the model with. This should be the same as the number of features
+used for the LogisticRegressionClassifier. The default value if none is provided will be 100.
+
+**numClasses:** The number of classes to use in training the model. It must be of type integer. Default is 2.
+
+Condition
```
albertshau: don't need this extra section, just have the sentences without numbering

spark-plugins/pom.xml
```diff
@@ -24,7 +24,7 @@
   </parent>
 
   <properties>
-    <app.parents>system:cdap-data-pipeline[4.0.0-SNAPSHOT,4.1.0-SNAPSHOT),system:cdap-data-streams[4.0.0-SNAPSHOT,4.1.0-SNAPSHOT)</app.parents>
+    <app.parents>system:cdap-data-pipeline[3.5.0-SNAPSHOT,3.5.0-SNAPSHOT),system:cdap-data-streams[3.5.0-SNAPSHOT,3.5.0-SNAPSHOT)</app.parents>
```
albertshau: this should not be changed

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
```diff
@@ -0,0 +1,176 @@
...
+      this.featureFieldsToExclude = featureFieldsToExclude;
+      this.predictionField = predictionField;
+      this.numFeatures = numFeatures;
+    }
+
+    private void validate(Schema inputSchema) {
+      Schema.Field predictionField = inputSchema.getField(this.predictionField);
+      Preconditions.checkArgument(predictionField == null, "Prediction field must not already exist in input schema.");
+
+      SparkUtils.validateConfigParameters(inputSchema, featureFieldsToInclude, featureFieldsToExclude,
```
albertshau: It would be better if there was a concrete class that the configs extended rather than a utils class. For example, MLConfig, which contains all common properties and a validate() method.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
```diff
@@ -0,0 +1,176 @@
...
+          outputSchema = (outputSchema != null) ? outputSchema :
+            SparkUtils.getOutputSchema(inputSchema, config.predictionField);
+
+          return SparkUtils.cloneRecord(structuredRecord, outputSchema, config.predictionField).
+            set(config.predictionField, prediction).build();
+        }
+      });
+      return output;
+    }
+    return null;
```
albertshau: `return context.getSparkContext().emptyRDD();`

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    pipelineConfigurer.createDataset(config.fileSetName, FileSet.class);
+
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
+    SparkUtils.validateConfigParameters(inputSchema, config.featureFieldsToInclude, config.featureFieldsToExclude,
```
albertshau: as mentioned above, this would be cleaner if it was in a concrete parent class instead of a utils class.

albertshau (changes requested):
most minor comments

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
romy-khetan: For parent ML Config class, current implementation would require two different parent classes for both Trainer and Classifier/Predictor plugins since description for the common properties like "fileSetName", "path" , "featuresToInclude" and "featuresToExclude" are different for the plugins.
For ex:  In Trainer plugin, fileSetName will have description like: The name of the FileSet to save the model to.
And, in Classifier/Predictor plugin, the description will be like: The name of the FileSet to load the model from.
Also in this case, the validation code and other functions like getFeatureList() or getCategoricalInfo() which uses these config details will get repeated. And only getCloneRecord() plus related method will remain in SparkUtils class.

That's why we kept all the required methods in SparkUtils class. Please let me know, if still the parent class needs to be created.

albertshau: You can have one for everything that is common across all (features, etc), one for trainers that extends that common one, one for predictors that extends that common one, then each individual plugin extends the trainer or predictor one. 

In any case, we can keep the utils for now and as we add more ML plugins, see which approach makes more sense.

romy-khetan: Updated the common config classes, one for Trainer and other for Classifiers/Predictors.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /**
+   * Configuration for the LogisticRegressionClassifier.
+   */
+  public static class Config extends MLPredictorConfig {
+    @Nullable
+    @Description("The number of features to use in training the model. It must be of type integer and equal to the" +
+                  " number of features used in LogisticRegressionTrainer. The default value if none is provided " +
+                  " will be 100.")
```
albertshau: nit: To avoid double spaces or missing spaces, we usually have the space at the end of the previous line, and don't start with a space at the start of the line. This, for example, will end up with 2 spaces.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
```diff
@@ -0,0 +1,147 @@
...
+      final JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
+        @Override
+        public StructuredRecord call(StructuredRecord structuredRecord) throws Exception {
+          List<Object> result = new ArrayList<>();
+          for (String column : featuresList.keySet()) {
+            if (structuredRecord.get(column) != null) {
+              result.add(structuredRecord.get(column));
+            }
+          }
+          double prediction = loadedModel.predict(tf.transform(result));
```
albertshau: should not be using hashing tf. Feature generation should happen before this stage.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * Configuration for the LogisticRegressionTrainer.
+   */
+  public static class Config extends MLTrainerConfig {
+
+    @Nullable
+    @Description("The number of features to use in training the model. It must be of type integer and equal to the" +
+      " number of features used in LogisticRegressionClassifier. The default value if none is provided will be" +
+      " 100.")
+    private final Integer numFeatures;
```
albertshau: this property should not exist, feature generation should be handled separately.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
```diff
@@ -0,0 +1,139 @@
...
+@Description("Uses Logistic Regression to train a model based upon whether messages are spam or not.")
+public class LogisticRegressionTrainer extends SparkSink<StructuredRecord> {
+  public static final String PLUGIN_NAME = "LogisticRegressionTrainer";
+
+  private Config config;
+
+  /**
+   * Configuration for the LogisticRegressionTrainer.
+   */
+  public static class Config extends MLTrainerConfig {
```
albertshau, line 48: if we're having this, it should be its own class and not an inner class. Otherwise there is something generic inside something specific.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,175 @@
...
+  protected String featureFieldsToExclude;
+
+  @Description("The field on which to set the prediction. It will be of type double.")
+  protected String predictionField;
+
+  public MLPredictorConfig() {
+  }
+
+  public MLPredictorConfig(String fileSetName, String path, @Nullable String featureFieldsToInclude,
+                         @Nullable String featureFieldsToExclude, String predictionField) {
```
albertshau: nit: fix indentation

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
albertshau: OK I see there is a separate class down below. Did you just forget to delete this?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,175 @@
...
+   * Get the feature list of the features that have to be used for training/prediction depending on the
+   * featuresToInclude or featuresToInclude list.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the label values.
+   * @return feature list to be used for training/prediction.
+   */
+  static Map<String, Integer> getFeatureList(Schema inputSchema, @Nullable String featuresToInclude,
```
albertshau: same comment for this, and for the other methods about not being static. Also, there is a lot of duplication with the trainer config. All the duplication should be in a common parent class.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Validate the config parameters for the spark sink class.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training.
+   * @param featuresToExclude features to be excluded when training.
+   * @param predictionField   field containing the label values.
+   */
+  public static void validateConfigParameters(Schema inputSchema, @Nullable String featuresToInclude,
```
albertshau: nit: just name this validate() to make it consistent with other configs.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,175 @@
...
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the label values.
+   * @return feature list to be used for training/prediction.
+   */
+  static Map<String, Integer> getFeatureList(Schema inputSchema, @Nullable String featuresToInclude,
+                                             @Nullable String featuresToExclude, String predictionField) {
+    if (!Strings.isNullOrEmpty(featuresToExclude) && !Strings.isNullOrEmpty(featuresToInclude)) {
```
albertshau: you can still call that utils class from here instead of copying all the code. Otherwise there is a lot of code duplication between this and the trainer config.

albertshau (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionClassifier.java
romy-khetan: Yes Albert. we can do it once after Feature generation PR is merged.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
romy-khetan: Yes Albert. we can remove it once after Feature generation PR is merged.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
romy-khetan: There is are properties for Logistic Trainer class other than the properties mentioned in MLTrainerConfig class (which is a common class for Spark ML Trainer Configs). That's why i extended MLTrainerConfig class to add numFeatures and numClass properties.

romy-khetan:
@albertshau 
Updated PR with the review comments.

albertshau:
now that the decision tree plugins have been merged, all the other ML plugins should be refactored so that all the common logic is shared rather than copy-pasted across a bunch of classes.

romy-khetan:
Logistic Regression already has the refactored code which share the common logic(Spark Utils). Please review.

albertshau:
That is not what I meant. If you look at these plugins and the decision tree plugins, they are almost the same. Configuring the pipeline is almost the same. Initialization is almost the same except for one line. Converting the record into a vector _should_ be almost the same, but it isn't because you haven't removed HashingTF like I mentioned earlier. The same is true for all the other PRs that are open.

When there is a lot of code duplication, things should be refactored. In this case, that means creating an abstract class for predictors and an abstract class for trainers. Those classes should contain all the duplicate code, with the actual plugins only implementing what is different.

romy-khetan:
@albertshau 
Added SparkMLTrainer as an abstract class which extends SparkSink. Methods added in SparkMLTrainer are mainly containing configuration, validation and extracting feature list.
However, could not add "convertion of record into a vector" functionality because they are somewhat different in each of the trainer plugins. Also saving the model into the fileset is different for each of the plugin as there are different models and does not have a common parent class. 
Please review.
Once review for SparkMLTrainer has been done, SparkMLPredictor PR can be raised.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    configure(pipelineConfigurer, config.fileSetName, config.featureFieldsToInclude, config.featureFieldsToExclude,
+              config.labelField, null);
+  }
+
+  @Override
+  public void run(SparkExecutionPluginContext context, JavaRDD<StructuredRecord> input) throws Exception {
+    final HashingTF tf = new HashingTF((config.numFeatures));
```
albertshau: I've mentioned it 2 times already, but you should _not_ be doing this to generate the feature vector. It is not useful for anything unless the incoming records have a bunch of string fields that have been pre-split.

Instead, you should be doing it _exactly_ the same way the decision tree plugin is doing it, and the code should be in the parent class.

Another reason they should all share the same code is that after the feature generator plugins are merged, we will likely just take the features from a single array or record field instead of how it is being done now. At that point, we should just have to change a single class instead of 10 classes.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
```diff
@@ -0,0 +1,126 @@
...
+                  @Nullable Integer numClasses) {
+      super(fileSetName, path, featureFieldsToInclude, featureFieldsToExclude, labelField);
+      this.numFeatures = numFeatures;
+      this.numClasses = numClasses;
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    configure(pipelineConfigurer, config.fileSetName, config.featureFieldsToInclude, config.featureFieldsToExclude,
```
albertshau: Rather than having a method that just calls a parent method, it is better for the parent to just implement the method and for the child to leave it out.

In this case, the parent should implement configurePipeline, and the method should not be implemented in this class.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Validate the config parameters for the spark sink class.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training.
+   * @param featuresToExclude features to be excluded when training.
+   * @param predictionField   field containing the label values.
+   */
+  public static void validate(Schema inputSchema, @Nullable String featuresToInclude,
```
albertshau: this should not be static. It should just use the fields in this class instead of passing everything in.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,98 @@
...
+                           @Nullable String featureFieldsToExclude, String labelField) {
+      this.fileSetName = fileSetName;
+      this.path = path;
+      this.featureFieldsToInclude = featureFieldsToInclude;
+      this.featureFieldsToExclude = featureFieldsToExclude;
+      this.labelField = labelField;
+    }
+  }
+
+  protected void configure(PipelineConfigurer pipelineConfigurer, String fileSetName,
```
albertshau: instead of doing this, the plugin should take an MLTrainerConfig in its constructor and store it as a private final field. This method should override configurePipeline instead of forcing child classes to implement it.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,98 @@
...
+                           @Nullable String featureFieldsToInclude, @Nullable String featureFieldsToExclude,
+                           String labelField, @Nullable String cardinalityMapping) {
+    pipelineConfigurer.createDataset(fileSetName, FileSet.class);
+
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
+    validate(inputSchema, featureFieldsToInclude, featureFieldsToExclude, labelField, cardinalityMapping);
+  }
+
+  protected void validate(Schema inputSchema, @Nullable String featuresToInclude,
```
albertshau: this should be a method in the config, and it shouldn't take any arguments.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/LogisticRegressionSpamMessageModel.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public static LogisticRegressionSpamMessageModel fromStructuredRecord(StructuredRecord structuredRecord) {
+    return new LogisticRegressionSpamMessageModel((Integer) structuredRecord.get(IMP_FIELD),
+                                                  (Double) structuredRecord.get(READ_FIELD),
+                                                  (Double) structuredRecord.get(SPAM_PREDICTION_FIELD));
+  }
+
+  @Override
+  public int hashCode() {
+    int result = text.hashCode();
```
albertshau: for consistency with other classes, use Objects.hash()

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/LogisticRegressionSpamMessageModel.java
```diff
@@ -0,0 +1,98 @@
...
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    LogisticRegressionSpamMessageModel that = (LogisticRegressionSpamMessageModel) o;
+
+    if (!text.equals(that.text)) {
```
albertshau: for consistency with other classes, and to protect against nulls, use Objects.equals()

```
return Objects.equals(text, that.text) &&
  Objects.equals(read, that.read) &&
  ...
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/LogisticRegressionTrainer.java
albertshau: In fact, your test case sends in a boolean and a integer field to be hashed. That makes no sense and should have been an indicator that this is not correct.

albertshau (changes requested):
The entire purpose of creating a parent class is so that these plugins can share common code with the existing decision tree plugins (and future plugins).  Please make changes so that both trainers extend the same class.

romy-khetan:
ITN results:
[INFO]
[INFO] --- maven-checkstyle-plugin:2.12.1:check (validate) @ long-running-test ---
[INFO]
[INFO] --- maven-surefire-plugin:2.18.1:test (default-test) @ long-running-test ---
[INFO] Tests are skipped.
[INFO] Reactor Summary:
[INFO]
[INFO] integration-test ................................... SUCCESS [  3.905 s]
[INFO] integration-test-core .............................. SUCCESS [  3.275 s]
[INFO] integration-test-remote ............................ SUCCESS [04:35 min]
[INFO] integration-test-standalone ........................ SUCCESS [  0.614 s]
[INFO] upgrade-test ....................................... SUCCESS [  1.242 s]
[INFO] long-running-test .................................. SUCCESS [  1.287 s]
[INFO] BUILD SUCCESS
[INFO] Total time: 04:46 min
[INFO] Finished at: 2016-11-03T16:03:27+05:30
[INFO] Final Memory: 70M/255M

romy-khetan:
@albertshau @abhinavbansal246 
Refactored code according to base class trainer and predictor model. Please review.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/MLPredictorConfig.java
```diff
@@ -0,0 +1,76 @@
...
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.plugin.PluginConfig;
+
+import javax.annotation.Nullable;
+
+/**
+ * Config class for Classifiers/Predictors. Contains common config properties and validation methods to be used
+ * in Classifiers/Predictors.
+ */
+public class MLPredictorConfig extends PluginConfig {
```
albertshau: there is already a class like this in SparkMLPredictor, you can remove this.

albertshau (changes requested):
just one comment about removing a redundant class

romy-khetan:
ITN results:
[INFO] --- maven-checkstyle-plugin:2.12.1:check (validate) @ long-running-test ---
[INFO]
[INFO] --- maven-surefire-plugin:2.18.1:test (default-test) @ long-running-test ---
[INFO] Tests are skipped.
[INFO] Reactor Summary:
[INFO]
[INFO] integration-test ................................... SUCCESS [04:31 min]
[INFO] integration-test-core .............................. SUCCESS [  4.272 s]
[INFO] integration-test-remote ............................ SUCCESS [04:39 min]
[INFO] integration-test-standalone ........................ SUCCESS [  0.624 s]
[INFO] upgrade-test ....................................... SUCCESS [  1.291 s]
[INFO] long-running-test .................................. SUCCESS [  1.417 s]
[INFO] BUILD SUCCESS
[INFO] Total time: 09:18 min
[INFO] Finished at: 2016-11-04T12:05:38+05:30
[INFO] Final Memory: 83M/343M

romy-khetan:
@albertshau @abhinavbansal246 
Removed unused class. Please review.

albertshau:
lgtm

PR #429, opened 2016-08-23
Tweet Post Action

knazir:
# Tweet Post Action Plugin

Cask Hydrator allows data scientists to create ETL pipelines through a simple drag-and-drop interface. However, is that really good enough?

The world of big data is indeed very, very big. The face of the field is evolving almost on a monthly basis. In this ever-changing world, the question is: how can a data scientist simultaneously do work while reminding those around them that they are, in fact, doing work? The answer is social media, which often goes hand-in-hand with big data. 

Now, with the Tweet post-action, pipeline operators can post a Tweet conditionally based on the result of an ETL pipeline. Here are some examples:
- On completion:
  `Just finished a run of an ETL Pipeline with #CaskHydrator! Boy, that was a lot of fun, y'all should try it out.`
- On success:
  `Successfully finished my very own ETL Pipeline with #CaskHydrator! Happy to throw more money at @CaskData.`
- On failure:
  `My f****** pipeline failed... thanks a lot @CaskData #FML`

As a company, Cask Data is committed to growing and evolving with the needs of its users. We believe that this Tweet post action is a testament to how hip, far out, and down with the kids we are.

> “If Paul Revere had been a modern day citizen, he wouldn't have ridden down Main Street. He would have tweeted." - @AlecJRoss

ᵀʰᶦˢ ᵐᵃʸ ᵒʳ ᵐᵃʸ ᶰᵒᵗ ᵇᵉ ᵃ ʲᵒᵏᵉ

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/TweetAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/TweetAction.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.action;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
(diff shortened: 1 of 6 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/TweetAction.java
```diff
@@ -0,0 +1,150 @@
...
+    @Macro
+    private String accessTokenSecret;
+
+    @Name("Tweet")
+    @Description("The message to post with the Tweet, do you need any more explanation?")
+    @Macro
+    @Nullable
+    private String tweet;
+
+    @Name("Namespace")
```
nitinmotgi, line 128: Remove namespace.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/TweetAction.java
```diff
@@ -0,0 +1,150 @@
...
+    @Description("The namespace this pipeline will be run in.")
+    @Macro
+    @Nullable
+    private String namespace;
+
+    @Name("PipelineName")
```
nitinmotgi, line 134: Remove this too

russorat:
Closing this for now. Will need to redesign not using selenium and firefox.

PR #430, opened 2016-08-23
RowDenormalier - Adding Error Dataset (HYDRATOR-522)

rashig:
Dropped records will be added to error dataset.
JIRA ticket: https://issues.cask.co/browse/HYDRATOR-522


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -280,12 +294,22 @@ public void testDenormalizerWithNullValues() throws Exception {
...
     }
+
+    DataSetManager<Table> recordsManager = getDataset(errorDatasetName);
+    Table dropRecordsTable = recordsManager.get();
+    Scanner scanner = dropRecordsTable.scan(null, null);
+    int counter = 0;
+    while (scanner.next() != null) {
+      counter++;
+    }
+    Assert.assertEquals("Expected drop records", 1, counter);
```
gokulavasan: Can we also check the errorDataset record to make sure it actually got written?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregatorTest.java
```diff
@@ -158,12 +161,22 @@ public void testDenormalizerWithMultipleKeyFieldValues() throws Exception {
         Assert.assertEquals("PQR1 place near XYZ1", record.get("Addr").toString());
       }
     }
+
+    DataSetManager<Table> recordsManager = getDataset(errorDatasetName);
```
gokulavasan: Looks like this functionality can be moved into a method? 
`private int getErrorDatasetSize(String errorDatasetName)`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
```diff
@@ -87,6 +104,11 @@ public void groupBy(StructuredRecord record, Emitter<String> emitter) throws Exc
         throw new IllegalArgumentException(
           String.format("Keyfield '%s' does not exist in input schema %s", keyField, record.getSchema()));
       }
+      StructuredRecord.Builder droppedRecordBuilder = StructuredRecord.builder(errorRecordSchema);
```
gokulavasan, line 106: So we are using recordCounter as the key? But that is unique within only one reducer right? Will the number of reducers always be one for the RowDenormalizer plugin? If that is not the case, these values can be overridden, isn't it?

gokulavasan:
Took a pass. Please address comments.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/RowDenormalizerAggregator.java
rashig: Tested this with multiple mappers, currently it seems to be working fine and not overriding the keys. However, I agree that in case of multiple reducers, these keys could be overridden.

Can we keep the record itself as a key or do you have any other suggestion that we can use to maintain the unique key.

gokulavasan:
Currently, the BatchAggregator's groupBy emitter's emitError method doesn't work as expected. This needs to be supported by changing the cdap-data-pipeline app logic in a way that is supported and works for both MR and Spark. We should be using the framework support to write errors instead of having a logic in every plugin. So I would suggest to hold off this PR until the support is available. CCing @albertshau

rashig:
Okay. Putting this on hold for now.
Please update once the support in framework is available.

russorat:
Closing for now

PR #432, opened 2016-08-23
Adding dynamic partitioning sink

russorat:
The purpose of this plugin is to allow users to use values in each record for partitioning. For example, if a user has records with "name" and "original_create_date" and they specify they want to partition on original_create_date, the resulting File Dataset will have a partition for each value in original_create_date. The use case for this is a user loading historical data from a DB or set of files and would like to keep the data partitioned on the original transaction or load date instead of the run date of the load.

### DIFF
--- a/core-plugins/docs/DynamicPFSParquet-batchsink.md
+++ b/core-plugins/docs/DynamicPFSParquet-batchsink.md
@@ -1,0 +1,53 @@
+# Dynamic PartitionedFileSet Parquet Batch Sink
+
+
+Description
+-----------
+Sink for a ``PartitionedFileSet`` that writes data in Parquet format
+and leverages one or more record field values for creating partitions.
+All data for the run will be written to a partition based on the
+specified fields and their value.
+
+
+Use Case
+--------
+This sink is used whenever you want to write to a ``PartitionedFileSet`` in Parquet format
+using a value from the record as a partition. For example, you might want to load historical
+data from a database and partition the dataset on the original creation date of the data.
+
+
+Properties
+----------
+**name:** Name of the ``PartitionedFileSet`` to which records are written.
+If it doesn't exist, it will be created.
+
+**schema:** The Avro schema of the record being written to the sink as a JSON Object.
+
+**basePath:** Base path for the ``PartitionedFileSet``. Defaults to the name of the dataset.
+
+Example
+-------
+This example will write to a ``PartitionedFileSet`` named ``'users'`` using the original
+``'create_date'`` of the record as the partition:
+
+    {
+        "name": "DynamicPFSParquet",
+        "type": "batchsink",
+        "properties": {
+            "name": "users",
+            "fieldNames": "create_date",
+            "schema": "{
+                \"type\":\"record\",
+                \"name\":\"user\",
+                \"fields\":[
+                    {\"name\":\"id\",\"type\":\"long\"},
+                    {\"name\":\"name\",\"type\":\"string\"},
+                    {\"name\":\"create_date\",\"type\":\"string\"}
+                ]
+            }"
+        }
+    }
+
+It will write data in Parquet format using the given schema. Every time the pipeline runs, 
+partitions will be determined in the ``PartitionedFileSet`` based on the ``create_date``
+of each record.
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
@@ -1,0 +1,202 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.sink;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.batch.Output;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.data.schema.UnsupportedTypeException;
+import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.lib.DynamicPartitioner;
+import co.cask.cdap.api.dataset.lib.FileSetProperties;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.api.dataset.lib.PartitionKey;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSet;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSetArguments;
+import co.cask.cdap.api.dataset.lib.PartitionedFileSetProperties;
+import co.cask.cdap.api.dataset.lib.Partitioning;
+import co.cask.cdap.api.mapreduce.MapReduceTaskContext;
+import co.cask.cdap.etl.api.Emitter;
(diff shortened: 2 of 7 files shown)


core-plugins/docs/DynamicPFSParquet-batchsink.md
```diff
@@ -0,0 +1,53 @@
...
+-------
+This example will write to a ``PartitionedFileSet`` named ``'users'`` using the original
+``'create_date'`` of the record as the partition:
+
+    {
+        "name": "DynamicPFSParquet",
+        "type": "batchsink",
+        "properties": {
+            "name": "users",
+            "fieldNames": "create_date",
```
awholegunch, line 38: This isn't listed as a property in the docs. Is this what is used to create the partitions? 

Why is the name of the property plural? Does that mean it can be a list? 
If so, can you change the example to show that, as it might not be obvious how to do that.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
```diff
@@ -0,0 +1,202 @@
...
+@Description("Sink for a PartitionedFileSet that writes data in Parquet format and uses a dynamic partition key.")
+public class DynamicPartitionFileSetParquetSink extends
+  PartitionedFileSetSink<Void, GenericRecord> {
+  public static final String NAME = "DynamicPFSParquet";
+  public static final String PARTITION_COL_PREFIX = "p_";
+
+  private static final Logger LOG = LoggerFactory.getLogger(DynamicPartitionFileSetParquetSink.class);
+  private static final String SCHEMA_DESC = "The Parquet schema of the record being written to the Sink as a JSON " +
+    "Object.";
+  private static final String FIELD_DESC = "The fields to be used for the partitions as comma separated values.";
```
awholegunch, line 68: `The fields to be used for the partitions as comma-separated values.`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
```diff
@@ -0,0 +1,202 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link BatchSink} to write Parquet records to a {@link PartitionedFileSet}.
+ */
+@Plugin(type = "batchsink")
+@Name("DynamicPFSParquet")
```
gokulavasan, line 58: Curious, why can't this be added as a feature to the existing FileSet sink?

gokulavasan:
Is there a JIRA for this PR?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
russorat: I don't know if there is a reason. What was the reason we broke out TimePartition file sets from the FileSet sink?

gokulavasan: @albertshau Do you think this would be better to have it as a feature in existing TPFS instead of having DynmaicPFSParquet, DynamicPFSAvro, etc.

albertshau: i haven't taken a super close look, but it seems like this is for a normal partitioned fileset and not a time partitioned one right? I believe TPFS only supports partitioning on time so it wouldn't apply there.

anwar6953: `FileSet` doesn't support dynamic partitioning, because it doesn't have partitions.
Technically, you could still create a path dynamically within a FileSet and put various records in different paths of the FileSet, but that's what the abstraction of `PartitionedFileSet` is for.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
```diff
@@ -0,0 +1,202 @@
...
+    String schema = config.schema.toLowerCase();
+    // parse to make sure it's valid
+    new org.apache.avro.Schema.Parser().parse(schema);
+    String hiveSchema;
+    Schema parsedSchema;
+    try {
+      parsedSchema = Schema.parseJson(schema);
+      hiveSchema = HiveSchemaConverter.toHiveSchema(parsedSchema);
+    } catch (UnsupportedTypeException | IOException e) {
+      throw new RuntimeException("Error: Schema is not valid ", e);
```
anwar6953, line 100: I think you should throw `IllegalArgumentException` instead.
That might indicate to the user that its a bad request.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    Partitioning.Builder partitionBuilder = Partitioning.builder();
+    String[] partitionFields = config.fieldNames.split(",");
+    for (int i = 0; i < partitionFields.length; i++) {
+      // We need to make the name of the partitioned column unique to avoid 'Column repeated in partitioning columns'
```
anwar6953, line 106: the partitioned column name must be unique for a single table, right?

anwar6953, line 106: I don't see how adding a static/fixed string as a prefix to every column name makes them unique.

anwar6953, line 106: Oh; nevermind, I think I understand now. The issue is that the partitioningFields would have the same name as the columns in the schema?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/PartitionedFileSetSinkConfig.java
```diff
@@ -0,0 +1,47 @@
...
+               "name of the dataset.")
+  @Nullable
+  protected String basePath;
+
+  public PartitionedFileSetSinkConfig(String name, @Nullable String basePath) {
+    this.name = name;
+    this.basePath = basePath;
+  }
+
+  public void validate() {
```
anwar6953, line 44: why empty? It needs to be defined?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLDynamicPFSTestRun.java
```diff
@@ -0,0 +1,136 @@
...
+    for (GenericRecord record : toWrite) {
+      parquetWriter.write(record);
+    }
+    parquetWriter.close();
+    inputManager.flush();
+
+    // run the pipeline
+    WorkflowManager workflowManager = appManager.getWorkflowManager(ETLWorkflow.NAME);
+    // add a minute to the end time to make sure the newly added partition is included in the run.
+    workflowManager.start(ImmutableMap.of("logical.start.time", String.valueOf(timeInMillis + 60 * 1000)));
```
anwar6953, line 117: why do we need to set this?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/DynamicPartitionFileSetParquetSink.java
russorat: yep. ideally, when someone specifies a column as a partitioned column, they either rename their existing column indicating it shouldn't be used for querying, or we have a way of not including partitioned fields in the table. see the hive docs for an example: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-PartitionedTables

I'm open to other ways to solve this problem, but i didn't see any other way.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLDynamicPFSTestRun.java
russorat: this is set because the TPFS source in the test i believe.

anwar6953: I'm not sure I follow.
I don't think we're using the workflow's logical start time for anything, are we?

russorat:
Closing for now. will need to be re-designed.

PR #439, opened 2016-08-25
[HYDRATOR-898]: enable Macros for referenceName in HDFSSink

thegreenfrog:


hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSink.java
```diff
@@ -38,8 +38,10 @@ public ReferenceBatchSink(ReferencePluginConfig config) {
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
-    // Verify that reference name meets dataset id constraints
-    @SuppressWarnings("unused")
-    DatasetId datasetId = new DatasetId(NamespaceId.DEFAULT.getNamespace(), config.referenceName);
+    if (!config.containsMacro(Constants.Reference.REFERENCE_NAME)) {
+      // Verify that reference name meets dataset id constraints
+      @SuppressWarnings("unused")
+      DatasetId datasetId = new DatasetId(NamespaceId.DEFAULT.getNamespace(), config.referenceName);
```
knazir, line 44: If `referenceName` contains a macro, does this need to be (conditionally) done at runtime?

thegreenfrog: I'll add something to `prepareRun`

sagarkapare:
I think this PR should be opened against develop and not against release/1.4. Only critical fixes in release/1.4 and doc changes.

chtyim:
Making this PR on-hold. Classes in "cdap-proto" shouldn't be used. Checking with @gokulavasan why it gets introduced in https://github.com/caskdata/hydrator-plugins/pull/235

russorat:
we don't allow macros for other reference name options, so closing this for now.

PR #441, opened 2016-08-25
HYDRATOR-913 Remove cdap-proto dependency from Hydrator Plugins

gokulavasan:
Build : http://builds.cask.co/browse/HYP-BAD97-1
JIRA : https://issues.cask.co/browse/HYDRATOR-913

### DIFF
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.common;
+
+import java.util.regex.Pattern;
+
+/**
+ * Utility class for Id related operations.
+ */
+public final class IdUtils {
+
+  private IdUtils() {
+  }
+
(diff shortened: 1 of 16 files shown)


hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+package co.cask.hydrator.common;
+
+import java.util.regex.Pattern;
+
+/**
+ * Utility class for Id related operations.
+ */
+public final class IdUtils {
+
```
bdmogal, line 25: add a private constructor.

hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
```diff
@@ -0,0 +1,34 @@
...
+  private static final Pattern datasetIdPattern = Pattern.compile("[$\\.a-zA-Z0-9_-]+");
```
bdmogal, line 29: is there no way that this can be gotten from CDAP? This code duplication could do bad things in future?

bdmogal:
1 comment, 1 question, rest LGTM

hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
gokulavasan: Hmm I don't think so. It is better this way than pulling in the entire artifact. Any thoughts @chtyim ?

bdmogal: Then can this check itself be done in CDAP somewhere instead of in every plugin? (I'm not sure it can, just confirming)

hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
gokulavasan: done

chtyim:
LGTM

hydrator-common/src/main/java/co/cask/hydrator/common/IdUtils.java
gokulavasan: We want to catch this before the pipeline is created. This will be caught anyway in CDAP but this is just for better user experience.

gokulavasan:
Build passed. Merging.

PR #442, opened 2016-08-27
changes to comment in pom if app.parents property is not present in plugin pom

CuriousVini:
changes to comment in pom if app.parents property is not present in plugin pom.


gsps1:
LGTM

PR #446, opened 2016-08-31
[HYDRATOR-909] NGramTransform implementation

shashankmoghe:
NGram implementation file
Junit file
Widget file
documentation
spark version change to 1.6.1 in pom.xml

### DIFF
--- a/spark-plugins/docs/NGramTransform-sparkcompute.md
+++ b/spark-plugins/docs/NGramTransform-sparkcompute.md
@@ -1,0 +1,53 @@
+# NGramTransform Spark Compute
+
+Description
+-----------
+Transforms the input features into n-grams, where n-gram is a sequence of n tokens (typically words) for some integer 'n'.
+
+Use Case
+--------
+A bio data scientist wants to  study the sequence of the nucleotides using the input stream of DNA sequencing to identify the bonds.
+The input Stream contains the DNA sequence eg AGCTTCGA. The output contains the bigram sequence AG, GC, CT, TT, TC, CG, GA.
+
+Properties
+----------
+**fieldToBeTransformed:** Field to be used to transform input features into n-grams.
+
+**ngramSize:** NGram size. (Macro-enabled)
+
+**outputField:** Transformed field for sequence of n-gram.
+
+Example
+-------
+This example transforms features from "tokens" field into n-grams(output field name is ngrams) using ngram size as "2".
+
+    {
+        "name": "NGramTransform",
+        "type": "sparkcompute",
+        "properties": {
+            "fieldToBeTransformed": "tokens",
+            "ngramSize": "2",
+            "outputField": "ngrams"
+        }
+    }
+
+
+For example, suppose the NGramTransform receives below input records:
+
+    +===================================+
+    | topic | tokens                    |
+    +===================================+
+    | java  | [hi,i,heard,about,spark]  |
+    | hdfs  | [hdfs,is,a,file,system]   |
+    | spark | [spark,is,an,engine]      |
+    +===================================+
+
+Output schema will contain only a single field "ngrams" having transformed ngrams in string array form:
+
+    +==========================================+
+    | ngrams                                   |
+    +==========================================+
+    | [hi i,i heard,heard about,about spark]   |
+    | [hdfs is,is a,a file,file system]        |
+    | [spark is,is an,an engine]               |
+    +==========================================+
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/NGramTransform.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/NGramTransform.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.spark;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.plugin.PluginConfig;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.batch.SparkCompute;
+import co.cask.cdap.etl.api.batch.SparkExecutionPluginContext;
+import com.google.common.base.Preconditions;
+import org.apache.spark.api.java.JavaRDD;
+import org.apache.spark.api.java.JavaSparkContext;
+import org.apache.spark.api.java.function.Function;
+import org.apache.spark.ml.feature.NGram;
+import org.apache.spark.sql.DataFrame;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.RowFactory;
+import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+
+/**
+ * NGramTransform - SparkCompute to transform input features into n-grams.
+ */
+@Plugin(type = SparkCompute.PLUGIN_TYPE)
+@Name(NGramTransform.PLUGIN_NAME)
+@Description("Used to transform input features into n-grams.")
+public class NGramTransform extends SparkCompute<StructuredRecord, StructuredRecord> {
+  public static final String PLUGIN_NAME = "NGramTransform";
+  private Config config;
+  public Schema outputSchema;
+
+  public NGramTransform(Config config) {
+    this.config = config;
+  }
+
+  /**
+   * Configuration for the NGramTransform Plugin.
+   */
+  public static class Config extends PluginConfig {
+
+    @Description("Field to be used to transform input features into n-grams.")
+    private final String fieldToBeTransformed;
+
+    @Description("N-Gram size.")
+    @Macro
+    private final Integer ngramSize;
+
+    @Description("Transformed field for sequence of n-gram.")
+    private final String outputField;
+
+    public Config(String fieldToBeTransformed, Integer ngramSize, String outputField) {
+      this.fieldToBeTransformed = fieldToBeTransformed;
+      this.ngramSize = ngramSize;
+      this.outputField = outputField;
+    }
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    super.configurePipeline(pipelineConfigurer);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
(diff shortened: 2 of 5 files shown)


spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,54 @@
...
+    | spark | [spark,is,an,engine]      |
+    +===================================+
+
+Output schema will contain only a single field "ngrams" having transformed ngrams in string array form:
+
+    +==========================================+
+    | ngrams                                   |
+    +==========================================+
+    | [hi i,i heard,heard about,about spark]   |
+    | [hdfs is,is file,file system]            |
```
knazir: nit: [hdfs is, is a, a file, file system] instead

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/NGramTransform.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Configuration for the NGramTransform Plugin.
+     */
+    public static class Config extends PluginConfig {
+
+        @Description("Field to be used to transform input features into n-grams.")
+        private final String fieldToBeTransformed;
+
+        @Description("Number of terms in each n-gram.")
+        private final Integer noOfTerms;
```
knazir: Better to name `numTerms` or `numberOfTerms`?

knazir: Also we can enable this as a macro with `@Macro` and then add (Macro-enabled) at the end of the description in the markdown file (not in this description).

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
+# NGramTransform Spark Compute
+
+Description
+-----------
+Transforms the input features into n-grams, where n-grams is a sequence of n tokens (typically words) for some integer n.
```
sagarkapare: `where n-gram is a sequence of n tokens (typically words) for some integer 'n'`

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
...
+
+Use Case
+--------
+User wants to use tokens such as bigrams/ngrams instead of just unigrams while developing features for supervised Machine Learning models.
+
+Properties
+----------
+**fieldToBeTransformed:** Field to be used to transform input features into n-grams.
+
+**numberOfTerms:** Number of terms in each n-gram. (Macro-enabled)
```
sagarkapare: should we call this field as `ngramSize` or just `size`?

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
...
+
+**outputField:** Transformed field for sequence of n-gram.
+
+Example
+-------
+This example transforms features from "tokens" field into n-grams(output field name is ngrams) using no of terms as "2".
```
sagarkapare: `using size as "2"`?

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,246 @@
...
+    // add the artifact for etl batch app
+    setupBatchArtifacts(DATAPIPELINE_ARTIFACT_ID, DataPipelineApp.class);
+    // add artifact for spark plugins
+    Set<ArtifactRange> parents = ImmutableSet.of(
+      new ArtifactRange(NamespaceId.DEFAULT.toId(), DATAPIPELINE_ARTIFACT_ID.getArtifact(),
+                        new ArtifactVersion(DATAPIPELINE_ARTIFACT_ID.getVersion()), true,
+                        new ArtifactVersion(DATAPIPELINE_ARTIFACT_ID.getVersion()), true)
+    );
+    addPluginArtifact(NamespaceId.DEFAULT.artifact("spark-plugins", "1.0.0"), parents,
+                      NGramTransform.class, TwitterStreamingSource.class);
```
sagarkapare, line 107: why do we need `TwitterStreamingSource.class`?

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,246 @@
...
+public class NGramTransformTest extends HydratorTestBase {
+
+  @ClassRule
+  public static final TestConfiguration CONFIG = new TestConfiguration("explore.enabled", false);
+
+  protected static final ArtifactId DATAPIPELINE_ARTIFACT_ID =
+    NamespaceId.DEFAULT.artifact("data-pipeline", "3.5.0");
+  protected static final ArtifactSummary DATAPIPELINE_ARTIFACT = new ArtifactSummary("data-pipeline", "3.5.0");
+
+  private static final String MULTIPLE_FIELDS_DATASET = "MultipleFields";
```
sagarkapare: better to move this to the method where its used since its used in only one method.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,246 @@
...
+  private static final String SINGLE_FIELD_DATASET = "SingleField";
```
sagarkapare: same for this field

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,246 @@
...
+  private static final String APP_NAME = "NGramTransformTest";
+  private static final String OUTPUT_FIELD = "ngrams";
+  private static final String FIELD_TO_BE_TRANSFORMED = "tokens";
+  private static final String ADDITIONAL_FIELD = "name";
+  private static final String[] TOKENS_DATASET1 = {"hi", "i", "am", "cdap"};
```
sagarkapare: why this field is named `TOKENS_DATASET1`?

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,246 @@
...
+  private static final String[] TOKENS_DATASET2 = {"how", "are", "you", "cdap"};
```
sagarkapare: similar to this field why named as `TOKENS_DATASET2`? name does not tell much about its usage..

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
sagarkapare: in general no need to use class level member variables if the variable is only used in method scope.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
shashankmoghe: There were some changes in co.cask.cdap.test.TestBase API post which we had to add this entry in all spark junits.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
shashankmoghe: Kept this variable as it is since it is used in multiple tests.

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
+# NGramTransform Spark Compute
+
+Description
+-----------
+Transforms the input features into n-grams, where n-grams is a sequence of n tokens (typically words) for some integer 'n'.
```
sagarkapare: `where n-gram` and not `where n-grams`

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
...
+
+Use Case
+--------
+User wants to use tokens such as bigrams/ngrams instead of just unigrams while developing features for supervised Machine Learning models.
```
sagarkapare: Better to explain actual use case. How user will use this particular plugin in the hydrator pipeline?

spark-plugins/docs/NGramTransform-sparkcompute.md
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+
+For example, suppose the NGramTransform receives below input records:
+
+    +===================================+
+    | topic | tokens                    |
+    +===================================+
+    | java  | [hi,i,heard,about,spark]  |
+    | hdfs  | [hdfs,is,a,file,system]     |
```
sagarkapare: nit: formatting of the table is missed.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
sagarkapare: Not sure if I understand this. If we remove the `TwitterStreamingSource.class` entry, does it fail? if so whats the cause of failure?

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,253 @@
...
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForFinish(5, TimeUnit.MINUTES);
+    DataSetManager<Table> nGrams = getDataset(multiFieldData);
+    List<StructuredRecord> output = MockSink.readOutput(nGrams);
+    Set<List> results = new HashSet<>();
+    for (StructuredRecord structuredRecord : output) {
+      results.add((ArrayList) structuredRecord.get(OUTPUT_FIELD));
+    }
+    StructuredRecord row = output.get(0);
```
sagarkapare: delay defining the variable `row` where its actually used.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
```diff
@@ -0,0 +1,253 @@
...
+    Assert.assertEquals(2, output.size());
+    Assert.assertEquals(getExpectedData(), results);
+    Assert.assertTrue(getExpectedData().containsAll(results));
```
sagarkapare: I think `Assert.assertEquals(getExpectedData(), results);` will check for both the size and contain. so you dont actually need the `containsAll` assert.

sagarkapare (commented):
few more comments.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/NGramTransformTest.java
shashankmoghe: If we remove TwitterStreamingSource.class entry,we face the issue.
Below are the findings: 
1.If we remove the TwitterStreamingSource.class entry  it throws 
"co.cask.cdap.common.InvalidArtifactException: Class could not be found while inspecting artifact for plugins. Please check dependencies are available, and that the correct parent artifact was specified. Error class: class java.lang.NoClassDefFoundError, message: co/cask/hydrator/common/ReferencePluginConfig."

2.It checks for ReferencePluginConfig for streaming source plugins.
3.If we remove streaming source classes/java files and remove this class file entry, the spark tests work fine.
4.Since SparkPluginTest is used for testing Naive Bayes and the streaming source plugins the entry for the classes is mentioned and the error is not caught.
However, since we have written separate tests for spark plugins other than naive bayes, we have to add this entry to run the tests.
5.I think we can get more details on this from the folks who worked on streaming source plugins.

spark-plugins/docs/NGramTransform-sparkcompute.md
prinam: @shashankmoghe please could you update this with the use case? the code looks fine, it is just missing the use case.

sagarkapare:
Please update the use case. rest LGTM

shashankmoghe:
@abhinavbansal246 Squashed the commits.Please see if you can merge

PR #448, opened 2016-08-31
HYDRATOR-922 remove usage of BadRequestException

albertshau:
The class causes problems with some versions of spark that bundle
jersey classes in the spark-assembly jar. In those situations,
the javax.ws.rs classes visible to the program are the jersey
classes (which dont include BadRequestException) and not the
rs-api-2.0 classes (which do include BadRequestException). Since
that package seems to exist in multiple dependencies, just removing
usage of BadRequestException for simplicity.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
@@ -55,10 +55,8 @@ import java.sql.Driver;
 import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.sql.SQLSyntaxErrorException;
 import java.sql.Statement;
 import javax.annotation.Nullable;
-import javax.ws.rs.BadRequestException;
 import javax.ws.rs.Path;
 
 /**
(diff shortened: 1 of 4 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -116,13 +114,12 @@ private String getJDBCPluginType() {
    * @throws SQLException
    * @throws InstantiationException
    * @throws IllegalAccessException
-   * @throws BadRequestException If executing the query threw {@link SQLSyntaxErrorException}
    * we get the message and throw it as BadRequestException
```
awholegunch: Need to remove this line, as it a continuation of the previous.

awholegunch:
One comment, otherwise LGTM

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
albertshau: oops, good catch

albertshau:
http://builds.cask.co/browse/HYP-BAD106

PR #454, opened 2016-09-02
HYDRATOR-979 - WindowShareCopy Action Plugin

nitinmotgi:


core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+**netBiosHostname:** Specifies the NetBios hostname to import files from.
+
+**netBiosUsername:** Specifies the NetBios username to user for importing files from share.
+
+**netBiosPassword:** Specifies the NetBios password for importing files from windows share.
+
+**netBiosSharename:** Specifies the NetBios share name.
+
+**numThreads:** Specifies the number of parellel tasks to execute the copy operation. Defaults to 1.
```
anwar6953: parallel

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+    sb.append(config.netBiosSharename);
+    sb.append("/");
+    if (config.sourceDirectory.startsWith("/")) {
+      if (config.sourceDirectory.length() > 1) {
+        sb.append(config.sourceDirectory.substring(1));
+      }
+    } else {
+      sb.append(config.sourceDirectory);
+    }
+    final String smbDirectory = sb.toString();
```
anwar6953, line 88: why make this final?

anwar6953:
Is the following use case not supported?
Renaming the file while moving it to hdfs (i.e. the target specified being a file name, instead of the directory name).

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+    final Path hdfsDir = new Path(config.destinationDirectory);
+    final FileSystem hdfs = hdfsDir.getFileSystem(new Configuration());
+    if (!hdfs.exists(hdfsDir)) {
+      hdfs.mkdirs(hdfsDir);
+    }
+
+    // Determines the buffer size to be used during writing.
+    Configuration conf = new Configuration();
+    conf.setLong("io.file.buffer.size", config.bufferSize);
+    String[] list = dir.list();
```
anwar6953: rename `list` to `files` or something more meaningful

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+            try {
+              if (smbDirectory.endsWith("/")) {
+                fileName = copyFileToHDFS(hdfs, smbDirectory + file, hdfsDir, auth);
+              } else {
+                fileName = copyFileToHDFS(hdfs, smbDirectory + "/" + file, hdfsDir, auth);
+              }
+            } catch (IOException e) {
+              LOG.warn("Exception while copying the file {}", file, e);
+            }
+            return fileName;
```
anwar6953: What happens with the return value?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+      }
+
+      for (int i = 0; i < list.length; i++) {
+        try {
+          Future<String> fileWritten = completionService.take();
+          String fileName = fileWritten.get();
+          LOG.info("{} is copied", fileName);
+        } catch (Throwable t) {
+          Throwable rootCause = Throwables.getRootCause(t);
+          if (rootCause instanceof ExecutionException) {
```
anwar6953: I don't find much value in catching the exception, and then logging+rethrowing the root cause.
In fact logging and rethrowing the exception result in redundant error messages (since the thrown exception will be caught and logged at a higher level).

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+      executorService.shutdownNow();
+      executorTerminateLatch.await();
+    }
+  }
+
+  private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest, NtlmPasswordAuthentication auth)
+    throws IOException {
+    SmbFile smbFile = new SmbFile(smbSourceFile, auth);
+    String name = smbFile.getName();
+    Path nDest = new Path(dest, name);
```
anwar6953: what is the 'n' in 'nDest' for?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+    // If file already exists, then we skip over.
+    if (hdfs.exists(nDest)) {
+      LOG.info("File {} already exists on HDFS, Skipping", nDest.getName());
+      return name;
+    }
+    LOG.info("Copying file {} to {}", smbSourceFile, nDest.toString());
+    try (InputStream in = smbFile.getInputStream();
+         BufferedOutputStream out = new BufferedOutputStream(hdfs.create(nDest))) {
+      ByteStreams.copy(in, out);
+    } catch (IOException e) {
```
anwar6953, line 172: This doesn't seem like `INFO`.
Also, why isn't this exception thrown (instead of just caught and logged)? What's the behavior for other files that we want?

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+**sourceDirectory:** Specifies the NetBios directory.
+
+**destinationDirectory:** The valid, full HDFS destination path in the same cluster where the file or files are to be
+moved. If a directory is specified with a file sourcePath, the file will be put into that directory. If sourcePath is a
+directory, it is assumed that destPath is also a directory. HDFSAction will not catch this inconsistency.
+
+**bufferSize:** The size of the buffer to be used for copying the files. Minimum Buffer size is 4096.
```
anwar6953: what about the default?

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+Use Case
+--------
+This action can be used when a file or files need to be moved from windows share to HDFS directory.
+
+
+Properties
+----------
+**netBiosDomainName:** Specifies the NetBios domain name.
```
anwar6953: whats the default?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,276 @@
...
+    };
+  }
+
+  /**
+   * Config class that contains all properties necessary to execute an HDFS move command.
+   */
+  public class WindowsShareCopyConfig extends PluginConfig {
+
+    @Description("Specifies the NetBios domain name.")
+    @Nullable
```
anwar6953, line 218: Its really nullable?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: This is being accessed from inner class, has to be final.

anwar6953: oh ok; I didn't notice that.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: changed to debug, Since this will be an exception for individual file and multiple files are being copied simultaneously, we dont want this to prevent the copying of all other files.

anwar6953: I think it should be a `WARN` or `ERROR`, since it at least shouldn't be hidden from users if files failed to be copied.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: yes

anwar6953:
lgtm

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+
+Use Case
+--------
+This action can be used when a file or files need to be moved from windows share to HDFS directory.
+
+
+Properties
+----------
+**netBiosDomainName:** Specifies the NetBios domain name. Default is null.
```
rohitsinha54: Why default is null ? Shouldn't it be a required field ?

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+**netBiosHostname:** Specifies the NetBios hostname to import files from.
+
+**netBiosUsername:** Specifies the NetBios username to user for importing files from share.
+
+**netBiosPassword:** Specifies the NetBios password for importing files from windows share.
+
+**netBiosSharename:** Specifies the NetBios share name.
+
+**numThreads:** Specifies the number of parallel tasks to execute the copy operation. Defaults to 1.
```
rohitsinha54, line 26: Is there any upper limit on this ?

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+**sourceDirectory:** Specifies the NetBios directory.
```
rohitsinha54: Give an example of how this should look like and all above one too.

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+
+**destinationDirectory:** The valid, full HDFS destination path in the same cluster where the file or files are to be
+moved. If a directory is specified with a file sourcePath, the file will be put into that directory. If sourcePath is a
+directory, it is assumed that destPath is also a directory. HDFSAction will not catch this inconsistency.
+
+**bufferSize:** The size of the buffer to be used for copying the files. Minimum Buffer size is 4096. Default is 4096.
+
+Example
+-------
+This example copies all the files from `10.150.10.10:/source/share/path` to `/dest/path`:
```
rohitsinha54, line 40: Is this a valid ip address to any of our machine or clusters ?

core-plugins/docs/WindowsShareCopy-action.md
```diff
@@ -0,0 +1,62 @@
...
+                "scope": "SYSTEM"
+            },
+            "properties": {
+                "netBiosDomainName": "domain",
+                "netBiosHostname": "10.150.10.10",
+                "netBiosUsername": "username",
+                "netBiosPassword": "password",
+                "netBiosSharename": "share",
+                "sourceDirectory": "source",
+                "destinationDirectory": "hdfs://example.com/dest/path",
```
rohitsinha54, line 59: is port needed here ?

core-plugins/pom.xml
```diff
@@ -179,6 +179,11 @@
       <version>0.6.3</version>
     </dependency>
     <!-- End for SQS -->
+    <dependency>
```
rohitsinha54, line 183: Put a comment as why this dependency is needed.

core-plugins/docs/WindowsShareCopy-action.md
abhinavbansal246: Domains can be null in netbios, it opens a null session.

core-plugins/docs/WindowsShareCopy-action.md
abhinavbansal246: No.

core-plugins/docs/WindowsShareCopy-action.md
abhinavbansal246: no, just made up a dummy ip address

core-plugins/docs/WindowsShareCopy-action.md
abhinavbansal246: Port is not needed, tested on dev reactor

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+
+      this.netBiosDomainName = netBiosDomainName;
+      this.netBiosHostname = netBiosHostname;
+      this.netBiosUsername = netBiosUsername;
+      this.netBiosPassword = netBiosPassword;
+      this.netBiosSharename = netBiosSharename;
+      this.sourceDirectory = sourceDirectory;
+      this.destinationDirectory = destinationDirectory;
+      this.bufferSize = (bufferSize < MIN_BUFFER_SIZE) || (bufferSize == null) ? MIN_BUFFER_SIZE : bufferSize;
+      this.numThreads = numThreads;
```
rohitsinha54, line 275: what if someone put num threads to 0 ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+                           String destinationDirectory, Long bufferSize, Integer numThreads) {
+
+      this.netBiosDomainName = netBiosDomainName;
+      this.netBiosHostname = netBiosHostname;
+      this.netBiosUsername = netBiosUsername;
+      this.netBiosPassword = netBiosPassword;
+      this.netBiosSharename = netBiosSharename;
+      this.sourceDirectory = sourceDirectory;
+      this.destinationDirectory = destinationDirectory;
+      this.bufferSize = (bufferSize < MIN_BUFFER_SIZE) || (bufferSize == null) ? MIN_BUFFER_SIZE : bufferSize;
```
rohitsinha54: What if someone puts buf size lower than the min ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
rohitsinha54: How ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: than it assigns the minimum buffer size

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: Windows allow you to create null sessions.

core-plugins/docs/WindowsShareCopy-action.md
rohitsinha54: There will be an upper limit from the OS where this will run right ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+        sb.append(config.sourceDirectory.substring(1));
+      }
+    } else {
+      sb.append(config.sourceDirectory);
+    }
+    final String smbDirectory = sb.toString();
+
+    // Register the SMB File handler.
+    jcifs.Config.registerSmbURLHandler();
+    jcifs.Config.setProperty("jcifs.util.loglevel", "4");
```
rohitsinha54, line 93: put a comment as what is this.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+              }
+            } catch (Exception e) {
+              LOG.warn("Exception while copying the file {}", file, e);
+              return null;
+            }
+          }
+        });
+      }
+
+      for (int i = 0; i < files.length; i++) {
```
rohitsinha54: replace for with foreach

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+
+  private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest, NtlmPasswordAuthentication auth)
+    throws IOException {
+    SmbFile smbFile = new SmbFile(smbSourceFile, auth);
+    String name = smbFile.getName();
+    Path destFile = new Path(dest, name);
+    LOG.debug("Thread {} is accessing source file {}, dest file {}", Thread.currentThread().getName(),
+              smbFile.getName(), destFile.getName());
+    // If file already exists, then we skip over.
+    if (hdfs.exists(destFile)) {
```
rohitsinha54: Shouldn't you allow user to overwrite ? Take a configuration to allow overwrite or not. I think it will be a common use case where you want to run copy again and again with change in data at windows share. With this the use has to go to hdfs and delete and then run the pipeline.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+
+      for (int i = 0; i < files.length; i++) {
+        try {
+          Future<String> fileWritten = completionService.take();
+          String fileName = fileWritten.get();
+          if (fileName != null) {
+            LOG.info("{} is copied", fileName);
+          }
+        } catch (Throwable t) {
+          Throwables.propagateIfPossible(t, Exception.class);
```
rohitsinha54: What is the use of this two separate propagation

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+    throws IOException {
+    SmbFile smbFile = new SmbFile(smbSourceFile, auth);
+    String name = smbFile.getName();
+    Path destFile = new Path(dest, name);
+    LOG.debug("Thread {} is accessing source file {}, dest file {}", Thread.currentThread().getName(),
+              smbFile.getName(), destFile.getName());
+    // If file already exists, then we skip over.
+    if (hdfs.exists(destFile)) {
+      LOG.info("File {} already exists on HDFS, Skipping", destFile.getName());
+      return name;
```
rohitsinha54: You return the name of the file which is being skipped here after logging thats its skipped and then in line 135 you go through all the filenames and print its copied.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+            } catch (Exception e) {
+              LOG.warn("Exception while copying the file {}", file, e);
+              return null;
+            }
+          }
+        });
+      }
+
+      for (int i = 0; i < files.length; i++) {
+        try {
```
rohitsinha54, line 139: Also the value of i is not used anywhere so don't see the need for loop. I guess you are trying to make sure that files.length job completed in which case keeping a running count till it reached files.length will be more suitable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+    // Create the HDFS directory if it doesn't exist.
+    final Path hdfsDir = new Path(config.destinationDirectory);
+    final FileSystem hdfs = hdfsDir.getFileSystem(new Configuration());
+    if (!hdfs.exists(hdfsDir)) {
+      hdfs.mkdirs(hdfsDir);
+    }
+
+    // Determines the buffer size to be used during writing.
+    Configuration conf = new Configuration();
+    conf.setLong("io.file.buffer.size", config.bufferSize);
```
rohitsinha54: What is the use of this and how does it affect the copying ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+      executorTerminateLatch.await();
+    }
+  }
+
+  private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest, NtlmPasswordAuthentication auth)
+    throws IOException {
+    SmbFile smbFile = new SmbFile(smbSourceFile, auth);
+    String name = smbFile.getName();
+    Path destFile = new Path(dest, name);
+    LOG.debug("Thread {} is accessing source file {}, dest file {}", Thread.currentThread().getName(),
```
rohitsinha54: accessing -> copying maybe ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,267 @@
...
+    if (hdfs.exists(destFile)) {
+      LOG.info("File {} already exists on HDFS, Skipping", destFile.getName());
+      return name;
+    }
+    LOG.info("Copying file {} to {}", smbSourceFile, destFile.toString());
+    try (InputStream in = smbFile.getInputStream();
+         BufferedOutputStream out = new BufferedOutputStream(hdfs.create(destFile))) {
+      ByteStreams.copy(in, out);
+    } catch (IOException e) {
+      LOG.warn("Exception in getting input stream for file {}", name, e);
```
rohitsinha54: Why this message ? Inside try block there are multiple things which can throw IOException
1. smbFile.getInputStream();
2. hdfs.create
3. ByteStreams.copy

rohitsinha54 (changes requested):

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,275 @@
...
+
+    // Authentication with NTLM and read the directory from the Windows Share.
+    final NtlmPasswordAuthentication auth = new NtlmPasswordAuthentication(config.netBiosDomainName,
+                                                                           config.netBiosUsername,
+                                                                           config.netBiosPassword);
+    SmbFile dir = new SmbFile(smbDirectory, auth);
+
+    // Determines the buffer size to be used during writing.
+    Configuration conf = new Configuration();
+    conf.setLong("io.file.buffer.size", config.bufferSize);
```
rohitsinha54, line 103: I don't really understand why is this done and how its being used to control the buffer size. Can you explain ?

abhinavbansal246: https://hadoop.apache.org/docs/r0.23.11/hadoop-project-dist/hadoop-common/core-default.xml
The size of buffer for use in sequence files. The size of this buffer should probably be a multiple of hardware page size (4096 on Intel x86), and it determines how much data is buffered during read and write operations.

rohitsinha54: What I meant was why is this needed in this plugin and how is it being used here. Earlier you just had io.file.buffer.size then you added file.stream-buffer-size why is this new one also needed in this plugin specifically and how it's used here ? If this plugin is needed then the PR can be merged but you should definitely revisit this and put proper comment as what this is doing. Open a jira to track it.

rohitsinha54:
LGTM if it has been tested with the new changes to overwrite files. Have one clarification which can be addressed later if needed.

abhinavbansal246:
https://issues.cask.co/browse/HYDRATOR-998

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -267,8 +267,8 @@ protected void terminated() {
       this.netBiosSharename = netBiosSharename;
       this.sourceDirectory = sourceDirectory;
       this.destinationDirectory = destinationDirectory;
-      this.bufferSize = (bufferSize < MIN_BUFFER_SIZE) || (bufferSize == null) ? MIN_BUFFER_SIZE : bufferSize;
-      this.numThreads = (numThreads < MIN_NUM_THREADS) || (numThreads == null) ? MIN_NUM_THREADS : numThreads;
+      this.bufferSize = (bufferSize == null) && (bufferSize < MIN_BUFFER_SIZE) ? MIN_BUFFER_SIZE : bufferSize;
```
rohitsinha54: What does this means ? 
The first part of  && is checking buffersize == null and since its && if its null the right hand side will be evaluated so you are camping null < MIN_BUFFER_SIZE

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -267,8 +267,8 @@ protected void terminated() {
...
+      this.numThreads = (numThreads == null) && (numThreads < MIN_NUM_THREADS) ? MIN_NUM_THREADS : numThreads;
```
rohitsinha54: Same as above

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
rohitsinha54: Ignore this I see that you changed this in another commit.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,289 @@
...
+        return MIN_NUM_THREADS;
+      }
+      return (numThreads < MIN_NUM_THREADS) ? MIN_NUM_THREADS : numThreads;
+    }
+
+    public Integer getBufferSize() {
+      if (bufferSize == null) {
+        return MIN_BUFFER_SIZE;
+      }
+      return (bufferSize < MIN_BUFFER_SIZE) ? MIN_BUFFER_SIZE : bufferSize;
```
rohitsinha54: Why not just 
(bufferSize == null || bufferSize < MIN_BUFFER_SIZE) ? MIN_BUFFER_SIZE : bufferSize

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -0,0 +1,289 @@
...
+      this.bufferSize = bufferSize;
+      this.numThreads = numThreads;
+      this.overwrite = !("false".equals(overwrite));
+    }
+
+    public Integer getNumThreads() {
+      if (numThreads == null) {
+        return MIN_NUM_THREADS;
+      }
+      return (numThreads < MIN_NUM_THREADS) ? MIN_NUM_THREADS : numThreads;
```
rohitsinha54: ditto

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -267,9 +267,23 @@ protected void terminated() {
...
       this.sourceDirectory = sourceDirectory;
       this.destinationDirectory = destinationDirectory;
-      this.bufferSize = (bufferSize == null) && (bufferSize < MIN_BUFFER_SIZE) ? MIN_BUFFER_SIZE : bufferSize;
-      this.numThreads = (numThreads == null) && (numThreads < MIN_NUM_THREADS) ? MIN_NUM_THREADS : numThreads;
+      this.bufferSize = bufferSize;
+      this.numThreads = numThreads;
       this.overwrite = !("false".equals(overwrite));
     }
+
+    public Integer getNumThreads() {
```
rohitsinha54: Also why do all this check every time when you need numthreads and buffersize. The value is not going to change between calls. Just do check before initialization and initialize with right value and use that.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -234,7 +237,8 @@ protected void terminated() {
 
     @Description("Specifies the number of threads to execute the copy operation")
     @Nullable
-    private final Integer numThreads;
+    @Macro
+    private Integer numThreads;
```
rohitsinha54, line 241: Why this is made a macro ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -253,9 +257,9 @@ protected void terminated() {
     @Description("The i/o buffer size used in copying bufferstream and to write to the hdfs file. The values should" +
       " be in the multiples of 4096. Default value is 4096")
     @Nullable
-    private final Integer bufferSize;
+    @Macro
```
rohitsinha54, line 260: Why ?

rohitsinha54 (commented):
2 comments on why numThreads, bufferSize are macros other than that LGTM if its tested on a secure cluster and standalone.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/WindowsShareCopy.java
abhinavbansal246: The user should not have to publish a new pipeline every time a different buffer size or thread count is needed. If these fields are macro that allows the user to take these configurations as runtime arguments

abhinavbansal246:
http://builds.cask.co/browse/HYP-BAD152-1

PR #443, opened 2016-08-30
[HYDRATOR-374] Move Mock APIs from Hydrator Plugins to CDAP

knazir:
Issue: https://issues.cask.co/browse/HYDRATOR-374
PR in CDAP: https://github.com/caskdata/cdap/pull/6610

Whenever there is an API change in CDAP, Hydrator Plugins will face compilation errors. This slows down development and leads to bad times in general. Moving the mock APIs used for Hydrator testing to the Hydrator-Test package in CDAP should resolve this issue as API changes leading to compilation errors will be picked up immediately when building CDAP.

PR #456, opened 2016-09-08
Moved CoreValidator Back to Original Package

knazir:
Moves `CoreValidator` back to package before Mock API removal caused by https://github.com/caskdata/hydrator-plugins/pull/443


knazir:
Some weird thing happened with reordering imports alphabetically then squashing 2 commits so it shows for a few files that I only added an import even though it was both a removal of the old and addition of the new...

anwar6953:
that's strange...

anwar6953:
Actually, it was not a removal and addition of import in those files; only an addition.
Previously, the import wasn't needed because it was in the same package.

knazir:
... I could've sworn I deleted an import from all of those files.. :| Clearly I'm just going insane then.

anwar6953:
check develop branch (prior to this PR). It probably doesn't have the imports in those classes.

albertshau:
lgtm

knazir:
Guild is breen: http://builds.cask.co/browse/HYP-BAD117-2 going to merge.

PR #459, opened 2016-09-12
Utility class for Spark ML plugins

ananya-bhattacharya:
This class contains the common logic for config validations and other methods like clone record, required for trainer and predictor for spark ml plugins.

### DIFF
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
@@ -1,0 +1,199 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.spark;
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import joptsimple.internal.Strings;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Spark plugin Utility class. Contains common code to be used in trainer and predictors.
+ */
+final class SparkUtils {
+
+  private SparkUtils() {
+  }
+
+  /**
+   * Validate the config parameters for the spark sink and spark compute classes.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the prediction values.
+   */
+  static void validateConfigParameters(Schema inputSchema, @Nullable String featuresToInclude,
+                                       @Nullable String featuresToExclude, String predictionField,
+                                       @Nullable String cardinalityMapping) {
+    if (!Strings.isNullOrEmpty(featuresToExclude) && !Strings.isNullOrEmpty(featuresToInclude)) {
+      throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
+                                           "Please specify fields for one.");
+    }
+    Map<String, Integer> fields = getFeatureList(inputSchema, featuresToInclude, featuresToExclude, predictionField);
+    for (String field : fields.keySet()) {
+      Schema.Field inputField = inputSchema.getField(field);
+      Schema schema = inputField.getSchema();
+      Schema.Type features = schema.isNullableSimple() ? schema.getNonNullable().getType() : schema.getType();
+      if (!(features.equals(Schema.Type.INT) || features.equals(Schema.Type.LONG) ||
+        features.equals(Schema.Type.FLOAT) || features.equals(Schema.Type.DOUBLE))) {
+        throw new IllegalArgumentException(String.format("Features must be of type : int, double, float, long but " +
+                                                           "was of type %s for field %s.", features, field));
+      }
+    }
+    getCategoricalFeatureInfo(inputSchema, featuresToInclude, featuresToExclude, predictionField, cardinalityMapping);
+  }
+
+  /**
+   * Get the feature list of the features that have to be used for training/prediction depending on the
+   * featuresToInclude or featuresToInclude list.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the prediction values.
+   * @return feature list to be used for training/prediction.
+   */
+  static Map<String, Integer> getFeatureList(Schema inputSchema, @Nullable String featuresToInclude,
+                                             @Nullable String featuresToExclude, String predictionField) {
+    if (!Strings.isNullOrEmpty(featuresToExclude) && !Strings.isNullOrEmpty(featuresToInclude)) {
+      throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
+                                           "Please specify fields for one.");
+    }
+    Map<String, Integer> fields = new HashMap<>();
+
+    if (!Strings.isNullOrEmpty(featuresToInclude)) {
+      Iterable<String> tokens = Splitter.on(',').trimResults().split(featuresToInclude);
+      String[] features = Iterables.toArray(tokens, String.class);
+      for (int i = 0; i < features.length; i++) {
+        String field = features[i];
+        Schema.Field inputField = inputSchema.getField(field);
+        if (!field.equals(predictionField) && inputField != null) {
+          fields.put(field, i);
+        }
+      }
+      return fields;
+    }
+
+    Set<String> excludeFeatures = new HashSet<>();
+    if (!Strings.isNullOrEmpty(featuresToExclude)) {
+      excludeFeatures.addAll(Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToExclude)));
+    }
+    Object[] inputSchemaFields = inputSchema.getFields().toArray();
+    for (int i = 0; i < inputSchemaFields.length; i++) {
+      String field = ((Schema.Field) inputSchemaFields[i]).getName();
+      if (!field.equals(predictionField) && !excludeFeatures.contains(field)) {
+        fields.put(field, i);
+      }
+    }
+    return fields;
+  }
+
+  /**
+   * Get the feature to cardinality mapping provided by the user.
+   *
+   * @param inputSchema        schema of the received record.
+   * @param featuresToInclude  features to be used for training/prediction.
+   * @param featuresToExclude  features to be excluded when training/predicting.
+   * @param labelField         field containing the prediction values.
+   * @param cardinalityMapping feature to cardinality mapping specified for categorical features.
+   * @return categoricalFeatureInfo for categorical features.
+   */
+  static Map<Integer, Integer> getCategoricalFeatureInfo(Schema inputSchema,
+                                                         @Nullable String featuresToInclude,
+                                                         @Nullable String featuresToExclude,
+                                                         String labelField,
+                                                         @Nullable String cardinalityMapping) {
+    Map<String, Integer> featureList = getFeatureList(inputSchema, featuresToInclude, featuresToExclude, labelField);
+    Map<Integer, Integer> outputFieldMappings = new HashMap<>();
+
+    if (Strings.isNullOrEmpty(cardinalityMapping)) {
+      return outputFieldMappings;
+    }
+    try {
+      Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(cardinalityMapping);
+      for (Map.Entry<String, String> field : map.entrySet()) {
+        String value = field.getValue();
+        try {
+          outputFieldMappings.put(featureList.get(field.getKey()), Integer.parseInt(value));
+        } catch (NumberFormatException e) {
+          throw new IllegalArgumentException(
+            String.format("Invalid cardinality %s. Please specify valid integer for cardinality.", value));
+        }
+      }
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(
+        String.format("Invalid categorical feature mapping. %s. Please make sure it is in the format " +
+                        "'feature':'cardinality'.", e.getMessage()), e);
+    }
+    return outputFieldMappings;
+  }
+
+  /**
+   * Validate label field for trainer.
+   *
+   * @param inputSchema schema of the received record.
+   * @param labelField  field from which to get the prediction.
+   */
+  static void validateLabelFieldForTrainer(Schema inputSchema, String labelField) {
+    Schema.Field prediction = inputSchema.getField(labelField);
+    if (prediction == null) {
+      throw new IllegalArgumentException(String.format("Label field %s does not exists in the input schema.",
+                                                       labelField));
+    }
+    Schema predictionSchema = prediction.getSchema();
+    Schema.Type predictionFieldType = predictionSchema.isNullableSimple() ?
+      predictionSchema.getNonNullable().getType() : predictionSchema.getType();
+    if (predictionFieldType != Schema.Type.DOUBLE) {
+      throw new IllegalArgumentException(String.format("Label field must be of type Double, but was %s.",
+                                                       predictionFieldType));
+    }
+  }
+
+  /**
+   * Creates a builder based off the given record. The record will be cloned without the prediction field.
+   */
+  static StructuredRecord.Builder cloneRecord(StructuredRecord record, Schema outputSchema,
+                                              String predictionField) {
+    List<Schema.Field> fields = new ArrayList<>(outputSchema.getFields());
+    fields.add(Schema.Field.of(predictionField, Schema.of(Schema.Type.DOUBLE)));
+    outputSchema = Schema.recordOf("records", fields);
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    for (Schema.Field field : outputSchema.getFields()) {
+      if (!predictionField.equals(field.getName())) {
+        builder.set(field.getName(), record.get(field.getName()));
+      }
+    }
+    return builder;
+  }
+
+  static Schema getOutputSchema(Schema inputSchema, String predictionField) {
+    List<Schema.Field> fields = new ArrayList<>(inputSchema.getFields());
+    fields.add(Schema.Field.of(predictionField, Schema.of(Schema.Type.DOUBLE)));
+    return Schema.recordOf(inputSchema.getRecordName() + ".predicted", fields);
+  }
+}
--- a/core-plugins/widgets/Email-postaction.json
+++ b/core-plugins/widgets/Email-postaction.json
@@ -49,7 +49,7 @@
           "name": "username"
         },
         {
-          "widget-type": "textbox",
+          "widget-type": "password",
           "label": "Password",
           "name": "password"
         },
--- a/pom.xml
+++ b/pom.xml
@@ -748,7 +748,7 @@
         <executions>
           <!-- Create the config file for artifact which can be used to deploy the artifact.
                Sets the parents field to system:cdap-etl-batch and system:cdap-etl-realtime with whatever
-               version range is set in the etl.versionRange property.
+               version range is set in the app.parents property.
                also sets a widget and doc property for each file contained in the widgets and docs directories. -->
           <execution>
             <id>create-artifact-config</id>


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import javax.ws.rs.Path;
+
+/**
+ * Spark plugin Utility class. Contains common code to be used in trainer and predictors.
+ */
+public class SparkUtils {
```
chtyim: Util class should be final and have a private constructor.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+   * Validate the config parameters for the spark sink and spark compute classes.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param labelField        field containing the prediction values.
+   */
+  public static void validateConfigParameters(Schema inputSchema, String featuresToInclude,
+                                              String featuresToExclude, String labelField) {
+    if (!Strings.isNullOrEmpty(featuresToExclude) && !Strings.isNullOrEmpty(featuresToInclude)) {
```
chtyim, line 53: What happen if both `featuresToExclude` and `featuresToInclude` are null? Is that allowed?

chtyim, line 53: Also, you should have `@Nullable` for the parameters that are allowed to be null.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  public static Schema getOutputSchema(Schema inputSchema, String predictionField) {
+    List<Schema.Field> fields = new ArrayList<>(inputSchema.getFields());
+    fields.add(Schema.Field.of(predictionField, Schema.of(Schema.Type.DOUBLE)));
+    return Schema.recordOf(inputSchema.getRecordName() + ".predicted", fields);
+  }
+
+  @Path("outputSchema")
+  private static Schema getOutputSchema(SparkUtils.GetSchemaRequest request) {
```
chtyim: I don't see this method is used any where.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    return getOutputSchema(request.inputSchema, request.predictionField);
+  }
+
+  /**
+   * Endpoint request for output schema.
+   */
+  private static final class GetSchemaRequest {
```
chtyim: Unused class (since `getOutputSchema(GetSchemaRequest)` is an unused method).

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  /**
+   * Creates a builder based off the given record.
+   */
+  public static StructuredRecord.Builder cloneRecord(StructuredRecord record, Schema outputSchema,
+                                                     String predictionField) {
+    List<Schema.Field> fields = new ArrayList<>();
+    fields.addAll(outputSchema.getFields());
```
chtyim: You can combine two lines into one.

``` java
List<Schema.Field> fields = new ArrayList<>(outputSchema.getFields());
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    Schema.Type predictionFieldType = predictionSchema.isNullableSimple() ?
+      predictionSchema.getNonNullable().getType() : predictionSchema.getType();
+    if (predictionFieldType != Schema.Type.DOUBLE) {
+      throw new IllegalArgumentException(String.format("Prediction field must be of type Double, but was %s.",
+                                                       predictionFieldType));
+    }
+  }
+
+  /**
+   * Creates a builder based off the given record.
```
chtyim: should state it clearly that the record is cloned without the prediction field.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * Validate label field for trainer.
+   *
+   * @param inputSchema schema of the received record.
+   * @param labelField  field from which to get the prediction.
+   */
+  public static void validateLabelFieldForTrainer(Schema inputSchema, String labelField) {
+    Schema.Field prediction = inputSchema.getField(labelField);
+    if (labelField == null) {
```
chtyim: Wrong condition. I think you like to check `if (predication == null)`. Also, better call the variable `predictionField`.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    return outputFieldMappings;
+  }
+
+  /**
+   * Validate prediction field value for Predictor.
+   *
+   * @param inputSchema     schema of the received record.
+   * @param predictionField field on which to set the prediction.
+   */
+  public static void validatePredictionFieldForRegressor(Schema inputSchema, String predictionField) {
```
chtyim: Not a big fan of this kind of one line helper, as it actually reduces readability of the caller code. Get rid of it please.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+        if (!fieldName.equals(labelField) && (excludeFeatures.isEmpty() || !excludeFeatures.contains(fieldName))) {
+          fields.add(fieldName);
+        }
+      }
+    }
+    return fields;
+  }
+
+  /**
+   * Get the getCategoricalFeatureInfo using the feature to cardinality mapping provided by the user.
```
chtyim: Do you mean "Get the feature to cardinality mapping provided by the user"?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+   * Get the feature list of the features that have to be used for training/prediction depending on the
+   * featuresToInclude or featuresToInclude list.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param labelField        field containing the prediction values.
+   * @return feature list to be used for training/prediction.
+   */
+  public static List<String> getFeatureList(Schema inputSchema, String featuresToInclude, String featuresToExclude,
```
chtyim: Why you don't need to check the include and exclude features as the one did above?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+                                            String labelField) {
+    List<String> fields = new LinkedList<>();
```
chtyim: Why use LinkedList vs ArrayList?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    List<String> excludeFeatures = new ArrayList<>();
+
+    if (!Strings.isNullOrEmpty(featuresToInclude)) {
+      fields = Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToInclude));
+      for (String field : fields) {
+        Schema.Field inputField = inputSchema.getField(field);
+        if (inputField == null) {
+          throw new IllegalArgumentException(String.format("Field %s does not exists in the input schema", field));
+        }
+      }
```
chtyim, line 98: Just return. This make the code less clutter.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    } else {
```
chtyim: If you return from the above if block, then no need to have else.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param labelField        field containing the prediction values.
+   * @return feature list to be used for training/prediction.
+   */
+  public static List<String> getFeatureList(Schema inputSchema, String featuresToInclude, String featuresToExclude,
+                                            String labelField) {
+    List<String> fields = new LinkedList<>();
+    List<Schema.Field> inputSchemaFields = inputSchema.getFields();
+    List<String> excludeFeatures = new ArrayList<>();
```
chtyim: Should be a Set, since you are using it to do lookup. Also, move it close to the place where it is being used (e.g. after the features to include logic).

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+          throw new IllegalArgumentException(String.format("Field %s does not exists in the input schema", field));
+        }
+      }
+    } else {
+      if (!Strings.isNullOrEmpty(featuresToExclude)) {
+        excludeFeatures.addAll(Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToExclude)));
+      }
+      for (Schema.Field field : inputSchemaFields) {
+        String fieldName = field.getName();
+        if (!fieldName.equals(labelField) && (excludeFeatures.isEmpty() || !excludeFeatures.contains(fieldName))) {
```
chtyim: No need to check `excludeFeatures.isEmpty()`.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+  public static List<String> getFeatureList(Schema inputSchema, String featuresToInclude, String featuresToExclude,
+                                            String labelField) {
+    List<String> fields = new LinkedList<>();
+    List<Schema.Field> inputSchemaFields = inputSchema.getFields();
+    List<String> excludeFeatures = new ArrayList<>();
+
+    if (!Strings.isNullOrEmpty(featuresToInclude)) {
+      fields = Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToInclude));
+      for (String field : fields) {
+        Schema.Field inputField = inputSchema.getField(field);
```
chtyim, line 94: What if the `featuresToInclude` contains the `labelField`? Is that allowed? I see you have logic to always exclude that in the exclusion logic below, but not in here.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+   * @param cardinalityMapping feature to cardinality mapping specified for categorical features.
+   * @return categoricalFeatureInfo for categorical features.
+   */
+  public static Map<Integer, Integer> getCategoricalFeatureInfo(Schema inputSchema, String featuresToInclude,
+                                                                String featuresToExclude, String labelField,
+                                                                String cardinalityMapping) {
+    List<String> featureList = getFeatureList(inputSchema, featuresToInclude, featuresToExclude, labelField);
+    Map<Integer, Integer> outputFieldMappings = new HashMap<>();
+
+    if (!Strings.isNullOrEmpty(cardinalityMapping)) {
```
chtyim: Better revert the logic:

``` java
if (Strings.isNullOfEmpty(cardinalityMapping)) {
  return outputFieldMappings;
}
for (String field : Splitter...) {
}
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+      for (String field : Splitter.on(',').trimResults().split(cardinalityMapping)) {
+        String[] value = field.split(":");
```
chtyim: Is the format of the `cardinalityMapping` as `key1:value1,key2:value2,...` ? If that's the case, better use the `MapSplitter`.

``` java
Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(':').split(cardinalityMapping);
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
chtyim: If this method is intended to be private, make it private.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,210 @@
...
+    if (!Strings.isNullOrEmpty(featuresToExclude) && !Strings.isNullOrEmpty(featuresToInclude)) {
+      throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
+                                           "Please specify fields for one");
+    }
+    List<String> fields = getFeatureList(inputSchema, featuresToInclude, featuresToExclude, labelField);
+    for (String field : fields) {
+      Schema.Field inputField = inputSchema.getField(field);
+      Schema schema = inputField.getSchema();
+      Schema.Type features = schema.isNullableSimple() ? schema.getNonNullable().getType() : schema.getType();
+      if (features.equals(Schema.Type.BYTES) || features.equals(Schema.Type.STRING) ||
```
chtyim: What happen in the type is a RECORD? It's better to check for what you allowed (INT, FLOAT, DOUBLE, LONG).

ananya-bhattacharya:
PR closed when reverting last commit.
Reopening PR.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: If both featuresToExclude and featuresToInclude are null, all fields except predictionField will be used for training/prediction.
Please refer design discussion- https://wiki.cask.co/display/CE/Decision+Tree+Regression

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: This method is intended to be public, as this will be used from the Trainer and Predictor classes during transform or run.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: The else block is used if featuresToExclude is specified. It will also be valid in case both featuresToInclude and featuresToExclude are null.
Please refer design discussion- https://wiki.cask.co/display/CE/Decision+Tree+Regression

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: We need to check excludeFeatures.isEmpty() to include all fields except prediction field for training/prediction(we've already verified that featuresToInclude is empty in the if statement).
Please refer design discussion- https://wiki.cask.co/display/CE/Decision+Tree+Regression

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: No features to be used for training/prediction cannot include label/prediction field. Updated code for this.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: For trainer, we are using the term labelField instead of predictionField as is mentioned in the design discussion https://wiki.cask.co/display/CE/Decision+Tree+Regression.

chtyim: What I am saying is labelfield can never be null in this code logic. I believe you want to check if the result of inputSchema.getField is null, which is the `prediction` variable

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+/**
+ * Spark plugin Utility class. Contains common code to be used in trainer and predictors.
+ */
+public final class SparkUtils {
```
chtyim: Are all the methods in this class intended to be public?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,194 @@
...
+
+    if (!Strings.isNullOrEmpty(featuresToInclude)) {
+      List<String> features = Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToInclude));
+      for (String field : features) {
+        Schema.Field inputField = inputSchema.getField(field);
+        if (!field.equals(predictionField) && inputField != null) {
+          fields.add(field);
+        }
+      }
+    } else {
```
chtyim: What I meant was if you structure you code like this:

``` java
if (!Strings.isNullOrEmpty(featuresToinclude)) {
  ...
  return fields;
}

Set<String> excludeFeatures = new HashSet<>();
...

return fields;
```

The code would be less indented, hence easier to read.

chtyim: That's what I meant a about the `else` block is not needed if you return from the `if` block.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,194 @@
...
+      Set<String> excludeFeatures = new HashSet<>();
+      if (!Strings.isNullOrEmpty(featuresToExclude)) {
+        excludeFeatures.addAll(Lists.newArrayList(Splitter.on(',').trimResults().split(featuresToExclude)));
+      }
+      for (Schema.Field field : inputSchemaFields) {
+        String fieldName = field.getName();
+        if (!fieldName.equals(predictionField) && (excludeFeatures.isEmpty() || !excludeFeatures.contains(fieldName))) {
```
chtyim: The reason why you don't need to check `excludeFeatures.isEmpty()` is because, if it is empty, the `excludeFeatures.contains` will always return false, hence no field will be exclude (i.e. all fields, except the `predictionField` will be included). I believe that's the logic you want, right?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,194 @@
...
+    List<String> featureList = getFeatureList(inputSchema, featuresToInclude, featuresToExclude, labelField);
+    Map<Integer, Integer> outputFieldMappings = new HashMap<>();
+
+    if (!Strings.isNullOrEmpty(cardinalityMapping)) {
+      return outputFieldMappings;
+    }
+    try {
+      Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(cardinalityMapping);
+      for (String field : map.keySet()) {
+        String value = map.get(field);
```
chtyim: Instead of looping on the key set and call get, iterate against the entry set.

``` java
for (Map.Entry<String, String> entry : map.entrySet()) {
   ...
}
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,194 @@
...
+        try {
+          outputFieldMappings.put(featureList.indexOf(field), Integer.parseInt(value));
```
chtyim: `indexOf` is quite a slow op, especially used inside a loop. Better have the `getFeatureList` method as `getFeatureMap` that return a `Map<String, Integer>`, containing the feature name to position. For places that only need the feature names, simply use the `keySet()` method to get it.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: The methods of this class will be usually used within the spark-plugins module.
Changed the method scopes to protected and class scope to default package.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,199 @@
...
+   * Get the feature list of the features that have to be used for training/prediction depending on the
+   * featuresToInclude or featuresToInclude list.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the prediction values.
+   * @return feature list to be used for training/prediction.
+   */
+  protected static Map<String, Integer> getFeatureList(Schema inputSchema, @Nullable String featuresToInclude,
```
abhinavbansal246: Will this method be used in the plugin classes also? If it will just be used inside the methods of this class (like validateConfigParameters ). Please make it private

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,199 @@
...
+   * Get the feature to cardinality mapping provided by the user.
+   *
+   * @param inputSchema        schema of the received record.
+   * @param featuresToInclude  features to be used for training/prediction.
+   * @param featuresToExclude  features to be excluded when training/predicting.
+   * @param labelField         field containing the prediction values.
+   * @param cardinalityMapping feature to cardinality mapping specified for categorical features.
+   * @return categoricalFeatureInfo for categorical features.
+   */
+  protected static Map<Integer, Integer> getCategoricalFeatureInfo(Schema inputSchema,
```
abhinavbansal246: Same as getFeatureList, if this will not be used in plugin classes, make this private

abhinavbansal246:
minor comments, otherwise LGTM

ananya-bhattacharya:
@abhinavbansal246 , all review comments from @chtyim have been addressed.
If there are no new review comments, please see if this can be merged since the utility class will be reused by other spark ml plugins.

abhinavbansal246:
@ananya-bhattacharya Can you please address the comments regarding access modifiers

abhinavbansal246 (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: This is intended to be used in the plugin transform and run methods.
Hence, keeping this protected

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: This is intended to be used in the plugin transform and run methods.
Hence, keeping this protected

abhinavbansal246 (approved):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+   * Validate the config parameters for the spark sink and spark compute classes.
+   *
+   * @param inputSchema       schema of the received record.
+   * @param featuresToInclude features to be used for training/prediction.
+   * @param featuresToExclude features to be excluded when training/predicting.
+   * @param predictionField   field containing the prediction values.
+   */
+  protected static void validateConfigParameters(Schema inputSchema, @Nullable String featuresToInclude,
```
chtyim: Doesn't make sense to have `protected` modified in util class, since there will be no one inheriting this class. Since this class is package private, non-private methods in this class should also be package private.

PR #461, opened 2016-09-13
CDAP-7248 While setting the properties for the Input/Output format providers only use the properties that are set explicitly in the plugin.

sagarkapare:


hydrator-common/src/main/java/co/cask/hydrator/common/batch/ConfigurationUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Utility class for providing operations on the {@link Configuration}.
+ */
+public final class ConfigurationUtils {
+  private ConfigurationUtils() {
+    // no-op
+  }
+
+  public static Map<String, String> getConfigurationDiff(Configuration originalConf, Configuration modifiedConf) {
```
chtyim: Not javadoc? What does this method do?

sagarkapare: Missed adding javadocs. added

hydrator-common/src/main/java/co/cask/hydrator/common/batch/ConfigurationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Get the {@link Map} of properties which are newly added or have updated in the configuration instance passed
+   * to this method as a parameter. The modified configurations are compared with the default configurations
+   * created using {@code new Configuation()}.
+   * @param modifiedConf the updated configurations
+   * @return the {@link Map} of properties which are newly added or have updated in the modifiedConf.
+   */
+  public static Map<String, String> getNonDefaultConfigurations(Configuration modifiedConf) {
+    MapDifference<String, String> mapDifference = Maps.difference(getConfigurationAsMap(new Configuration()),
```
albertshau, line 41: is new Configuration() always the same as Job.getInstance().getConfiguration()? It seems safer to always pass in the original conf.

albertshau: It seems safer to create something like:

```
public class TrackingConfiguration extends Configuration {
  private final Configuration original;

  public TrackingConfiguration(Configuration original) {
    super(original);
    this.original = original;
  }

  public Map<String, String> getConfigurationChanges() {
    MapDifference<String, String> mapDifference = Maps.difference(getConfigurationAsMap(original), 
                                                                                                            getConfigurationAsMap(this));
    ...
  }
}
```

albertshau, line 41: I think its safer to make this a non-utils class that takes an original Configuration in its constructor and clones it, then uses that original instead of new Configuration() here.

albertshau, line 41: oh nevermind, ignore the above comments. Just saw the implementation of Job.getInstance().

albertshau:
lgtm assuming its been tested on a secure cluster.

sagarkapare:
Thanks @albertshau for reviewing. Verified the fix on the HDInsight cluster where it was breaking by running the CopyBookReader integration test. Will merge this and kick ITN so that its tested on other distros as well.

sagarkapare:
@albertshau Please take a look at last commit. Thanks!

albertshau:
lgtm

sagarkapare:
Build is green here - https://builds.cask.co/browse/HYP-BAD121-6

sagarkapare:
Squashing and merging.

PR #462, opened 2016-09-13
HYDRATOR-950 changing password field from string to password type.

abhinavbansal246:
Simple change to make password field masked
https://issues.cask.co/browse/HYDRATOR-950


anwar6953:
lgtm

PR #466, opened 2016-09-19
Linear Regression

rashig:
Spark plugin that trains and predicts the label data based on the Linear Regression.
For design, please refer: https://wiki.cask.co/display/CE/Linear+Regression


albertshau:
All these ML plugins are almost identical. The only difference between the predictors is a couple lines around loading the model, and the only difference between the trainers is a couple lines around training the model. Should get one merged, then look at how we can extract all the common logic into a base class that everything extends.

rashig:
@albertshau 
This PR has the refactored code which shares the common SparkUtils code. 
Please review.

russorat:
closing due to age

PR #471, opened 2016-09-21
add createConnection method for compatibility with versions >= HDP-2.3.4

gsps1:
JIRA : https://issues.cask.co/browse/HYDRATOR-791


albertshau:
lgtm assuming its been tested.

gsps1:
yes, tested the fix on HDP cluster. which doesn't work otherwise with a DBSource

gsps1:
build green http://builds.cask.co/browse/HYP-BAD133/latest

PR #473, opened 2016-09-23
snapshot file sink for text format

gsps1:
JIRA : https://issues.cask.co/browse/HYDRATOR-970

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.sink;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.dataset.lib.FileSetProperties;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.batch.BatchSink;
+import co.cask.hydrator.plugin.common.StructuredToTextTransformer;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
+import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+
+import javax.annotation.Nullable;
+
+
+/**
+ * {@link SnapshotFileBatchSink} that stores data in text format.
+ */
+@Plugin(type = BatchSink.PLUGIN_TYPE)
+@Name("SnapshotText")
+@Description("Sink for a SnapshotFileSet that writes data in Text format.")
+public class SnapshotFileBatchTextSink extends SnapshotFileBatchSink<String, NullWritable> {
+  private StructuredToTextTransformer recordTransformer;
+  private SnapshotFileBatchTextSinkConfig config;
+
+  public SnapshotFileBatchTextSink(SnapshotFileBatchTextSinkConfig config) {
+    super(config);
+    this.config = config;
+  }
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    super.initialize(context);
+    recordTransformer = new StructuredToTextTransformer(config.delimiter);
+  }
+
+  @Override
+  public void transform(StructuredRecord input,
+                        Emitter<KeyValue<String, NullWritable>> emitter) throws Exception {
+    emitter.emit(new KeyValue<>(recordTransformer.transform(input), NullWritable.get()));
+  }
+
+  @Override
+  protected void addFileProperties(FileSetProperties.Builder propertiesBuilder) {
+    propertiesBuilder
+      .setInputFormat(TextInputFormat.class)
+      .setOutputFormat(TextOutputFormat.class)
+      .setEnableExploreOnCreate(true)
+      .setExploreFormat("text")
+      .setExploreSchema("text string");
+  }
+
+  /**
+   * Config for SnapshotFileBatchTextSink.
+   */
+  public static class SnapshotFileBatchTextSinkConfig extends SnapshotFileSetBatchSinkConfig {
+    @Description("The Delimiter used to combine the Structured Record fields, Defaults to tab")
+    @Nullable
+    @Macro
+    private String delimiter;
+
+    public SnapshotFileBatchTextSinkConfig() {
+      this.delimiter = "\t";
+    }
+
+    public SnapshotFileBatchTextSinkConfig(String name, @Nullable String basePath, @Nullable String delimiter) {
+      super(name, basePath, null);
+      this.delimiter = delimiter;
+    }
+  }
+}
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
@@ -1,0 +1,62 @@
+
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.lib.output.TextOutputFormat;
+
+import javax.annotation.Nullable;
+
+
+/**
+ * {@link SnapshotFileBatchSink} that stores data in text format.
+ */
+@Plugin(type = "batchsink")
```
albertshau: use BatchSink.PLUGIN_TYPE

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
```diff
@@ -0,0 +1,86 @@
...
+                        Emitter<KeyValue<String, NullWritable>> emitter) throws Exception {
+    emitter.emit(new KeyValue<>(recordTransformer.transform(input, input.getSchema()),
+                                NullWritable.get()));
+  }
+
+  @Override
+  protected void addFileProperties(FileSetProperties.Builder propertiesBuilder) {
+    // parse it to make sure its valid
+    propertiesBuilder
+      .setInputFormat(AvroKeyInputFormat.class)
```
albertshau: should use TextInputFormat

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
```diff
@@ -0,0 +1,86 @@
...
+      .setEnableExploreOnCreate(true)
+      .setExploreFormat("text")
+      .setExploreSchema("text string");
+  }
+
+  /**
+   * Config for SnapshotFileBatchTextSink.
+   */
+  public static class SnapshotFileBatchTextSinkConfig extends SnapshotFileSetBatchSinkConfig {
+    @Description("The Delimiter used to combine the Structured Record fields")
```
albertshau: lets make this nullable and have it default to a tab

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.hydrator.common.RecordConverter;
+
+import java.io.IOException;
+
+/**
+ * Creates Single String from StructuredRecord
+ */
+public class StructuredToTextTransformer extends RecordConverter<StructuredRecord, String> {
```
albertshau: don't think there is any need to extend RecordConverter, since its not using any of the methods in RecordConverter

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+  private final String delimiter;
+  private static final String FIELD_NAME = "record";
+
+  public StructuredToTextTransformer(String delimiter) {
+    this.delimiter = delimiter;
+  }
+
+  @Override
+  public String transform(StructuredRecord structuredRecord,
+                          co.cask.cdap.api.data.schema.Schema schema) throws IOException {
```
albertshau: doesn't need to be the fully qualified name

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+    String recordString = "";
+    for (co.cask.cdap.api.data.schema.Schema.Field field : schema.getFields()) {
+      Object recordField = structuredRecord.get(field.getName());
+      String fieldValue = recordField == null ? "" : recordField.toString();
+      recordString += fieldValue + delimiter;
```
albertshau: use a StringBuilder instead of concatenating this way

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+    return recordString.substring(0, recordString.length());
```
albertshau: It's better not to substring, since that's another pass through the string. if using a StringBuilder, you can do deleteCharAt() instead. Or you can restructure the loop so that you do the first field, then loop with appending delimiter then next field value.

albertshau (changes requested):
please add a unit test

gsps1:
@albertshau thanks for the review. addressed the comments.

gsps1:
http://builds.cask.co/browse/HYP-BAD135-3

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
```diff
@@ -75,12 +75,14 @@ protected void addFileProperties(FileSetProperties.Builder propertiesBuilder) {
...
+    @Description("The Delimiter used to combine the Structured Record fields, Defaults to tab")
+    @Nullable
+    @Macro
     private String delimiter;
 
-    public SnapshotFileBatchTextSinkConfig(String name, @Nullable String basePath, String delimiter) {
+    public SnapshotFileBatchTextSinkConfig(String name, @Nullable String basePath, @Nullable String delimiter) {
       super(name, basePath, null);
-      this.delimiter = delimiter;
+      this.delimiter = delimiter == null ? "\t" : delimiter;
```
albertshau: its a little cleaner to have a no-args constructor that calls this one with tab as the delimiter. then you don't need to do the null check in initialize()

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
```diff
@@ -18,31 +18,42 @@
...
+   * @param structuredRecord
+   * @return String
+   * @throws IOException
+   */
+  public String transform(StructuredRecord structuredRecord) throws IOException {
+    StringBuilder builder = new StringBuilder();
+    List<Schema.Field> fields = structuredRecord.getSchema().getFields();
+    builder.append(getFieldAsString(structuredRecord.get(fields.get(0).getName())));
+
+    for (int index = 1; index < fields.size(); index++) {
```
albertshau: I think its always an ArrayList, but it's still better to use an iterator rather than get(index) since there is no guarantee that it's O(1).

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
```diff
@@ -195,6 +259,21 @@ private void testSource(String sourceETLPlugin, String sourceName, Map<String, I
...
+    String data = "";
+    for (Location file : outputLocation.list()) {
+      String fileName = file.getName();
+      // text output format doesn't add extension by default, so looking for all part files.
+      if (fileName.startsWith("part")) {
+        String partData = CharStreams.toString(new InputStreamReader(file.getInputStream(), Charsets.UTF_8));
+        data += partData;
+      }
+    }
+    List<String> records = new ArrayList<>();
```
albertshau: its a little better to read through line by line and add to the collection rather than getting the entire string then splitting it

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
```diff
@@ -195,6 +259,21 @@ private void testSource(String sourceETLPlugin, String sourceName, Map<String, I
     Assert.assertEquals(expected, actual);
   }
 
+  private List<String> readTextOutput(Location outputLocation) throws IOException {
```
albertshau: you probably want a Set here, since there is no guarantee on which lines are written to which files.

albertshau (changes requested):
a few minor comments

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
gsps1: how to read line by line when using input stream to get as a string, that string could cover multiple lines.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
gsps1: true

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
gsps1: but name is a non-nullable field.. no-args constructor won't be able to call this then..

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToTextTransformer.java
gsps1: yea, updated it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/SnapshotFileBatchTextSink.java
gsps1: setting the defaults in no-args constructor directly.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
gsps1: using scanner to read by line.

gsps1:
@albertshau addressed comments. please review again.

gsps1:
Build green -http://builds.cask.co/browse/HYP-BAD135-6

albertshau:
lgtm

albertshau (approved):

PR #475, opened 2016-09-23
HYDRATOR-971 adding a file streaming source

albertshau:
The source allows the user to configure a path, format, schema,
a whitelist of file extensions, and whether to ignore existing
files or not. It will then monitor the directory and read any
new files it finds.

### DIFF
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.spark;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.annotation.Name;
+import co.cask.cdap.api.annotation.Plugin;
+import co.cask.cdap.api.data.format.FormatSpecification;
+import co.cask.cdap.api.data.format.RecordFormat;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.flow.flowlet.StreamEvent;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+import co.cask.cdap.etl.api.streaming.StreamingContext;
+import co.cask.cdap.etl.api.streaming.StreamingSource;
+import co.cask.cdap.format.RecordFormats;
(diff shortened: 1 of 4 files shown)


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
```diff
@@ -0,0 +1,225 @@
...
+  public static class Conf extends ReferencePluginConfig {
+    private static final Set<String> FORMATS = ImmutableSet.of("text", "csv", "tsv", "clf", "grok", "syslog");
+
+    @Macro
+    @Description("The format of the source files. Must be text, csv, tsv, clf, grok, or syslog. Defaults to text.")
+    @Nullable
+    private String format;
+
+    @Description("The schema of the source files.")
+    @Nullable
```
CuriousVini: Why do we want schema to be nullable? If schema is not specified `conf.getSchema()` will throw NPE

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
```diff
@@ -0,0 +1,225 @@
...
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    super.configurePipeline(pipelineConfigurer);
+    conf.validate();
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(conf.getSchema());
+  }
+
+  @Override
+  public JavaDStream<StructuredRecord> getStream(StreamingContext context) throws Exception {
+    conf.validate();
```
CuriousVini, line 79: Why do we need to call `conf.validate()` twice?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
```diff
@@ -0,0 +1,225 @@
...
+   * Transforms kafka key and message into a structured record when message format and schema are given.
+   * Everything here should be serializable, as Spark Streaming will serialize all functions.
+   */
+  private static class FormatFunction implements FlatMapFunction<Tuple2<LongWritable, Text>, StructuredRecord> {
+    private final String format;
+    private final String schemaStr;
+    private transient Schema schema;
+    private transient RecordFormat<StreamEvent, StructuredRecord> recordFormat;
+
+    FormatFunction(String format, String schemaStr) {
```
CuriousVini, line 130: Make `schemaStr` nullable?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
```diff
@@ -0,0 +1,225 @@
...
+      this.format = format;
+      this.schemaStr = schemaStr;
+    }
+
+    @Override
+    public Iterable<StructuredRecord> call(Tuple2<LongWritable, Text> in) throws Exception {
+      // first time this was called, initialize schema and time, key, and message fields.
+      if (recordFormat == null) {
+        schema = Schema.parseJson(schemaStr);
```
CuriousVini, line 139: Again what happens when `schemaStr` is null

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
albertshau: you are right, it should not be nullable. will change.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/FileStreamingSource.java
albertshau: this is because macros are enabled, so if there is a macro, it won't get checked at configure time since substitution has not happened yet.

CuriousVini:
lgtm

PR #478, opened 2016-09-24
HYDRATOR-973 add http poll streaming source

albertshau:
Adding the equivalent of the HTTP Poller source from the old
realtime framework into the new streaming framework. The source
uses the exact same config properties.

### DIFF
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HTTPPollerSource.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HTTPPollerSource.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
+ *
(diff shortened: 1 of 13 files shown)


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HTTPPollerSource.java
```diff
@@ -0,0 +1,95 @@
...
+              while (!isStopped()) {
+
+                try {
+                  store(httpRequestor.get());
+                } catch (Exception e) {
+                  LOG.error("Error getting content from {}.", conf.getUrl(), e);
+                }
+
+                try {
+                  TimeUnit.SECONDS.sleep(conf.getInterval());
```
gsps1, line 75: might be better to store interval and url in variable during initialization.

sreevatsanraman:
LGTM

PR #483, opened 2016-10-06
Avoid suppressing HCatException. Do proper equality check on hive types.

anwar6953:
1. Avoid suppressing HCatException. Otherwise, it seems like the pipeline execution will continue, and transfer only a subset of the fields (ignoring the field(s) that had issues).
2. Do proper equality check on hive types. Otherwise, the WARN message will ALWAYS be logged.

### DIFF
--- a/hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/commons/HiveSchemaConverter.java
+++ b/hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/commons/HiveSchemaConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 Cask Data, Inc.
+ * Copyright © 2015-2016 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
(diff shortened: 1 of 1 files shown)


hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/commons/HiveSchemaConverter.java
```diff
@@ -57,14 +58,15 @@ public static HCatSchema toHiveSchema(Schema schema, HCatSchema tableSchema) {
...
           LOG.warn("The given schema {} for the field {} does not match the schema {} from the table. " +
-                     "The schema {} for field {} be used.", type, name, hiveType, hiveType, name);
+                     "The schema {} for field {} will be used.", type, name, hiveType, hiveType, name);
         }
         fields.add(hCatFieldSchema);
       } catch (HCatException e) {
         LOG.error("Failed to create HCatFieldSchema field {} of type {} from schema", name,
-                  field.getSchema().getType());
+                  field.getSchema().getType(), e);
+        Throwables.propagate(e);
```
albertshau: I think we should just throw an exception with the message that is being logged, and with e as its parent instead of logging and throwing a RuntimeException. Perhaps as an IllegalArgumentException since the caller is, sending in bad input?

Though it's kind of a moot point. Looking at the code, tableSchema.get(name) declares that it throws HCatException, but actually will never throw an HCatException. Good job Hive...

anwar6953: Done.

albertshau:
lgtm

PR #486, opened 2016-10-07
Fixing [HYDRATOR-995] test case failure in DecisionTreeConfigTest

ananya-bhattacharya:
Changed map type in getFeatureList() in SparkUtils.java from HashMap() to LinkedHashMap() to maintain the insertion order of fields from schema for different reducers, and avoid inconsistent test behavior.

### DIFF
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
@@ -26,6 +26,7 @@ import joptsimple.internal.Strings;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -84,7 +85,7 @@ final class SparkUtils {
       throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
                                            "Please specify fields for one.");
     }
-    Map<String, Integer> fields = new HashMap<>();
+    Map<String, Integer> fields = new LinkedHashMap<>();
 
     if (!Strings.isNullOrEmpty(featuresToInclude)) {
       Iterable<String> tokens = Splitter.on(',').trimResults().split(featuresToInclude);
--- a/spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/DecisionTreeConfigTest.java
+++ b/spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/DecisionTreeConfigTest.java
@@ -20,7 +20,6 @@ import co.cask.cdap.api.data.schema.Schema;
 import co.cask.hydrator.plugin.spark.DecisionTreePredictor;
 import co.cask.hydrator.plugin.spark.DecisionTreeTrainer;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
 public class DecisionTreeConfigTest {
@@ -72,8 +71,6 @@ public class DecisionTreeConfigTest {
   }
 
   @Test
-  @Ignore
-  // Ignoring this test temporarily. Enable this test once HYDRATOR-995 is fixed.
   public void testIncludeAllFeatures() throws Exception {
     DecisionTreeTrainer.DecisionTreeTrainerConfig config =
       new DecisionTreeTrainer.DecisionTreeTrainerConfig("decision-tree-regression-model", "decisionTreeRegression",


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -84,7 +85,7 @@ static void validateConfigParameters(Schema inputSchema, @Nullable String featur
       throw new IllegalArgumentException("Cannot specify values for both featuresToInclude and featuresToExclude. " +
                                            "Please specify fields for one.");
     }
-    Map<String, Integer> fields = new HashMap<>();
+    Map<String, Integer> fields = new LinkedHashMap<>();
```
CuriousVini, line 88: Can you explain the cause of the flakiness and how this change solves it? Since we iterate over [inputSchemaFields](https://github.com/caskdata/hydrator-plugins/blob/develop/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java#L106) the same order as schema

ananya-bhattacharya: The test checks the schema for field types and throws error if the type is not in : int, long, float or double.
In the input schema the "tailNum" is the first field that does not satisfy the condition, and therefore is expected to throw error on this field.
However, since HashMap does not maintain the insertion order the keySet received from getFeatureList() was in random order even if "inputSchemaFields"  had the correct order of elements and caused the test to fail at field other than "tailNum".
Since, LinkedList maintains the insertion order the keySet retrieved from the getFeatureList() was in the same order as is present in "inputSchemaFields" and hence the test case passes as it now throws error on the first non-numeric field in the schema i.e. "tailNum" as is expected by the test.
Please let me know if this answers the question.

CuriousVini: I see. This is a util class and usage of the method `getFeatureList()` does not depend on insertion order of fields. So I think it would be better if we make changes into `DecisionTreeConfigTest` rather than into `SparkUtils`

ananya-bhattacharya: Agreed. But the HashMap, would never guarantee the insertion order in case of multiple reducers, and changing the field from `tailNum` to origin in the test case may not be guarantee the test case would pass in all scenarios. This is because for multiple reducers we can have different sequence and the test may fail for fields other than tailNum or origin.

CuriousVini: In the test case, instead of using schema with more than one string types, may be we can use only 1 invalid field with string type in the schema and change `Assert.assertEquals("Features must be of type : int, double, float, long but was of type STRING for " +
                            "field tailNum.", e.getMessage());` message accordingly?

ananya-bhattacharya: The SparkUtils class is being used by the other spark ml plugins and the order error will occur for every plugin. Also, the sequence of fields in the output schema created using the `getFeatureList()` will differ for different reducers for the current implementation.
The test `testIncludeAllFeatures()` checks for a scenario where the `featuresToInclude` and `featuresToInclude` parameters are both empty and hence all the fields will be checked for the data types in the sequence they appear in the input schema. Not sure if changing the test case will solve the basic issue.
Please suggest.

CuriousVini: There are 4 places where `SparkUtils#getFeatureList()` method is used:
- `DecisionTreePredictor#transform()`
- `DecisionTreeTrainer#run()`
- `SparkUtils#getCategoricalFeatureInfo()`
- `SparkUtils#validateConfigParameters()`

Wondering in which method do we need ordering in any of the above usages?

ananya-bhattacharya: We are using it in the `DecisionTreePredictor#transform()` and `DecisionTreeTrainer#run()` to create the sparse vector of the features to be used for training/prediction.For every structured record the [featureList ](https://github.com/caskdata/hydrator-plugins/blob/5057d5af2ab52ec2f5fef432df6fc8f01d04cebb/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java#L90) and [featureIndex ](https://github.com/caskdata/hydrator-plugins/blob/5057d5af2ab52ec2f5fef432df6fc8f01d04cebb/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java#L91) are created for the fields in the sequence retrieved [getFeatureList().keySet()](https://github.com/caskdata/hydrator-plugins/blob/5057d5af2ab52ec2f5fef432df6fc8f01d04cebb/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreeTrainer.java#L79).Random order returned from `SparkUtils#getFeatureList()` will result in incorrect sequence of values in the sparse vector in case of multiple reducers.

CuriousVini: I see. In that case changes to SparkUtils makes sense.

CuriousVini:
Lgtm.

CuriousVini:
http://builds.cask.co/browse/HYP-BAD150-1

CuriousVini:
Build is green. Merging.

PR #491, opened 2016-10-17
Adding base class structure for Spark ML Trainers and Predictors/Classifiers

romy-khetan:


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,152 @@
...
+    pipelineConfigurer.createDataset(config.fileSetName, FileSet.class);
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
+    config.validate(inputSchema);
+  }
+
+  @Override
+  public void prepareRun(SparkPluginContext context) throws Exception { }
+
+  protected Map<String, Integer> getFeatures(Schema inputSchema, @Nullable String featuresToInclude,
```
albertshau: this should only need the inputSchema. Everything else is already in the config.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,152 @@
...
+                                             @Nullable String featuresToExclude, String labelField) {
+    return SparkUtils.getFeatureList(inputSchema, featuresToInclude, featuresToExclude, labelField);
+  }
+
+  public static JavaRDD<LabeledPoint> getTrainingData(final Map<String, Integer> fields,
```
albertshau: this is written like its meant to be used as a static helper method. Should not need to do it that way, as the purpose of this abstract class is to define the basic code structure and leave as little to the children as possible.

Instead, this method should override the run method from SparkSink, and at the end call a trainModel() method that the child must implement. So something like this:

```
public abstract trainModel(SparkExecutionPluginContext context, JavaRDD<LabeledPoint> trainingData, String outputPath); 

@Override
public void run(final SparkExecutionPluginContext context, final JavaRDD<StructuredRecord> input) throws Exception {

  if (input == null || input.isEmpty()) {
    return;
  }
  Schema inputSchema = input.first().getSchema();
  final Map<String, Integer> fields = SparkUtils.getFeatureList(inputSchema, config.featuresToInclude, config.featuresToExclude, config.labelField);

  [existing code to generate JavaRDD<LabeledPoint>]

  FileSet outputFS = context.getDataset(config.fileSetName);
  String outputPath = outputFS.getBaseLocation().append(config.path).toURI().getPath();

  trainModel(context, trainingData, outputPath);
}
```

This way, the child basically just has a constructor and trainModel().

albertshau: Please refactor DecisionTreeTrainer to extend this class to validate that it works as expected and make sure there isn't anything missing.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
```diff
@@ -0,0 +1,137 @@
...
+    return JavaSparkContext.toSparkContext(javaSparkContext);
+  }
+
+  public Schema getOutputSchema(Schema inputSchema) {
+    return outputSchema = (outputSchema != null) ? outputSchema :
+      SparkUtils.getOutputSchema(inputSchema, config.predictionField);
+  }
+
+  public Vector getVector(StructuredRecord record, final Map<String, Integer> featuresList) {
+        List<Double> featureList = new ArrayList<>();
```
albertshau, line 135: inconsistent indentation

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
```diff
@@ -0,0 +1,137 @@
...
+    JavaSparkContext javaSparkContext = context.getSparkContext();
+    return JavaSparkContext.toSparkContext(javaSparkContext);
+  }
+
+  public Schema getOutputSchema(Schema inputSchema) {
+    return outputSchema = (outputSchema != null) ? outputSchema :
+      SparkUtils.getOutputSchema(inputSchema, config.predictionField);
+  }
+
+  public Vector getVector(StructuredRecord record, final Map<String, Integer> featuresList) {
```
albertshau: featuresList doesn't have to be final

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
```diff
@@ -0,0 +1,137 @@
...
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
+    config.validate(inputSchema);
+    stageConfigurer.setOutputSchema(SparkUtils.getOutputSchema(inputSchema, config.predictionField));
+  }
+
+  public String getModelPath(SparkExecutionPluginContext context) throws IOException {
```
albertshau: instead of supplying methods that are meant to be used by the children, it is better for this class to define the basic code structure and leave as little to the child as possible. 

this means it should override the initialize and transform methods from SparkCompute and define abstract methods to fill in the gaps. Something like:

```

  public abstract void initialize(SparkExecutionPluginContext context, String modelPath);

  public abstract double predict(Vector vector);

  @Override
  public void initialize(SparkExecutionPluginContext context) throws Exception {
    super.initialize(context);
    FileSet fileSet = context.getDataset(config.fileSetName);
    Location modelLocation = fileSet.getBaseLocation().append(config.path);
    if (!modelLocation.exists()) {
      throw new IllegalArgumentException(String.format("Failed to find model to use for Regression. Location does "
                                                         "not exist: %s.", modelLocation));
    }
    initialize(context, modelLocation.toURI().getPath());
  }

  @Override
  public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext context,
                                             JavaRDD<StructuredRecord> input) throws Exception {
    if (input == null || input.isEmpty()) {
      return context.getSparkContext().emptyRDD();
    }
    Schema schema = input.first().getSchema();
    outputSchema = (outputSchema != null) ? outputSchema : SparkUtils.getOutputSchema(schema, config.predictionField);
    final Map<String, Integer> fields = SparkUtils.getFeatureList(schema, config.featuresToInclude,
                                                                  config.featuresToExclude,
                                                                  config.predictionField);
    JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
      @Override
      public StructuredRecord call(StructuredRecord record) throws Exception {
        Vector vector = getVector(record, )
        double prediction = loadedModel.predict(vector, fields);

        return SparkUtils.cloneRecord(record, outputSchema, config.predictionField)
          .set(config.predictionField, prediction)
          .build();
      }
    });
    return output;
  }
```

This way, the child is basically a constructor and two very short methods.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
```diff
@@ -0,0 +1,137 @@
...
+    Location modelLocation = fileSet.getBaseLocation().append(config.path);
+
+    if (!modelLocation.exists()) {
+      throw new IllegalArgumentException("Failed to find model to use for classification. " +
+                                           "Location does not exist: " + modelLocation, null);
+    }
+    return modelLocation.toURI().getPath();
+  }
+
+  public SparkContext getSparkContext(SparkExecutionPluginContext context) {
```
albertshau: this can be removed as it doesn't provide much utility. Users can just do `context.getSparkContext().sc()`

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,152 @@
...
+                                                           labelField));
+        }
+        return new LabeledPoint(prediction, Vectors.sparse(counter, Ints.toArray(featureIndex),
+                                                           Doubles.toArray(featureList)));
+      }
+    });
+    return trainingData;
+  }
+
+  public SparkContext getSparkContext(SparkExecutionPluginContext context) {
```
albertshau: this can be removed as it doesn't provide much utility. Users can just do context.getSparkContext().sc()

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
albertshau: please refactor DecisionTreePredictor to extend this class to make sure everything works and nothing is missing.

albertshau (changes requested):

romy-khetan:
@albertshau 
Updated code with the review comments. Please review.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
```diff
@@ -0,0 +1,153 @@
...
+    final Map<String, Integer> featuresList = SparkUtils.getFeatureList(inputSchema,
+                                                                        config.featureFieldsToInclude,
+                                                                        config.featureFieldsToExclude,
+                                                                        config.predictionField);
+    JavaRDD<StructuredRecord> output = input.map(new Function<StructuredRecord, StructuredRecord>() {
+      @Override
+      public StructuredRecord call(StructuredRecord record) throws Exception {
+        List<Double> featureList = new ArrayList<>();
+        List<Integer> featureIndex = new ArrayList<>();
+        int counter = 0;
```
albertshau, line 137: size can be found by just doing featuresList.size(). The featureList and featureIndex lists can then be arrays. That way we don't have to turn the lists into arrays and save two traversals

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,145 @@
...
+      final Schema inputSchema = input.first().getSchema();
+      final Map<String, Integer> featuresList = SparkUtils.getFeatureList(inputSchema, config.featureFieldsToInclude,
+                                                                          config.featureFieldsToExclude,
+                                                                          config.labelField);
+      JavaRDD<LabeledPoint> trainingData = input.map(new Function<StructuredRecord, LabeledPoint>() {
+        @Override
+        public LabeledPoint call(StructuredRecord record) throws Exception {
+          List<Double> featureList = new ArrayList<>();
+          List<Integer> featureIndex = new ArrayList<>();
+          int counter = 0;
```
albertshau, line 131: size can be found by just doing featuresList.size(). The featureList and featureIndex lists can then be arrays. That way we don't have to turn the lists into arrays and save two traversals

albertshau (changes requested):
mentioned it last time, but please refactor the DecisionTree classes to extend these classes to make sure they work properly.

romy-khetan:
@albertshau 
Refactored code for Decision Tree Regression.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
romy-khetan: Yes, in that way we can save two traversals. But the size of featureList and featureIndex could not be defined using featuresList.size(). It's because a record may have NULL values from the fields contained in featuresList. So if we create an array of  size 5 for 5 features, and if one of the input records have 2 features then unoccupied elements will be treated as 0 in case of integers. We are using Sparse Vectors in which generally 0's are avoided.

Please suggest for which approach should we go for.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
romy-khetan: Yes, in that way we can save two traversals. But the size of featureList and featureIndex could not be defined using featuresList.size(). It's because a record may have NULL values from the fields contained in featuresList. So if we create an array of  size 5 for 5 features, and if one of the input records have 2 features then unoccupied elements will be treated as 0 in case of integers. We are using Sparse Vectors in which generally 0's are avoided.

Please suggest for which approach should we go for.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,145 @@
...
+          for (String field : featuresList.keySet()) {
+            if (record.get(field) != null) {
+              featureList.add(((Number) record.get(field)).doubleValue());
+              featureIndex.add(counter);
+            }
+            counter++;
+          }
+          Double prediction = record.get(config.labelField);
+          if (prediction == null) {
+            throw new IllegalArgumentException(String.format("Value of Label field %s value must not be null.",
```
chtyim: Is it the intend that if any input is invalid (missing label field), the whole Spark job will fail? Or shall we skip the invalid input since it is invalid, it can't affect the prediction anyway.

romy-khetan: Filtered out invalid input when label is missing. Updated PR.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/DecisionTreePredictor.java
```diff
@@ -50,112 +32,32 @@
...
-
-    public DecisionTreePredictorConfig(String fileSetName, String path, @Nullable String featuresToInclude,
-                                       @Nullable String featuresToExclude, String predictionField) {
-      this.fileSetName = fileSetName;
-      this.path = path;
-      this.featuresToInclude = featuresToInclude;
-      this.featuresToExclude = featuresToExclude;
-      this.predictionField = predictionField;
-    }
+  public static class DecisionTreeConfig extends MLPredictorConfig {
```
albertshau: don't need this class if there are no additional fields and methods. Just use a MLPredictorConfig.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLPredictor.java
albertshau: ok I see, that makes sense. Let's keep what you have then.

albertshau (changes requested):
Just one minor comment for cleanup, rest LGTM.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
albertshau: ok I see, that makes sense. Let's keep what you have then.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkMLTrainer.java
```diff
@@ -0,0 +1,153 @@
...
+      final Schema inputSchema = input.first().getSchema();
+      final Map<String, Integer> featuresList = SparkUtils.getFeatureList(inputSchema, config.featureFieldsToInclude,
+                                                                          config.featureFieldsToExclude,
+                                                                          config.labelField);
+
+      // Filter out the records for which label is invalid
+      JavaRDD<StructuredRecord> filteredData = input.filter(new Function<StructuredRecord, Boolean>() {
+        @Override
+        public Boolean call(StructuredRecord record) throws Exception {
+          if(record.get(config.labelField) == null) {
```
chtyim: missing space after `if`. This should be a checkstyle error.

romy-khetan:
@albertshau 
Removed DecisionTreeConfig from DecisionTreeRegression class. Please review.

albertshau:
lgtm

romy-khetan:
Decision-Tree Regression ITN results:

[INFO]
[INFO] --- maven-checkstyle-plugin:2.12.1:check (validate) @ long-running-test ---
[INFO]
[INFO] --- maven-surefire-plugin:2.18.1:test (default-test) @ long-running-test ---
[INFO] Tests are skipped.
[INFO] Reactor Summary:
[INFO]
[INFO] integration-test ................................... SUCCESS [  4.081 s]
[INFO] integration-test-core .............................. SUCCESS [  3.934 s]
[INFO] integration-test-remote ............................ SUCCESS [05:40 min]
[INFO] integration-test-standalone ........................ SUCCESS [  0.797 s]
[INFO] upgrade-test ....................................... SUCCESS [  1.100 s]
[INFO] long-running-test .................................. SUCCESS [  1.415 s]
[INFO] BUILD SUCCESS
[INFO] Total time: 05:52 min
[INFO] Finished at: 2016-10-27T12:35:56+05:30
[INFO] Final Memory: 79M/238M

PR #311, opened 2016-07-13
Fix compilations errors

anwar6953:
Implement newPluginInstance(String,MacroEvaluator) in MockRealtimeContext and MockTransformContext, to match new APIs in cdap.

`mvn clean package -DskipTests` passed locally.

PR #492, opened 2016-10-17
[HYDRATOR-983] Feature Generator Plugins

ananya-bhattacharya:
Feature generator plugin can be used to generate text based feature from a string field. Text based feature generation can be done using either HashingTF technique or Skip-Gram model (Spark's Word2Vec) model.
For design and use case please refer - https://wiki.cask.co/display/CE/Feature+Generator


spark-plugins/docs/HashingTFFeatureGenerator-sparkcompute.md
```diff
@@ -0,0 +1,53 @@
...
+-----------
+SparkCompute to generate text based feature from string using Hashing TF technique.
+
+Use Case
+--------
+This transform can be used when user wants to generate text based feature using Hashing TF technique.
+
+Properties
+----------
+**numFeatures:** The number of features to use in training the model. It must be of type integer. Deafult is 100.
```
albertshau: Deafult -> Default

albertshau: 100 is usually too small to be useful, as there will be a ton of different strings mapping to the same feature. Why not use Spark's default (2^20)?

spark-plugins/docs/HashingTFFeatureGenerator-sparkcompute.md
```diff
@@ -0,0 +1,53 @@
...
+
+**outputColumnMapping:** A comma-separated list of the input fields to map to the transformed output fields. The key
+specifies the name of the field to generate feature vector from, with its corresponding value specifying the output
+columns(size, indices and value) to emit the sparse vector.
+
+
+Example
+-------
+This example transforms column ``text`` to generate fixed length vector of size 10 and emit the generated sparse vector
+as a cobination of three columns: result_size, result_indices, result_value.
```
albertshau: cobination -> combination

spark-plugins/docs/SkipGramFeatureGenerator-sparkcompute.md
```diff
@@ -0,0 +1,57 @@
+# SkipGramFeatureGenerator
+
+
+Description
+-----------
+SparkCompute to generate text based feature from string using stored skip-gram model (Spark's Word2Vec).
+
+Use Case
+--------
+This transform can be used when user has a saved Word2Vec model and wants to genearte text based feature.
```
albertshau: genearte -> generate

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Class to generate text based features using Hashing TF.
+ */
+@Plugin(type = SparkCompute.PLUGIN_TYPE)
+@Name(HashingTFFeatureGenerator.PLUGIN_NAME)
+@Description("Trains a regression model based upon a particular label and features of a record.")
+
```
albertshau: nit: remove extra newline

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+  @Override
+  public void initialize(SparkExecutionPluginContext context) throws Exception {
+    super.initialize(context);
+    hashingTF = new HashingTF().setNumFeatures(config.numFeatures);
+  }
+
+  @Override
+  public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext context, JavaRDD<StructuredRecord> input)
+    throws Exception {
+    if (input == null) {
```
albertshau: I don't think input can ever be null, but it can be empty. Did you notice cases where it is null? If not, this should be changed to input.isEmpty()

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+  private StructType schema;
+  private Schema outputSchema;
+  private HashingTF hashingTF;
+  private List<String> fieldList = new ArrayList<>();
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
```
albertshau, line 67: should be performing more validation here. If the input schema is not null, need to check that the fields specified in the config exist in the input schema and are strings or nullable strings.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+    JavaRDD<StructuredRecord> output = transformedRDD.map(new Function<Row, StructuredRecord>() {
+      @Override
+      public StructuredRecord call(Row row) throws Exception {
+        StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+        String[] fields = fieldList.toArray(new String[fieldList.size()]);
+        for (int i = 0; i < fields.length; i++) {
+          String fieldName = fields[i];
+          if (config.getFeatureListMapping().values().contains(fields[i])) {
+            SparseVector vector = ((Vector) row.get(i)).toSparse();
+            builder.set(fieldName + "_size", vector.size());
```
albertshau: should not be using 3 separate fields to represent a SparseVector. Instead, it should be a single field that is a record of 3 fields (int size, int[] indices, double[] values).  May be worthwhile to have a VectorUtils utility class, with SPARSE_SCHEMA as a constant:

```
Schema.recordOf("sparseVector",
  Schema.Field.of("size", Schema.of(Schema.Type.INT)),
  Schema.Field.of("indices", Schema.arrayOf(Schema.of(Schema.Type.INT))),
  Schema.Field.of("values", Schema.arrayOf(Schema.of(Schema.Type.DOUBLE))));
```

and methods to convert a SparseVector to a StructuredRecord and vice versa:

```
public static StructuredRecord asRecord(SparseVector vector) {
  return StructuredRecord.builder(SPARSE_SCHEMA)
    .set("size", vector.size())
    .set("indices", vector.indices())
    .set("values", vector.values())
    .build();
}

public static SparseVector fromRecord(StructuredRecord record) {
  int size = record.get("size");
  int[] indices = record.get("indices");
  int[] values = record.get("values"); 
  return new SparseVector(size, indices, values);
}
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+    outputSchema = outputSchema != null ? outputSchema : config.getOutputSchema(input.first().getSchema());
+    List<Column> requiredFields = new ArrayList<>();
+    for (String field : fieldList) {
+      requiredFields.add(new Column(field));
+    }
+    JavaRDD<Row> transformedRDD =
+      javaSparkContext.parallelize(result.select(JavaConversions.asScalaBuffer(requiredFields).toSeq())
+                                     .collectAsList());
+    JavaRDD<StructuredRecord> output = transformedRDD.map(new Function<Row, StructuredRecord>() {
```
albertshau: All this logic looks much more complex than it needs to be. You can get rid of all the stuff with DataFrames and parallelize and instead do something like:

```
...
    final Map<String, String> mapping = config. getFeatureListMapping();
    final Splitter splitter = config.getSplitter();

    return input.map(new Function<StructuredRecord, StructuredRecord>() {
      @Override
      public StructuredRecord call(StructuredRecord input) throws Exception {
        StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
        for (Schema.Field field : input.getSchema().getFields()) {
          builder.set(fieldName, input.get(fieldName));
        }
        for (Map.Entry<String, String> mapEntry : mapping.entrySet()) {
          String inputField = mapEntry.getKey();
          String outputField = mapEntry.getValue();
          String text = input.get(inputField);
          SparseVector vector = text == null ?
            Vectors.sparse(0, new int[], new double[]) : tf.transform(Lists.newArrayList(splitter.split(text)));
          builder.set(outputField, VectorUtils.toRecord(vector);
        }
        return builder.build();
      }
    });
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+  public static class HashingTFConfig extends PluginConfig {
+    @Nullable
+    @Description("The number of features to use in training the model. It must be of type integer. The default value " +
+      "if none is provided will be 100.")
+    private final Integer numFeatures;
+
+    @Description("A comma-separated list of the input fields to map to the transformed output fields. The key " +
+      "specifies the name of the field to generate feature vector from, with its corresponding value specifying the " +
+      "output columns(size, indices and value) to emit the sparse vector.")
+    private final String outputColumnMapping;
```
albertshau: there should be a nullable property for the pattern to split the input string fields on. It should default to split on whitespace `\\s+`

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,188 @@
...
+      this.outputColumnMapping = outputColumnMapping;
+    }
+
+    private Map<String, String> getFeatureListMapping() {
+      try {
+        Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(outputColumnMapping);
+        return map;
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+          String.format("Invalid categorical feature mapping. %s. Please make sure it is in the format " +
```
albertshau: this exception message is wrong

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramTrainer.java
```diff
@@ -0,0 +1,172 @@
...
+    private String fileSetName;
+
+    @Description("Path of the FileSet to save the model to.")
+    private String path;
+
+    @Description("Field to be used for training.")
+    private String inputCol;
+
+    @Nullable
+    @Description("The dimension of codes after transforming from words. Default 3.")
```
albertshau: Default of 3 seems way too small. A lot of data is lost with a vector size of 3. Why not use Spark's default (100)?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramTrainer.java
```diff
@@ -0,0 +1,172 @@
...
+      if (field == null) {
+        throw new IllegalArgumentException(String.format("Input column %s does not exist in the input schema",
+                                                         inputCol));
+      }
+      Schema schema = field.getSchema();
+      Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+      if (!(type.equals(Schema.Type.STRING) || type.equals(Schema.Type.BOOLEAN) || (type.equals(Schema.Type.ARRAY) &&
+        (schema.getComponentSchema().getType().equals(Schema.Type.STRING) ||
+          schema.getComponentSchema().getType().equals(Schema.Type.BOOLEAN))))) {
+        throw new IllegalArgumentException(String.format("Input column can be of type string, boolean or array of " +
```
albertshau: it doesn't make any sense to allow boolean fields, as this model is for text. It should be a string or nullable string. Don't need to support an array of strings either, should just have an optional split pattern as a property that defaults to whitespace.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramTrainer.java
```diff
@@ -0,0 +1,172 @@
...
+      return;
+    }
+
+    JavaSparkContext javaSparkContext = context.getSparkContext();
+    SQLContext sqlContext = new SQLContext(javaSparkContext);
+    Set<String> inputCols = new LinkedHashSet<>();
+    inputCols.add(config.inputCol);
+    List<StructField> structField = SparkUtils.getStructFieldList(input.first().getSchema().getFields(), inputCols);
+    schema = new StructType(structField.toArray(new StructField[structField.size()]));
+    JavaRDD<Row> rowRDD = SparkUtils.convertJavaRddStructuredRecordToJavaRddRows(input, schema);
```
albertshau: you don't need to use the DataFrame version and do all this conversion. Can just do something like:

```
    final String fieldName = config.inputCol;
    final Splitter splitter = config.getSplitt();
    JavaRDD<Iterable<String>> textRDD = input.map(new Function<StructuredRecord, Iterable<String>>() {
      @Override
      public Iterable<String> call(StructuredRecord input) throws Exception {
        String text = input.get(fieldName);
        return splitter.split(text);
      }
    });
    Word2Vec word2Vec = new Word2Vec()
     .setVectorSize(config.vectorSize)
     ...
    Word2VecModel model = word2Vec.fit(textRDD);
    model.save(...)

```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+    @Description("The name of the FileSet to load the model from.")
+    private final String fileSetName;
+
+    @Description("Path of the FileSet to load the model from.")
+    private final String path;
+
+    @Description("List of the input fields to map to the transformed output fields. The key specifies the name of " +
+      "the field to generate feature vector from, with its corresponding value the output column in which the vector " +
+      "would be emitted.")
+    private final String outputColumnMapping;
```
albertshau: there should be a nullable property for the split pattern to use to split input fields. It should default to whitespace.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+      this.outputColumnMapping = outputColumnMapping;
+    }
+
+    private Map<String, String> getFeatureListMapping() {
+      try {
+        Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(outputColumnMapping);
+        return map;
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+          String.format("Invalid categorical feature mapping. %s. Please make sure it is in the format " +
```
albertshau: wrong exception message

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+
+    List<Column> requiredFields = new ArrayList<>();
+    for (Schema.Field field : outputSchema.getFields()) {
+      requiredFields.add(new Column(field.getName()));
+    }
+
+    JavaRDD<Row> transformedRDD =
+      javaSparkContext.parallelize(result.select(JavaConversions.asScalaBuffer(requiredFields).toSeq())
+                                     .collectAsList());
+    JavaRDD<StructuredRecord> output = transformedRDD.map(new Function<Row, StructuredRecord>() {
```
albertshau: don't need to convert to DataFrame and all this other weird stuff. Collecting things in the driver should pretty much always be avoided. Basically, if anything other than a function on an RDD is happening, you should reconsider and see if it is really the right way to do it.

This can be done by doing something like:

```

    final Map<String, String> mapping = config.getFeatureListMapping();
    final Splitter splitter = config.getSplitter();
    final int vectorSize = model.wordVectors().length / model.wordIndex().size();
    return input.map(new Function<StructuredRecord, StructuredRecord>() {
      @Override
      public StructuredRecord call(StructuredRecord input) throws Exception {
        StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
        for (Schema.Field field : input.getSchema().getFields()) {
          builder.set(fieldName, input.get(fieldName));
        }
        for (Map.Entry<String, String> mapEntry : mapping.entrySet()) {
          String inputField = mapEntry.getKey();
          String outputField = mapEntry.getValue();
          String text = input.get(inputField);

          int numWords = 0;
          double[] vectorValues = new double[vectorSize];
          for (String word : splitter.split(text)) {
            double[] wordVector = model.transform(word).toArray();
            for (int i = 0; i < vectorSize; i++) {
              vectorValues[i] += wordVector[i];
            }
            numWords++;
          }
          for (int i = 0; i < vectorSize; i++) {
            vectorValues[i] /= (double) numWords;
          }
          builder.set(outputField, vectorValues);
        }
        return builder.build();
      }
    });
```

albertshau (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramTrainer.java
ananya-bhattacharya: Agreed. Will remove the boolean types.
And regarding the array of strings:
The array of strings was kept considering scenarios where the user has already converted the data into tokens(array of string - maybe created using the tokenizer plugin), which can be directly used as an input for the text based feature generation.
Please let me know if we can have the field to be of type string, nullable string or array of strings or this needs to be changed to accept only string or nullable string.

albertshau: ok I see, an array makes sense as well. We should be sure to document the types that are supported. Does the tokenizer output an array of strings or an array of nullable strings? If the strings in the array are nullable you might need to filter out the nulls before fitting the model.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
ananya-bhattacharya: `(mllib.feature.Word2VecModel)model.transform(word)`  ( in `double[] wordVector = model.transform(word).toArray();`) requires the word to be present in the vocabulary of the model, otherwise the Word2VecModel throws IllegalStateException exception with message - `<word>` not in vocabulary.
Just wanted to confirm if this was the expected behaviour of the plugin. And, should we catch and ignore such words or throw error to the user?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,157 @@
...
+      this.outputColumnMapping = outputColumnMapping;
+    }
+
+    private Map<String, String> getFeatureListMapping() {
+      try {
+        Map<String, String> map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(outputColumnMapping);
+        return map;
+      } catch (IllegalArgumentException e) {
+        throw new IllegalArgumentException(
+          String.format("Invalid output feature mapping. %s. Please make sure it is in the format " +
```
albertshau: Put the exception message at the end instead of in the middle.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,157 @@
...
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    Preconditions.checkArgument(inputSchema != null, "Input Schema must be a known constant.");
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getOutputSchema(inputSchema));
+    SparkUtils.validateFeatureGeneratorConfig(inputSchema, config.outputColumnMapping);
```
albertshau: its kind of strange for this to be a static call. Seems to make more sense for it to just be a method in the config. In addition, you need to validate that numFeatures is positive, and that the pattern is valid (use Pattern.compile()).

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
```diff
@@ -0,0 +1,173 @@
...
+            double[] wordVector = loadedModel.transform(word).toArray();
+            for (int i = 0; i < vectorSize; i++) {
+              vectorValues[i] += wordVector[i];
+            }
+            numWords++;
+          }
+          for (int i = 0; i < vectorSize; i++) {
+            vectorValues[i] /= (double) numWords;
+          }
+          builder.set(outputField, new ArrayList<>(Arrays.asList(ArrayUtils.toObject(vectorValues))));
```
albertshau: you don't need to make this a List, just set it to the double[].

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
     fields.add(Schema.Field.of(predictionField, Schema.of(Schema.Type.DOUBLE)));
     return Schema.recordOf(inputSchema.getRecordName() + ".predicted", fields);
   }
+
+  /**
+   * Validate config parameters for feature generator classes.
+   *
+   * @param inputSchema         input Schema
```
albertshau: For consistency, one space between param name and description

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+   * @param input      input Structured Record
+   * @param inputField field to use for feature generation
+   * @param pattern pattern to use for splitting the string
+   * @return
+   */
+  static List<String> getInputFieldValue(StructuredRecord input, String inputField, String pattern) {
+    List<String> text = new ArrayList<>();
+    Schema schema = input.getSchema().getField(inputField).getSchema();
+    if (!schema.isNullable() && input.get(inputField) == null) {
+      throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
```
albertshau: You don't need to check for this, StructuredRecord.Builder checks for this.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+    }
+    Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+    if (type.equals(Schema.Type.ARRAY)) {
+      schema = schema.getComponentSchema();
+      if (!schema.isNullable() && ((ArrayList) input.get(inputField)).isEmpty()) {
+        throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
```
albertshau: a null is different than an empty list and should not be treated the same. You also don't need to check that a non-nullable field is not null.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+  static List<String> getInputFieldValue(StructuredRecord input, String inputField, String pattern) {
+    List<String> text = new ArrayList<>();
+    Schema schema = input.getSchema().getField(inputField).getSchema();
+    if (!schema.isNullable() && input.get(inputField) == null) {
+      throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
+    }
+    Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+    if (type.equals(Schema.Type.ARRAY)) {
+      schema = schema.getComponentSchema();
+      if (!schema.isNullable() && ((ArrayList) input.get(inputField)).isEmpty()) {
```
albertshau: can't just cast it to an ArrayList. An array type means it can be a List or an array.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -0,0 +1,157 @@
...
+        throw new IllegalArgumentException(
+          String.format("Invalid output feature mapping. %s. Please make sure it is in the format " +
+                          "'input-column':'transformed-output-column'.", e.getMessage()), e);
+      }
+    }
+
+    private Schema getOutputSchema(Schema inputSchema) {
+      List<Schema.Field> fields = new ArrayList<>(inputSchema.getFields());
+      for (Map.Entry<String, String> entry : getFeatureListMapping().entrySet()) {
+        fields.add(Schema.Field.of(entry.getValue(), Schema.recordOf(
```
albertshau: use VectorUtils.SPARSE_SCHEMA

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramTrainer.java
```diff
@@ -0,0 +1,164 @@
...
+    public void validate(Schema inputSchema) {
+      Schema.Field field = inputSchema.getField(inputCol);
+      if (field == null) {
+        throw new IllegalArgumentException(String.format("Input column %s does not exist in the input schema",
+                                                         inputCol));
+      }
+      Schema schema = field.getSchema();
+      Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+      if (!(type.equals(Schema.Type.STRING) || (type.equals(Schema.Type.ARRAY) &&
+        schema.getComponentSchema().getType().equals(Schema.Type.STRING)))) {
```
albertshau: this misses the case where the column is an array of nullable strings. component schema in that case is a union. Seems like you could just call the common validation method like the others.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
```diff
@@ -0,0 +1,330 @@
...
+    workflowManager.waitForFinish(5, TimeUnit.MINUTES);
+
+    DataSetManager<Table> labeledTexts = getDataset(sink);
+    List<StructuredRecord> structuredRecords = MockSink.readOutput(labeledTexts);
+    Assert.assertEquals(3, structuredRecords.size());
+    List<Double> result;
+    for (StructuredRecord record : structuredRecords) {
+      result = (ArrayList) (record.get("result"));
+      if ((Integer) record.get("offset") == 1) {
+        Assert.assertArrayEquals(new double[]{-0.05517900735139847, -0.13193030655384064, 0.14834064245224},
```
albertshau: are these tests consistent? The model generates the exact same results every time? It seems very brittle.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+  /**
+   * Validate config parameters for feature generator classes.
+   *
+   * @param inputSchema         input Schema
+   * @param outputColumnMapping List of the input fields to map to the transformed output fields.
+   */
+  static void validateFeatureGeneratorConfig(Schema inputSchema, String outputColumnMapping) {
+    Map<String, String> map;
+    try {
+      map = Splitter.on(',').trimResults().withKeyValueSeparator(":").split(outputColumnMapping);
```
albertshau: just pass in the map. The logic shouldn't be in many different places.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+      if (inputSchema.getField(key) == null) {
+        throw new IllegalArgumentException(String.format("Input field %s does not exist in the input schema %s.",
+                                                         key, inputSchema.toString()));
+      }
+      Schema schema = inputSchema.getField(key).getSchema();
+      Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+      Schema.Type componentSchemaType = null;
+      if (type.equals(Schema.Type.ARRAY)) {
+        Schema componentSchema = schema.getComponentSchema();
+        componentSchemaType = componentSchema.isNullable() ? componentSchema.getNonNullable().getType() :
```
albertshau: simpler if you assign this to the 'type' variable at this point. Then you only need to check type below and not type and componentSchemaType.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+          componentSchema.getType();
+      }
+      if (!(type.equals(Schema.Type.STRING) || (type.equals(Schema.Type.ARRAY) &&
+        componentSchemaType.equals(Schema.Type.STRING)))) {
+        throw new IllegalArgumentException(String.format("Field to be transformed should be of type String or " +
+                                                           "Nullable String or Array of type String or Nullable " +
+                                                           "String . But was %s for field %s.", type, key));
+      }
+      if (value == null) {
```
albertshau, line 218: Is this even possible with the splitter? In any case, don't think this method needs to check for it. Should be the responsibility of the caller to make sure the map passed in is valid.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+      if (type.equals(Schema.Type.STRING)) {
+        text = (ArrayList) input.get(inputField);
+      } else {
+        throw new IllegalArgumentException(
+          String.format("Field to be used for text based feaure generation should be of type string, nullable " +
+                          "string or array of strings or nullable string. But was ARRAY of %s for field %s.",
+                        schema.getComponentSchema(), inputField));
+      }
+    } else if (type.equals(Schema.Type.STRING)) {
+      text = Lists.newArrayList(Splitter.on(Pattern.compile(pattern)).split((String) input.get(inputField)));
```
albertshau: should pass in the Splitter to this method instead of compiling a Pattern and creating a Splitter every call, because compiling is expensive.  Plugins should compile it once in their initialize methods.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+    }
+    Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+    if (type.equals(Schema.Type.ARRAY)) {
+      schema = schema.getComponentSchema();
+      if (!schema.isNullable() && ((ArrayList) input.get(inputField)).isEmpty()) {
+        throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
+      }
+      type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+      if (type.equals(Schema.Type.STRING)) {
+        text = (ArrayList) input.get(inputField);
```
albertshau: can't assume its an ArrayList. The array schema type can be an actual array or it can be a List.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+      } else if (inputSchema.getField(value) != null) {
+        throw new IllegalArgumentException(
+          String.format("Output column name %s for column %s is already present in the input schema. Please " +
+                          "provide a different output field name.", value, key));
+      }
+    }
+  }
+
+  /**
+   * Get the valid of input field which is to be used for text based feature generation.
```
albertshau: Gets the input field for feature generation. If the field is an array, returns the field. Otherwise, splits the field based on the given pattern. Assumes the input field is an array of strings or a string.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
albertshau: what happens here if the input value is null?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -197,4 +198,87 @@ static Schema getOutputSchema(Schema inputSchema, String predictionField) {
...
+   *
+   * @param input      input Structured Record
+   * @param inputField field to use for feature generation
+   * @param pattern pattern to use for splitting the string
+   * @return
```
albertshau: document the return value

albertshau (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
ananya-bhattacharya: Directly setting the field to double[] throws `java.io.IOException: Incomplete document`
Therefore, converting double[] to Objects[].
Please let me know if this is okay or if I'm missing anything.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: The Splitter allows the value to be null, if the key is present.
This is used to validate, if a valid (not null) output field name is provided or not, to hold the generated feature vector value.
Please let me know if this is okay or should we throw a runtime exception during StructuredRecord.Builder()

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: Initializing splitter in the initialize method throws : Task is not serializable exception.
Changed the code to create the splitter object only if it is null within the map() function otherwise use splitter.
`splitter = (splitter == null) ? Splitter.on(Pattern.compile(config.pattern)) : splitter`
Please let me know if this approach is okay.

ananya-bhattacharya: Also, regarding input value null : Changed code to split the input, if it is not null.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
ananya-bhattacharya: The test generates different results on multiple runs. However, during testing found that the results always lie in the vicinity of the expected values, and the test works with the delta 0.3.
Please let me know if the this is okay or the test needs to be changed.

albertshau: 0.3 is very large compared to these numbers. I don't think this is actually testing anything worthwhile, especially since the results are based on a 3rd party library that we don't control. 

I would remove this and add more tests for the utility code that extracts words from a StructuredRecord. Anything beyond that point is really just testing the spark ml library.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
```diff
@@ -0,0 +1,330 @@
...
+        Assert.assertEquals(expected, actual);
+      } else {
+        expected = new SparseVector(10, new int[]{0, 2, 4, 5, 8}, new double[]{1.0, 1.0, 1.0, 1.0, 1.0});
+        Assert.assertEquals(expected, actual);
+      }
+    }
+  }
+
+  @Test
+  public void testHashingTFFeatureGeneratorWithArrayString() throws Exception {
```
albertshau: we shouldn't need to create a pipeline and run it in order to test the string array logic. This is a more roundabout, and much more expensive way to test a small, isolated piece of code. Instead, there should be test cases for the utility code, that build input records and just call the static methods.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
albertshau: 0.3 is very large compared to these numbers. I don't think this is actually testing anything worthwhile, especially since the results are based on a 3rd party library that we don't control.

I would remove this and add more tests for the utility code that extracts words from a StructuredRecord. Anything beyond that point is really just testing the spark ml library.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
albertshau: The reason you get that error is that spark is complaining that the plugin is not serializable, because the Splitter is not serializable. Instead, you could just compile the Pattern in initialize and store it, since that is the expensive part, though what you have done is fine as well.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
```diff
@@ -79,11 +81,12 @@ public void initialize(SparkExecutionPluginContext context) throws Exception {
       return context.getSparkContext().emptyRDD();
     }
     outputSchema = outputSchema != null ? outputSchema : config.getOutputSchema(input.first().getSchema());
-    final Map<String, String> mapping = config.getFeatureListMapping();
+    final Map<String, String> mapping = SparkUtils.getFeatureListMapping(config.outputColumnMapping);
```
albertshau: Sorry if I was unclear about this, did not mean to create a utility function to parse the string. It is fine to have the parse method in the individual plugin configs, since you need to give a plugin-specific error message. Just meant you should be passing the Map into the validation method instead of the string.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -248,37 +229,76 @@ static void validateFeatureGeneratorConfig(Schema inputSchema, String outputColu
...
+                                                       key, inputSchema.toString()));
+    }
+    Schema schema = inputSchema.getField(key).getSchema();
+    Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
+    if (type.equals(Schema.Type.ARRAY)) {
+      Schema componentSchema = schema.getComponentSchema();
+      type = componentSchema.isNullable() ? componentSchema.getNonNullable().getType() :
+        componentSchema.getType();
+    }
+    if (!(type.equals(Schema.Type.STRING) || type.equals(Schema.Type.ARRAY))) {
```
albertshau: this should just be `type != Schema.Type.STRING` (you can compare directly for enums). It is not valid for it to be an array. If it is, it means the field is an array of arrays.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -248,37 +229,76 @@ static void validateFeatureGeneratorConfig(Schema inputSchema, String outputColu
...
-      throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
-    }
     Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
     if (type.equals(Schema.Type.ARRAY)) {
       schema = schema.getComponentSchema();
-      if (!schema.isNullable() && ((ArrayList) input.get(inputField)).isEmpty()) {
-        throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
-      }
       type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
       if (type.equals(Schema.Type.STRING)) {
```
albertshau: it's easier to read if you do:

```
if (type != Schema.Type.String)) {
  throw new ...
}

//logic here
```

as it removes one layer of indentation. In any case, I'm not sure if you even need the check. Isn't this validated at configure time?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -248,37 +229,76 @@ static void validateFeatureGeneratorConfig(Schema inputSchema, String outputColu
...
       } else {
         throw new IllegalArgumentException(
-          String.format("Field to be used for text based feaure generation should be of type string, nullable " +
+          String.format("Field to be used for text based feature generation should be of type string, nullable " +
                           "string or array of strings or nullable string. But was ARRAY of %s for field %s.",
                         schema.getComponentSchema(), inputField));
       }
     } else if (type.equals(Schema.Type.STRING)) {
-      text = Lists.newArrayList(Splitter.on(Pattern.compile(pattern)).split((String) input.get(inputField)));
+      if (inputField != null) {
```
albertshau: inputField should never be null. I meant what happens if `input.get(inputField)` is null. Does split throw an exception? Does it return an empty iterable?

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -248,37 +229,76 @@ static void validateFeatureGeneratorConfig(Schema inputSchema, String outputColu
...
     Schema.Type type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
     if (type.equals(Schema.Type.ARRAY)) {
       schema = schema.getComponentSchema();
-      if (!schema.isNullable() && ((ArrayList) input.get(inputField)).isEmpty()) {
-        throw new IllegalArgumentException(String.format("Null value found for non-nullable field %s", inputField));
-      }
       type = schema.isNullable() ? schema.getNonNullable().getType() : schema.getType();
       if (type.equals(Schema.Type.STRING)) {
-        text = (ArrayList) input.get(inputField);
+        text = input.get(inputField);
```
albertshau: Since an array value can be either a List or an array, you need to check what type it is. If it's a String[], you need to convert it to a List.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
```diff
@@ -20,24 +20,22 @@
...
     Schema.recordOf("sparseVector", Schema.Field.of("size", Schema.of(Schema.Type.INT)),
                     Schema.Field.of("indices", Schema.arrayOf(Schema.of(Schema.Type.INT))),
                     Schema.Field.of("vectorValues", Schema.arrayOf(Schema.of(Schema.Type.DOUBLE))));
 
   public static StructuredRecord asRecord(SparseVector vector) {
     return StructuredRecord.builder(SPARSE_SCHEMA)
       .set("size", vector.size())
-      .set("indices", new ArrayList<>(Arrays.asList(ArrayUtils.toObject(vector.indices()))))
-      .set("vectorValues", new ArrayList<>(Arrays.asList(ArrayUtils.toObject(vector.values()))))
+      .set("indices", ArrayUtils.toObject(vector.indices()))
```
albertshau: you can just pass it vector.indices().

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
```diff
@@ -20,24 +20,22 @@
...
+      .set("vectorValues", ArrayUtils.toObject(vector.values()))
```
albertshau: just pass it vector.values().

albertshau (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
albertshau: I don't understand, what part of the code is complaining about incomplete document? StructuredRecord should work fine with a double[].

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
ananya-bhattacharya: Validation were added in the last updated PR on line [#68](https://github.com/caskdata/hydrator-plugins/pull/492/files/0ade175af8e6ea6a61627332bb07cc93468881c2#diff-a4c33ebf066fe93c6a027acac168d929R68).
Validation specific to the model will be done by the model itself.
Please let me know if I'm missing anything.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SkipGramFeatureGenerator.java
ananya-bhattacharya: We are able to build the StructuredRecord using int[] or double[]. But the `StructuredRecordStringConverter`(being called internally by MockSink) on line #311 and `RecordConverter` on line #71 directly try to type cast the value to Object[] using `(Object[]) values` which throws `ClassCastException`. Therefore, converting this to Object before setting it in the structured record.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
ananya-bhattacharya: We are able to build the StructuredRecord using int[] or double[]. But the StructuredRecordStringConverter(being called internally by MockSink) on line #311 and RecordConverter on line #71 directly try to type cast the value to Object[] using (Object[]) values which throws ClassCastException. Therefore, converting this to Object before setting it in the structured record.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
ananya-bhattacharya: We are able to build the StructuredRecord using int[] or double[]. But the StructuredRecordStringConverter(being called internally by MockSink) on line #311 and RecordConverter on line #71 directly try to type cast the value to Object[] using (Object[]) values which throws ClassCastException. Therefore, converting this to Object before setting it in the structured record.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
ananya-bhattacharya: Agreed. Added SparkUtilsFeatureGenerator to test code specific to the utility methods.
Also, since the SprkUtils class is package private, thus creating the test within the package co.cask.hydrator.plugin.spark.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/test/FeatureGeneratorTest.java
ananya-bhattacharya: Agreed. Added SparkUtilsFeatureGenerator to test code specific to the utility methods.
Also, since the SprkUtils class is package private,thus creating the test within the package co.cask.hydrator.plugin.spark.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
ananya-bhattacharya: The getFeatureListMapping() is common for HashingTFFeatureGenerator and SkipGramFeatureGenerator and was being used in 4 different places.
Therefore, shifted the method to utils to avoid redundancy.
And regarding validation, we are now passing the map in the validation method directly line [#147](https://github.com/caskdata/hydrator-plugins/pull/492/files/0ade175af8e6ea6a61627332bb07cc93468881c2#diff-a4c33ebf066fe93c6a027acac168d929R147).

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: Agreed. Changed the logic to remove extra indentation.
And regarding the check: 
The utility method can be used later, by other methods or plugins which do not validate the type during configurePipeline().
The validation present here is only to check if the type is string, nullable string or array of string or nullable string before converting and returning them as List<String>.
Please let me know if this is okay or we we still need to remove this.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
ananya-bhattacharya: Okay. Changed code to split string only if `input.get(inputField)` is not null.
Splitter throws NullPointer exception in case `input.get(inputField)` is null. 
However, in this case the plugin would return an empty list if the value is null.
Please let me know if this approach is okay.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
albertshau: It is not related to spark though. it is actually just a single line that parses a string into a map, wrapped around some error message.  In general, we want to avoid utility methods wherever possible, as they tend to indicate logic is not in the correct place. I don't see much value in this method, as the error message should be specific to the plugin that uses it.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
albertshau: but then it is a programming bug if somebody calls this without performing the right checks at configure time.  I think we should add to the javadoc what types are supported and mention that it should be checked at configure time.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/HashingTFFeatureGenerator.java
albertshau: right, the comment still exists because this line didn't change. no action needed here.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
albertshau: that is fine, though you should add to the javadocs that this returns an empty list if the value is null.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
albertshau: ok I see. This is a bug (opened https://issues.cask.co/browse/HYDRATOR-1062 to fix it). You shouldn't need this workaround once it is merged.

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * Unit test for {@Link SparkUtils}
+ */
+public class SparkUtilsFeatureGeneratorTest {
```
albertshau: rename to SparkUtilsTest, since it contains more than just stuff required for feature gen

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+import com.google.common.base.Splitter;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * Unit test for {@Link SparkUtils}
```
albertshau: lowercase l. L -> l

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+public class SparkUtilsFeatureGeneratorTest {
+  @Test
+  public void testInputFieldStringForFeatureGeneration() throws Exception {
```
albertshau: testGetStringInputField is a better name, since its not necessarily just feature gen

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    record = StructuredRecord.builder(input).set("offset", 2).set("body", "Classes in Java").build();
+    expected = new ArrayList<>();
+    expected.add("Classes");
+    expected.add("in");
+    expected.add("Java");
+    Assert.assertEquals(expected, SparkUtils.getInputFieldValue(record, "body", splitter));
+  }
+
+  @Test
+  public void testInputFieldArrayForFeatureGeneration() throws Exception {
```
albertshau: testGetArrayInputField

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+      .build();
+    expected = new ArrayList<>();
+    expected.add("Classes");
+    expected.add("in");
+    expected.add("Java");
+    Assert.assertEquals(expected, SparkUtils.getInputFieldValue(record, "body", splitter));
+  }
+
+  @Test
+  public void testPatternFeatureGeneration() throws Exception {
```
albertshau: testSplitterGetInputFieldValue

spark-plugins/src/test/java/co/cask/hydrator/plugin/spark/SparkUtilsFeatureGeneratorTest.java
```diff
@@ -0,0 +1,145 @@
...
+    expected.add("Hi heard about Spark");
+    Assert.assertEquals(expected, SparkUtils.getInputFieldValue(record, "body", splitter));
+    record = StructuredRecord.builder(input).set("offset", 2).set("body", "Classes in Java").build();
+    expected = new ArrayList<>();
+    expected.add("Classes in Java");
+    Assert.assertEquals(expected, SparkUtils.getInputFieldValue(record, "body", splitter));
+  }
+
+  @Test
+  public void testContinuousSpacePatternFeatureGeneration() throws Exception {
```
albertshau: testWhitespaceSplitterGetInputFieldValue

albertshau (changes requested):
changes lgtm. Let's wait until https://github.com/caskdata/cdap/pull/7029 is merged, then use the double[] and int[].

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/VectorUtils.java
ananya-bhattacharya: The fix for bug HYDRATOR-1062 ensures that the unit test cases run successfully for StructuredRecord using int[] or double[]. 
However, since RecordConverter still tries to directly type cast value to Object[]  : `Object[] valuesArr = (Object[]) values;` ([line #71](https://github.com/caskdata/hydrator-plugins/blob/aa1e4caafe2b603bffe7714125b0e421d22294a4/hydrator-common/src/main/java/co/cask/hydrator/common/RecordConverter.java#L71) ), and the sinks like TPFSAvro internally call the RecordConverter, the pipelines for the feature generator plugins are still failing on UI with ClassClastException.

albertshau:
I see. That is a separate bug. Can you open a jira with the details and fix it in a separate PR? It should be pretty simple and similar to the fix for HYDRATOR-1062. Something like:

```
      int length = Array.getLength(values);
      output = Lists.newArrayListWithCapacity(length);
      for (int i = 0; i < length; i++) {
        output.add(convertField(Array.get(values, i), elementSchema));
      }
```

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/SparkUtils.java
```diff
@@ -265,49 +268,24 @@ static void validateTextField(Schema inputSchema, String key) {
...
+        if (value != null) {
+          text = Lists.newArrayList(splitter.split(value));
+        }
       }
-    } else {
-      throw new IllegalArgumentException(
-        String.format("Field to be used for text based feature generation should be of type string, nullable " +
-                        "string or array of strings or nullable string. But was of type %s for field %s.", type,
-                      inputField));
+    } catch (ClassCastException e) {
```
albertshau, line 285: how come this is required? I don't think there should be any class cast problems unless the values don't match their schemas? But if that is the case, the error message will be confusing.

ananya-bhattacharya:
Raised PR https://github.com/caskdata/hydrator-plugins/pull/503 to resolve type cast issues on UI with sinks like TPFSAvro .

albertshau:
lgtm

PR #493, opened 2016-10-21
[HYDRATOR-1002] Converted Rows limit input widget type from number to textbox in excel plugin.

romy-khetan:


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -493,6 +493,10 @@ public void validate() {
         throw new IllegalArgumentException("Value for Table Expiry period is valid only when file tracking table " +
                                              "is specified. Please specify file tracking table name.");
       }
+
+      if (!containsMacro("rowsLimit") && !StringUtils.isNumeric(rowsLimit)) {
```
abhinavbansal246: This might fail as isNumeric returns true for empty string based on what version of apache commons you are using.

abhinavbansal246: Add a unit test for this check

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -493,6 +493,10 @@ public void validate() {
...
+        throw new IllegalArgumentException("Invalid row limit. The value should be greater than zero.");
+      }
     }
```
abhinavbansal246, line 502: What happens if user gives string as a runtime argument using macro?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputReader.java
```diff
@@ -493,6 +493,10 @@ public void validate() {
         throw new IllegalArgumentException("Value for Table Expiry period is valid only when file tracking table " +
                                              "is specified. Please specify file tracking table name.");
       }
+
+      if (!containsMacro("rowsLimit") && !StringUtils.isNumeric(rowsLimit)) {
+        throw new IllegalArgumentException("Invalid row limit. The value should be greater than zero.");
```
abhinavbansal246: incorrect message should be something like this
String.format("Invalid row limit, numeric value expected. value provided = {}", rowLimit)

abhinavbansal246 (changes requested):

romy-khetan:
@abhinavbansal246 
Updated PR with review comments. Please review.

abhinavbansal246:
LGTM

abhinavbansal246:
Please squash the commits, Paste last 20 lines of ITN and verify by running on a cluster

romy-khetan:
@abhinavbansal246
Squashed the commits. However, due to FileSet issues in Excel Reader plugin, tests including ITN tests has been ignored until the issue is resolved. Although verified the fix through creating pipeline in the cluster using UI.

abhinavbansal246 (approved):

PR #496, opened 2016-10-26
Fixing the directory structure for solr search plugins

abhinavbansal246:


abhinavbansal246:
https://issues.cask.co/browse/HYDRATOR-1033

PR #499, opened 2016-10-28
Fixing generic fields visible on UI - Decision Tree Regression[HYDRATOR-1060]

ananya-bhattacharya:
Replacing all occurrences of `featuresToInclude` and `featuresToExclude` with`featureFieldsToInclude` and `featureFieldsToExclude` (field names as present in the common spark ml base class). 
Changed the field values in widget jsons, docs and in description in the classes.


abhinavbansal246:
LGTM

abhinavbansal246:
http://builds.cask.co/browse/HYP-BAD160

PR #500, opened 2016-10-28
HYDRATOR-1061 Do not exceed Travis resources

wolf31o2:
Travis container infrastructure has 4G of memory. We're allocating 6G. Also, we're exiting on OOM using `kill -9` on the process, causing builds to fail in a way that hides the actual problem.

Fixes HYDRATOR-1061

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ language: java
 jdk:
   - oraclejdk7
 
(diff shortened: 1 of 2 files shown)


sreevatsanraman:
LGTM

wolf31o2:
Waiting for the build to pass since this is a build fix. :-)

PR #501, opened 2016-10-31
Fixing XMLRecordReader to avoid divide by zero exception

CuriousVini:
XMLRecordReader was using `InputStream.available()` method which does not guarantee to return file size. The effect was causing `InputStream.available()` to return zero for MapR 4.1. This PR should fix this issue.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
@@ -77,7 +77,7 @@ public class XMLRecordReader extends RecordReader<LongWritable, Map<String, Stri
     XMLInputFactory factory = XMLInputFactory.newInstance();
     FSDataInputStream fdDataInputStream = fs.open(file);
     inputStream = new TrackingInputStream(fdDataInputStream);
-    availableBytes = inputStream.available();
+    availableBytes = split.getLength();
     try {
       reader = factory.createXMLStreamReader(inputStream);
     } catch (XMLStreamException exception) {


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
```diff
@@ -77,7 +77,7 @@ public XMLRecordReader(FileSplit split, Configuration conf) throws IOException {
     XMLInputFactory factory = XMLInputFactory.newInstance();
     FSDataInputStream fdDataInputStream = fs.open(file);
```
bdmogal, line 78: unrelated to this PR, but this should be closed, right?

anwar6953:
In the `getProgress` method, can we also replace `inputStream.getTotalBytes()` with `fdDataInputStream.getPos()`?
It's not an issue, but seems unnecessary to have to have a TrackingInputStream when the getPos method is available.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
anwar6953: I think its closed when the xmlReader is closed (in the close method of the XMLRecordReader):
https://github.com/caskdata/hydrator-plugins/blob/5afa1ea0ac1e5de6f623a5adb21e7a910fa42eda/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java#L105

CuriousVini:
@anwar6953 We are using `XMLStreamReader#next()` to read data from file, instead of using `FSDataInputStream` directly. Looks like thats why we are keeping track of `totalBytes` read using `TrackingInputStream`. Current implementation looks right in that case.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
bdmogal: that closes the reader, not the `FSDataInputStream`, right?

chtyim:
Who wrote this class? It is not great at so many places. For one, the initialization of record reader should be done in the initialize method, not the constructor.

CuriousVini:
@chtyim @bdmogal I will create separate jira for the points mentioned above.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
anwar6953: I think the reader internally closes the InputStream that it was built with.

anwar6953:
Also add the point of https://github.com/caskdata/hydrator-plugins/pull/501#issuecomment-257432894 to the JIRA. The existing code to get bytes read is probably correct, but does unnecessary tracking.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLRecordReader.java
bdmogal: Likely. I didn't notice that the reader was created using this input stream.

bdmogal: Perhaps just add a comment.

CuriousVini:
@anwar6953 The `RecordReader` inteface requires implementation of `getProgress()`. Not sure  what do you mean by unnecessary tracking?

CuriousVini:
Created jira: https://issues.cask.co/browse/HYDRATOR-1065 to addressed points raised above

anwar6953:
LGTM

PR #502, opened 2016-11-03
Adding Kinesis batch sink HYDRATOR-235

abhinavbansal246:
Adding Kinesis batch sink. 
This will work for both batch and realtime.
This will replace the existing deprecated RealTimeKinesisSink


abhinavbansal246:
https://issues.cask.co/browse/HYDRATOR-235

core-plugins/widgets/KinesisSink-batchsink.md
```diff
@@ -0,0 +1,42 @@
...
+
+**distribute:** Boolean to decide if the data has to be sent to a single shard or has to be uniformly distributed
+among all the shards. Default value is true
+
+**shardCount:** Number of shards to be created, each shard has input of 1mb/s. Default value is 1.
+
+Example
+-------
+This example will write to a kinesis stream named 'MyKinesisStream'. The kinesis stream will be created if it does not
+exists already. Each record it receives will be written as a single stream event. @ shards for this stream will be
```
anwar6953: What's the `@` for?

anwar6953: Oh, I think you meant `2`.

anwar6953: Rather, use `Two`.

core-plugins/widgets/KinesisSink-batchsink.md
```diff
@@ -0,0 +1,42 @@
...
+Properties
+----------
+**name:** The name of the kinesis stream to output to. Must be a valid kinesis stream name. The kinesis stream
+will be created if it does not exist.
+
+**accessID:** The access Id provided by AWS required to access the kinesis streams.
+
+**accessKey:** AWS access key secret having access to Kinesis streams.
+
+**distribute:** Boolean to decide if the data has to be sent to a single shard or has to be uniformly distributed
```
anwar6953: nit - just for consistency, end the description w/ a period

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -0,0 +1,117 @@
...
+    inputTable.put(put3);
+    inputManager.flush();
+
+    long startTs = System.currentTimeMillis() / 1000;
+
+    MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+    mrManager.start();
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
+
+    long endTs = System.currentTimeMillis() / 1000;
```
anwar6953: this seems incomplete. What's the endTs used for?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -0,0 +1,117 @@
...
+    ETLStage sink = new ETLStage("KinesisSinkTest", sinkConfig);
+    ETLBatchConfig etlConfig = ETLBatchConfig.builder("* * * * *")
+      .setSource(source)
+      .addSink(sink)
+      .addConnection(source.getName(), sink.getName())
+      .build();
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(ETLBATCH_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "testKinesis");
+    // This will throw illegal argument exception because of invalid keys
```
anwar6953: Is this comment wrong?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -0,0 +1,117 @@
...
+import co.cask.hydrator.plugin.common.Properties;
+import com.google.common.collect.ImmutableMap;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Test for Kinesis configuration
```
anwar6953: seems its testing more than just configuration

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,152 @@
...
+      sendKinesisRecord(buffer, putRecordRequest);
+    }
+
+    private void sendKinesisRecord(ByteBuffer data, PutRecordRequest putRecordRequest) {
+
+      putRecordRequest.setData(data);
+      try {
+        kinesisClient.putRecord(putRecordRequest);
+      } catch (ProvisionedThroughputExceededException ex) {
+        LOG.debug("Throughput exceeded. Sleeping for 10 ms and retrying");
```
anwar6953: Where is the retry happening?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,152 @@
...
+                                                          conf.get(Properties.KinesisRealtimeSink.ACCESS_KEY));
+
+    if (kinesisClient == null) {
+      LOG.info("Creating amazon kinesis client for stream {}", Properties.KinesisRealtimeSink.NAME);
+      kinesisClient = new AmazonKinesisClient(awsCred);
+      try {
+        KinesisUtil.createAndWaitForStream(kinesisClient, conf.get(Properties.KinesisRealtimeSink.NAME),
+                                           Integer.valueOf(conf.get(Properties.KinesisRealtimeSink.SHARD_COUNT)));
+      } catch (AmazonServiceException e) {
+        throw new IllegalArgumentException("AWS credentials are not correct", e);
```
anwar6953: seems like we're giving a potentially misleading/inaccurate error message here.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
abhinavbansal246: Since we can not check in the api keys for aws accounts this test contains invalid keys and will prevent the code to create kinesis client

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -126,17 +131,27 @@ public void write(NullWritable key, Text value) throws IOException, InterruptedE
...
-        Thread.sleep(10);
-      } catch (InterruptedException e) {
-        LOG.info("Retry wait interrupted", e);
+      boolean done = false;
+      int retry = 0;
+      while (!done) {
+        putRecordRequest.setData(data);
+        try {
+          kinesisClient.putRecord(putRecordRequest);
+          done = true;
```
anwar6953: why need to set `done = true`? Can you just `return`?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -126,17 +131,27 @@ public void write(NullWritable key, Text value) throws IOException, InterruptedE
...
+        } catch (ProvisionedThroughputExceededException ex) {
+          retry++;
+          if (retry < MAX_RETRIES) {
+            LOG.debug("Throughput exceeded. Sleeping for 10 ms and retrying. Try # {}", retry + 1);
+          } else {
+            LOG.error("Maximum retries exhausted");
```
anwar6953: I think you should also log the `ex`. Pass it as the second parameter.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -76,7 +77,11 @@ public void checkOutputSpecs(JobContext context) throws IOException, Interrupted
         KinesisUtil.createAndWaitForStream(kinesisClient, conf.get(Properties.KinesisRealtimeSink.NAME),
                                            Integer.valueOf(conf.get(Properties.KinesisRealtimeSink.SHARD_COUNT)));
       } catch (AmazonServiceException e) {
-        throw new IllegalArgumentException("AWS credentials are not correct", e);
+        if (e.getStatusCode() == 400 && e.getErrorCode().equalsIgnoreCase("UnrecognizedClientException")) {
```
anwar6953: does `e.getCause()` return anything meaningful,useful?

Do you have the full stacktrace that this would print, as an example?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -81,37 +77,13 @@ public void test() throws Exception {
...
-    inputTable.put(put3);
-    inputManager.flush();
-
-    long startTs = System.currentTimeMillis() / 1000;
+    ApplicationManager appManager = deployApplication(appId, appRequest);
 
     MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+    // Throws illegal argument exception because of invalid AWS keys
     mrManager.start();
     mrManager.waitForFinish(5, TimeUnit.MINUTES);
```
anwar6953, line 90: can you assert that the MR failed?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -137,13 +137,13 @@ private void sendKinesisRecord(ByteBuffer data, PutRecordRequest putRecordReques
...
-          done = true;
+          return;
         } catch (ProvisionedThroughputExceededException ex) {
           retry++;
           if (retry < MAX_RETRIES) {
             LOG.debug("Throughput exceeded. Sleeping for 10 ms and retrying. Try # {}", retry + 1);
           } else {
-            LOG.error("Maximum retries exhausted");
+            LOG.error("Maximum retries exhausted", ex);
             done = true;
```
anwar6953: can we return here also? Completely remove the `done` variable?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(ETLBATCH_ARTIFACT, etlConfig);
+    Id.Application appId = Id.Application.from(Id.Namespace.DEFAULT, "testKinesis");
+
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+
+    MapReduceManager mrManager = appManager.getMapReduceManager(ETLMapReduce.NAME);
+    // Throws illegal argument exception because of invalid AWS keys
+    mrManager.start();
+    mrManager.waitForFinish(5, TimeUnit.MINUTES);
```
anwar6953, line 90: can you assert that the MR status is failed?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -84,6 +86,6 @@ public void test() throws Exception {
     // Throws illegal argument exception because of invalid AWS keys
     mrManager.start();
     mrManager.waitForFinish(5, TimeUnit.MINUTES);
-
+    Assert.assertEquals(mrManager.getHistory().get(0).getStatus(), ProgramRunStatus.FAILED);
```
anwar6953, line 92: just as a sanity check, assert that there's only 1 record in history first.
Otherwise, you might get a NoSuchElementException, if list size is 0, and you try to access the 0th index.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,166 @@
...
+      // single shard
+      String dist = distribute ? String.valueOf(Math.random()) : String.valueOf(distribute);
+      putRecordRequest.setPartitionKey(dist);
+      ByteBuffer buffer = ByteBuffer.wrap(value.toString().getBytes());
+      sendKinesisRecord(buffer, putRecordRequest);
+    }
+
+    private void sendKinesisRecord(ByteBuffer data, PutRecordRequest putRecordRequest) {
+      int retry = 0;
+      while (true) {
```
anwar6953: I have a feeling that a `for (int i = 0; i < MAX_RETRIES; i++) {` loop will look safer here.
I know this while loop will only iterate up to MAX_RETRIES times, but upon first glance, it can look like there's a risk of infinite-while loop here.

core-plugins/docs/KinesisSink-batchsink.md
```diff
@@ -0,0 +1,42 @@
...
+Properties
+----------
+**name:** The name of the kinesis stream to output to. Must be a valid kinesis stream name. The kinesis stream will be
+created if it does not exist.
+
+**accessID:** The access Id provided by AWS required to access the kinesis streams.
+
+**accessKey:** AWS access key secret having access to Kinesis streams.
+
+**distribute:** Boolean to decide if the data has to be sent to a single shard or has to be uniformly distributed among
```
anwar6953: I think you should reorder the two choices, since true indicates that it will be uniformly distributed, so I think that should be mentioned first.
I think it can be confusing otherwise.

core-plugins/docs/KinesisSink-batchsink.md
```diff
@@ -0,0 +1,42 @@
...
+all the shards. Default value is true.
+
+**shardCount:** Number of shards to be created, each shard has input of 1mb/s. Default value is 1.
```
anwar6953: it's actually 1 MB/s. There's a difference.

anwar6953: Also, we should probably have the default to be much higher, so as to support the common use case by default.
100? 1000?
Does this have to be defined on the server side (for instance, in the Amazon Kinesis stream configuration)?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KinesisSink.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ *
+ */
+@Plugin(type = BatchSink.PLUGIN_TYPE)
+@Name("KinesisSink")
+@Description("Sink that outputs to a specified AWS Kinesis stream.")
+public class KinesisSink extends ReferenceBatchSink<StructuredRecord, NullWritable, Text> {
+
+  private final KinesisConfig config;
```
anwar6953: put non-static below static fields.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/KinesisSink.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    public Integer getShardCount() {
+      return shardCount == null ? DEFAULT_SHARD_COUNT : shardCount;
+    }
+
+
+    public String getDistribute() {
+      //ensuring that distribute can have only 2 possible values "true" or "false". defaults to true
+      return String.valueOf("true".equals(distribute));
```
anwar6953: I think its cleaner to do Boolean.toString(Boolean.parseBoolean(distribute)).
That way, you're not having the string "true" in your code.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,160 @@
...
+
+  private class KinesisRecordWriter extends RecordWriter<NullWritable, Text> {
+    private final AmazonKinesisClient kinesisClient;
+    private final TaskAttemptContext context;
+    private final boolean distribute;
+
+    KinesisRecordWriter(AmazonKinesisClient kinesisClient, TaskAttemptContext context) {
+      this.kinesisClient = kinesisClient;
+      this.context = context;
+      distribute = "true".equals(context.getConfiguration().get(Properties.KinesisRealtimeSink.DISTRIBUTE));
```
anwar6953: I think you should use `context.getconfiguration().getBoolean`

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,160 @@
...
+   * @param context information about the job
+   * @throws IOException when output should not be attempted
+   */
+  @Override
+  public void checkOutputSpecs(JobContext context) throws IOException, InterruptedException {
+    Configuration conf = context.getConfiguration();
+    BasicAWSCredentials awsCred = new BasicAWSCredentials(conf.get(Properties.KinesisRealtimeSink.ACCESS_ID),
+                                                          conf.get(Properties.KinesisRealtimeSink.ACCESS_KEY));
+
+    if (kinesisClient == null) {
```
anwar6953, line 70: What's the point of this check? To make multiple calls to this method a no-op?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,160 @@
...
+      LOG.info("Creating amazon kinesis client for stream {}", Properties.KinesisRealtimeSink.NAME);
+      kinesisClient = new AmazonKinesisClient(awsCred);
+      try {
+        KinesisUtil.createAndWaitForStream(kinesisClient, conf.get(Properties.KinesisRealtimeSink.NAME),
```
anwar6953: This method simply returns if the stream already exists? If so, does the shard count take effect or no?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,160 @@
...
+     * Writes a key/value pair.
+     *
+     * @param key   the key to write.
+     * @param value the value to write.
+     * @throws IOException
+     */
+    @Override
+    public void write(NullWritable key, Text value) throws IOException, InterruptedException {
+      PutRecordRequest putRecordRequest = new PutRecordRequest();
+      putRecordRequest.setStreamName(context.getConfiguration().get(Properties.KinesisRealtimeSink.NAME));
```
anwar6953: put the stream name in a class field, since it wont change each call. (no need to get the property each time).

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/sink/KinesisSinkTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Test for Kinesis batch sink
+ */
+public class KinesisSinkTest extends ETLBatchTestBase {
+
+
+
+  @Test
+  public void test() throws Exception {
```
anwar6953: add a comment or rename the test method to briefly describe what you're testing

core-plugins/docs/KinesisSink-batchsink.md
abhinavbansal246: Kinesis  service charges by the shard hour, IMO it is better to keep default as 1 and scale up if user needs it explicitly.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
abhinavbansal246: The check is to make sure that the kinesis client is created only once

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
abhinavbansal246: No, If the stream already exists the number of shards will not get modified.

anwar6953: I think we should update our documentation to reflect this.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -0,0 +1,160 @@
...
+
+    if (kinesisClient == null) {
+      LOG.info("Creating amazon kinesis client for stream {}", Properties.KinesisRealtimeSink.NAME);
+      kinesisClient = new AmazonKinesisClient(awsCred);
+      try {
+        KinesisUtil.createAndWaitForStream(kinesisClient, conf.get(Properties.KinesisRealtimeSink.NAME),
+                                           Integer.valueOf(conf.get(Properties.KinesisRealtimeSink.SHARD_COUNT)));
+      } catch (AmazonServiceException e) {
+        if (e.getStatusCode() == 400 && e.getErrorCode().equalsIgnoreCase("UnrecognizedClientException")) {
+          throw new IllegalArgumentException("AWS credentials are not correct", e);
```
anwar6953: is the original exception not containing enough information? Or why do we need to wrap it?

anwar6953:
I suspect the `distribute` config property is actually defaulting to false, whereas docs says it defaults to true. Can you check that?

abhinavbansal246:
`distribute` is `true` by default

core-plugins/src/main/java/co/cask/hydrator/plugin/common/KinesisOutputFormat.java
```diff
@@ -73,16 +72,8 @@ public void checkOutputSpecs(JobContext context) throws IOException, Interrupted
     if (kinesisClient == null) {
       LOG.info("Creating amazon kinesis client for stream {}", Properties.KinesisRealtimeSink.NAME);
       kinesisClient = new AmazonKinesisClient(awsCred);
-      try {
         KinesisUtil.createAndWaitForStream(kinesisClient, conf.get(Properties.KinesisRealtimeSink.NAME),
```
anwar6953: dedent

anwar6953:
https://github.com/caskdata/hydrator-plugins/pull/502#issuecomment-265632706
You've verified that (deploying a pipeline and checking the value)?

anwar6953:
LGTM.
Its worth testing this plugin on a cluster again, since there's been quite a few changes recently.

PR #503, opened 2016-11-04
[HYDRATOR-1071] Fix type casting error for primitive arrays

ananya-bhattacharya:
Fixing type cast issues when using primitive arrays in sink like TPFSAvro.


albertshau:
lgtm

PR #505, opened 2016-11-08
Adding a FastFilter

russorat:
This adds a filter plugin based in Java that runs much faster than the script version.


russorat:
Closing in favor of new PR.

PR #507, opened 2016-11-10
Tokenizer - Code changes for HYDRATOR-1006 and HYDRATOR-1031

ananya-bhattacharya:
Code changes for
https://issues.cask.co/browse/HYDRATOR-1006 and
https://issues.cask.co/browse/HYDRATOR-1031

### DIFF
--- a/spark-plugins/docs/Tokenizer-sparkcompute.md
+++ b/spark-plugins/docs/Tokenizer-sparkcompute.md
@@ -2,20 +2,21 @@
 
 Description
 -----------
-Tokenization is the process of taking text (such as a sentence)
-and breaking it into individual terms (usually words) on the basis of pattern.
+Tokenization is the process of taking text (such as a sentence) and breaking it into individual terms (usually words) 
+on the basis of pattern.
 
 Tokenizer splits data on the basis of specified pattern and emits the output as string array of tokens.
 
 Use Case
 --------
-User wants to extract the hashtags from the twitter feeds.User would tokenize the words based on space and then can identify the words that start with hashtags.
+User wants to extract the hashtags from the twitter feeds.User would tokenize the words based on space and then can 
+identify the words that start with hashtags.
 
 Properties
 ----------
 **columnToBeTokenized:** Column on which tokenization is to be done.
 
-**pattern:** Pattern for tokenization.
+**patternSeparator:** Pattern separator for tokenization.
 
 **outputColumn:** Output column name for tokenized data.
 
@@ -28,7 +29,7 @@ This example tokenizes "sentence" column into output column "words" using patter
         "type": "sparkcompute",
         "properties": {
             "columnToBeTokenized": "sentence",
-            "pattern": "/",
+            "patternSeparator": "/",
             "outputColumn": "words"
         }
     }
@@ -44,13 +45,12 @@ For example, suppose the tokenizer receives below input records:
     | Spark | Spark /is an engine for /bigdata processing   |
     +=======================================================+
 
-Output schema will contain only a single column "words" having tokenized data in string array form:
+Output schema will contain additional column "words" having tokenized data in string array form:
 
-    +===============================================+
-    | words                                         |
-    +===============================================+
-    | [hello world, is the, basic application]      |
-    | [hdfs, is a ,file system]                     |
-    | [spark ,is an engine for ,bigdata processing] |
-    +===============================================+
-
+    +=====================================================================================================+
+    | topic | sentence                                     | words                                        |
+    +=====================================================================================================+
+    | java  | Hello world / is the /basic application      | [hello world, is the, basic application]     |
+    | HDFS  | HDFS/ is a /file system                      | [hdfs, is a ,file system]                    |
+    | Spark | Spark /is an engine for /bigdata processing  | [spark ,is an engine for ,bigdata processing]|
+    +=====================================================================================================+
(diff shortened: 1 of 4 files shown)


spark-plugins/docs/Tokenizer-sparkcompute.md
```diff
@@ -2,20 +2,21 @@
...
-and breaking it into individual terms (usually words) on the basis of pattern.
+Tokenization is the process of taking text (such as a sentence) and breaking it into individual terms (usually words) 
+on the basis of pattern.
 
 Tokenizer splits data on the basis of specified pattern and emits the output as string array of tokens.
 
 Use Case
 --------
-User wants to extract the hashtags from the twitter feeds.User would tokenize the words based on space and then can identify the words that start with hashtags.
+User wants to extract the hashtags from the twitter feeds.User would tokenize the words based on space and then can
```
albertshau: Insert a space between 'feeds.' and 'User'

spark-plugins/docs/Tokenizer-sparkcompute.md
```diff
@@ -44,13 +45,12 @@ For example, suppose the tokenizer receives below input records:
...
-    | [hello world, is the, basic application]      |
-    | [hdfs, is a ,file system]                     |
-    | [spark ,is an engine for ,bigdata processing] |
-    +===============================================+
+Output schema will contain additional column "words" having tokenized data in string array form:
 
+    +=====================================================================================================+
+    | topic | sentence                                     | words                                        |
+    +=====================================================================================================+
+    | java  | Hello world / is the /basic application      | [hello world, is the, basic application]     |
```
albertshau: this isn't correct right? Won't the words contain spaces in them?

albertshau: Also, it lowercases everything? Should be sure to document that fact, and perhaps include an option to not lowercase.

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/Tokenizer.java
```diff
@@ -49,80 +46,83 @@
...
-      public Row call(StructuredRecord rec) throws Exception {
-        return RowFactory.create(rec.get(config.columnToBeTokenized));
+      public String call(StructuredRecord structuredRecord) throws Exception {
+        return StructuredRecordStringConverter.toJsonString(structuredRecord);
       }
     });
-    DataFrame sentenceDataFrame = sqlContext.createDataFrame(rowRDD, schema);
-    RegexTokenizer tokenizer = new RegexTokenizer().setInputCol(config.columnToBeTokenized)
-      .setOutputCol(config.outputColumn).setPattern(config.pattern);
+    DataFrame sentenceDataFrame = sqlContext.read().json(javardd);
```
albertshau, line 91: all of this is really overkill to do a simple tokenization. We are converting each record to a json string, converting that RDD into a DataFrame and parsing the json back into columns, tokenizing on the data frame, then transforming it back to json then back to an RDD of records.

Instead, just do something like:

```
  @Override
  public void initialize(SparkExecutionPluginContext context) throws Exception {
    super.initialize(context);
    separator = Pattern.compile(conf.separatorPattern);
  }

  public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext context,
                                             JavaRDD<StructuredRecord> input) throws Exception {
      return input.map(new Function<StructuredRecord, StructuredRecord>() {
        @Override
        public StructuredRecord call(StructuredRecord inputRecord) throws Exception {
          String val = inputRecord.get(conf.columnToBeTokenized);

          StructuredRecord.Builder outputBuilder = StructuredRecord.builder(outputSchema);
          for (Schema.Field field : inputRecord.getSchema().getFields()) {
            outputBuilder.set(field.getName(), inputRecord.get(field.getName()));
          }

          List<String> tokens = new ArrayList<>();
          if (val != null) {
            for (String token : separator.split(val)) {
              tokens.add(token);
            }
          }
          outputBuilder.set(conf.outputColumn, tokens);

          return outputBuilder.build();
        }
      });
  }
```

In fact, you can just do whatever is inside the map function in a Transform instead of a SparkCompute.

albertshau (changes requested):

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/Tokenizer.java
ananya-bhattacharya: Agreed. Using Splitter will reduce the plugin overhead as compared to the Spark Tokenizer API.
However, the Tokenizer plugin was already using the spark API org.apache.spark.ml.feature.RegexTokenizer for converting the string to tokens(as per the design discussions and merged code).
And as spark Tokenizer API requires dataframe for transformation, changed the StructuredRecords into JSONstring to maintain the structure in the dataframe, as per the requirement mentioned in the HYDRATOR-1006 ("Do not drop fields. The output schema should contain all the fields that were in the input schema and not only the column that is being tokenized" - https://issues.cask.co/browse/HYDRATOR-1006)

@albertshau @abhinavbansal246  @sagarkapare 
Should we change this to a transform plugin, using Splitter for tokenization?
Or should we just change the implementation in the current SparkCompute plugin to use Splitter instead of the Spark Tokenizer API?
Please confirm.

albertshau: Yeah I understand why we were changing it to a DataFrame, and I think that may make sense for using more complex Spark libraries that only work on DataFrames, but for something like splitting on a pattern, I don't think it is necessary. I think we should make it a Transform.

nitinmotgi:
Would allow this for now. If anyone has any questions, please talk to me.

PR #510, opened 2016-11-14
CSVParser - Changes for HYDRATOR-1069

ananya-bhattacharya:
Code changes for - https://issues.cask.co/browse/HYDRATOR-1069

### DIFF
--- a/transform-plugins/docs/CSVParser-transform.md
+++ b/transform-plugins/docs/CSVParser-transform.md
@@ -5,12 +5,15 @@ Description
 -----------
 Parses an input field as a CSV Record into a Structured Record. Supports multi-line CSV Record parsing
 into multiple Structured Records. Different formats of CSV Record can be parsed using this plugin.
-Supports these CSV Record types: ``DEFAULT``, ``EXCEL``, ``MYSQL``, ``RFC4180``, and ``TDF``.
+Supports these CSV Record types: ``DEFAULT``, ``EXCEL``, ``MYSQL``, ``RFC4180``, ``TDF`` and ``Custom``.
 
 
 Configuration
 -------------
 **format:** Specifies the format of the CSV Record the input should be parsed as.
+
+**delimiter:** Custom delimiter to be used for parsing the fields. The custom delimiter can only be specified by 
+selecting the option 'Custom' from the format drop-down.
 
 **field:** Specifies the input field that should be parsed as a CSV Record. 
 Input records with a null input field propagate all other fields and set fields that
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
@@ -30,8 +30,6 @@ import co.cask.cdap.etl.api.TransformContext;
 import com.google.common.base.Throwables;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVRecord;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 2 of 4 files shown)


transform-plugins/docs/CSVParser-transform.md
```diff
@@ -5,13 +5,17 @@ Description
 -----------
 Parses an input field as a CSV Record into a Structured Record. Supports multi-line CSV Record parsing
 into multiple Structured Records. Different formats of CSV Record can be parsed using this plugin.
-Supports these CSV Record types: ``DEFAULT``, ``EXCEL``, ``MYSQL``, ``RFC4180``, and ``TDF``.
+Supports these CSV Record types: ``DEFAULT``, ``EXCEL``, ``MYSQL``, ``RFC4180``, and ``TDF``. 
+User also has the option to select custom delimiter to parse the data.
```
abhinavbansal246: just add custom at the end of the list

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -42,28 +40,29 @@
...
+ * by selecting tht option "Other".
  * <ul>
  *   <li>DEFAULT</li>
  *   <li>EXCEL</li>
  *   <li>RFC4180</li>
  *   <li>MYSQL</li>
  *   <li>TDF and</li>
- *   <li>PDL</li>   
+ *   <li>PDL</li>
+ *   <li>Other</li>
```
abhinavbansal246: use "Custom" instead of other

abhinavbansal246:
Add a validation check in configure pipeline for if the delimiter type is other (or custom after recommended change) delimiter should not be null.

ananya-bhattacharya:
We are already validating the custom delimiter in config.validate() [line #296](https://github.com/ananya-bhattacharya/hydrator-plugins/blob/a7044072a95b482fef3f85e2b13c683ad50b92eb/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java#L296) , which is being called inside the configurePipeline() at line [#89](https://github.com/ananya-bhattacharya/hydrator-plugins/blob/a7044072a95b482fef3f85e2b13c683ad50b92eb/transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java#L89).
Please let me know if this is okay or am I missing something.

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -271,15 +282,22 @@ private void validate() {
...
+        !format.equalsIgnoreCase("TDF") && !format.equalsIgnoreCase("PDL") && !format.equalsIgnoreCase("Custom")) {
         throw new IllegalArgumentException("Format specified is not one of the allowed values. Allowed values are " +
-                                             "DEFAULT, EXCEL, MYSQL, RFC4180, PDL & TDF");
+                                             "DEFAULT, EXCEL, MYSQL, RFC4180, PDL, TDF or Custom");
+      }
+
+      if (format.equalsIgnoreCase("Custom") && (delimiter == null || delimiter == 0)) {
+        throw new IllegalArgumentException("Please specify the delimiter for format option 'Custom'.");
+      }
+      if (!format.equalsIgnoreCase("Custom") && delimiter != null && delimiter != 0) {
```
abhinavbansal246, line 299: Why are we checking for 0 here?

ananya-bhattacharya: `delimiter` is a Character. We are checking delimiter equal to zero to check if it is an empty character , since the default value for each char element is '\u0000' (the null character) which is also equal to 0. 
Please let me know if this is okay.

abhinavbansal246:
LGTM
Please squash commits and paste ITN results

abhinavbansal246:
nit: use better detail for PR message
eg. Instead of "CSVParser - Changes for HYDRATOR-1069" use "CSVParser - Adding custom delimiter"

shashankmoghe:
Squashed the commits.There is no ITN for this plugin.We have verified the changes on the cluster.

PR #516, opened 2016-11-18
adding record splitter transform

russorat:
A plugin to take a record delimited by a delimiter and split it into multiple messages. The use case for this would be if someone had a message with n lines, and wanted to split each one into a single record. Could be used to split HL7 messages into individual segments.


transform-plugins/src/main/java/co/cask/hydrator/plugin/RecordSplitter.java
```diff
@@ -0,0 +1,141 @@
...
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Unable to parse output schema.", e);
+    }
+  }
+
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+    try {
+      config.validate();
```
abhinavbansal246: Do we need to validate the config once in configurePipeline() and again here?

transform-plugins/src/main/java/co/cask/hydrator/plugin/RecordSplitter.java
```diff
@@ -0,0 +1,141 @@
...
+  public void transform(StructuredRecord in, Emitter<StructuredRecord> emitter) throws Exception {
+    List<Schema.Field> fields = in.getSchema().getFields();
+    String[] records = String.valueOf(in.get(config.fieldToSplit)).split(config.delimiter);
+    for (String record : records) {
+      StructuredRecord.Builder builder = StructuredRecord.builder(outSchema);
+      for (Schema.Field field : fields) {
+        String name = field.getName();
+        if (outSchema.getField(name) != null && !name.equals(config.fieldToSplit)) {
+          builder.set(name, in.get(name));
+        }
```
abhinavbansal246, line 94: You can move the inner loop out of the outer loop so that you dont have to set all the fields for each split. That way you will be able to reuse the builder

transform-plugins/src/main/java/co/cask/hydrator/plugin/RecordSplitter.java
```diff
@@ -0,0 +1,141 @@
...
+      this.outputField = outputField;
+      this.schema = schema;
+    }
+
+    public void validate() throws IllegalArgumentException {
+      try {
+        Schema outSchema = Schema.parseJson(schema);
+        if (outSchema.getField(outputField) == null) {
+          throw new IllegalArgumentException("Output Schema must contain the specified outputField.");
+        }
```
abhinavbansal246, line 137: Add a check if field to split is present in the input schema too

transform-plugins/widgets/RecordSplitter-transform.json
```diff
@@ -0,0 +1,43 @@
...
+    }
+  ],
+  "outputs": [
+    {
+      "name": "schema",
+      "label": "schema",
+      "widget-type": "schema"
+    }
+  ]
+}
```
abhinavbansal246: Add new line at the end

transform-plugins/src/main/java/co/cask/hydrator/plugin/RecordSplitter.java
russorat: i think it's bad practice to reuse the builder after the object is built.

transform-plugins/src/main/java/co/cask/hydrator/plugin/RecordSplitter.java
russorat: i can't get the input schema here :(

transform-plugins/widgets/RecordSplitter-transform.json
russorat: fixed

russorat:
closing in favor of https://github.com/hydrator/record-splitter-transform/pull/1

PR #518, opened 2016-11-18
Fixing Tokernizer plugin so that spark-plugins are loaded on all the distros

CuriousVini:
Issue: https://issues.cask.co/browse/HYDRATOR-1116
Build: http://builds.cask.co/browse/HYP-BAD172-2

Tested it on CDH 5.4 cluster.

### DIFF
--- a/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/Tokenizer.java
+++ b/spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/Tokenizer.java
@@ -47,7 +47,8 @@ public class Tokenizer extends SparkCompute<StructuredRecord, StructuredRecord> 
   public static final String PLUGIN_NAME = "Tokenizer";
   private Config config;
   private Schema outputSchema;
-  private RegexTokenizer tokenizer;
+  // Work around because of HYDRATOR-1116. Remove this change once its fixed
+  private Object tokenizer;
 
   public Tokenizer(Config config) {
     this.config = config;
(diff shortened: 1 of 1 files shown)


albertshau:
How does this fix the problem?

albertshau:
Oh it will deploy properly, but this plugin will still be broken right?

CuriousVini:
@albertshau The problem was because `RegexTokenizer` was an instance variable of `Tokenizer` class, the spark-plugins jar was not getting loaded  So when `Tokenizer` is loaded  `RegexTokenizer` class was also getting loaded. We changed it to use `Object` and type cast it when needed so the spark-plugins jar is loaded and other tests are not affected on CDH 5.4 and HDP 2.2

anwar6953:
"Oh it will deploy properly, but this plugin will still be broken right?"
Yep. It's just to unblock the other spark plugins which normally work on certains distros (CDH 5.2, CDH 2.4, I believe).

albertshau:
ok I see. lgtm

spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/Tokenizer.java
```diff
@@ -47,7 +47,8 @@
   public static final String PLUGIN_NAME = "Tokenizer";
   private Config config;
   private Schema outputSchema;
-  private RegexTokenizer tokenizer;
+  // Work around because of HYDRATOR-1116. Remove this change once its fixed
+  private Object tokenizer;
```
chtyim: Seems like this doesn't need to be a field at all. Just new it in the RDD `transform` function.

CuriousVini:
@chtyim Addressed comments and tested on cdh 5.4 cluster. Please take another pass when you get a chance.

CuriousVini:
Build is green

PR #520, opened 2016-11-24
[HYDRATOR-1124] Fixed record length calculation for Copybook reader plugin

romy-khetan:


abhinavbansal246:
https://github.com/caskdata/hydrator-plugins/pull/531

PR #522, opened 2016-11-28
HYDRATOR 471 - conditions clause is excluded from the query in get schema request in database source

shashankmoghe:
Tested from UI. No ITN for DbSource.
If use enters $CONDITIONS in the query,though it would be visible in the UI,internally back end logic would exclude the $CONDITIONS clause and execute the manipulated query.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
@@ -123,9 +123,13 @@ public class DBSource extends ReferenceBatchSource<LongWritable, DBRecord, Struc
     try {
       driverCleanup = loadPluginClassAndGetDriver(request, pluginContext);
       try (Connection connection = getConnection(request.connectionString, request.user, request.password)) {
+        String query = request.query;
         Statement statement = connection.createStatement();
         statement.setMaxRows(1);
-        ResultSet resultSet = statement.executeQuery(request.query);
+        if (query.contains("$CONDITIONS")) {
+          query = removeConditionsClause(query);
+        }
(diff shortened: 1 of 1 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -136,6 +140,17 @@ public Schema getSchema(GetSchemaRequest request,
     }
   }
 
+  private static String removeConditionsClause(String importQuerySring) {
+    if (importQuerySring.toUpperCase().contains("WHERE $CONDITIONS AND")) {
```
nitinmotgi: What happens if there are two spaces between $CONDITIONS and AND. Can you please make this robust.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/source/DBSource.java
```diff
@@ -136,6 +140,17 @@ public Schema getSchema(GetSchemaRequest request,
...
+      importQuerySring = importQuerySring.toUpperCase().replace(" $CONDITIONS AND", "");
+    } else if (importQuerySring.toUpperCase().contains("WHERE $CONDITIONS")) {
+      importQuerySring = importQuerySring.toUpperCase().replace(" WHERE $CONDITIONS", "");
+    } else if (importQuerySring.toUpperCase().contains("AND $CONDITIONS")) {
```
nitinmotgi: overall this method is fragile. Could you please make it robust.

nitinmotgi (changes requested):

shashankmoghe:
Closing this PR and will open a new PR against release/1.5

PR #524, opened 2016-11-29
[HYDRATOR-282] - Changing option names to "Tab Delimited" and "Pipe Delimited"

ananya-bhattacharya:
Changing the option name from TDF to "Tab Delimited" and PDL to "Pipe Delimited"  - https://issues.cask.co/browse/HYDRATOR-282


abhinavbansal246:
LGTM

PR #526, opened 2016-12-01
Fixing simple type decoding for python evaluator

abhinavbansal246:
https://issues.cask.co/browse/HYDRATOR-1010

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
@@ -134,6 +134,52 @@ public class PythonEvaluatorTest {
   }
 
   @Test
+  public void testDecodingSimpleTypes() throws Exception {
+    PythonEvaluator.Config config = new PythonEvaluator.Config(
+      "def transform(x, emitter, context):\n" +
+        "  x['floatField'] *= 1.01\n" +
+        "  x['doubleField'] *= 1.01\n" +
+        "  x['longField'] *= 2\n" +
+        "  emitter.emit(x)",
+      null);
+    Transform<StructuredRecord, StructuredRecord> transform = new PythonEvaluator(config);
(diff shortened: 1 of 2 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -229,27 +229,6 @@ public void testComplex() throws Exception {
       Schema.Field.of("inner1", inner1Schema)
     );
 
-    /*
-    {
-      "complex": {
-        "num": 8,
-        "inner1": {
```
gokulavasan, line 236: any reason for removing this comment?

gokulavasan:
@abhinavbansal246 Can you add unit tests to test the scenario mentioned in the JIRA?

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
abhinavbansal246: Since there was no TODO associated with the comment so it was not clear why the comment was there

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -134,6 +134,30 @@ public void testSimple() throws Exception {
   }
 
   @Test
+  public void testDecodingSimpleTypes() throws Exception {
```
gokulavasan, line 137: This is great. Can we add one more test which corresponds to the exact scenario that failed (refer to the json file attached to the JIRA)?

Python Transform Code:
```
def transform(record, emitter, context):
   output = {}  
   output['body'] = record['body']
   emitter.emit(output)
```

gokulavasan:
LGTM 👍 Squash and merge.

PR #533, opened 2016-12-05
HYDRATOR-699 remove use of etlbatch from tests

albertshau:
Replace all usages of old etlbatch app with datapipeline app in
unit tests. Also replacing deprecated Id.Application with new
ApplicationId.


sreevatsanraman (approved):
LGTM

PR #534, opened 2016-12-05
[HYDRATOR-1004] - Ignore non-existing files/folders  in File Source

ananya-bhattacharya:
https://issues.cask.co/browse/HYDRATOR-1004

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
@@ -43,12 +43,16 @@ import com.google.common.collect.Lists;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.lib.input.CombineTextInputFormat;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.lang.reflect.Type;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -66,7 +70,6 @@ import javax.annotation.Nullable;
 @Name("File")
 @Description("Batch source for File Systems")
 public class FileBatchSource extends ReferenceBatchSource<LongWritable, Object, StructuredRecord> {
-
   public static final String INPUT_NAME_CONFIG = "input.path.name";
   public static final String INPUT_REGEX_CONFIG = "input.path.regex";
   public static final String LAST_TIME_READ = "last.time.read";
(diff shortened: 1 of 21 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -169,7 +175,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
     conf.set(CUTOFF_READ_TIME, dateFormat.format(prevHour));
     FileInputFormat.setInputPathFilter(job, BatchFileFilter.class);
-    FileInputFormat.addInputPath(job, new Path(config.path));
+
+    if (!fs.exists(new Path(config.path)) && config.ignoreNonExistingFolders) {
+      path = new Path(fs.getWorkingDirectory() + "/tmp");
+      fs.mkdirs(path);
+      conf.set(INPUT_NAME_CONFIG, path.toUri().getPath());
+      FileInputFormat.addInputPath(job, path);
+      LOG.warn("File/Folder specified in path does not exists. Setting input path to " +
```
russorat: Should you log the warning first, in case something fails when trying to create the /tmp dir?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -169,7 +175,19 @@ public void prepareRun(BatchSourceContext context) throws Exception {
...
+                 "<FileSystem-WorkingDirectory>/tmp.");
```
russorat: Also, you know the working directory, so put the full path to the /tmp dir instead of a placeholder.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -198,6 +216,13 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
       failed.addAll(existing);
       table.write(LAST_TIME_READ, GSON.toJson(failed, ARRAYLIST_DATE_TYPE));
     }
+    try {
+      if (fs.exists(path)) {
+        fs.delete(path, true);
+      }
+    } catch (IOException e) {
+      throw new IllegalArgumentException("Error deleting temporary file.");
```
russorat: You should include the message from the original IOException here.

core-plugins/widgets/FTP-batchsource.json
```diff
@@ -30,6 +30,18 @@
...
           "label": "File System Properties",
           "name": "fileSystemProperties"
+        },
+        {
+          "widget-type": "select",
+          "label": "Ignore Non-Existing Folders",
+          "name": "ignoreNonExistingFolders",
+          "widget-attributes": {
+            "values": [
+              "false",
```
russorat: While we have the values listed both ways in our plugins, let's keep the ordering as true, false instead of false, true. I'll file a jira to fix the other ones. https://issues.cask.co/browse/HYDRATOR-1144

russorat:
few comments, but otherwise looks pretty good.

core-plugins/widgets/FTP-batchsource.json
ananya-bhattacharya: Changed the order in json in the latest commit.

russorat:
@ananya-bhattacharya please make the changes for ordering to all the widget .jsons so that all of them have the ordering true, false

russorat:
LGTM.

PR #536, opened 2016-12-06
HYDRATOR-1153 Fixing the documentation with JSON Parser Plugin

nitinmotgi:


transform-plugins/src/main/java/co/cask/hydrator/plugin/JSONParser.java
```diff
@@ -167,11 +167,12 @@ public void transform(StructuredRecord input, Emitter<StructuredRecord> emitter)
...
   public static class Config extends PluginConfig {
     @Name("field")
-    @Description("Input Field")
+    @Description("Input field to be parsed as JSON")
     private String field;
 
     @Name("mapping")
-    @Description("Mapping complex JSON to fields using JSON Path expressions")
+    @Description("Maps complex JSON to output fields using JSON path expressions. First field defines the output " +
+      "field name and the second field specifies the JSON path expression.")
```
anew: how about adding an example? :)

abhinavbansal246:
LGTM
We should merge it in release/1.5

anew:
I still think this should have en example in the documentation. It is very hard to understand without an example.

anew:
LGTM

PR #543, opened 2016-12-07
HYDRATOR-1160 fix failing orc test

albertshau:


anwar6953:
The plugin has the necessary dependency otherwise defined in the pom file, so that it has access to the classes in distributed/SDK?

If so, LGTM.

Edit: I guess the change which caused this change to be needed (https://issues.cask.co/browse/CDAP-7256) shouldn't have affected distributed, and in order to work there, the necessary dependencies must already have been defined.

albertshau:
I'll test that part out next, but I don't see how it would have worked before if it didn't, since that class wouldn't have been accessible before either I think...

anwar6953:
Right, appended an edit to my original comment.

albertshau:
cool, will double check before merging

albertshau:
http://builds.cask.co/browse/HYP-BAD188

albertshau:
verified it still works on sdk and on cluster.

PR #548, opened 2016-12-13
Add hydrator/wrangler-transform as submodule

wolf31o2:
This adds `hydrator/wrangler-transform` as a submodule, which will be fetched with Git and built along with the rest of this repository when creating CDAP artifacts.

Build: http://builds.cask.co/browse/HYP-BAD191-2

(This is just including it, so build should include it... other failures unrelated)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,21 @@
+#
+# Copyright © 2016 Cask Data, Inc.
+#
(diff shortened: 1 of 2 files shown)


sreevatsanraman:
LGTM. We will branch and update the version before release

PR #554, opened 2016-12-14
HYDRATOR-1163 fix topic partition offset lookup

albertshau:
Fixed a bug where partitions configured to use the latest or
earliest offset were not querying all brokers to find the
information.


sreevatsanraman:
LGTM

PR #555, opened 2016-12-15
Update docs for ignoreNonExistingFolders. It is not nullable.

anwar6953:
Update docs for ignoreNonExistingFolders. It is not nullable/default-able from the plugin perspective. The default is in the UI widget, actually.


albertshau (approved):
lgtm

anwar6953:
Hmm... maybe we should mark it nullable from the plugin perspective, instead...
Edit: This is fine for now - at least clarifying in the documentation.

PR #556, opened 2016-12-15
Making Ignore Non-Existing Folders optional in all file batch batch sources

abhinavbansal246:
The Ignore Non-Existing Folders was not optional config parameter, this might break the older pipelines

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
@@ -264,6 +264,7 @@ public class FileBatchSource extends ReferenceBatchSource<LongWritable, Object, 
     @Macro
     public Long maxSplitSize;
 
+    @Nullable
     @Description("Identify if path needs to be ignored or not, for case when directory or file does not exists. If " +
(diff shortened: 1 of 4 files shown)


sreevatsanraman:
LGTM

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -279,7 +280,7 @@ public FileBatchConfig() {
 
     public FileBatchConfig(String referenceName, String path, @Nullable String fileRegex, @Nullable String timeTable,
                            @Nullable String inputFormatClass, @Nullable String fileSystemProperties,
-                           @Nullable Long maxSplitSize, Boolean ignoreNonExistingFolders) {
```
anwar6953, line 282: hydrator calls this constructor? or the no-arg ctor?

abhinavbansal246: The batch sources that extend filebtach source calls this

anwar6953: LGTM

PR #557, opened 2016-12-15
Update information on Git Submodules

wolf31o2:
Update the information on Git Submodules to make it more succinct and to include `--remote` which will update the submodule to the latest HEAD on the configured branch.


awholegunch:
LGTM

PR #558, opened 2016-12-15
[HYDRATOR-1181] Making format field optional in CSV parser

abhinavbansal246:


russorat:
NPE if format is null in the validate method.

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -272,27 +273,23 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
     @Description("Specifies the schema that has to be output.")
     private final String schema;
 
-    public Config(String format, Character delimiter, String field, String schema) {
-      this.format = format;
+    public Config(@Nullable String format, @Nullable Character delimiter, String field, String schema) {
+      this.format = format == null ? "DEFAULT" : format;
```
albertshau, line 286: are you trying to default format to "DEFAULT"? If so, this is not the way to do it. You have to do it in a no-args constructor. This will throw NPE if you actually try to use it.

albertshau (changes requested):

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
anwar6953: what's this ctor for? only for unit tests?

abhinavbansal246:
https://issues.cask.co/browse/HYDRATOR-1181

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -282,11 +284,19 @@ public Config(@Nullable String format, @Nullable Character delimiter, String fie
...
+        }
+        return;
+      }
+
       // Check if format is one of the allowed types.
       if (!format.equalsIgnoreCase("DEFAULT") && !format.equalsIgnoreCase("EXCEL") &&
         !format.equalsIgnoreCase("MYSQL") && !format.equalsIgnoreCase("RFC4180") &&
         !format.equalsIgnoreCase("Tab Delimited") && !format.equalsIgnoreCase("Pipe Delimited")
-        && !format.equalsIgnoreCase("Custom")) {
+        && !format.equalsIgnoreCase("Custom") && !format.equalsIgnoreCase("PDL") &&
```
awholegunch: style nit: should not the leading `&&` be placed at the end of the previous line?

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
abhinavbansal246: Yes, The plugin in the pipelines is instantiated through injection. Since the fields in the config are final, This ctor initializes them.

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -272,27 +275,31 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
...
       this.schema = schema;
     }
 
     private void validate() {
-      // Check if the format specified is valid.
-      if (format == null || format.isEmpty()) {
-        throw new IllegalArgumentException("Format is not specified. Allowed values are DEFAULT, EXCEL, MYSQL," +
-                                             " RFC4180, Pipe Delimited, Tab Delimited or Custom");
+
+      if (format == null) {
```
albertshau: based on the error message, this should just be
```
if (!"custom".equalsIgnoreCase(format) && delimiter != null)
```

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -272,27 +275,31 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
...
       }
 
       // Check if format is one of the allowed types.
       if (!format.equalsIgnoreCase("DEFAULT") && !format.equalsIgnoreCase("EXCEL") &&
         !format.equalsIgnoreCase("MYSQL") && !format.equalsIgnoreCase("RFC4180") &&
         !format.equalsIgnoreCase("Tab Delimited") && !format.equalsIgnoreCase("Pipe Delimited")
-        && !format.equalsIgnoreCase("Custom")) {
-        throw new IllegalArgumentException("Format specified is not one of the allowed values. Allowed values are " +
-                                             "DEFAULT, EXCEL, MYSQL, RFC4180, Pipe Delimited, Tab Delimited or Custom");
+        && !format.equalsIgnoreCase("Custom") && !format.equalsIgnoreCase("PDL") &&
```
albertshau: nit: be consistent with when to start a new line and put the && at the end of the last line

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -272,27 +275,31 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
...
+        !format.equalsIgnoreCase("TDF")) {
+        throw new IllegalArgumentException(String.format("Format %s is not one of the allowed values. Allowed values " +
```
albertshau, line 305: better to keep a set of allowed formats. Then you can check that the format is in the set, and do `Joiner.on(',').join(FORMATS)` here and won't have to worry about forgetting to update this message if a new format is added or removed

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
albertshau: oh but this check already happens below...

albertshau: ok this is duplicating a check that happens later so you don't have to worry about null pointers... Just set the format to "default" in a no-argument constructor and you don't have to worry about it ever being null and don't need this part at all.

albertshau (changes requested):

transform-plugins/src/main/java/co/cask/hydrator/plugin/CSVParser.java
```diff
@@ -62,6 +66,10 @@
   // Output Schema associated with transform output.
   private Schema outSchema;
 
+  static final Set<String> FORMATS = new HashSet<>(Arrays.asList("DEFAULT", "EXCEL", "MYSQL", "RFC4180",
```
albertshau: nit: `private static final Set<String> FORMATS = ImmutableSet.of("DEFAULT", ...)`

albertshau (approved):
lgtm

russorat:
lgtm

PR #559, opened 2016-12-15
Revert "Update docs for ignoreNonExistingFolders. It is not nullable."

abhinavbansal246:
Reverts caskdata/hydrator-plugins#555


anwar6953:
This is because we have actually made the parameter in the config optional now:
https://github.com/caskdata/hydrator-plugins/pull/556

LGTM

PR #571, opened 2016-12-21
[HYDRATOR-1158] JsonParser should not fail if the path has no match

ananya-bhattacharya:
https://issues.cask.co/browse/HYDRATOR-1158


russorat:
LGTM

PR #572, opened 2016-12-22
[HYDRATOR-1141] Field type in Projection should be a drop down

rashig:
JIRA Link:
https://issues.cask.co/browse/HYDRATOR-1141


russorat:
LGTM

PR #573, opened 2016-12-22
Fix develop hydrator-plugin to use 4.1.0-SNAPSHOT cdap version

deepakwadhwani:
Fix develop hydrator-plugin to use 4.1.0-SNAPSHOT cdap version


sreevatsanraman:
LGTM

PR #574, opened 2016-12-22
Fix develop hydrator-plugin to use 4.1.0-SNAPSHOT app parents version

deepakwadhwani:
Fix develop hydrator-plugin to use 4.1.0-SNAPSHOT app parents version


gokulavasan:
LGTM 👍

PR #579, opened 2016-12-29
[HYDRATOR-984] Fixing Bulk file reading functionality in XMLReader plugin

romy-khetan:


russorat:
@romy-khetan please add a unit test to verify the functionality

romy-khetan:
@russorat Added and updated unit tests to search the pattern recursively. Please review.

russorat:
@romy-khetan I believe you have misunderstood the requirements for this issue. I don't see recursive directory searching anywhere in the original issue and i'm not clear on how that would solve the problem where the XMLReader plugin does not accept any regex pattern other than .*  Maybe we can discuss on a call so that we are all on the same page as to what the original issue was.

romy-khetan:
@russorat 
When i started debugging this issue where plugin was not able to read the file when file pattern includes '_' in the input, i was not able to read the file in any of the clusters (although this works fine when tested using junit) when mentioned exact file name in file pattern input widget and directory path in path widget. 

To fix this issue, i provided path object of the input path in the job configurations as new Path(config.path).toString(). After that, i tried to reproduce the issue with the file pattern which includes '_' as the input and i was able to read the file.

Could you please suggest if i should raise a PR with the current fix?

russorat:
@romy-khetan yes please open a new PR with the fix you suggest. could you investigate why that solves the problem?

PR #580, opened 2016-12-29
[HYDRATOR-1149 & HYDRATOR-1151] - File regex in FileBatchSource 

ananya-bhattacharya:
Changes for: 
https://issues.cask.co/browse/HYDRATOR-1151 and 
https://issues.cask.co/browse/HYDRATOR-1149

### DIFF
--- a/core-plugins/docs/AzureBlobStore-batchsource.md
+++ b/core-plugins/docs/AzureBlobStore-batchsource.md
@@ -16,7 +16,7 @@ Properties
 ----------
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
-**path:** The path on Microsoft Azure Blob Storage to use as input.
+**path:** The path on Microsoft Azure Blob Storage to use as input. Also, the path uses Globbing to read files.
 
 **account:** The Microsoft Azure Blob Storage account to use. This is of the form 
 `mystorageaccount.blob.core.windows.net`, where `mystorageaccount` is the Microsoft 
@@ -29,6 +29,8 @@ Must be a valid base64 encoded storage key provided by Microsoft Azure. (Macro-e
 
 **ignoreNonExistingFolders:** Identify if path needs to be ignored or not, for case when directory or file does not
 exists. If set to true it will treat the not present folder as 0 input and log a warning. Default is false.
+
+**recursive:** Boolean value to determine if files are to be read recursively from the path. Default is false.
 
 Example
 -------
@@ -46,6 +48,7 @@ specified directory. This example uses the 'hydrator' container at
             "account": "hydratorstorage.blob.core.windows.net",
             "container": "hydrator",
             "storageKey": "XXXXXEEESSS/YYYY=",
-            "ignoreNonExistingFolders": "false"
+            "ignoreNonExistingFolders": "false",
+            "recursive": "false"
         }
     }
--- a/core-plugins/docs/AzureBlobStore-batchsource.md
+++ b/core-plugins/docs/AzureBlobStore-batchsource.md
@@ -16,7 +16,8 @@ Properties
 ----------
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
-**path:** The path on Microsoft Azure Blob Storage to use as input. Also, the path uses Globbing to read files.
+**path:** The path on Microsoft Azure Blob Storage to use as input. The path uses filename expansion (globbing) to read
+files. (Macro-enabled)
 
 **account:** The Microsoft Azure Blob Storage account to use. This is of the form 
 `mystorageaccount.blob.core.windows.net`, where `mystorageaccount` is the Microsoft 
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/BatchFileFilter.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/BatchFileFilter.java
@@ -21,6 +21,7 @@ import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.conf.Configured;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
@@ -63,6 +64,7 @@ public class BatchFileFilter extends Configured implements PathFilter {
   private String pathName;
   private String lastRead;
   private Date prevHour;
+  private Pattern pattern;
 
   /*
    * dateRangesToRead is an odd length List of Dates. The non terminal elements are tuples of Dates that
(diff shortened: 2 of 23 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -146,4 +148,47 @@ public void testNotPresentFolder() throws Exception {
...
+
+  @Test
+  public void testRecursiveFolders() throws Exception {
+    Map<String, String> sourceProperties = new ImmutableMap.Builder<String, String>()
+      .put(Constants.Reference.REFERENCE_NAME, "TestCase")
+      .put(Properties.File.PATH, "src/test/resources/")
+      .put(Properties.File.FILE_REGEX, "fileBatchSource.*")
+      .put(Properties.File.IGNORE_NON_EXISTING_FOLDERS, "false")
+      .put(Properties.File.FILESYSTEM_PROPERTIES,
+           "{\"mapreduce.input.fileinputformat.input.dir.recursive\": \"true\" }")
```
anew: We should have a Properties.File.RECURSIVE property here, because we don't want expose Hadoop properties here. Recursive read is common enough to justify our own property.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/BatchFileFilter.java
```diff
@@ -73,9 +75,13 @@
   @Override
   public boolean accept(Path path) {
     String filePathName = path.toString();
-    //The path filter will first check the directory if a directory is given
-    if (filePathName.equals(pathName) || filePathName.equals(pathName + "/")) {
-      return true;
+    try {
+      FileSystem fileSystem = path.getFileSystem(new Configuration());
```
anew, line 81: so this will now accept any directory, even directories that cannot match the pattern. That can be significant overhead

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -146,4 +148,47 @@ public void testNotPresentFolder() throws Exception {
     workflowManager.waitForFinish(5, TimeUnit.MINUTES);
     Assert.assertEquals("FAILED", workflowManager.getHistory().get(0).getStatus().name());
   }
+
+  @Test
+  public void testRecursiveFolders() throws Exception {
+    Map<String, String> sourceProperties = new ImmutableMap.Builder<String, String>()
+      .put(Constants.Reference.REFERENCE_NAME, "TestCase")
+      .put(Properties.File.PATH, "src/test/resources/")
+      .put(Properties.File.FILE_REGEX, "fileBatchSource.*")
```
anew: I don't think this test case is complete. I want to be able to do something like this: My base directory is /tmp/data, it contains a/x/some.txt, a/y/other.txt, b/x/this.txt, b/y/that.txt, and I want to read recursively and accept all files that match */x/*.txt. Is that possible?

anew:
Perhaps we can agree to document the semantics before implementing. What exactly will be the behavior of this source, given a filter, reading recursively, or a combination of both. Can the filter regex apply to the path, too? Or only to the file name? And what are reasonable use cases that help us define this behavior?

russorat:
So @abhinavbansal246 and I discussed with @anew about this. We would like to add a config option to the file batch source so that the user can select recursive directory scanning or not. Default to false. Also, we would like to document that the file path uses Globbing and the File RegEx uses RegEx. We can use this issue: https://issues.cask.co/browse/HYDRATOR-1152 for the docs improvements. Finally, we should verify that the regex works as expected https://issues.cask.co/browse/HYDRATOR-1151.

ananya-bhattacharya:
@anew , @russorat , @abhinavbansal246 
So the config parameter ``recursive`` will accept Boolean values true or false and would determine if the files are to be read recursively within a directory. The source, will accept file path Globbing through the ``path`` parameter and the regex expressions for the file name and path through ``fileRegex``.

So if the user has base directory /tmp/data, and it contains a/x/some.txt, a/y/other.txt, b/x/this.txt, b/y/that.txt, he can recursively read and accept all files that match /x/.txt.
The plugin in this case would split the regex at the last occurrence of ``/``, and set ``regex`` as ``.txt`` and the directory ``pattern`` as ``/x`` using the setConf() in BatchFileFilter . 
``/x`` would be used to validate the directory name before reading through it or returning true as:
	
      FileSystem fileSystem = path.getFileSystem(new Configuration());
      FileStatus[] fileStatus = fileSystem.globStatus(path);
      if (fileStatus != null && fileSystem.isDirectory(path) &&
        (fileSystem.getContentSummary(path).getDirectoryCount() > 1 || filePathName.endsWith(pattern) ||
          filePathName.endsWith(pattern + "/"))) {
        return true;
      }

 And the regex ``.txt`` would be used to match the file names. 

      String fileName = path.getName();
      Matcher matcher = regex.matcher(fileName);
      return matcher.find() && path.getParent().toString().endsWith(pattern);



User can also use path globbing in combination with a regex. So for the above case, user can set ``recursive`` as ``false`` and use path globbing as ``"path"="tmp/data/*/x"`` and ``"fileRegex"=".txt"`` to directly read the file ``tmp/data/a/x/some.txt`` and ``tmp/data/b/x/this.txt``.

Please let me know if this approach is okay.

russorat:
@ananya-bhattacharya so here are my comments:
- adding an option for recursively reading files makes sense.
- It seems like a bad idea for a user to mix globbing and regular expressions in the same field. The file path should always use globbing, and the file regex should always use regular expressions.
- The big question here is when someone uses both globbing and regex, what happens? What specific part of the path is the regex applied to? The regex should be applied to the full path and should be retitled "Regex Path Filter".
- ".txt" would only match things like "atxt, btxt". You would need something like ".+\.txt" to match things like something.txt and else.txt.
- only applying the regular expression to the end of a file might cause some strange side effects. I think a better approach would be to allow the user to enter regular expressions such as ".+\.txt$" which would mean "only match files ending in .txt"

ananya-bhattacharya:
@russorat 
Will make the suggested changes and update document to retitle "File Regex" to "Regex Path Filter".
So now fileRegex would match the entire ``filePathName``, instead of just the ``fileName`` and ``path`` would accept globbing.
So, if a user wants to recursively read and accept all files that match /x/.txt in the above discussed scenario, he would have to specify the complete regex like: ``"[a-zA-Z/]*/x/[a-z_]*.txt$"`` and ``"path=/tmp/data"``, instead of just specifying ``"/x/.txt"`` to read the files ``tmp/data/a/x/some.txt`` and ``tmp/data/b/x/this.txt``.
Or mention ``".+.txt$"`` if he wants to read all the text files.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/BatchFileFilter.java
ananya-bhattacharya: Updated code to only accept directories that match part of the pattern provided by user.

core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -16,7 +16,7 @@ Properties
 ----------
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
-**path:** The path on Microsoft Azure Blob Storage to use as input.
+**path:** The path on Microsoft Azure Blob Storage to use as input. Also, the path uses Globbing to read files.
```
awholegunch: The path on Microsoft Azure Blob Storage to use as input. The path uses filename expansion (globbing) to read files.

russorat:
LGTM

PR #581, opened 2017-01-06
Moving SparkUtils and VectorUtils to hydrator-common

ananya-bhattacharya:
Moving common code i.e. SparkUtils and VectorUtils to hydrator-common, so that it can be available to all spark plugins in their individual hydrator-org repos.
https://issues.cask.co/browse/HYDRATOR-1227


russorat:
LGTM

PR #583, opened 2017-01-10
[HYDRATOR-1247] Bump versions of cdap in app parents of plugins

abhinavbansal246:
fix to prevent ITNs from failing


Nishith:
LGTM

PR #586, opened 2017-01-20
[HYDRATOR-1288] Removing spark ml plugins from hydrator-plugins

ananya-bhattacharya:
https://issues.cask.co/browse/HYDRATOR-1288


russorat:
LGTM

PR #590, opened 2017-02-03
Fixed flaky tests.

sagarkapare:
`ProgramManager.waitForFinish()` methods calls the `status` endpoint to figure out if program is running. `status` endpoint checks if runtime info is available in the in-memory map. The runtime info for the program is added asynchronously through the listener mechanism, so it is possible that when program is started, status endpoint gets called before its runtime info is added to the in-memory map. Then the `ProgramManager.waitForFinish()` return immediately without running the program causing test case failures. 

This issue is more evident now since we added new listener for the Workflow(https://issues.cask.co/browse/CDAP-6008) which gets executed before the listener which adds runtime info to in-memory map.

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/DistinctTestRun.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/DistinctTestRun.java
@@ -28,6 +28,7 @@ import co.cask.cdap.etl.api.batch.BatchSource;
 import co.cask.cdap.etl.proto.v2.ETLBatchConfig;
 import co.cask.cdap.etl.proto.v2.ETLPlugin;
 import co.cask.cdap.etl.proto.v2.ETLStage;
+import co.cask.cdap.proto.ProgramRunStatus;
 import co.cask.cdap.proto.artifact.AppRequest;
 import co.cask.cdap.proto.id.ApplicationId;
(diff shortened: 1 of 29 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/DistinctTestRun.java
```diff
@@ -60,12 +61,12 @@ public void testDistinct() throws Exception {
     Schema inputSchema = Schema.recordOf(
       "purchase",
       Schema.Field.of("ts", Schema.of(Schema.Type.LONG)),
-      Schema.Field.of("user", Schema.of(Schema.Type.STRING)),
+      Schema.Field.of("user_name", Schema.of(Schema.Type.STRING)),
```
sagarkapare, line 64: added this change since table does not get created in hive because of `user` keyword.

sagarkapare:
Verified that the tests are passing locally.

poornachandra:
Changes LGTM. However, it looks like writing tests has become a little more difficult now since we have to check for cumulative runs. Can you file a JIRA to simplify the API?

sagarkapare:
Thanks @poornachandra for the review.  Agree that its extra work on client to maintain the counts of runs. In this JIRA https://issues.cask.co/browse/CDAP-7641, we would return the runid of the program that is started, which client can use to wait on.

PR #592, opened 2017-02-03
[HYDRATOR-1008] Adding proper validation messages in Projection transform

romy-khetan:


core-plugins/src/main/java/co/cask/hydrator/plugin/transform/ProjectionTransform.java
```diff
@@ -125,7 +125,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
   @Override
   public void initialize(TransformContext context) throws Exception {
     super.initialize(context);
-    init();
+    init(null);
```
CuriousVini, line 128: validation can be done at `configurePipeline` level right?

romy-khetan: Yes. init() is called both from configurePipeline to validate and initialize to initialize the variables with the inputs.

russorat:
LGTM

PR #594, opened 2017-02-03
Bumping version from 4.1 to 4.2

Nishith:


CuriousVini:
we also need to bump hydrator version right?

CuriousVini:
@Nishith https://github.com/caskdata/hydrator-plugins/pull/596 changed versions right?

CuriousVini (approved):

PR #598, opened 2017-02-15
Feature release/hydrator 1323 add error flattener

albertshau:


albertshau:
should not be merged until https://github.com/caskdata/cdap/pull/7969 is merged

elfenheart:
@albertshau just a thought, would it be possible to add a getSchema method for this plugin, so that the user can propagate the correct schema?

albertshau:
ok can add that

core-plugins/docs/ErrorCollector-errortransform.md
```diff
@@ -0,0 +1,61 @@
...
+Use Case
+--------
+The plugin is used when you want to capture errors emitted from another stage and pass them along
+with all the error information flattened into the record. For example, you may want to connect a sink
+to this plugin in order to store and later examine the error records.
+
+Properties
+----------
+**messageField:** The name of the error message field to use in the output schema. Defaults to 'errMsg'.
+If this is not specified, the error message will be dropped.
```
anew: Pick one: Does it default to errMsg, or will the messages be dropped?

core-plugins/docs/ErrorCollector-errortransform.md
```diff
@@ -0,0 +1,61 @@
...
+
+**codeField:** The name of the error code field to use in the output schema. Defaults to 'errCode'.
+If this is not specified, the error code will be dropped.
+
```
anew: same here

core-plugins/docs/ErrorCollector-errortransform.md
```diff
@@ -0,0 +1,61 @@
...
+**stageField:** The name of the error stage field to use in the output schema. Defaults to 'errStage'.
+If this is not specified, the error stage will be dropped.
+
```
anew: and here

core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
```diff
@@ -0,0 +1,144 @@
...
+import co.cask.cdap.etl.api.ErrorTransform;
+import co.cask.cdap.etl.api.PipelineConfigurer;
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+import javax.ws.rs.Path;
+
+/**
+ * Adds the error code and error message to each record, then emits it.
```
anew: and error stage?

core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
```diff
@@ -0,0 +1,144 @@
...
+
+  public ErrorCollector(Config config) {
+    this.config = config;
+  }
+
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
+    Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
+    if (inputSchema != null) {
+      if (inputSchema.getField(config.messageField) != null) {
```
anew: can this produce NPE if config.messageField is not set? and the same for the next two

anew (commented):
I left a few comments. Can't really speak to the widget, but it seems that @elfenheart has taken a look at that

core-plugins/docs/ErrorCollector-errortransform.md
albertshau: in the UI, it has a default value of 'errMsg'. If the user chooses to delete it, the message will be dropped. I can probably make this clearer.

anew (approved):
LGTM

PR #599, opened 2017-02-21
HYDRATOR-1384 fix json parser doc

albertshau:
When specifying paths with brackets, elements in the bracket
need to be enclosed in single quotes.


awholegunch:
LGTM

PR #604, opened 2017-02-24
Include submodules in build

bdmogal:
On hold, because the realtime stream source does not build correctly right now.


sreevatsanraman:
LGTM. Let's wait until upstream changes are made.

PR #606, opened 2017-02-26
Removed -SNAPSHOT from 1.6

Nishith:


database-plugins/pom.xml
```diff
@@ -20,12 +20,12 @@
   <parent>
     <artifactId>hydrator-plugins</artifactId>
     <groupId>co.cask.hydrator</groupId>
-    <version>1.6.0-SNAPSHOT</version>
+    <version>1.6.0</version>
   </parent>
 
   <properties>
     <!-- properties for script build step that creates the config files for the artifacts -->
```
anwar6953, line 27: any idea why some pom files have this `app.parents`, but like cassandra-plugins/pom.xml doesn't?

Nishith: Don't know

sreevatsanraman: Let's open a JIRA to cleanup for next release.

sreevatsanraman:
LGTM

database-plugins/pom.xml
anwar6953: https://issues.cask.co/browse/HYDRATOR-1400

PR #608, opened 2017-02-28
[HYDRATOR-1326 and HYDRATOR-1373] Added support for IAM authentication to S3batchsource and S3batchsink.Added support for server side encryption in S3 batchsink

shashankmoghe:
Design wiki: https://wiki.cask.co/display/CE/S3+IAM+Roles+Authentication+and+server+side+encryption
JIRA:https://issues.cask.co/browse/HYDRATOR-1326

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
@@ -60,6 +60,12 @@ public abstract class S3BatchSink<KEY_OUT, VAL_OUT> extends ReferenceBatchSink<S
   private static final String FILESYSTEM_PROPERTIES_DESCRIPTION = "A JSON string representing a map of properties " +
     "needed for the distributed file system.";
   private static final String DEFAULT_PATH_FORMAT = "yyyy-MM-dd-HH-mm";
+  private static final String ACCESS_KEY = "fs.s3n.awsAccessKeyId";
+  private static final String SECRET_KEY = "fs.s3n.awsSecretAccessKey";
+  private static final String S3N_ENCRYPTION = "fs.s3n.server-side-encryption-algorithm";
+  private static final String S3A_ENCRYPTION = "fs.s3a.server-side-encryption-algorithm";
+  private static final String ACCESS_CREDENTIALS = "Access Credentials";
+  private static final String IAM = "IAM";
   private static final Gson GSON = new Gson();
   private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() {
   }.getType();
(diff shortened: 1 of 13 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
```diff
@@ -88,15 +106,26 @@ public final void prepareRun(BatchSinkContext context) {
...
     } else {
       providedProperties = GSON.fromJson(fileSystemProperties, MAP_STRING_STRING_TYPE);
     }
-    providedProperties.put("fs.s3n.awsAccessKeyId", accessID);
-    providedProperties.put("fs.s3n.awsSecretAccessKey", accessKey);
+    boolean encryptionEnabled = enableEncryption != null && enableEncryption.equalsIgnoreCase("True");
+    if (authenticationMethod != null && authenticationMethod.equalsIgnoreCase("Access Credentials")) {
+      providedProperties.put("fs.s3n.awsAccessKeyId", accessID);
+      providedProperties.put("fs.s3n.awsSecretAccessKey", accessKey);
+      if (encryptionEnabled) {
```
russorat: can this be extracted outside the parent if statement so it doesn't need to be repeated?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
```diff
@@ -72,25 +84,47 @@ private static String updateFileSystemProperties(@Nullable String fileSystemProp
...
     private final String accessID;
 
     @Description(ACCESS_KEY_DESCRIPTION)
     @Macro
+    @Nullable
     private final String accessKey;
 
-    public S3BatchConfig(String referenceName, String accessID, String accessKey, String path) {
-      this(referenceName, accessID, accessKey, path, null, null, null, null, null, false, false);
+    @Description(AUTHENTICATION_METHOD)
```
russorat, line 99: Why doesn't this plugin let me enter the IAMRole name like the other S3 plugins?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
```diff
@@ -48,21 +52,29 @@ public S3BatchSource(S3BatchConfig config) {
...
                               config.inputFormatClass, updateFileSystemProperties(
-                                config.fileSystemProperties, config.accessID, config.accessKey),
+      config.fileSystemProperties, config.accessID, config.accessKey, config.authenticationMethod),
                               config.maxSplitSize, config.ignoreNonExistingFolders, config.recursive));
     this.config = config;
   }
 
+  @Override
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    config.validate(config.authenticationMethod);
```
russorat, line 66: could you add a validate check for s3a:// when using IAMRoles?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
```diff
@@ -140,17 +181,35 @@ public S3BatchSinkConfig() {
...
       this.accessID = accessID;
       this.accessKey = accessKey;
-      this.fileSystemProperties = updateFileSystemProperties(fileSystemProperties, accessID, accessKey);
+      this.authenticationMethod = authenticationMethod;
+      this.enableEncryption = enableEncryption;
+      this.fileSystemProperties = updateFileSystemProperties(fileSystemProperties, accessID, accessKey,
+                                                             authenticationMethod, enableEncryption);
+    }
+
+    public void validate(String authenticationMethod) {
```
russorat, line 206: Add validation of s3a:// when using IAMRoles

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
```diff
@@ -88,15 +106,26 @@ public final void prepareRun(BatchSinkContext context) {
...
     if (fileSystemProperties == null) {
       providedProperties = new HashMap<>();
     } else {
       providedProperties = GSON.fromJson(fileSystemProperties, MAP_STRING_STRING_TYPE);
     }
-    providedProperties.put("fs.s3n.awsAccessKeyId", accessID);
-    providedProperties.put("fs.s3n.awsSecretAccessKey", accessKey);
+    boolean encryptionEnabled = enableEncryption != null && enableEncryption.equalsIgnoreCase("True");
+    if (authenticationMethod != null && authenticationMethod.equalsIgnoreCase("Access Credentials")) {
+      providedProperties.put("fs.s3n.awsAccessKeyId", accessID);
```
russorat: can these property keys be extracted to static constants? same for line 120 and 122

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
shashankmoghe: For Access Credentials,properties- awsAccessKeyId and awsSecretAccessKey needs to be set irrespective of encryptionEnabled so need this check inside the parent if.
In else if ,encryptionEnabled condition can be added instead of a child if.e.g
else if (authenticationMethod != null && authenticationMethod.equalsIgnoreCase(IAM) && encryptionEnabled)

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
shashankmoghe: IAM role is not needed when accessing S3 bucket using IAM role based authentication from hadoop.We just need to omit the key properties.
https://hadoop.apache.org/docs/r3.0.0-alpha1/hadoop-aws/tools/hadoop-aws/index.html

IAMRole configuration is needed in redshift copy command while accessing S3.
http://docs.aws.amazon.com/redshift/latest/dg/t_loading-tables-from-s3.html

shashankmoghe:
@russorat  Please review

russorat:
LGTM

PR #610, opened 2017-03-08
HYDRATOR-703 exception if diver class is not loaded and its null

gsps1:
currently when we can't load the driver class (null), we still try to ensure it exists and throw NPE during the process. rather we should do null check earlier.
https://issues.cask.co/browse/HYDRATOR-703


albertshau (approved):
lgtm

PR #618, opened 2017-03-22
CDAP-7884 modify several plugins to reflect change in cdap

yaojiefeng:
Note this cannot be merged until https://github.com/caskdata/cdap/pull/8405 is merged. 
Also fix a TODO in hive source/sink, since cdap-4132 is resolved long time ago.
Also found in ExcelInputReader source, we are getting dataset in the transform method and tries to write to it in the transform, which is not an expected behavior.


CuriousVini:
@yaojiefeng Is this an api change in cdap? Do we also need corresponding changes to plugins in hydrator repo?

yaojiefeng:
@CuriousVini It is an ETL api change in cdap. The plugins are not needed to change if they are not trying to create cdap datasets and write to them during the runtime of the pipeline. The api changes is to prevent the plugin to write on real space. You can have more info on https://wiki.cask.co/display/CE/Figure+out+the+real+datasets+for+preview

yaojiefeng:
Closing, we need a better approach

PR #620, opened 2017-03-29
CDAP-9170 fix bad xml parser unit tests

albertshau:
Fixing a unit test that was not changed when array handling was
changed in the plugin. Also performing various unit test
improvements. None of these tests need to deploy and run pipelines
so changed them to just call transform methods. This reduces
run time from minutes to one second. Also changing tests to
compare records directly instead of comparing subsets of record
fields and doing silly things like looping through a list and
having an if statement for every different record in the list.


yaojiefeng:
LGTM

PR #621, opened 2017-04-05
making schema macro enabled in multiple sinks

albertshau:


gsps1:
is there a JIRA for this ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
```diff
@@ -220,6 +227,13 @@ public void validate(String authenticationMethod) {
           throw new IllegalArgumentException("Path must start with s3a:// for IAM based authentication.");
         }
       }
+      if (schema != null) {
```
gsps1, line 230: should  this also be `if (!containsMacro("schema")) {`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TPFSSinkConfig.java
```diff
@@ -101,5 +107,16 @@ public void validate() {
         throw new IllegalArgumentException("Cannot clean partitions from less than 1 minute ago.");
       }
     }
+    if (schema != null) {
```
gsps1, line 110: `!containsMacro("schema")` ?

gsps1 (commented):
couple comments.

gsps1:
should this be merged to release branch or develop

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/S3BatchSink.java
albertshau: this only happens if it contains a macro because it is not nullable.

albertshau: should be consistent though, will change the other places to check for null

gsps1:
LGTM.

PR #629, opened 2017-04-11
Updated branch for wrangler transform

bdmogal:
Verified that the right version of wrangler gets created.


deepakwadhwani (approved):
lgtm

PR #636, opened 2017-04-18
HYDRATOR-488 and HYDRATOR-465 Upgrade jython to 2.7.0

Denton-L:
JIRA: https://issues.cask.co/browse/HYDRATOR-488
JIRA: https://issues.cask.co/browse/HYDRATOR-465

This upgrades jython to 2.7.0 and, in addition to that, reverts 7834ed8.

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -90,7 +90,6 @@
     <dependency>
       <groupId>org.python</groupId>
       <artifactId>jython-standalone</artifactId>
-      <version>2.5.2</version>
     </dependency>
     <!-- Start for JMS Source -->
     <dependency>
@@ -245,6 +244,36 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.0.0</version>
+        <configuration>
+          <createDependencyReducedPom>false</createDependencyReducedPom>
+        </configuration>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer">
+                </transformer>
+              </transformers>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+          </archive>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
         <version>2.5.4</version>
@@ -265,16 +294,13 @@
               <!-- Used by S3 source/sink -->
               org.apache.http.*
             </_exportcontents>
-            <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
-            <Embed-Transitive>true</Embed-Transitive>
-            <Embed-Directory>lib</Embed-Directory>
           </instructions>
         </configuration>
         <executions>
           <execution>
-            <phase>package</phase>
+            <phase>process-classes</phase>
             <goals>
-              <goal>bundle</goal>
+              <goal>manifest</goal>
             </goals>
           </execution>
         </executions>
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
@@ -160,6 +160,8 @@ public class PythonEvaluator extends Transform<StructuredRecord, StructuredRecor
     } else {
       pipelineConfigurer.getStageConfigurer().setOutputSchema(pipelineConfigurer.getStageConfigurer().getInputSchema());
     }
+    // try evaluating the script to fail application creation if the script is invalid
+    init();
   }
 
   @Override
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
@@ -46,6 +46,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import javax.annotation.Nullable;
 
 /**
@@ -344,6 +345,12 @@ public class PythonEvaluator extends Transform<StructuredRecord, StructuredRecor
   }
 
   private void init() {
+
+    // This is necessary to prevent 'ImportError' being thrown by Jython
+    Properties properties = new Properties();
+    properties.put("python.import.site", "false");
+    PythonInterpreter.initialize(System.getProperties(), properties, new String[0]);
+
     interpreter = new PythonInterpreter();
     interpreter.set(CONTEXT_NAME, new ScriptContext(
       logger, metrics,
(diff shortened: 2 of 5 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
```diff
@@ -344,6 +345,10 @@ private Object decodeUnion(Object object, List<Schema> schemas) {
   }
 
   private void init() {
+    Properties properties = new Properties();
```
anwar6953: add a comment about why this is needed?

anwar6953:
LGTM, assuming its been tested on SDK and cluster for regression.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
```diff
@@ -344,6 +347,12 @@ private Object decodeUnion(Object object, List<Schema> schemas) {
   }
 
   private void init() {
+
+    // This is necessary to prevent 'ImportError' being thrown by Jython
+    Properties properties = new Properties();
+    properties.put("python.import.site", "false");
```
gsps1, line 353: should we document this ? because site packages were available earlier ?

Denton-L: From what location were site-packages loaded from before?

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -189,7 +189,7 @@ public void testSchemaValidation() throws Exception {
     PythonEvaluator.Config config = new PythonEvaluator.Config(
       "def transform(x, emitter, context):\n" +
         " y = {}\n" +
-        "  y['intField'] *= 1024\n" +
+        "y['intField'] *= 1024\n" +
```
awholegunch: These lines (191 to 193) aren't lining up correctly.

Denton-L: Which part isn't lining up correctly?

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -204,12 +204,37 @@ public void testSchemaValidation() throws Exception {
...
+      Schema.Field.of("intField", Schema.of(Schema.Type.INT)),
+      Schema.Field.of("longField", Schema.of(Schema.Type.LONG)));
+
+    PythonEvaluator.Config config = new PythonEvaluator.Config(
+      "def transform(x, emitter, context):\n" +
+        " y = {}\n" +
+        " y['intField'] *= 1024\n" +
+        " y['longField'] *= 1024\n" +
+        " emitter.emit(y)",
+      outputSchema.toString());
```
awholegunch, line 222: I see this is a variation of the above...but is this going to work? If `y` is an empty dict, and then you multiply an element that doesn't exist by 1024, isn't this going to happen?:
```
>>> y = {}
>>> y['intField'] *= 1024
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'intField'
```
as the field doesn't exist in `y`, though it does in `x`.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
Denton-L: Fixed.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
Denton-L: Hmm, it passes the tests but I was just reverting Ali's commit. @anwar6953, could you please answer this?

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
```diff
@@ -189,9 +189,9 @@ public void testSchemaValidation() throws Exception {
     PythonEvaluator.Config config = new PythonEvaluator.Config(
       "def transform(x, emitter, context):\n" +
         " y = {}\n" +
-        "  y['intField'] *= 1024\n" +
-        "  y['longField'] *= 1024\n" +
-        "  emitter.emit(y)",
+        "y['intField'] *= 1024\n" +
```
awholegunch, line 192: missing a leading space. Also, should this not be `y` is a (shallow?) copy of `x`? Those fields are not defined in `y`.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
awholegunch: I'm thinking that `y` needs to be a shallow copy of `x`. But if you look at the docs for the plugin (http://docs.cask.co/cdap/4.1.1/en/developers-manual/pipelines/plugins/transforms/pythonevaluator-transform.html) you will see this:
```
'def transform(record, emitter, context): record['count'] *= 1024; emitter.emit(record)'
```
which operates on the input record directly.

core-plugins/pom.xml
```diff
@@ -90,7 +90,6 @@
     <dependency>
       <groupId>org.python</groupId>
       <artifactId>jython-standalone</artifactId>
-      <version>2.5.2</version>
```
awholegunch, line 93: Does this need a version, and should it be 2.7.0? Or, should the dependency in the pom.xml be
```
<jython-standalone.version>2.7.0</jython-standalone.version>

awholegunch:
I would like to see two tests added: a test of loading user code (a custom `.py` file) so we know how and where it gets loaded from, and an example showing an exception thrown; where and how do those get handled? Does the developer need to do anything special, or does CDAP handle those? Can an exception thrown by the Evaluator affect the system?

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
gsps1: since you are setting this to false, i am wondering if they were available before, am not sure if they were.

awholegunch: Not sure. But what does setting "python.import.site", "false" mean?

core-plugins/pom.xml
Denton-L: We should inherit the version from the parent pom since there's no reason to duplicate or override the value.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
Denton-L: Seems like this was intentional by @gsps1: https://github.com/caskdata/hydrator-plugins/commit/7d4eb4d9cb445133b3277227a3d9abf896ce63b1#diff-9c235278978bbd8d97a04a785720692bR145

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
Denton-L: @awholegunch It prevents the `site` module from being imported automatically.
@gsps1 I looked into the [2.7.0 release notes](https://wiki.python.org/jython/JythonReleaseNotes). Seems like site modules are only automatically imported beginning from version `2.7b3`.

core-plugins/src/test/java/co/cask/hydrator/plugin/transform/PythonEvaluatorTest.java
Denton-L: Sorry, my bad. I just looked at the name of the test case. It's called `testScriptCompilationValidation` and it expects an `Exception` to be thrown. As a result, I think this "syntax error" is expected.

Denton-L:
@awholegunch 
* The scripts that are evaluated are entered into the UI so the plugin never reads files directly. 
* When a Python exception is raised, the pipeline simply fails. It's no different from when a Java exception is thrown. Also, we already have a test case for this. `testScriptCompilationValidation` should be a syntax error and syntax errors are exceptions in Python.

Denton-L:
This has been tested on a cluster and on the SDK. Seems to work for me.

awholegunch:
LGTM

PR #641, opened 2017-04-21
Feature/fix unit test

sreevatsanraman:
Failure before the fix
```java.lang.ClassCastException: java.nio.HeapByteBuffer cannot be cast to [B
	at co.cask.hydrator.plugin.db.batch.sink.DBSourceTestRun.testDBSource(DBSourceTestRun.java:168)```


albertshau (approved):
lgtm

PR #643, opened 2017-04-24
Revert "HYDRATOR-488 Create fat JAR with shade"

Denton-L:
This reverts commit ca99ef9. This
should solve the issue where the JAR produced is massive compared to
before this commit.

At some point, we should probably move the PythonEvaluator plugin into
its own package, though.


anwar6953:
Related to: https://github.com/caskdata/hydrator-plugins/pull/636

anwar6953:
LGTM

PR #644, opened 2017-04-25
CDAP-8132 add file path as an output field in file source

albertshau:
Adds an optional property for the file source to include the file
path that the record came from. The property cannot be used if
the input format class is set.

Also contains some refactoring with children of the FileBatchSource
to override config methods instead of some odd constructor json
string manipulation that was happening.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
@@ -27,6 +27,7 @@ import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.DatasetProperties;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.lib.KeyValueTable;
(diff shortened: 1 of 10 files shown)


albertshau:
https://builds.cask.co/browse/HYP-BAD284

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileBatchSource.java
```diff
@@ -91,7 +92,8 @@
   protected static final String TABLE_DESCRIPTION = "Name of the Table that keeps track of the last time files " +
     "were read in. If this is null or empty, the Regex is used to filter filenames.";
   protected static final String INPUT_FORMAT_CLASS_DESCRIPTION = "Name of the input format class, which must be a " +
-    "subclass of FileInputFormat. Defaults to CombineTextInputFormat.";
+    "subclass of FileInputFormat. Defaults to a CombinePathTrackingInputFormat, which is a customized version of " +
+    "CombineTextInputFormat that records the file path each record was read from.";
```
gokulavasan, line 96: customized version of CombineFileInputFormat?

gokulavasan:
Just one minor comment. LGTM 👍

PR #649, opened 2017-05-02
Fixing AzureBatchSource to work with Filesets

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-11384
Build: https://builds.cask.co/browse/HYP-BAD289-16

Tested on cluster:

image

image

### DIFF
--- a/core-plugins/docs/AzureBlobStore-batchsource.md
+++ b/core-plugins/docs/AzureBlobStore-batchsource.md
@@ -17,15 +17,13 @@ Properties
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
 **path:** The path on Microsoft Azure Blob Storage to use as input. The path uses filename expansion (globbing) to read
-files. (Macro-enabled)
+files. The path must start with `wasb://` or `wasbs://`, for example, `wasb://mycontainer@mystorageaccount.blob.core.windows.net/filename.txt`. (Macro-enabled)
 
 **account:** The Microsoft Azure Blob Storage account to use. This is of the form 
 `mystorageaccount.blob.core.windows.net`, where `mystorageaccount` is the Microsoft 
 Azure Storage account name. (Macro-enabled)
 
-**container:** The container to use on the specified Microsoft Azure Storage account. (Macro-enabled)
-
-**storageKey:** The storage key for the specified container on the Microsoft Azure Storage account. 
+**storageKey:** The storage key for the container on the Microsoft Azure Storage account. 
 Must be a valid base64 encoded storage key provided by Microsoft Azure. (Macro-enabled)
 
 **ignoreNonExistingFolders:** Identify if path needs to be ignored or not, for case when directory or file does not
@@ -37,17 +35,15 @@ Example
 -------
 
 This example connects to Microsoft Azure Blob Storage and reads in files found in the
-specified directory. This example uses the 'hydrator' container at
-'hydratorstorage.blob.core.windows.net' (Microsoft Azure Storage), using the
-'hydratorstorage' account name:
+specified directory. This example uses Microsoft Azure Storage 'mystorageaccount.blob.core.windows.net', using the
+'mystorageaccount' account name:
 
     {
         "name": "AzureBlobStore",
         "type": "batchsource",
         "properties": {
-            "path": "/path/to/input",
-            "account": "hydratorstorage.blob.core.windows.net",
-            "container": "hydrator",
+            "path": "`wasb://mycontainer@mystorageaccount.blob.core.windows.net/filename.txt",
+            "account": "mystorageaccount.blob.core.windows.net",
             "storageKey": "XXXXXEEESSS/YYYY=",
             "ignoreNonExistingFolders": "false",
             "recursive": "false"
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
@@ -31,8 +31,7 @@ import java.util.Map;
 @Plugin(type = "batchsource")
 @Name("AzureBlobStore")
 @Description("Batch source to read from Azure Blob Storage.")
-public class AzureBatchSource extends FileBatchSource {
-
+public class AzureBatchSource extends AbstractFileBatchSource {
   @SuppressWarnings("unused")
   private final AzureBatchConfig config;
 
@@ -44,14 +43,15 @@ public class AzureBatchSource extends FileBatchSource {
   /**
    * Plugin config for {@link AzureBatchSource}.
    */
-  public static class AzureBatchConfig extends FileBatchConfig {
+  public static class AzureBatchConfig extends FileSourceConfig {
+    @Description("Path to file(s) to be read. If a directory is specified,terminate the path name with a '/'. " +
+      "The path must start with `wasb://` or `wasbs://`.")
+    @Macro
+    public String path;
+
     @Description("The Microsoft Azure Storage account to use.")
     @Macro
     private String account;
-
-    @Description("The container to use on the specified Microsoft Azure Storage account.")
-    @Macro
-    private String container;
 
     @Description("The storage key for the specified container on the specified Azure Storage account. Must be a " +
       "valid base64 encoded storage key provided by Microsoft Azure.")
@@ -61,11 +61,15 @@ public class AzureBatchSource extends FileBatchSource {
     @Override
     protected Map<String, String> getFileSystemProperties() {
       Map<String, String> properties = new HashMap<>(super.getFileSystemProperties());
-      properties.put("fs.defaultFS", String.format("wasb://%s@%s/", container, account));
       properties.put("fs.wasb.impl", "org.apache.hadoop.fs.azure.NativeAzureFileSystem");
       properties.put("fs.AbstractFileSystem.wasb.impl", "org.apache.hadoop.fs.azure.Wasb");
       properties.put(String.format("fs.azure.account.key.%s", account), storageKey);
       return properties;
+    }
+
+    @Override
+    protected String getPath() {
+      return path;
     }
   }
 }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
@@ -34,7 +34,7 @@ import javax.annotation.Nullable;
 @Plugin(type = "batchsource")
 @Name("S3")
 @Description("Batch source to use Amazon S3 as a source.")
-public class S3BatchSource extends FileBatchSource {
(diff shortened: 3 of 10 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -61,7 +61,6 @@ public AzureBatchSource(AzureBatchConfig config) {
     @Override
     protected Map<String, String> getFileSystemProperties() {
       Map<String, String> properties = new HashMap<>(super.getFileSystemProperties());
-      properties.put("fs.defaultFS", String.format("wasb://%s@%s/", container, account));
```
albertshau, line 64: doesn't this mean container is no longer used?

CuriousVini: so we can add container and corresponding account while accessing the files, similar to S3 source plugin. For example:

`wasb://<container>@<account>/filename.txt`

CuriousVini: Another option is to prefix the file names with `wasb://<container>@<account>` and only take file names without file system in the config. I guess this would be the better option right @albertshau ?

albertshau: I was just saying we should remove 'container' as a property from the PluginConfig if it is no longer being used.

If the path always needs to be of the form 'wasb://<container>@<account>/...', then I think what you mentioned about prefixing makes sense. If we do that, should also update the documentation, and add a validation check to make sure it's just a path and not prefixed with 'wasb...'

YanAtCask: Here is "[Use HDFS-compatible storage with Hadoop in HDInsight](https://docs.microsoft.com/en-us/azure/hdinsight/hdinsight-hadoop-use-blob-storage)" .  The  syntax is wasb[s]://<containername>@<accountname>.blob.core.windows.net/<path>.  The recommendation is to use wasbs wherever possible.

CuriousVini:
@albertshau Addressed comments. Please take another pass when you get a chance.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -32,15 +34,27 @@
 @Name("AzureBlobStore")
 @Description("Batch source to read from Azure Blob Storage.")
 public class AzureBatchSource extends FileBatchSource {
+  private static final Logger LOG = LoggerFactory.getLogger(AzureBatchSource.class);
```
albertshau: unused

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -32,15 +34,27 @@
...
-    super(config);
+    super(new FileBatchConfig(config.referenceName, getPrefixedPath(config.account, config.container, config.path),
+                              config.fileRegex, config.timeTable, config.inputFormatClass,
+                              config.fileSystemProperties, config.maxSplitSize, config.ignoreNonExistingFolders,
+                              config.recursive));
     this.config = config;
   }
 
+  // add prefix "wasb://<container>@<account>" to the path if path does not start with "wasb://"
+  private static String getPrefixedPath(String account, String container, String path) {
```
albertshau: Modifying behavior in child classes is cleaner using inheritance rather than modifying properties in a config file. In fact, I think we should refactor how path is handled in the File sources since they all use the description from FileBatchSource, which is inaccurate and confusing.

One way to do this is to create a 'FileSourceConfig' that they all extend that has a getPath() method but no path property. Each source can then have its own path property with its own description, and AzureBatchConfig can do this logic in getPath() instead of creating a modified config object.

albertshau:
Please check with Yan about the wasbs protocol. Seems like we may need to tweak the logic a little to support it.

CuriousVini:
@albertshau To keep this plugin consistent with S3 source, I have removed container property instead of adding prefix to the path.

CuriousVini:
Build is green: https://builds.cask.co/browse/HYP-BAD289-13

core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -17,7 +17,7 @@ Properties
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
 **path:** The path on Microsoft Azure Blob Storage to use as input. The path uses filename expansion (globbing) to read
-files. (Macro-enabled)
+files. The path must strat with `wasb://` or `wasbs://`. (Macro-enabled)
 
 **account:** The Microsoft Azure Blob Storage account to use. This is of the form 
 `mystorageaccount.blob.core.windows.net`, where `mystorageaccount` is the Microsoft
```
YanAtCask: It might be better to change account to `mystorageaccount` only and later append '.blob.core.windows.net'.  It can be more consistent with Azure site and the usage of other plug-ins (when I put this plug-in source with other actions together,  it looks better to use one macro across the pipeline).

core-plugins/docs/AzureBlobStore-batchsource.md
```diff
@@ -17,7 +17,7 @@ Properties
 **referenceName:** This will be used to uniquely identify this source for lineage, annotating metadata, etc.
 
 **path:** The path on Microsoft Azure Blob Storage to use as input. The path uses filename expansion (globbing) to read
-files. (Macro-enabled)
+files. The path must strat with `wasb://` or `wasbs://`. (Macro-enabled)
```
albertshau: typo: start

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -32,26 +32,28 @@
...
+    @Description("Path to file(s) to be read. If a directory is specified,terminate the path name with a '/'. " +
+      "The path must start with `wasb://` or `wasbs://`.")
     @Macro
-    private String account;
+    public String path;
 
-    @Description("The container to use on the specified Microsoft Azure Storage account.")
+    @Description("The Microsoft Azure Storage account to use.")
     @Macro
-    private String container;
```
albertshau, line 54: also remove this from the docs

core-plugins/docs/AzureBlobStore-batchsource.md
albertshau: Also, can you give an example of what it's expected to be?

core-plugins/docs/AzureBlobStore-batchsource.md
CuriousVini: @YanAtCask I am not sure if we should do that because Azure also supports custom domains on http protocol https://docs.microsoft.com/en-us/azure/storage/storage-custom-domain-name#register-a-custom-domain. Not sure if it is also supported for `wasb` protocol though.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
```diff
@@ -48,16 +48,23 @@
   private final S3BatchConfig config;
 
   public S3BatchSource(S3BatchConfig config) {
-    super(config);
+    super(new FileBatchConfig(config.referenceName, config.fileRegex, config.timeTable, config.inputFormatClass,
```
albertshau: I think you should be able to pass config in directly

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/S3BatchSource.java
```diff
@@ -48,16 +48,23 @@
...
   }
 
   /**
    * Config class that contains properties needed for the S3 source.
    */
-  public static class S3BatchConfig extends FileBatchConfig {
+  public static class S3BatchConfig extends FileSourceConfig {
     private static final Gson GSON = new Gson();
 
+    @Description("Path to file(s) to be read. If a directory is specified,terminate the path name with a '/'. " +
```
albertshau: add a space between specified, and terminate

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -32,26 +32,28 @@
 @Name("AzureBlobStore")
 @Description("Batch source to read from Azure Blob Storage.")
 public class AzureBatchSource extends FileBatchSource {
-
   @SuppressWarnings("unused")
   private final AzureBatchConfig config;
 
   public AzureBatchSource(AzureBatchConfig config) {
-    super(config);
+    super(new FileBatchConfig(config.referenceName, config.fileRegex, config.timeTable, config.inputFormatClass,
```
albertshau: i think you should be able to pass in config directly

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
```diff
@@ -32,26 +32,28 @@
...
   }
 
   /**
    * Plugin config for {@link AzureBatchSource}.
    */
-  public static class AzureBatchConfig extends FileBatchConfig {
-    @Description("The Microsoft Azure Storage account to use.")
+  public static class AzureBatchConfig extends FileSourceConfig {
+    @Description("Path to file(s) to be read. If a directory is specified,terminate the path name with a '/'. " +
+      "The path must start with `wasb://` or `wasbs://`.")
```
albertshau, line 48: include an example of what's expected. Should also validate this at configure time.

albertshau (commented):
minor comments

core-plugins/docs/AzureBlobStore-batchsource.md
YanAtCask: Thank you.  It is fine to me to keep this way right now.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AzureBatchSource.java
CuriousVini: Since `AzureBatchConfig` and `FileBatchConfig` are type of `FileSourceConfig`, passing the same config throws compilation error. 

image

albertshau: hm, I see. Can you do another refactor to change the current FileBatchSource to:

```
abstract class AbstractFileBatchSource<T extends FileSourceConfig> extends ReferenceBatchSource<Object, Object, StructuredRecord>
```

Then FileBatchSource, AzureBatchSource, and S3BatchSource all extend from the abstract class

albertshau (approved):
lgtm

CuriousVini:
Build is green. Squashing and merging.

PR #650, opened 2017-05-05
CDAP-11417 fix a bug with byte fields in parquet and avro sinks

albertshau:
Fixed a bug introduced when a fix was added to convert
byte[] fields to ByteBuffers, but which caused ClassCastExceptions
when the field actually contained a ByteBuffer and not a byte[].

Also cleaned up the logic to avoid a special case for bytes.
Also improved a weird unit test that was testing conversion
using empty byte arrays.


sreevatsanraman:
LGTM 👍

albertshau:
https://builds.cask.co/browse/HYP-BAD290

PR #651, opened 2017-05-12
adding validations and documentation

russorat:
This is meant to solve https://issues.cask.co/browse/CDAP-10241 and https://issues.cask.co/browse/CDAP-10158

Since this uses the default FileInputFormat, I am working around some limitations when using patterns in the config. Hopefully, the added validations and docs will help others in the future.

No unit tests since the unit tests for this plugin are ignored right now anyway

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLReaderBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLReaderBatchSource.java
@@ -349,8 +349,8 @@ public class XMLReaderBatchSource extends ReferenceBatchSource<LongWritable, Obj
         if (path.endsWith("/") ||
           !(path.contains("*") || path.contains("?") || path.contains("{") || path.contains("["))) {
           throw new IllegalArgumentException("When filtering with regular expressions, " +
-                                               "the Path must be a directory and leverage glob syntax. Most likely, " +
-                                               "you just need to end your folder path with a /*");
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/XMLReaderBatchSource.java
```diff
@@ -336,6 +337,22 @@ void validate() {
...
+          throw new IllegalArgumentException(String.format("The regular expression '%s' is not valid.", pattern), e);
+        }
+        // By default, the Hadoop FileInputFormat won't return any files when using a regex pattern unless the path
+        // has globs in it. Checking for that scenario.
+        if (path.endsWith("/") ||
+          !(path.contains("*") || path.contains("?") || path.contains("{") || path.contains("["))) {
+          throw new IllegalArgumentException("When filtering with regular expressions, " +
+                                               "the Path must be a directory and leverage glob syntax. Most likely, " +
+                                               "you just need to end your folder path with a /*");
+        }
```
awholegunch:
```
          throw new IllegalArgumentException("When filtering with regular expressions, " +
                                                "the path must be a directory and leverage glob syntax. Usually " +
                                                "the folder path needs to end with '/*'.");

awholegunch:
I've updated the documentation. Please check my changes carefully. With the changes in the comment above, LGTM.

russorat:
LGTM

PR #652, opened 2017-05-17
CDAP-9476 remove flatmap from File streaming source

albertshau:
There is no need to use flatmap in the File streaming source, so
changing it to use map. This also makes the code compatible with
both spark1 and spark2.


yaojiefeng:
LGTM

albertshau:
https://builds.cask.co/browse/HYP-BAD292

albertshau:
build failure from an unrelated module, merging.

PR #657, opened 2017-05-22
CDAP-9334 update KeyValueListParser error message

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-9334

### DIFF
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/KeyValueListParser.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/KeyValueListParser.java
@@ -82,11 +82,15 @@ public class KeyValueListParser {
       Iterator<String> keyValIter = Splitter.on(keyValDelimiter).trimResults().split(pair).iterator();
       String key = keyValIter.next();
       if (!keyValIter.hasNext()) {
-        throw new IllegalArgumentException("Invalid syntax for key-value pair in list: " + pair);
+        throw new IllegalArgumentException(String.format("Invalid syntax for key-value pair in list: %s. " +
+                                                           "It is expected to be a string separated " +
+                                                           "by exactly one %s", pair, keyValDelimiter));
       }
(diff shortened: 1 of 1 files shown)


hydrator-common/src/main/java/co/cask/hydrator/common/KeyValueListParser.java
```diff
@@ -82,11 +82,15 @@ public boolean hasNext() {
       Iterator<String> keyValIter = Splitter.on(keyValDelimiter).trimResults().split(pair).iterator();
       String key = keyValIter.next();
       if (!keyValIter.hasNext()) {
-        throw new IllegalArgumentException("Invalid syntax for key-value pair in list: " + pair);
+        throw new IllegalArgumentException(String.format("Invalid syntax for key-value pair in list: %s. " +
```
albertshau, line 85: don't think we need to put the pair in the message twice, right after each other.

Invalid syntax for key-value pair: %s. It is expected to be...

hydrator-common/src/main/java/co/cask/hydrator/common/KeyValueListParser.java
```diff
@@ -82,11 +82,15 @@ public boolean hasNext() {
...
+                                                           "%s is expected to to be a string separated " +
```
albertshau: typo: to to -> to

hydrator-common/src/main/java/co/cask/hydrator/common/KeyValueListParser.java
```diff
@@ -82,11 +82,15 @@ public boolean hasNext() {
...
+                                                           "by one and only one %s", pair, pair, keyValDelimiter));
```
albertshau: nit: less wordy if we change 'one and only one' to 'exactly'

albertshau (approved):
lgtm

PR #658, opened 2017-05-23
Bump version to 1.8.0 on develop

sreevatsanraman:


anwar6953:
I think we should also bump `cdap.version`:
https://github.com/caskdata/hydrator-plugins/blob/9cbb0608665e30875b81b32604f43b1b5c499ea9/pom.xml#L109

and `app.parents`:
https://github.com/caskdata/hydrator-plugins/blob/9cbb0608665e30875b81b32604f43b1b5c499ea9/pom.xml#L135

sreevatsanraman:
Good catch. I fixed app.parents. I will bump to 4.3.0 once the artifacts are available in maven central which requires a BUT.

anwar6953:
LGTM

PR #660, opened 2017-05-24
Move transformValue function from DBRecord to DBUtils as public stati…

rohitsinha54:
…c method

- Moving it to use in SQLServerCDC source
- Ran unitTest after change which passes.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
@@ -23,7 +23,6 @@ import com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -230,6 +234,57 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
...
+          } finally {
+            blob.free();
+          }
+          return toReturn;
+        case Types.CLOB:
+          String s;
+          StringBuilder sbf = new StringBuilder();
+          Clob clob = (Clob) original;
+          try {
+            try (BufferedReader br = new BufferedReader(clob.getCharacterStream(1, (int) clob.length()))) {
```
chtyim: Why so complicated?? Isn't it just `clob.getSubString(1, (int) clob.length())`??

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -230,6 +234,57 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
...
+        case Types.DECIMAL:
+          return ((BigDecimal) original).doubleValue();
+        case Types.DATE:
+          return resultSet.getDate(fieldName).getTime();
+        case Types.TIME:
+          return resultSet.getTime(fieldName).getTime();
+        case Types.TIMESTAMP:
+          return resultSet.getTimestamp(fieldName).getTime();
+        case Types.BLOB:
+          Object toReturn;
```
chtyim: This is not needed, just do `return blog.getBytes(...);` inside the try.

chtyim:
LGTM

PR #661, opened 2017-05-25
CDAP-11615 macro enable table schema

albertshau:
Enabling macros for the schema property of tables.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
@@ -57,7 +57,8 @@ public class TableSink extends BatchWritableSink<StructuredRecord, byte[], Put> 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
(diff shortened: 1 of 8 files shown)


albertshau:
https://builds.cask.co/browse/HYP-BAD303

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/TableSink.java
```diff
@@ -74,6 +75,12 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   }
 
   @Override
+  protected boolean shouldSkipCreateAtConfigure() {
+    return tableSinkConfig.containsMacro(Properties.Table.PROPERTY_SCHEMA) ||
```
gokulavasan, line 79: How does these two values being macros affect the creation of the dataset during configure time? Doesn't it only affect the way StructuredRecord->Put transform happens?

albertshau: if we create them at configure time, schema will be unavailable and they will not be explorable

gokulavasan: Oh yes ok, it is a CDAP table not HBase table. Thanks for the clarification!

gokulavasan:
LGTM 👍

PR #662, opened 2017-05-25
Moving Azure blob store out of hydrator plugins

CuriousVini:
Moving azure blob store to separate repo: https://github.com/hydrator/file-batch-source-plugins/tree/master/azure-blob-store

Issue: https://issues.cask.co/browse/CDAP-11599
Build: https://builds.cask.co/browse/HYP-BAD304-1


abhinavbansal246:
Do we have any ITNs for this plugin?

anwar6953:
LGTM

CuriousVini:
build is green, there is not ITN for it, merging.

PR #663, opened 2017-05-25
Moving S3 plugins to separate repo

CuriousVini:
Moving S3 Source and Sink to separate repo https://github.com/hydrator/amazon-s3-plugins to add them into market.

Build: https://builds.cask.co/browse/HYP-BAD306-1

There are no ITN present for these plugins

Tested on cluster and sdk


albertshau (approved):
lgtm

PR #664, opened 2017-05-26
Feature release/cdap 11558 fix multiple mongo sinks

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-11558
Build: https://builds.cask.co/browse/HYP-BAD307-2

The MongoOutputCommitter generated a path to store the temp mongo result using the mapreduce task id, so if multiple sinks are present, only one path is created and therefore we can only see data written to one sink, fix is to override the path with the UUID for each sink plugin.

### DIFF
--- a/mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/MongoDBBatchSink.java
+++ b/mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/MongoDBBatchSink.java
@@ -33,10 +33,12 @@ import co.cask.hydrator.common.ReferencePluginConfig;
 import com.mongodb.BasicDBObjectBuilder;
 import com.mongodb.hadoop.MongoOutputFormat;
 import com.mongodb.hadoop.io.BSONWritable;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 
 /**
  * A {@link BatchSink} that writes data to MongoDB.
@@ -57,7 +59,13 @@ public class MongoDBBatchSink extends ReferenceBatchSink<StructuredRecord, NullW
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    context.addOutput(Output.of(config.referenceName, new MongoDBOutputFormatProvider(config)));
+    Configuration conf = new Configuration();
+    String path = conf.get(
+      "mapreduce.task.tmp.dir",
+      conf.get(
+        "mapred.child.tmp",
+        conf.get("hadoop.tmp.dir", System.getProperty("java.io.tmpdir")))) + "/" + UUID.randomUUID().toString();
+    context.addOutput(Output.of(config.referenceName, new MongoDBOutputFormatProvider(config, path)));
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/MongoDBBatchSink.java
```diff
@@ -76,6 +77,7 @@ public void transform(StructuredRecord input, Emitter<KeyValue<NullWritable, BSO
     MongoDBOutputFormatProvider(MongoDBSinkConfig config) {
       this.conf = new HashMap<>();
       conf.put("mongo.output.uri", config.connectionString);
+      conf.put("mapreduce.task.tmp.dir", "/tmp/" + UUID.randomUUID().toString());
```
albertshau: i don't understand how this fixes the problem. won't everything still write to the same directory?

albertshau: oh i see, because this is different for each sink... but this will break in setups where mapred-site.xml has been modified to something custom. You will then override it to /tmp/, which may not be available or writable.

yaojiefeng: Updated to use the path from mapred-site.xml if the path is present. If it is not, use /tmp/

mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/MongoDBBatchSink.java
```diff
@@ -57,7 +59,13 @@ public MongoDBBatchSink(MongoDBSinkConfig config) {
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    context.addOutput(Output.of(config.referenceName, new MongoDBOutputFormatProvider(config)));
+    Configuration conf = new Configuration();
+    String path = conf.get(
+      "mapreduce.task.tmp.dir",
+      conf.get(
+        "mapred.child.tmp",
+        conf.get("hadoop.tmp.dir", "/tmp"))) + "/" + UUID.randomUUID().toString();
```
chtyim: Should never use `/tmp` hardcoded. Use `System.getProperty("java.io.tmpdir")` instead. It will be set to local container directory.

mongodb-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/MongoDBBatchSink.java
```diff
@@ -57,7 +59,13 @@ public MongoDBBatchSink(MongoDBSinkConfig config) {
 
   @Override
   public void prepareRun(BatchSinkContext context) throws Exception {
-    context.addOutput(Output.of(config.referenceName, new MongoDBOutputFormatProvider(config)));
+    Configuration conf = new Configuration();
+    String path = conf.get(
+      "mapreduce.task.tmp.dir",
```
albertshau, line 64: now im even more confused, doesn't this mean if any of these setting exist (don't they always exist?), then the path will be the same for all sinks?

yaojiefeng: This is copied from https://github.com/mongodb/mongo-hadoop/blob/master/core/src/main/java/com/mongodb/hadoop/output/MongoOutputCommitter.java#L236, it is using this path as the base path for the tmp dir for the mongo sink.

yaojiefeng: The mongo sink will append the path with maptaskid/_MONGO_OUT_TEMP, all the temporary mongo data will in this directory.

albertshau: oh ok i see, parentheses were confusing. uuid is added to that entire thing.

albertshau (approved):
lgtm

PR #665, opened 2017-05-26
CDAP-11525 fix incorrect error message in decompression transform

gsps1:
fix error message.


albertshau (approved):
lgtm

transform-plugins/src/main/java/co/cask/hydrator/plugin/Decompressor.java
```diff
@@ -114,7 +114,8 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
           deCompMap.containsKey(field.getName()) && deCompMap.get(field.getName()) != DecompressorType.NONE &&
           !Schema.Type.BYTES.equals(field.getSchema().getType())) {
           throw new IllegalArgumentException(
-            String.format("Input field  %s should be of type bytes or string. It is currently of type %s",
+            String.format("Decompression is selected for the input field %s, but the field is of type %s " +
```
deepakwadhwani: missing . at the end. 
also better to say 'for type Bytes'
@awholegunch may be has more comments about the language.

deepakwadhwani: I meant a . after " but the field is of type %s " since the next statement is a new sentence and starts with a capital letter. lgtm otherwise.

PR #672, opened 2017-06-06
Use the SparkInterpreter abstraction everywhere.

chtyim:


spark-plugins/src/main/java/co/cask/hydrator/plugin/spark/dynamic/ScalaSparkProgram.java
```diff
@@ -52,13 +51,13 @@ public ScalaSparkProgram(Config config) throws CompilationFailureException {
       // Since we don't really be able to distinguish whether it is configure time or runtime,
       // we have to compile here using an explicitly constructed SparkCompiler and then compile again
```
sagarkapare: `explicitly created SparkInterpreter`?

sagarkapare:
minor comment, else LGTM 👍

chtyim:
fixed

PR #677, opened 2017-06-16
The rowField of TableSourceConfig need to be mandatory

xho22:
The description say “must not be nullable” , but here  still contain an
annotation of “@Nullable”. It’s opposite.
And after my testing, if pipeline doesn’t contain this field, result
will failed.


xho22:
Please close this PR. I found some issues in unit test.
I thought there was more complex than I expected. I will comfirm the key of it.

PR #683, opened 2017-07-05
Fix/tpfs source

DShau18:
Issue: https://issues.cask.co/browse/CDAP-10269
Build: https://builds.cask.co/browse/HYP-BAD327-4

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -181,9 +181,15 @@
     </dependency>
     <!-- End for EMail -->
     <dependency>
-      <groupId>com.twitter</groupId>
+      <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-avro</artifactId>
-      <version>1.6.0</version>
+      <version>1.8.1</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>net.sf.uadetector</groupId>
@@ -260,7 +266,8 @@
               co.cask.hydrator.plugin.validator.*;
               co.cask.hydrator.plugin.error.*;
               org.apache.avro.mapreduce;
-              parquet.avro.*;
+              org.apache.parquet.avro.*;
+              org.apache.parquet.hadoop.*;
               org.apache.orc.*;
               <!-- Used by S3 source/sink -->
               org.apache.http.*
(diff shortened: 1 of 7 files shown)


core-plugins/pom.xml
```diff
@@ -239,6 +245,58 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-core_2.10</artifactId>
```
albertshau: why was spark added?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -166,7 +166,7 @@ protected void cleanPartitions(TimePartitionedFileSet fileSet) throws IOExceptio
       if (locName.endsWith(".parquet")) {
         Path parquetFile = new Path(file.toString());
         AvroParquetReader.Builder<GenericRecord> genericRecordBuilder = AvroParquetReader.builder(parquetFile);
-        ParquetReader<GenericRecord> reader = genericRecordBuilder.build();
+        org.apache.parquet.hadoop.ParquetReader<GenericRecord> reader = genericRecordBuilder.build();
```
albertshau: how come the full name is being used? Can we just remove the parquet.hadoop.ParquetReader import and import this class?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLSnapshotTestRun.java
```diff
@@ -295,7 +295,7 @@ private void testSource(String sourceETLPlugin, String sourceName, Map<String, I
       } else if (fileName.endsWith(".parquet")) {
         Path parquetFile = new Path(file.toString());
         AvroParquetReader.Builder<GenericRecord> genericRecordBuilder = AvroParquetReader.builder(parquetFile);
-        ParquetReader<GenericRecord> reader = genericRecordBuilder.build();
+        org.apache.parquet.hadoop.ParquetReader<GenericRecord> reader = genericRecordBuilder.build();
```
albertshau: same question about the full class name

core-plugins/pom.xml
DShau18: it was added to exclude parquet-hadoop on line 288.

albertshau: you shouldn't add a new dependency to exclude a transitive one. If you need to exclude a transitive one, it should be excluded from the dependency that pulls it in. For example, if this is getting pulled in by cdap-unit-test, it should be excluded from there.

albertshau: Also, why does it need to be excluded? Does it cause test failures?

albertshau (approved):
lgtm

PR #684, opened 2017-07-10
Fix TPFS source and sink

DShau18:
Build: https://builds.cask.co/browse/HYP-BAD329-1


CuriousVini:
lgtm once it is tested on release/4.2 sdk and cluster.

PR #686, opened 2017-07-13
CDAP-12123 remove etl-batch as a parent

albertshau:


database-plugins/pom.xml
```diff
@@ -23,11 +23,6 @@
     <version>1.8.0-SNAPSHOT</version>
   </parent>
 
-  <properties>
-    <!-- properties for script build step that creates the config files for the artifacts -->
-    <app.parents>system:cdap-etl-batch[4.2.0,4.3.0-SNAPSHOT),system:cdap-data-pipeline[4.2.0,4.3.0-SNAPSHOT),system:cdap-data-streams[4.2.0,4.3.0-SNAPSHOT)</app.parents>
```
CuriousVini, line 28: Why are we removing `cdap-data-pipeline` as parent artifacts?

albertshau: this will make it use the value from the parent pom. This was only required here when we still had cdap-etl-realtime in the parent pom, which has now been removed.

CuriousVini: I see

CuriousVini (approved):

CuriousVini:
lgtm

PR #687, opened 2017-07-18
Read Avro and Parquet formats from file batch source

sheivin:


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -50,6 +50,7 @@
     "which indicates that no files will be filtered.";
   protected static final String FILESYSTEM_PROPERTIES_DESCRIPTION = "A JSON string representing a map of properties " +
     "needed for the distributed file system.";
+  protected static final String FILE_SOURCE_FORMAT_DESCRIPTION = "Format of the file source. Defaults to text.";
```
albertshau: also include what the possible valid options are.

'Format of the file. Must be 'text', 'parquet', or 'avro'. Defaults to 'text'.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -41,14 +47,20 @@
   private static final String PATH_FIELD = "path.tracking.path.field";
   private static final String FILENAME_ONLY = "path.tracking.filename.only";
 
+  private static String fileSourceFormat = "text";
```
albertshau: don't store static state like this.

Configuration options are set in the Configuration object, then mapreduce takes care of serializing and distributing that Configuration to all mappers/reducers. The process that calls configure will not be the same process that calls createRecordReader.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -41,14 +47,20 @@
...
    * Configure the input format to use the specified schema and optional path field.
    */
-  public static void configure(Configuration conf, @Nullable String pathField, boolean filenameOnly) {
+  public static void configure(Configuration conf, @Nullable String pathField, boolean filenameOnly,
+                               String format) {
     if (pathField != null) {
       conf.set(PATH_FIELD, pathField);
     }
     conf.setBoolean(FILENAME_ONLY, filenameOnly);
+    if (format != null) {
```
albertshau: should be setting a property in the conf, not in a local variable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -70,14 +82,23 @@ public static Schema getOutputSchema(@Nullable String pathField) {
...
     }
-
-    RecordReader<LongWritable, Text> delegate = (new TextInputFormat()).createRecordReader(split, context);
     FileSplit fileSplit = (FileSplit) split;
     String pathField = context.getConfiguration().get(PATH_FIELD);
     boolean filenameOnly = context.getConfiguration().getBoolean(FILENAME_ONLY, false);
     String path = filenameOnly ? fileSplit.getPath().getName() : fileSplit.getPath().toUri().toString();
-
-    return new TrackingTextRecordReader(delegate, pathField, path);
+    fileSourceFormat = fileSourceFormat.toLowerCase();
```
albertshau: this should always throw a null pointer exception in distributed mode. The only reason your tests are passing is because they are running locally in the same JVM. Please make sure your plugin works on a cluster.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -70,14 +82,23 @@ public static Schema getOutputSchema(@Nullable String pathField) {
...
+    if (Objects.equals(fileSourceFormat, "avro")) {
```
albertshau: there is no need to use Objects.equals. "avro".equals(fileSourceFormat)

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -70,14 +82,23 @@ public static Schema getOutputSchema(@Nullable String pathField) {
...
+      RecordReader<AvroKey<GenericRecord>, NullWritable> delegate = (new AvroKeyInputFormat<GenericRecord>())
+        .createRecordReader(split, context);
+      return new TrackingAvroRecordReader(delegate, pathField, path);
+    } else if (Objects.equals(fileSourceFormat, "parquet")) {
```
albertshau: similar comment here

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
albertshau: you should be reading the format from the configuration, similar to how the path field and filenameOnly are being read.

albertshau: you should be reading the format from the configuration, similar to how the path field and filenameOnly are being read.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -124,6 +140,103 @@ public StructuredRecord getCurrentValue() throws IOException, InterruptedExcepti
...
+  }
+
+  private static class TrackingParquetRecordReader extends RecordReader<NullWritable, StructuredRecord> {
+    private final RecordReader<Void, GenericRecord> delegate;
+    private final Schema schema;
+    private final String pathField;
+    private final String path;
+
+    private TrackingParquetRecordReader(RecordReader<Void, GenericRecord> delegate,
+                                     @Nullable String pathField, String path) {
```
albertshau: fix indentation

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -124,6 +140,103 @@ public StructuredRecord getCurrentValue() throws IOException, InterruptedExcepti
...
+    }
+
+    @Override
+    public NullWritable getCurrentKey() throws IOException, InterruptedException {
+      return NullWritable.get();
+    }
+
+    @Override
+    public StructuredRecord getCurrentValue() throws IOException, InterruptedException {
+      AvroToStructuredTransformer recordTransformer = new AvroToStructuredTransformer();
```
albertshau: don't create a transformer for each record. Instead, create it once in initialize().

albertshau: don't create a transformer for each record. Instead, create it once in initialize().

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -124,6 +140,103 @@ public StructuredRecord getCurrentValue() throws IOException, InterruptedExcepti
...
+      return delegate.getProgress();
+    }
+
+    @Override
+    public void close() throws IOException {
+      delegate.close();
+    }
+  }
+
+  private static class TrackingAvroRecordReader extends RecordReader<NullWritable, StructuredRecord> {
```
albertshau: these classes have a lot of code that is exactly the same. This is usually a sign that they should have a common parent. Something like:

```
private abstract static class TrackingReader<K, V> extends RecordReader<NullWritable, StructuredRecord> {
  protected final RecordReader<K, V> delegate;
  protected final Schema schema;
  protected final String pathField;
  protected final String path;

  protected TrackingReader(RecordReader<K, V> delegate, @Nullable String pathField, String path) {
    ..
  }

  // common methods (initialize, getProcess, close, getCurrentKey, nextKeyValue) here
  
  protected abstract StructuredRecord getCurrentValue(StructuredRecord.Builder recordBuilder);

  public StructuredRecord getCurrentValue() {
    StructuredRecord.Builder recordBuilder = StructuredRecord.builder(schema);
    if (pathField != null) {
      recordBuilder.set(pathField, path);
    }
    return getCurrentValue(recordBuilder);
  }
}
```

subclasses would then only be responsible for filling out the fields in the builder.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -124,6 +140,103 @@ public StructuredRecord getCurrentValue() throws IOException, InterruptedExcepti
...
+
+    @Override
+    public NullWritable getCurrentKey() throws IOException, InterruptedException {
+      return NullWritable.get();
+    }
+
+    @Override
+    public StructuredRecord getCurrentValue() throws IOException, InterruptedException {
+      AvroToStructuredTransformer recordTransformer = new AvroToStructuredTransformer();
+      return recordTransformer.transform(delegate.getCurrentKey().datum());
```
albertshau: this won't add the path field.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
albertshau: instead of creating a transformer for each record, create just a single one in initialize().

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
albertshau: this would also fix the bug where path is never added for avro or parquet

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +475,171 @@ public void testRecursiveTimeFilterRegex() throws Exception {
     Assert.assertTrue(outputValue.contains("Hello,World"));
     Assert.assertTrue(outputValue.contains("CDAP,Platform"));
   }
+
+  @Test
+  public void testFileBatchInputFormatText() throws Exception {
+    String filePath = "file:///tmp/test/test.txt";
```
albertshau: hardcoding a directory like this is bad practice, as it assumes the system has permission to write to that location, will not clean up after itself, and will break if some other test writes to the same directory. Use the TemporaryFolder variable in this class to create a new temporary folder just for this test case.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +475,171 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+    String testData = "String for testing purposes";
+
+    Path textFile = new Path(filePath);
+    Configuration conf = new Configuration();
+    FileSystem fs = FileSystem.get(conf);
+    FSDataOutputStream writeData = fs.create(textFile);
+    writeData.write(testData.getBytes());
```
albertshau: don't need to write through the Hadoop FileSystem abstraction, just do it through normal java File.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +475,171 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+    Schema recordSchema = Schema.recordOf("record",
+                                          Schema.Field.of("i", Schema.of(Schema.Type.INT)),
+                                          Schema.Field.of("l", Schema.of(Schema.Type.LONG)));
+
+    String filePath = "file:///tmp/test/test.avro";
+    Path avroFilePath = new Path(filePath);
+    Configuration conf = new Configuration();
+    FileSystem fs = FileSystem.get(conf);
+
+    ETLStage source = new ETLStage(
```
albertshau: half of these test cases are exactly the same and can be refactored into a common method. Each test case should only deploy the app, write input, then call that common method.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +475,171 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+    ApplicationManager appManager = deployApplication(appId, appRequest);
+
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+
+    DataSetManager<Table> outputManager = getDataset(outputDatasetName);
+    List<StructuredRecord> output = MockSink.readOutput(outputManager);
+
+    Assert.assertEquals(1, output.size());
```
albertshau: this is an incomplete comparison. It only compares a single field. You should compare List<StructuredRecord> to another List<StructuredRecord>.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +475,171 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+    inputManager.flush();
+
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+
+    DataSetManager<Table> outputManager = getDataset(outputDatasetName);
+    List<StructuredRecord> output = MockSink.readOutput(outputManager);
+    Assert.assertEquals(1, output.size());
+    Assert.assertEquals(Integer.MAX_VALUE, output.get(0).get("i"));
```
albertshau: as mentioned before, this is an incomplete comparison. Fields are missing and schema is wrong, but the test is not catching that. It should be comparing object directly instead of a couple fields.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -123,6 +128,7 @@ public FileSourceConfig(String referenceName, @Nullable String fileRegex, @Nulla
     this.timeTable = timeTable;
     this.inputFormatClass = inputFormatClass == null ?
       CombinePathTrackingInputFormat.class.getName() : inputFormatClass;
+    this.fileSourceFormat = fileSourceFormat == null ? "text" : fileSourceFormat;
```
albertshau: you need to add some validation logic, to make sure the given format is a valid one.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -77,6 +78,10 @@
   public String inputFormatClass;
 
   @Nullable
+  @Description(FILE_SOURCE_FORMAT_DESCRIPTION)
+  public String fileSourceFormat;
```
albertshau: just name it format.

albertshau (commented):
Documentation and widget configs also need to be modified.

sheivin:
@albertshau addressed comments

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -143,6 +144,10 @@ protected void validate() {
       throw new IllegalArgumentException("pathField can only be used if inputFormatClass is " +
                                            CombinePathTrackingInputFormat.class.getName());
     }
+    if (!format.matches("[a-zA-Z]+") && !(format.equalsIgnoreCase("text") || (format.equalsIgnoreCase("avro")) ||
```
albertshau: don't need the first comparison, just the comparisons to text, avro, and parquet.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -125,125 +121,88 @@ public NullWritable getCurrentKey() throws IOException, InterruptedException {
...
     @Override
-    public float getProgress() throws IOException, InterruptedException {
-      return delegate.getProgress();
+    public boolean nextKeyValue() throws IOException, InterruptedException {
+      return delegate.nextKeyValue();
     }
 
     @Override
     public void close() throws IOException {
+      System.out.println("* * * ** * * * * " + schema.toString());
```
albertshau: assuming this was for debugging, should remove.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -125,125 +121,88 @@ public NullWritable getCurrentKey() throws IOException, InterruptedException {
...
-      return recordTransformer.transform(delegate.getCurrentKey().datum());
-    }
+  private static class TrackingAvroRecordReader extends TrackingRecordReader<AvroKey<GenericRecord>, NullWritable> {
+    private static AvroToStructuredTransformer recordTransformer;
 
-    @Override
-    public boolean nextKeyValue() throws IOException, InterruptedException {
-      return delegate.nextKeyValue();
+    private TrackingAvroRecordReader(RecordReader<AvroKey<GenericRecord>, NullWritable> delegate,
+                                 @Nullable String pathField, String path) {
```
albertshau: nit: fix indentation

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -478,88 +479,28 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
-                              ImmutableMap.<String, String>builder()
-                                .put(Constants.Reference.REFERENCE_NAME, "TestFile")
-                                .put(Properties.File.FILESYSTEM, "Text")
-                                .put(Properties.File.PATH, filePath)
-                                .put(Properties.File.FILE_SOURCE_FORMAT, "text")
-                                .put(Properties.File.IGNORE_NON_EXISTING_FOLDERS, "false")
-                                .build(),
-                              null));
-
+    File fileText = temporaryFolder.newFolder("test2").toPath().resolve(fileName + "-test5.txt").toFile();
```
albertshau: This is a weird way to get a File, just do something like

 `File fileText = new File(temporaryFolder.newFolder(), "test.txt");`

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -478,88 +479,28 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
-    workflowManager.start();
-    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+    workflowStartAndWait(appManager);
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
     List<StructuredRecord> output = MockSink.readOutput(outputManager);
 
-    Assert.assertEquals(1, output.size());
-    Assert.assertEquals(testData, output.get(0).get("body"));
+    Assert.assertEquals("Hello,World", output.get(0).get("body"));
```
albertshau: you shouldn't be testing equality by looking up a single field in a record. Create a `List<StructuredRecord> expected` variable and do `Assert.assertEquals(expected, output)`

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -478,88 +479,28 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
-                                    ImmutableMap.<String, String>builder()
-                                      .put(Constants.Reference.REFERENCE_NAME, "TestFile")
-                                      .put(Properties.File.FILESYSTEM, "Text")
-                                      .put(Properties.File.PATH, filePath)
-                                      .put(Properties.File.FILE_SOURCE_FORMAT, "Avro")
-                                      .put(Properties.File.IGNORE_NON_EXISTING_FOLDERS, "false")
-                                      .build(),
-                                      null));
-
+    File fileAvro = temporaryFolder.newFolder("test2").toPath().resolve(fileName + "-test6.avro").toFile();
```
albertshau: you shouldn't be relying on a hardcoded filename for uniqueness, as that's pretty brittle and requires somebody to look at all tests in order to know how to name their file.

Instead, do `temporaryFolder.newFolder()` to create a new, unique folder for each test.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -569,45 +510,72 @@ public void testFileBatchInputFormatAvro() throws Exception {
...
     dataFileWriter.append(record);
     dataFileWriter.close();
     inputManager.flush();
 
-    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
-    workflowManager.start();
-    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+    workflowStartAndWait(appManager);
+
+    List<StructuredRecord> input = ImmutableList.of(
```
albertshau: input is a bad name for this, can name is 'expected'

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -569,45 +510,72 @@ public void testFileBatchInputFormatAvro() throws Exception {
...
-    Assert.assertEquals(Long.MAX_VALUE, output.get(0).get("l"));
+    Assert.assertEquals(input, output);
   }
 
   @Test
   public void testFileBatchInputFormatParquet() throws Exception {
-    Schema recordSchema = Schema.recordOf("record",
-                                          Schema.Field.of("i", Schema.of(Schema.Type.INT)),
-                                          Schema.Field.of("l", Schema.of(Schema.Type.LONG)));
+    File fileParquet = temporaryFolder.newFolder("test2").toPath().resolve(fileName + "-test7.parquet").toFile();
```
albertshau: similar comment here

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -617,29 +585,15 @@ public void testFileBatchInputFormatParquet() throws Exception {
...
     workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
-
-    DataSetManager<Table> outputManager = getDataset(outputDatasetName);
-    List<StructuredRecord> output = MockSink.readOutput(outputManager);
-    Assert.assertEquals(1, output.size());
-    Assert.assertEquals(Integer.MAX_VALUE, output.get(0).get("i"));
-    Assert.assertEquals(Long.MAX_VALUE, output.get(0).get("l"));
   }
-}
+}
```
albertshau: files need to be ended with a newline

core-plugins/widgets/File-batchsource.json
```diff
@@ -23,6 +23,11 @@
         },
         {
           "widget-type": "textbox",
+          "label": "Format",
```
albertshau: this should be a dropdown instead of a textbox. You can do this with a 'select' widget type. See Stream-batchsource.json for an example.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -569,45 +510,72 @@ public void testFileBatchInputFormatAvro() throws Exception {
...
     dataFileWriter.close();
     inputManager.flush();
 
-    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
-    workflowManager.start();
-    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+    workflowStartAndWait(appManager);
+
+    List<StructuredRecord> input = ImmutableList.of(
+      StructuredRecord.builder(RECORD_SCHEMA)
```
albertshau: this is the wrong output. The pipeline is deployed with 'pathField' as 'file', which means the output is supposed to contain a 'file' field whose value is the URI of the file.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -125,125 +121,88 @@ public NullWritable getCurrentKey() throws IOException, InterruptedException {
...
+    public void initialize(InputSplit split, TaskAttemptContext context) throws IOException, InterruptedException {
+      super.initialize(split, context);
+      recordTransformer = new AvroToStructuredTransformer();
     }
 
-    @Override
-    public void close() throws IOException {
-      delegate.close();
+    public StructuredRecord getCurrentValue(StructuredRecord.Builder recordBuilder) throws IOException, InterruptedException {
+      return recordTransformer.transform(delegate.getCurrentKey().datum());
```
albertshau: the point of passing in the recordBuilder is for you to use it here. The builder already has the path field set, and it is supposed to have been setup with the correct output schema. Without using it, you will output the wrong record. You can add a new method to AvroToStructuredTransformer that lets you pass in a GenericRecord, Schema, StructuredRecord.Builder, and a field to skip.

albertshau: this also made me realize the getOutputSchema method in this class has not been changed, which means the schema in the builder is only correct for the 'text' format. 

Now that we support avro and parquet, there needs to be an additional 'schema' property added to the plugin's config. The schema minus the path field should be the schema used to read the data. 

For AvroKeyInputFormat, this means setting the 'avro.schema.input.key'. For AvroParquetInputFormat, this means setting the 'parquet.avro.read.schema'. You can see where those come from by reading the respective input format. If the source is not given a schema, it should use the current behavior, which uses the write schema as the read schema. 

You can handle the case where schema is not specified in a separate PR, as it involves modifying schemas on the fly, since schema is not longer constant.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
sheivin: Oh yeah! I forgot to remove this

PR #688, opened 2017-07-19
CDAP-12137 add null field splitter

albertshau:
Adding a SplitterTransform that has two output ports, one for
records with a null field value and one for records that have
a non-null value.

### DIFF
--- a/transform-plugins/src/main/java/co/cask/hydrator/plugin/NullFieldSplitter.java
+++ b/transform-plugins/src/main/java/co/cask/hydrator/plugin/NullFieldSplitter.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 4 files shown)


transform-plugins/src/main/java/co/cask/hydrator/plugin/NullFieldSplitter.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+    return Schema.recordOf(nullableSchema.getRecordName() + ".nonnull", fields);
+  }
+
+  /**
+   * Configuration for the plugin.
+   */
+  public static class Conf extends PluginConfig {
+    @Description("Which field should be checked for null values.")
+    private final String field;
```
gsps1: should we allow this to be a macro ?

gsps1 (commented):
just one question, LGTM otherwise.

transform-plugins/src/main/java/co/cask/hydrator/plugin/NullFieldSplitter.java
albertshau: can make it macro enabled, will update.

gsps1:
LGTM

PR #689, opened 2017-07-20
CDAP-12136 add union splitter transform

albertshau:
The union splitter will split input into multiple outputs
depending on the values in a union field. Each possible schema
in the union will have its own output port. If no field is given,
each schema received as input will have its own output port.

### DIFF
--- a/transform-plugins/docs/UnionSplitter-splittertransform.md
+++ b/transform-plugins/docs/UnionSplitter-splittertransform.md
@@ -1,0 +1,132 @@
+# Union Splitter
+
+
+Description
+-----------
+The union splitter is used to split data by a union schema, so that type specific logic can be done downstream.
+
+The union splitter will emit records to different ports depending on the schema of a particular field, or of
+the entire record. If no field is specified, each record will be emitted to a port named after the name of the
+record schema. If a field is specified, the schema for that field must be a union of supported schemas. All schemas
+except maps, arrays, unions, and enums are supported. For each input record, the value of that field will be examined
+and emitted to a port corresponding to its schema in the union.
+
+For record schemas, the output port will be the name of the record schema. For simple types, the output port will
+be the schema type in lowercase ('null', 'bool', 'bytes', 'int', 'long', 'float', 'double', or 'string').
+
+
+Properties
+----------
+**unionField:** The union field to split on. If specified, the schema for the field must be a union of
+supported schemas. All schemas except maps, arrays, unions, and enums are supported. Note that nulls are supported,
+which means all nulls will get sent to the 'null' port.
+
+**modifySchema:** Whether to modify the output schema to remove the union. For example, suppose the field 'x'
+is a union of int and long. If modifySchema is true, the schema for field 'x' will be just an int for
+the 'int' port and just a long for the 'long' port. If modifySchema is false, the output schema for each port
+will be the same as the input schema. Defaults to true.
+
+
+Example 1: Splitting on the entire record
+-----------------------------------------
+Suppose the union splitter is configured to split on the entire record:
+
+    {
+        "name": "UnionSplitter",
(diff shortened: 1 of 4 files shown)


transform-plugins/docs/UnionSplitter-splittertransform.md
```diff
@@ -0,0 +1,130 @@
...
+and emitted to a port corresponding to its schema in the union.
+
+For record schemas, the output port will be the name of the record schema. For simple types, the output port will
+be the schema type in lowercase ('null', 'bool', 'bytes', 'int', 'long', 'float', 'double', or 'string').
+
+
+Properties
+----------
+**field:** The union field to split on. If specified, the schema for the field must be a union of
+supported schemas. All schemas except maps, arrays, unions, and enums are supported.
```
CuriousVini: should we add some information on what happens when the union schema has `null` type?

transform-plugins/widgets/UnionSplitter-splittertransform.json
```diff
@@ -0,0 +1,30 @@
...
+    "spec-version": "1.4"
+  },
+  "configuration-groups": [
+    {
+      "label": "Union Splitter",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Union field to split on",
+          "name": "field"
```
CuriousVini: should we make this `unionField`?

transform-plugins/docs/UnionSplitter-splittertransform.md
```diff
@@ -0,0 +1,130 @@
...
+except maps, arrays, unions, and enums are supported. For each input record, the value of that field will be examined
+and emitted to a port corresponding to its schema in the union.
+
+For record schemas, the output port will be the name of the record schema. For simple types, the output port will
+be the schema type in lowercase ('null', 'bool', 'bytes', 'int', 'long', 'float', 'double', or 'string').
+
+
+Properties
+----------
+**field:** The union field to split on. If specified, the schema for the field must be a union of
```
CuriousVini: should we call it `unionField`?

transform-plugins/src/main/java/co/cask/hydrator/plugin/UnionSplitter.java
```diff
@@ -0,0 +1,245 @@
...
+
+  public UnionSplitter(Conf conf) {
+    this.conf = conf;
+  }
+
+  @Override
+  public void configurePipeline(MultiOutputPipelineConfigurer multiOutputPipelineConfigurer) {
+    MultiOutputStageConfigurer stageConfigurer = multiOutputPipelineConfigurer.getMultiOutputStageConfigurer();
+    Schema inputSchema = stageConfigurer.getInputSchema();
+    if (inputSchema == null) {
```
CuriousVini, line 60: should we throw error if inputSchema is null?

transform-plugins/docs/UnionSplitter-splittertransform.md
```diff
@@ -0,0 +1,130 @@
...
+
+**modifySchema:** Whether to modify the output schema to remove the union. For example, suppose the field 'x'
+is a union of int and long. If modifySchema is true, the schema for field 'x' will be just an int for
+the 'int' port and just a long for the 'long' port. If modifySchema is false, the output schema for each port
+will be the same as the input schema. Defaults to true.
+
+
+Example 1: Splitting on the entire record
+-----------------------------------------
+Suppose the union splitter is configured to split on the entire record:
```
CuriousVini, line 32: For this usecase, if we have 2 of same plugins (eg. projection) connected to splitter, then the output schema name will be same for them, in that case the records will go to same output port. We should document that here.

CuriousVini (commented):
couple of minor changes. Rest looks good.

PR #690, opened 2017-07-20
bump parent artifact

elfenheart:


spark-plugins/pom.xml
```diff
@@ -24,7 +24,7 @@
   </parent>
 
   <properties>
```
albertshau, line 26: this can actually be removed, it's the same as the parent now that etl-batch and etl-realtime are gone.

albertshau (commented):
one comment

albertshau (approved):
lgtm

PR #691, opened 2017-07-21
Fix compilation errors caused by new emitAlert APIs.

anwar6953:
Fix compilation. New API was introduced in CDAP in https://github.com/caskdata/cdap/pull/9273.


albertshau (approved):
lgtm

PR #693, opened 2017-07-22
Add S3 to HDFS File copy functionality.

caskchester:
1. Adds a file metadata source template that can be extended to read
   file metadata from any source filesystem.

2. Adds a file copy sink template that can be exended to copy files
   from a remote source given file metadata as inputs.

3. Added InputFormat and OutputFormat to support source and sink plugins.

### DIFF
--- a/core-plugins/docs/HDFSFileCopySink-batchsink.md
+++ b/core-plugins/docs/HDFSFileCopySink-batchsink.md
@@ -35,7 +35,7 @@ This sink plugin only reads StructuredRecords with the following schema. Each re
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
+| **relativePath**       | String | The relative path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
 | **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
 | **permission**         | int    | The file's access permission                                                                                                                   |
 | **Credentials**        | Record | Additional information required to connect to the source Filesystem.                                                                           |
--- a/core-plugins/docs/HDFSFileCopySink-batchsink.md
+++ b/core-plugins/docs/HDFSFileCopySink-batchsink.md
@@ -1,0 +1,37 @@
+# HDfS File Copy Batch Sink
+
+Description
+-----------
+The HDFS File Copy plugin is a sink plugin that takes file metadata records as inputs and copies the files into local HDFS
+
+
+Use Case
+--------
+Use this sink to copy files from any source to your local HDFS.
+You may want to periodically sync your HDFS with some remote filesystem. Schedule to run a pipeline with this plugin to copy new files periodically.
+
+
+Properties
+----------
+| Configuration          | Required | Default   | Description                                                                                 |
+| :--------------------- | :------: | :------   | :------------------------------------------------------------------------------------------ |
+| **Reference Name**     |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.   |
+| **Base Path**          |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.   |
+| **Enable Overwrite**   |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                           |
+| **Buffer Size**        |  **Y**   | True      | The size of the buffer that temporarily stores data from file input stream while copying.   |
+
+Usage Notes
+-----------
+This sink plugin only reads StructuredRecords with the follwing schema. Each record should contain the metadata for a file to be copied
+
+| Field                  | Type   | Description                                                                                                                                    |
+| :--------------------- | :----- | :-------------------------                                                                                                                     |
+| **File Name**          | String | Only contains the name of the file.                                                                                                            |
+| **Full Path**          | String | Contains the full path of the file in the source file system.                                                                                  |
+| **File Size**          | long   | Specifies the number of fi                                                                                                                     |
+| **Timestamp**          | long   | The modification timestamp of the file.                                                                                                        |
+| **Owner**              | String | The owner of the file.                                                                                                                         |
+| **Is Folder**          | Boolean| Whether or not the file is a folder.                                                                                                           |
+| **Base Path**          | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **Data Base Type**     | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
+| **Credentials**        | Record | Additional information required to connect to the source Filesystem.                                                                           |
--- a/core-plugins/docs/HDFSFileCopySink-batchsink.md
+++ b/core-plugins/docs/HDFSFileCopySink-batchsink.md
@@ -13,14 +13,13 @@ You may want to periodically sync your HDFS with some remote filesystem. Schedul
 
 Properties
 ----------
-| Configuration                            | Required | Default   | Description                                                                                            |
-| :--------------------------------------- | :------: | :------   | :----------------------------------------------------------------------------------------------------- |
-| **Reference Name**                       |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.              |
-| **Base Path**                            |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.              |
-| **Enable Overwrite**                     |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                                      |
-| **Preserve File Owner**                  |  **Y**   | False     | Whether or not to preserve the owner of the file from source filesystem.                               |
-| **Enable Filesystem Connection Caching** |  **Y**   | True      | Whether or not for splits to use cached connections to filesystems.                                    |
-| **Buffer Size**                          |  **Y**   | 1 MB      | The size of the buffer (in Bytes) that temporarily stores data from file input stream while copying.   |
+| Configuration                            | Required | Default   | Description                                                                                                                  |
+| :--------------------------------------- | :------: | :------   | :--------------------------------------------------------------------------------------------------------------------------- |
+| **Reference Name**                       |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.                                    |
+| **Base Path**                            |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.                                    |
+| **Enable Overwrite**                     |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                                                            |
+| **Preserve File Owner**                  |  **Y**   | False     | Whether or not to preserve the owner of the file from source filesystem.                                                     |
+| **Buffer Size**                          |  **Y**   | 1 MB      | The size of the buffer (in MegaBytes) that temporarily stores data from file input stream while copying. Defaults to 1 MB.   |
 
 Usage Notes
 -----------
@@ -36,7 +35,7 @@ This sink plugin only reads StructuredRecords with the following schema. Each re
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
 | **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
 | **permission**         | int    | The file's access permission                                                                                                                   |
 | **Credentials**        | Record | Additional information required to connect to the source Filesystem.                                                                           |
--- a/core-plugins/docs/HDFSFileCopySink-batchsink.md
+++ b/core-plugins/docs/HDFSFileCopySink-batchsink.md
@@ -13,25 +13,30 @@ You may want to periodically sync your HDFS with some remote filesystem. Schedul
 
 Properties
 ----------
-| Configuration          | Required | Default   | Description                                                                                 |
-| :--------------------- | :------: | :------   | :------------------------------------------------------------------------------------------ |
-| **Reference Name**     |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.   |
-| **Base Path**          |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.   |
-| **Enable Overwrite**   |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                           |
-| **Buffer Size**        |  **Y**   | True      | The size of the buffer that temporarily stores data from file input stream while copying.   |
+| Configuration                            | Required | Default   | Description                                                                                            |
+| :--------------------------------------- | :------: | :------   | :----------------------------------------------------------------------------------------------------- |
+| **Reference Name**                       |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.              |
+| **Base Path**                            |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.              |
+| **Enable Overwrite**                     |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                                      |
+| **Preserve File Owner**                  |  **Y**   | False     | Whether or not to preserve the owner of the file from source filesystem.                               |
+| **Enable Filesystem Connection Caching** |  **Y**   | True      | Whether or not for splits to use cached connections to filesystems.                                    |
+| **Buffer Size**                          |  **Y**   | 1 MB      | The size of the buffer (in Bytes) that temporarily stores data from file input stream while copying.   |
 
 Usage Notes
 -----------
-This sink plugin only reads StructuredRecords with the follwing schema. Each record should contain the metadata for a file to be copied
+This sink plugin only reads StructuredRecords with the following schema. Each record should contain the metadata for a file to be copied
 
 | Field                  | Type   | Description                                                                                                                                    |
 | :--------------------- | :----- | :-------------------------                                                                                                                     |
-| **File Name**          | String | Only contains the name of the file.                                                                                                            |
-| **Full Path**          | String | Contains the full path of the file in the source file system.                                                                                  |
-| **File Size**          | long   | Specifies the number of fi                                                                                                                     |
-| **Timestamp**          | long   | The modification timestamp of the file.                                                                                                        |
-| **Owner**              | String | The owner of the file.                                                                                                                         |
-| **Is Folder**          | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **Base Path**          | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
-| **Data Base Type**     | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
+| **fileName**           | String | Only contains the name of the file.                                                                                                            |
+| **fullPath**           | String | Contains the full path of the file in the source file system.                                                                                  |
+| **fileSize**           | long   | File size in bytes.                                                                                                                            |
+| **hostURI**            | String | The source filesystem's URI.                                                                                                                   |
+| **modificationTime**   | long   | The modification timestamp of the file.                                                                                                        |
+| **group**              | String | The group that the of the file belongs to.                                                                                                     |
+| **owner**              | String | The owner of the file.                                                                                                                         |
+| **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
+| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
+| **permission**         | int    | The file's access permission                                                                                                                   |
 | **Credentials**        | Record | Additional information required to connect to the source Filesystem.                                                                           |
--- a/core-plugins/docs/S3FileMetadataSource-batchsource.md
+++ b/core-plugins/docs/S3FileMetadataSource-batchsource.md
@@ -18,10 +18,10 @@ Properties
 | **Reference Name**     |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.                                                                                                                                                                                       |
 | **Source Paths**       |  **Y**   | None      | Path(s) to file(s) to be read. If a directory is specified, terminate the path name with a '/'.                                                                                                                                                                                 |
 | **Max Split Size**     |  **Y**   | None      | Specifies the number of files that are controlled by each split. The number of splits created will be the total number of files divided by Max Split Size. The InputFormat will assign roughly the same number of bytes to each split.                                          |
+| **Host URI**           |  **Y**   | None      | URI of the bucket which you want to read from. For example, if you want to read from s3 bucket `example.bucket.co`, this configuration would be `s3a://example.bucket.co`                                                                                                       |
 | **Copy Recursively**   |  **Y**   | True      | Whether or not to copy recursively. Similar to the '-r' option in the linux cp command. Set this to true if you want to copy the entire directory.                                                                                                                              |
 | **Access Key ID**      |  **Y**   | None      | Your Amazon S3 access key ID.                                                                                                                                                                                                                                                   |
 | **Secret Key ID**      |  **Y**   | None      | Your Amazon S3 secret key ID.                                                                                                                                                                                                                                                   |
-| **Bucket Name**        |  **Y**   | None      | Address of the bucket which you want to read from.                                                                                                                                                                                                                              |
 | **Amazon Region**      |  **N**   | us-east-1 | Region of the bucket which you want to read from.                                                                                                                                                                                                                               |
 
 Usage Notes
@@ -31,14 +31,17 @@ A structured record with the following schema is emitted for each file it reads.
 
 | Field                  | Type   | Description                                                                                                                                    |
 | :--------------------- | :----- | :-------------------------                                                                                                                     |
-| **File Name**          | String | Only contains the name of the file.                                                                                                            |
-| **Full Path**          | String | Contains the full path of the file in the source file system.                                                                                  |
-| **File Size**          | long   | Specifies the number of fi                                                                                                                     |
-| **Timestamp**          | long   | The modification timestamp of the file.                                                                                                        |
-| **Owner**              | String | The owner of the file.                                                                                                                         |
-| **Is Folder**          | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **Base Path**          | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
-| **Data Base Type**     | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
-| **Access Key ID**      | String | Access Key ID for the source Filesystem.                                                                                                       |
-| **Secret Key ID**      | String | Secret Key ID for the source Filesystem.                                                                                                       |
-| **Bucket Name**        | String | The bucket which this file came from.                                                                                                          |
+| **fileName**           | String | Only contains the name of the file.                                                                                                            |
+| **fullPath**           | String | Contains the full path of the file in the source file system.                                                                                  |
+| **fileSize**           | long   | File Szie in bytes.                                                                                                                            |
+| **hostURI**            | String | The source filesystem's URI.                                                                                                                   |
+| **modificationTime**   | long   | The modification timestamp of the file.                                                                                                        |
+| **group**              | String | The group that the of the file belongs to.                                                                                                     |
+| **owner**              | String | The owner of the file.                                                                                                                         |
+| **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
+| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
+| **permission**         | int    | The file's access permission                                                                                                                   |
+| **accessKeyID**        | String | Access Key ID for the source Filesystem.                                                                                                       |
+| **secretKeyID**        | String | Secret Key ID for the source Filesystem.                                                                                                       |
+| **region**             | String | Region of the source Filesystem.                                                                                                               |
--- a/core-plugins/docs/S3FileMetadataSource-batchsource.md
+++ b/core-plugins/docs/S3FileMetadataSource-batchsource.md
@@ -39,7 +39,7 @@ A structured record with the following schema is emitted for each file it reads.
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
 | **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
 | **permission**         | int    | The file's access permission                                                                                                                   |
 | **accessKeyID**        | String | Access Key ID for the source Filesystem.                                                                                                       |
--- a/core-plugins/docs/S3FileMetadataSource-batchsource.md
+++ b/core-plugins/docs/S3FileMetadataSource-batchsource.md
@@ -39,7 +39,7 @@ A structured record with the following schema is emitted for each file it reads.
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
+| **relativePath**       | String | The relative path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
 | **filesystem**         | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
 | **permission**         | int    | The file's access permission                                                                                                                   |
 | **accessKeyID**        | String | Access Key ID for the source Filesystem.                                                                                                       |
--- a/core-plugins/docs/S3FileMetadataSource-batchsource.md
+++ b/core-plugins/docs/S3FileMetadataSource-batchsource.md
@@ -1,0 +1,44 @@
+# S3 File Metadata Batch Source
+
+Description
+-----------
+The S3 File Metadata plugin is a source plugin that allows users to read file metadata from an S3 Filesystem.
+
+
+Use Case
+--------
+Use this source to extract the metadata of files under specified paths. The metadata can be passed to the
+HDFSFileCopySink and the files will be copied from S3 to HDFS
+
+
+Properties
+----------
+| Configuration          | Required | Default   | Description                                                                                                                                                                                                                                                                     |
+| :--------------------- | :------: | :------   | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Reference Name**     |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.                                                                                                                                                                                       |
+| **Source Paths**       |  **Y**   | None      | Path(s) to file(s) to be read. If a directory is specified, terminate the path name with a '/'.                                                                                                                                                                                 |
+| **Max Split Size**     |  **Y**   | None      | Specifies the number of files that are controlled by each split. The number of splits created will be the total number of files divided by Max Split Size. The InputFormat will assign roughly the same number of bytes to each split.                                          |
+| **Copy Recursively**   |  **Y**   | True      | Whether or not to copy recursively. Similar to the '-r' option in the linux cp command. Set this to true if you want to copy the entire directory.                                                                                                                              |
+| **Access Key ID**      |  **Y**   | None      | Your Amazon S3 access key ID.                                                                                                                                                                                                                                                   |
+| **Secret Key ID**      |  **Y**   | None      | Your Amazon S3 secret key ID.                                                                                                                                                                                                                                                   |
+| **Bucket Name**        |  **Y**   | None      | Address of the bucket which you want to read from.                                                                                                                                                                                                                              |
+| **Amazon Region**      |  **N**   | us-east-1 | Region of the bucket which you want to read from.                                                                                                                                                                                                                               |
+
+Usage Notes
+-----------
+This source plugin only reads filemetadata from the source. To copy files, pass its outputs to a FileCopySink of the desired Filesystem.
+A structured record with the following schema is emitted for each file it reads.
+
+| Field                  | Type   | Description                                                                                                                                    |
+| :--------------------- | :----- | :-------------------------                                                                                                                     |
+| **File Name**          | String | Only contains the name of the file.                                                                                                            |
+| **Full Path**          | String | Contains the full path of the file in the source file system.                                                                                  |
+| **File Size**          | long   | Specifies the number of fi                                                                                                                     |
+| **Timestamp**          | long   | The modification timestamp of the file.                                                                                                        |
+| **Owner**              | String | The owner of the file.                                                                                                                         |
+| **Is Folder**          | Boolean| Whether or not the file is a folder.                                                                                                           |
+| **Base Path**          | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **Data Base Type**     | String | Contains the string "amazons3". Used to identify the type of filesystem this record originated from.                                           |
+| **Access Key ID**      | String | Access Key ID for the source Filesystem.                                                                                                       |
+| **Secret Key ID**      | String | Secret Key ID for the source Filesystem.                                                                                                       |
+| **Bucket Name**        | String | The bucket which this file came from.                                                                                                          |
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
@@ -147,6 +147,7 @@ public abstract class AbstractFileCopySink
       FileCopyOutputFormat.setBasePath(conf, config.basePath);
       FileCopyOutputFormat.setEnableOverwrite(conf, config.enableOverwrite.toString());
       FileCopyOutputFormat.setPreserveFileOwner(conf, config.preserveFileOwner.toString());
+      FileCopyOutputFormat.setFilesystemScheme(conf, config.getScheme());
 
       if (config.bufferSize != null) {
         // bufferSize is in megabytes
@@ -154,9 +155,6 @@ public abstract class AbstractFileCopySink
       } else {
         FileCopyOutputFormat.setBufferSize(conf, String.valueOf(FileCopyRecordWriter.DEFAULT_BUFFER_SIZE));
       }
-
-      // always disable caching
-      conf.put(String.format("fs.%s.impl.disable.cache", config.getScheme()), String.valueOf(true));
 
       // set the URI for the destination filesystem if it's provided
       if (config.getHostUri() != null) {
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright © 2017 Cask Data, Inc.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *  
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *  
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.hydrator.plugin.batch.file;
+
+import co.cask.cdap.api.annotation.Description;
+import co.cask.cdap.api.annotation.Macro;
+import co.cask.cdap.api.data.batch.OutputFormatProvider;
+import co.cask.cdap.api.data.format.StructuredRecord;
+import co.cask.cdap.api.dataset.lib.KeyValue;
+import co.cask.cdap.etl.api.Emitter;
+import co.cask.hydrator.common.ReferenceBatchSink;
+import co.cask.hydrator.common.ReferencePluginConfig;
+import co.cask.hydrator.plugin.batch.file.s3.S3FileMetadata;
+import co.cask.hydrator.plugin.batch.file.s3.S3FileMetadata.S3Credentials;
+import com.sun.istack.Nullable;
+import org.apache.hadoop.io.NullWritable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Abstract template for a FileCopySink. The transform method converts a structured record
+ * to AbstractFileMetadata class.
+ */
+public abstract class AbstractFileCopySink
+  extends ReferenceBatchSink<StructuredRecord, NullWritable, AbstractFileMetadata> {
+  protected final AbstractFileCopySinkConfig config;
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractFileCopySink.class);
+
+  public AbstractFileCopySink(AbstractFileCopySinkConfig config) {
+    super(config);
+    this.config = config;
+  }
+
+  /**
+   * Converts input StructuredRecord to AbstractFileMetadata class. Loads credentials and
+   * file metadata from the input.
+   * @param input The input structured record that contains credentials and file metadata.
+   * @param emitter
+   * @throws Exception
+   */
+  @Override
+  public void transform(StructuredRecord input,
+                        Emitter<KeyValue<NullWritable, AbstractFileMetadata>> emitter)
+    throws Exception {
+    AbstractFileMetadata output;
+    switch ((String) input.get(AbstractFileMetadata.Credentials.DATA_BASE_TYPE)) {
+      case S3FileMetadata.DATA_BASE_NAME :
+        output = new S3FileMetadata((String) input.get(AbstractFileMetadata.FILE_NAME),
+                                    (String) input.get(AbstractFileMetadata.FULL_PATH),
+                                    (long) input.get(AbstractFileMetadata.TIMESTAMP),
+                                    (String) input.get(AbstractFileMetadata.OWNER),
+                                    (long) input.get(AbstractFileMetadata.FILE_SIZE),
+                                    (Boolean) input.get(AbstractFileMetadata.IS_FOLDER),
+                                    (String) input.get(AbstractFileMetadata.BASE_PATH),
+                                    (short) input.get(AbstractFileMetadata.PERMISSION),
+                                    new S3FileMetadata.S3Credentials((String) input.get(S3Credentials.ACCESS_KEY_ID),
+                                                                     (String) input.get(S3Credentials.SECRET_KEY_ID),
+                                                                     (String) input.get(S3Credentials.REGION),
+                                                                     (String) input.get(S3Credentials.BUCKET_NAME))
+
+          );
+        break;
+
+      default:
+        throw new IllegalArgumentException("unrecognized database type");
+
+    }
+    emitter.emit(new KeyValue<NullWritable, AbstractFileMetadata>(null, output));
+  }
+
+  /**
+   * Abstract class for the configuration of FileCopySink
+   */
+  public abstract class AbstractFileCopySinkConfig extends ReferencePluginConfig {
+
+    @Macro
+    @Description("The destination path. Will be created if it doesn't exist.")
+    public String basePath;
+
+    @Description("Whether or not to overwrite if the file already exists.")
+    public Boolean enableOverwrite;
+
+    @Macro
+    @Nullable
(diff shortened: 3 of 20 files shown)


caskchester:
Design doc: https://wiki.cask.co/pages/viewpage.action?pageId=4364388

core-plugins/docs/HDFSFileCopySink-batchsink.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+Properties
+----------
+| Configuration          | Required | Default   | Description                                                                                 |
+| :--------------------- | :------: | :------   | :------------------------------------------------------------------------------------------ |
+| **Reference Name**     |  **Y**   | None      | This will be used to uniquely identify this source for lineage, annotating metadata, etc.   |
+| **Base Path**          |  **Y**   | None      | The folder where the copied files will be placed. It will be created if it doesn't exist.   |
+| **Enable Overwrite**   |  **Y**   | False     | Specifies whether or not to overwrite files if it already exists.                           |
+| **Buffer Size**        |  **Y**   | True      | The size of the buffer that temporarily stores data from file input stream while copying.   |
```
albertshau: default should be a number right?

core-plugins/docs/HDFSFileCopySink-batchsink.md
```diff
@@ -0,0 +1,37 @@
...
+
+Usage Notes
+-----------
+This sink plugin only reads StructuredRecords with the follwing schema. Each record should contain the metadata for a file to be copied
```
albertshau: typo: 'following'

core-plugins/docs/HDFSFileCopySink-batchsink.md
```diff
@@ -0,0 +1,37 @@
...
+
+| Field                  | Type   | Description                                                                                                                                    |
+| :--------------------- | :----- | :-------------------------                                                                                                                     |
+| **File Name**          | String | Only contains the name of the file.                                                                                                            |
+| **Full Path**          | String | Contains the full path of the file in the source file system.                                                                                  |
+| **File Size**          | long   | Specifies the number of fi                                                                                                                     |
```
albertshau: incomplete description

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
```diff
@@ -0,0 +1,144 @@
...
+    @Description("The destination path. Will be created if it doesn't exist.")
+    public String basePath;
+
+    @Description("Whether or not to overwrite if the file already exists.")
+    public Boolean enableOverwrite;
+
+    @Macro
+    @Nullable
+    @Description("The size of the buffer that temporarily stores data from file input stream.")
+    public int bufferSize;
```
albertshau: in order for this to be Nullable, it should be an Integer instead of an int.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+  public static final String FILE_SIZE = "fileSize";
+  public static final String TIMESTAMP = "timeStamp";
+  public static final String OWNER = "owner";
+  public static final String FULL_PATH = "fullPath";
+  public static final String IS_FOLDER = "isFolder";
+  public static final String BASE_PATH = "basePath";
+  public static final String PERMISSION = "permission";
+
+  // contains only the name of the file
+  protected String fileName;
```
albertshau: when possible, these should be final. It's easier to maintain and use objects if they are immutable.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+  protected long fileSize;
+
+  // modification time of file
+  protected long timeStamp;
+
+  // owner of file
+  protected String owner;
+
+  // whether or not the file is a folder
+  protected Boolean isFolder;
```
albertshau: this can be a boolean instead of Boolean so there is no null possibility.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public String getOwner() {
+    return owner;
+  }
+
+  public abstract Credentials getCredentials();
+
+  public Boolean getIsFolder() {
+    return this.isFolder;
```
albertshau: nit: for consistency, `return isFolder `

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+    return timeStamp;
+  }
+
+  public String getOwner() {
+    return owner;
+  }
+
+  public abstract Credentials getCredentials();
+
+  public Boolean getIsFolder() {
```
albertshau: nit: can name this isFolder()

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+
+  // Credentials needed to connect to the filesystem that contains this file
+  protected final Credentials credentials;
+
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractFileMetadata.class);
+
+  public AbstractFileMetadata(FileStatus fileStatus, String sourcePath, Credentials credentials) {
+    String fullPath = fileStatus.getPath().toString();
+    String[] paths = fullPath.split("/");
+    this.fileName = paths[paths.length - 1];
```
albertshau: the Path object from fileStatus.getPath() has a getName(). No need to split on the full path.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+  protected String fileName;
+
+  // full path of the file in the source filesystem
+  protected String fullPath;
+
+  // file size
+  protected long fileSize;
+
+  // modification time of file
+  protected long timeStamp;
```
albertshau: nit: should probably name this modificationTime or something more descriptive.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
```diff
@@ -0,0 +1,144 @@
...
+   * @throws Exception
+   */
+  @Override
+  public void transform(StructuredRecord input,
+                        Emitter<KeyValue<NullWritable, AbstractFileMetadata>> emitter)
+    throws Exception {
+    AbstractFileMetadata output;
+    switch ((String) input.get(AbstractFileMetadata.Credentials.DATA_BASE_TYPE)) {
+      case S3FileMetadata.DATA_BASE_NAME :
+        output = new S3FileMetadata((String) input.get(AbstractFileMetadata.FILE_NAME),
```
albertshau: I think it's better to put the AbstractFileMetadata to and from StructuredRecord conversion logic inside the AbstractFileMetadata class, through a `toRecord()` method and a `static fromRecord(StructuredRecord record)` method.
Then the logic is in a single place, it can be re-used in other classes, and the field names can be private

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+
+  public short getPermission() {
+    return permission;
+  }
+
+  /**
+   * extend from this class to create a credential class for each filesystem
+   */
+  public abstract static class Credentials {
+    public String databaseType;
```
albertshau: this isn't really a credential, it seems like another attribute of the file. 'database' is also a loaded term, as people usually think of sql based relational databases. We should use 'fileSystem' instead.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // generate file folder
+    this.fullPath = fullPath;
+
+    // check if file is a folder
+    this.isFolder = fileStatus.isDirectory();
+
+    // TODO: investigate how to cleanly set basePath
+    if (sourcePath.equals("")) {
+      basePath = this.fileName;
```
albertshau: what does this mean? I'm confused about what basePath is. It's not the directory the files came from?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+        sourcePath = sourcePath.substring(0, sourcePath.length() - 1);
+      } else {
+        numStrip = numSourcePaths - 1;
+      }
+
+      /*
+       * this block of code drops the URI and reconstructs basePath by
+       * concatenating the folders with "/"
+       */
+      String pathWithoutURI = fullPath.substring(fullPath.indexOf(sourcePath));
```
albertshau: is this trying to strip the scheme (change hdfs:///abc to /abc)? you can use the Path object's toURI().getPath()

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+      String[] pathsWithoutURI = pathWithoutURI.split("/");
+      basePath = "";
+      for (int i = numStrip; i < pathsWithoutURI.length; i++) {
+        basePath = basePath.concat(pathsWithoutURI[i] + "/");
+      }
+      if (basePath.length() > 0) {
```
albertshau: and basePath is the sourcePath with the front stripped as well?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+        basePath = basePath.substring(0, basePath.length() - 1);
+      }
+    }
+  }
+
+  public AbstractFileMetadata(String fileName, String fullPath, long timeStamp, String owner,
```
albertshau: abstract classes should have protected constructors.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+                              Long fileSize, Boolean isFolder, String basePath,
```
albertshau: Long and Boolean should be long and boolean

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadataSource.java
```diff
@@ -0,0 +1,114 @@
...
+    public Boolean recursiveCopy;
+
+    public AbstractFileMetadataSourceConfig(String name, String sourcePaths, int maxSplitSize) {
+      super(name);
+      this.sourcePaths = sourcePaths;
+      this.maxSplitSize = maxSplitSize;
+    }
+
+    public void validate() {
+      }
```
albertshau: fix indentation

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadataSource.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  public abstract class AbstractFileMetadataSourceConfig extends ReferencePluginConfig {
+
+    @Macro
+    @Description("Collection of sourcePaths separated by \",\" to read files from")
+    public String sourcePaths;
+
+    @Macro
+    @Description("The number of files each split manipulates")
+    public int maxSplitSize;
```
albertshau: Macro enabled properties should be Integer, as it will be empty at configure time when there is nothing to substitute.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+ * read file metadata. The getsplit method creates splits according to user-set configuration
+ * and tries to assign files to each split such that every split copies roughly the same number of bytes.
+ * @param <KEY>
+ * @param <VALUE>
+ */
+public abstract class AbstractMetadataInputFormat<KEY, VALUE> extends InputFormat<KEY, VALUE> {
+
+  protected static final String SOURCE_PATHS = "source.paths";
+  protected static final String MAX_SPLIT_SIZE = "max.split.size";
+  protected static final String FS_URI = "uri";
```
albertshau: filesystem.uri would be more clear

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+  public AbstractMetadataInputFormat() {
+  }
+
+  @Override
+  public List<InputSplit> getSplits(JobContext jobContext) throws IOException, InterruptedException {
+    // connect to the source filesystem
+    Configuration conf = jobContext.getConfiguration();
+    URI uri = URI.create(conf.get(FS_URI));
+    String[] sourcePaths = conf.get(SOURCE_PATHS).split(",");
+    Boolean recursiveCopy = conf.getBoolean(RECURSIVE_COPY, true);
```
albertshau: boolean, since it can't be null

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Override
+  public RecordReader createRecordReader(InputSplit inputSplit, TaskAttemptContext taskAttemptContext)
+    throws IOException, InterruptedException {
+    MetadataRecordReader recordReader = new MetadataRecordReader();
+    recordReader.initialize(inputSplit, taskAttemptContext);
+    return recordReader;
+  }
+
+  private void recursivelyAddFileStatus (List<FileStatus> fileStatuses, List<String> baseFolders,
```
albertshau: remove extra space after AddFileStatus

albertshau: FileSystem also has a listFiles(Path, boolean) method that can list recursively. Should use that instead of your own unless there is some missing functionality with it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+    List<FileStatus> fileStatuses = new ArrayList<>();
+    List<String> baseFolders = new ArrayList<>();
+    for (String prefix : sourcePaths) {
+      try {
+        FileStatus fileStatus = fileSystem.getFileStatus(new Path(prefix));
+        if (fileStatus.isFile()) {
+          prefix = "";
+        }
+        recursivelyAddFileStatus(fileStatuses, baseFolders, prefix, fileStatus, recursiveCopy, fileSystem);
+      } catch (Exception e) {
```
albertshau: exceptions need to be handled better. For example, you should probably let a FileNotFoundException on the source path bubble up and fail the job. In general, it seems like the IOExceptions that can be thrown here should cause the job to fail, not logged and skipped.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+
+    fileSystem.close();
+
+    // create a filemetadata list of all the files
+    List<AbstractFileMetadata> fileMetaDataList = getFileMetadaDataList(fileStatuses, baseFolders,
+                                                                        fileStatuses.size(),
+                                                                        getCredentialsFromConf(conf));
+
+    // compute number of splits and instantiate the splits
+    int numSplits = (fileMetaDataList.size() - 1) / maxSplitSize + 1;
```
albertshau: how come it's subtracting and adding? if there are 10 files and maxSplitSize is 2, shouldn't there be 5 splits and not 3?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+    PriorityQueue<AbstractMetadataInputSplit> abstractInputSplits = new PriorityQueue<>(numSplits);
+    for (int i = 0; i < numSplits; i++) {
+      abstractInputSplits.add(getInputSplit());
+    }
+
+    // assign each split approximately the same number of bytes (2-approx)
+    for (int i = 0; i < fileMetaDataList.size(); i++) {
+      AbstractMetadataInputSplit minInputSplit = abstractInputSplits.poll();
+      minInputSplit.addFileMetadata(fileMetaDataList.get(i));
+      abstractInputSplits.add(minInputSplit);
```
albertshau: this doesn't enforce maxSplitSize. Should check the number of files in the split before adding it back to the PriorityQueue. If it's greater than max, add it to inputSplits (from below) and not back to the PriorityQueue.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+    if (enable && fileStatus.isDirectory()) {
+      RemoteIterator<LocatedFileStatus> iter = fileSystem.listLocatedStatus(fileStatus.getPath());
+      while (iter.hasNext()) {
+        LocatedFileStatus fStatus = iter.next();
+        recursivelyAddFileStatus(fileStatuses, baseFolders, prefix, fStatus, enable, fileSystem);
+      }
+    }
+  }
+
+  private List<AbstractFileMetadata> getFileMetadaDataList (List<FileStatus> fileStatuses,
```
albertshau: typo: Metada

Also, remove extra space after DataList

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+        recursivelyAddFileStatus(fileStatuses, baseFolders, prefix, fileStatus, recursiveCopy, fileSystem);
+      } catch (Exception e) {
+        LOG.warn(e.getMessage());
+      }
+    }
+
+    fileSystem.close();
+
+    // create a filemetadata list of all the files
+    List<AbstractFileMetadata> fileMetaDataList = getFileMetadaDataList(fileStatuses, baseFolders,
```
albertshau: seems like you should be able to build this list while getting each FileStatus instead of making another pass. Would also make the logic simpler, as you wouldn't have to maintain an extra baseFolders list.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -0,0 +1,171 @@
...
+
+  protected abstract AbstractFileMetadata.Credentials getCredentialsFromConf(Configuration conf);
+
+  protected abstract AbstractMetadataInputSplit getInputSplit();
+
+  protected abstract AbstractFileMetadata getFileMetaData(FileStatus fileStatus,
+                                                          String sourcePath,
+                                                          AbstractFileMetadata.Credentials credentials);
+
+  public static void setSourcePaths(Job job, String value) {
```
albertshau: nit: can make it take a Configuration object instead.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -0,0 +1,162 @@
...
+  public void addFileMetadata(AbstractFileMetadata fileMetaData) {
+    fileMetaDataList.add(fileMetaData);
+  }
+
+  @Override
+  public int compareTo(Object o) {
+    AbstractMetadataInputSplit other = (AbstractMetadataInputSplit) o;
+    long myLength = getTotalSize();
+    long otherLength = other.getTotalSize();
+    if (myLength < otherLength) {
```
albertshau: slightly easier to do `Long.compare(x, y)`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -0,0 +1,162 @@
...
+        dataOutput.writeUTF(fileMetaData.getOwner());
+        dataOutput.writeLong(fileMetaData.getFileSize());
+        dataOutput.writeBoolean(fileMetaData.getIsFolder());
+        dataOutput.writeUTF(fileMetaData.getBasePath());
+        dataOutput.writeShort(fileMetaData.getPermission());
+
+        writeCredentials(dataOutput, fileMetaData.getCredentials());
+      }
+
+    } catch (Exception e) {
```
albertshau: shouldn't log and skip. Just let the exception propagate, it should be handled by mapreduce.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -0,0 +1,162 @@
...
+  public List<AbstractFileMetadata> getFileMetaDataList() {
+    return this.fileMetaDataList;
+  }
+
+  @Override
+  public void write(DataOutput dataOutput) throws IOException {
+    try {
+      // write obj summaries
+      dataOutput.writeLong(this.getLength());
+      for (int i = 0; i < this.getLength(); i++) {
```
albertshau: In general, to iterate through collections you want to do

```
for (AbstractFileMetadata fileMetaData : fileMetaDataList) {
```

For example, if it's not an ArrayList, `.get(i)` could be more expensive than iterating through items.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -0,0 +1,162 @@
...
+        // convert each filestatus (serializable) to byte array
+        AbstractFileMetadata fileMetaData = fileMetaDataList.get(i);
+
+        dataOutput.writeUTF(fileMetaData.getFileName());
```
albertshau: seems like AbstractFileMetadata should also be Writable, and you just call its write(dataOutput) method here. Then the logic will be contained in AbstractFileMetadata. Otherwise, changing AbstractFileMetadata would also require changing this class.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+    throws Exception {
+    this.destFileSystem = FileSystem.get(conf);
+    this.basePath = conf.get(FileCopyOutputFormat.BASE_PATH);
+    this.enableOverwrite = conf.getBoolean(FileCopyOutputFormat.ENABLE_OVERWRITE, false);
+    this.bufferSize = conf.getInt(FileCopyOutputFormat.BUFFER_SIZE, defaultBufferSize);
+    this.sourceFilesystemMap = new HashMap<>();
+  }
+
+  // source filesystems
+  private Map<AbstractFileMetadata.Credentials, FileSystem> sourceFilesystemMap;
```
albertshau: nit: put this before the constructor.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/s3/S3FileMetadataSource.java
```diff
@@ -0,0 +1,153 @@
...
+    "metadata",
+    Schema.Field.of(S3FileMetadata.S3Credentials.DATA_BASE_TYPE, Schema.of(Schema.Type.STRING)),
+    Schema.Field.of(S3FileMetadata.S3Credentials.ACCESS_KEY_ID, Schema.of(Schema.Type.STRING)),
+    Schema.Field.of(S3FileMetadata.S3Credentials.SECRET_KEY_ID, Schema.of(Schema.Type.STRING)),
+    Schema.Field.of(S3FileMetadata.S3Credentials.REGION, Schema.of(Schema.Type.STRING)),
+    Schema.Field.of(S3FileMetadata.S3Credentials.BUCKET_NAME, Schema.of(Schema.Type.STRING))
+  );
+
+  private Schema outputSchema;
+
```
albertshau: there should be some validation done on the config in the configurePipeline method. This ensures that invalid configs cause the pipeline to fail to deploy instead of waiting until runtime to fail.

albertshau: validation should be basic things like checking max size is not negative, checking the path starts with a valid scheme, etc.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Override
+  public void write(NullWritable key, AbstractFileMetadata fileMetadata) throws IOException, InterruptedException {
+    // get source database connection
+    AbstractFileMetadata.Credentials sourceCredentials = fileMetadata.getCredentials();
```
albertshau: it seems insecure to have the credentials in every record. We can address this in a follow up PR, but maybe the sink should require the source credentials in its configuration as well.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+    if (!sourceFilesystemMap.containsKey(sourceCredentials)) {
+      sourceFilesystemMap.put(sourceCredentials, getSourceFilesystemFromCredentials(sourceCredentials));
+    }
+    FileSystem sourceFilesystem = sourceFilesystemMap.get(sourceCredentials);
+
+    // construct file paths for source and destination
+    Path srcPath = new Path(fileMetadata.fullPath);
+    Path destPath = new Path(basePath + "/" + fileMetadata.basePath);
```
albertshau: shouldn't manually concatenate, use `new Path(basePath, fileMetadata.basePath)`

albertshau: the terminology is also confusing. So fileMetadata.basePath is the directory relative to the path given to the source? And basePath here is the destination directory?

albertshau: can address this in a follow up PR, but data should be written to a temporary directory that the OutputCommitter renames at the very end. This is because MapReduce tries to be atomic if possible.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+        destFileSystem.mkdirs(destPath, permission);
+      }
+      return;
+    } else if (!sourceFilesystem.exists(srcPath)) {
+      // file doesn't exist in source, return immediately
+      return;
+    }
+
+    // immediately return if we dont want to overwrite and file exists in destination
+    if (!enableOverwrite && destFileSystem.exists(destPath)) {
```
albertshau: you have to be a little careful here. If the mapper died early and is reprocessing this file, the file may exist but not be complete. Similarly, if speculative execution is enabled, mapreduce will launch multiple mappers on the same input split, and they should not conflict with each other. Basically, each mapper should write to its own unique directory in order to handle these scenarios.

This can be addressed in a follow up PR.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // data streaming
+    FSDataInputStream inputStream = sourceFilesystem.open(srcPath);
+    FSDataOutputStream outputStream = FileSystem.create(destFileSystem, destPath, permission);
+    byte[] buf = new byte[bufferSize];
+    int len;
+    while ((len = inputStream.read(buf)) >= 0) {
+      outputStream.write(buf, 0, len);
+    }
+    inputStream.close();
```
albertshau: in general, close should always be done in a finally block, in case an exception is thrown earlier.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -0,0 +1,130 @@
...
+    outputStream.close();
+  }
+
+  @Override
+  public void close(TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
+    /*
+     * TODO: investigate when to close the destination filesystem and the source filesystems
```
albertshau: this is closing the record writer, so it should be safe to close everything here. The record writer won't be used anymore after this.

albertshau (commented):
looks pretty good, have some comments plus some points we should address in future work

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
caskchester: Yeah if you plug those numbers in you will get 5.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
caskchester: Yeah that's one of the concerns I had when I wrote this method. Consider mapping the following list of files to only 2 splits.
one 100GB file ....... 1000 50B files ......... one 100GB file.
If we enforce the number of files in maxSplitSize, the two big files would end up in the same split.

I agree that the pipeline should behave how the users expect it to behave. Perhaps sorting the files by size first would be a better idea?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
caskchester: Example: suppose there is a file in `/a/b/c/d/haha.txt`
If you do `cp  -r   /a/b/c   /dest/`
Base path in this case would be `c/d/haha.txt`. Therefore the resulting file path would be `/dest/c/d/haha.txt`

I'll add a comment in the code to explain it more clearly.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
caskchester: Doesn't [FileOutputCommitter](https://hadoop.apache.org/docs/r2.7.2/api/org/apache/hadoop/mapred/FileOutputCommitter.html) prevent failed jobs from being committed?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
caskchester: When I close connections in the `close()` method, a pipeline with multiple splits would fail. It seems like some splits are sharing one `RecordWriter` instantiation. So if one split calls this method to close the connection, another split that's also using this `RecordWriter` would fail during its call to the `write()` method.

The connections are stored in a `Map<Credentials, Filesystem>` object

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
albertshau: right, i seem to have inserted parentheses in my head

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
albertshau: True, you can sort those first and it would probably be more evenly distributed.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
albertshau: I think it does some setup and cleanup, but you may need to make sure you're using whatever directories it is setting/cleaning up instead of the destination one specified by the plugin.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
albertshau: hm I think that means the FileSystem you're getting is a cached instance, not that the RecordWriter is shared between splits. If you look at FileSystem.get(), it keeps an internal cache as well, which can be turned off depending on the value of a config value:

```
    String disableCacheName = String.format("fs.%s.impl.disable.cache", scheme);
    if (conf.getBoolean(disableCacheName, false)) {
      return createFileSystem(uri, conf);
    }
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
caskchester: The information required to access the source filesystem will be stored in its subclass, so this class is called credentials.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
caskchester: `listFiles(Path, boolean)` ignores empty folders, so I wrote my own method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
caskchester: This is a design choice. Ideally a file copy sink can take inputs from multiple sources (different filesystems). That's why we attach the credentials to every file metadata record we send. But yeah I agree it could be a security concern.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
caskchester: I see. I guess this means we have some questions to figure out:
1. If we turn caching off, is it okay to have 100 splits all try to initiate connection to the filesystem at the same time?
2. If we leave caching on, does that mean file transmission resources (like bandwidth and other things) would be divided among the splits?

Maybe we can address these problems in a future PR.

caskchester:
@albertshau @sagarkapare 
Addressed comments. Please take another look. Thanks!

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
albertshau: Were you seeing the issue in a cluster or just in the sandbox/unit test? I wouldn't expect to see this in a cluster where the code is run in separate processes.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
```diff
@@ -85,32 +80,60 @@ public void transform(StructuredRecord input,
...
+    @Description("Whether or not for splits to use cached connections to the filesystems.")
+    public Boolean filesystemCaching;
+
+    @Description("Whether or not to preserve the owner of the file from source filesystem.")
+    public Boolean preserveFileOwner;
+
+    // TODO: figure out why this still shows up as required in configuration UI
     @Macro
     @Nullable
     @Description("The size of the buffer that temporarily stores data from file input stream.")
```
albertshau: the description should mention what unit this is in. MB? If it's currently in bytes, we should probably change it to megabytes, as it's pretty annoying putting in byte values.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
```diff
@@ -85,32 +80,60 @@ public void transform(StructuredRecord input,
...
-
     @Macro
     @Description("The destination path. Will be created if it doesn't exist.")
     public String basePath;
 
     @Description("Whether or not to overwrite if the file already exists.")
     public Boolean enableOverwrite;
 
+    @Description("Whether or not for splits to use cached connections to the filesystems.")
+    public Boolean filesystemCaching;
```
albertshau: I don't think we should have this property. It's misleading, a FileSystem doesn't represent a connection to the filesystem.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
albertshau: Description should also mention what it defaults to.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
albertshau: I think relativePath would be a better name for this. I would expect a base path to be the left part not the right part. It would also help to have a comment.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
caskchester: Yeah the issue only happens in sandbox if caching is not disabled. Doesn't happen in cluster

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
albertshau: The logic seems like it can be simpler. I believe you can do `fileStatus.getPath().toURI().getPath()` to get the path without scheme, host, port. If I understand correctly, that path should always be prefixed by sourcePath? If so, you can do `fileStatus.getPath().toURI().getPath().substring(sourcePath.length())` (might be off by one).

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -160,11 +233,78 @@ public short getPermission() {
...
+    dataOutput.writeBoolean(isFolder());
+    dataOutput.writeUTF(getBasePath());
+    dataOutput.writeShort(getPermission());
+    dataOutput.writeUTF(getFilesystem());
+    dataOutput.writeUTF(getHostURI());
+  }
+
   /**
-   * extend from this class to create a credential class for each filesystem
+   * Don't use this to deserialize AbstractFileMetadata. Use the constructor
```
albertshau: implementing this is part of the contract for the Writable interface. Hadoop calls this method to deserialize the class when it is used as the mapper output key or value, or when SequenceFileOutputFormat is used. If this isn't implemented, it won't work correctly when used in that way.

Since we're not using this as a mapper output key or value, I think it's fine to remove WritableComparable as a parent interface so you aren't forced to implement this unused method.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -118,54 +122,37 @@ public RecordReader createRecordReader(InputSplit inputSplit, TaskAttemptContext
     return recordReader;
   }
 
-  private void recursivelyAddFileStatus (List<FileStatus> fileStatuses, List<String> baseFolders,
+  private void recursivelyAddFileStatus(List<AbstractFileMetadata> fileMetadataList,
```
albertshau: lets add a comment saying this is used instead of FileSystem.listFiles() because that method doesn't list empty directories.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputFormat.java
```diff
@@ -118,54 +122,37 @@ public RecordReader createRecordReader(InputSplit inputSplit, TaskAttemptContext
...
                                          String prefix, FileStatus fileStatus, Boolean enable,
```
albertshau: nit: rename enable to 'recursive'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -52,56 +52,24 @@ public void write(DataOutput dataOutput) throws IOException {
...
-        dataOutput.writeUTF(fileMetaData.getBasePath());
-        dataOutput.writeShort(fileMetaData.getPermission());
-
-        writeCredentials(dataOutput, fileMetaData.getCredentials());
+        fileMetaData.write(dataOutput);
       }
 
-    } catch (Exception e) {
-      LOG.error("serialization failed");
+    } catch (IOException e) {
```
albertshau: don't need this catch clause

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -41,88 +41,128 @@
...
-  private Map<AbstractFileMetadata.Credentials, FileSystem> sourceFilesystemMap;
+  private Map<String, FileSystem> sourceFilesystemMap;
 
-  @Override
-  public void write(NullWritable key, AbstractFileMetadata fileMetadata) throws IOException, InterruptedException {
-    // get source database connection
-    AbstractFileMetadata.Credentials sourceCredentials = fileMetadata.getCredentials();
-    if (!sourceFilesystemMap.containsKey(sourceCredentials)) {
-      sourceFilesystemMap.put(sourceCredentials, getSourceFilesystemFromCredentials(sourceCredentials));
+  public FileCopyRecordWriter(Configuration conf) throws Exception {
```
albertshau: should throw as specific an exception as possible. I think this should be IOException.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyOutputFormat.java
```diff
@@ -74,7 +112,7 @@ public RecordWriter getRecordWriter(TaskAttemptContext taskAttemptContext) throw
     try {
       return new FileCopyRecordWriter(conf);
     } catch (Exception e) {
```
albertshau: don't need this try catch if FileCopyRecordWriter throws IOException

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -41,88 +41,128 @@
...
+  public void write(NullWritable key, AbstractFileMetadata fileMetadata) throws IOException, InterruptedException {
     // construct file paths for source and destination
-    Path srcPath = new Path(fileMetadata.fullPath);
-    Path destPath = new Path(basePath + "/" + fileMetadata.basePath);
+    Path srcPath = new Path(fileMetadata.getFullPath());
+    Path destPath;
+    if (fileMetadata.getBasePath().isEmpty()) {
+      // nothing to create
+      return;
+    } else {
```
albertshau: don't need to put this in an else clause, can just declare and assign it here.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -41,88 +41,128 @@
...
+      byte[] buf = new byte[bufferSize];
+      int len;
+      while ((len = inputStream.read(buf)) >= 0) {
+        outputStream.write(buf, 0, len);
+      }
+    } finally {
+      if (preserveOwner) {
+        destFileSystem.setOwner(destPath, fileMetadata.getOwner(), fileMetadata.getGroup());
+      }
+      inputStream.close();
```
albertshau: if this throws an exception, outputStream won't get closed. So you need to wrap this in a try catch finally as well.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -41,88 +41,128 @@
...
-      case S3FileMetadata.DATA_BASE_NAME :
-        S3FileMetadata.S3Credentials s3Credentials = (S3FileMetadata.S3Credentials) credentials;
-        conf.set("fs.s3a.access.key", s3Credentials.accessKeyId);
-        conf.set("fs.s3a.secret.key", s3Credentials.secretKeyId);
+
+    // turn filesystem caching off
+    URI uri = URI.create(metadata.getHostURI());
+
+    // whether or not to disable caching for source filesystems
+    String disableCacheName = String.format("fs.%s.impl.disable.cache", uri.getScheme());
```
albertshau: lets just always disable cache for now.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/s3/S3FileMetadataSource.java
```diff
@@ -137,17 +101,22 @@ public void transform(KeyValue<NullWritable, S3FileMetadata> input, Emitter<Stru
...
       this.secretKeyId = secretKeyId;
       this.region = region;
-      this.bucketName = bucketName;
+    }
+
+    @Override
+    public void validate() {
+      super.validate();
+      if (!this.containsMacro("filesystemURI")) {
+        if (!filesystemURI.startsWith("s3a://")) {
```
albertshau: s3n is an older, but still supported scheme. should allow it too.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileMetadata.java
```diff
@@ -160,11 +233,78 @@ public short getPermission() {
...
   }
+
+  /**
+   * @return Credential schema for different filesystems
+   */
+  protected abstract Schema getCredentialSchema();
+
+  protected abstract void addCredentialsToBuilder(StructuredRecord.Builder builder);
+
+  protected abstract String getFSName();
```
albertshau: where/how is this used? Is there a reason this is used instead of the scheme for the filesystem uri? (hdfs, s3a, s3n, ftp, etc.)

caskchester: yeah it's redundant. will remove

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/s3/S3FileMetadataSource.java
caskchester: Can I just replace all `s3n` URIs with `s3a`? This is from the [AWS doc:](https://aws.amazon.com/premiumsupport/knowledge-center/emr-file-system-s3/) 

> For Apache Hadoop, S3A is the successor to S3N and is backward compatible with S3N. Using Apache Hadoop, all objects accessible from s3n:// URLs should also be accessible from S3A by replacing the URL scheme.

albertshau: I'm not sure when they replaced it, but we have to support some old Hadoop distributions so you can only do that if we verify that s3a is supported in all of our supported Hadoop distributions.

caskchester:
@albertshau @sagarkapare 
Addressed comments and added support for s3n.
Please review. Thanks!

caskchester:
@albertshau @sagarkapare
Addressed comments and added support for s3n.
Please review. Thanks!

core-plugins/docs/HDFSFileCopySink-batchsink.md
```diff
@@ -36,7 +35,7 @@ This sink plugin only reads StructuredRecords with the following schema. Each re
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
```
albertshau: typo, relavite -> relative

core-plugins/docs/S3FileMetadataSource-batchsource.md
```diff
@@ -39,7 +39,7 @@ A structured record with the following schema is emitted for each file it reads.
 | **group**              | String | The group that the of the file belongs to.                                                                                                     |
 | **owner**              | String | The owner of the file.                                                                                                                         |
 | **isFolder**           | Boolean| Whether or not the file is a folder.                                                                                                           |
-| **basePath**           | String | The base path of the file. This path will be appended to the base path specified in the FileCopySink to create the file in the destination.    |
+| **relativePath**       | String | The relavite path is constructed by deleting the portion of the source path that comes before the last path separator ("/") from the full path.|
```
albertshau: typo, relavite

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractFileCopySink.java
```diff
@@ -147,8 +147,16 @@ public FileCopyOutputFormatProvider(AbstractFileCopySink.AbstractFileCopySinkCon
...
+
+      if (config.bufferSize != null) {
+        // bufferSize is in megabytes
+        FileCopyOutputFormat.setBufferSize(conf, String.valueOf(config.bufferSize << 20));
+      } else {
+        FileCopyOutputFormat.setBufferSize(conf, String.valueOf(FileCopyRecordWriter.DEFAULT_BUFFER_SIZE));
+      }
+
+      // always disable caching
+      conf.put(String.format("fs.%s.impl.disable.cache", config.getScheme()), String.valueOf(true));
```
albertshau: We only want our specific OutputFormat to disable caching, not everything else in Hadoop. So we shouldn't set this here, but only in the OutputFormat when we get a new FileSystem object.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -50,41 +56,53 @@ public AbstractMetadataInputSplit() {
   @Override
   public void write(DataOutput dataOutput) throws IOException {
     try {
-      // write obj summaries
+      // write number of files
       dataOutput.writeLong(this.getLength());
+
+      // write total number of file bytes
+      dataOutput.writeLong(this.getTotalBytes());
```
albertshau: this works, but most of the time you write the length of a collection and immediately follow it with the items in the collection. Let's do that here too and write the total bytes before we write the number of files.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -50,41 +56,53 @@ public AbstractMetadataInputSplit() {
...
     }
   }
 
   @Override
   public void readFields(DataInput dataInput) throws IOException {
+    // read number of files
     long numObjects = dataInput.readLong();
+
+    // get total number of bytes
+    totalBytes = dataInput.readLong();
```
albertshau: move this above reading the number of files when write() is changed.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/AbstractMetadataInputSplit.java
```diff
@@ -50,41 +56,53 @@ public AbstractMetadataInputSplit() {
...
+
     fileMetaDataList = new ArrayList<>();
     for (long i = 0; i < numObjects; i++) {
-      fileMetaDataList.add(readFileMetaData(dataInput));
+      AbstractFileMetadata metadata = readFileMetaData(dataInput);
+      fileMetaDataList.add(metadata);
+      totalBytes += metadata.getFileSize();
```
albertshau: how come you need to add it here? doesn't this mean totalBytes will be doubled?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -121,50 +120,72 @@ public void write(NullWritable key, AbstractFileMetadata fileMetadata) throws IO
...
+      safelyCloseSourceFilesystems(sourceFilesystemMap.entrySet().iterator());
+    }
+  }
+
+  /**
+   * this method attempts to close every filesystem in the list
+   * @param fs The iterator over all the filesystems we wish to close.
+   * @throws IOException
+   */
+  private void safelyCloseSourceFilesystems(Iterator<Map.Entry<String, FileSystem>> fs) throws IOException {
```
albertshau: since the map key isn't used, you can take an `Iterator<FileSystem>` instead and pass in `sourceFilesystemMap.values().iterator()`.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/file/FileCopyRecordWriter.java
```diff
@@ -121,50 +120,72 @@ public void write(NullWritable key, AbstractFileMetadata fileMetadata) throws IO
...
+    if (fs.hasNext()) {
+      try {
+        fs.next().getValue().close();
+      } finally {
+        safelyCloseSourceFilesystems(fs);
```
albertshau: its a little weird to go through an iterator using recursion instead of a while loop. Let's just do
```
while (fs.hasNext()) {
  try {
    fs.next().close();
  } catch () {
    // log some warning
  }
}
```

core-plugins/widgets/HDFSFileCopySink-batchsink.json
```diff
@@ -41,20 +41,8 @@
...
-            ],
-            "default": "true"
-          }
-        },
-        {
           "widget-type": "textbox",
-          "label": "Buffer Size",
+          "label": "Buffer Size (MB)",
           "name": "bufferSize",
           "default": "4096"
```
albertshau: default it to 1 to be consistent with code

albertshau (commented):
just a few style comments remaining

caskchester:
@albertshau 
Addressed comments. Please review. Thanks!

albertshau (approved):
lgtm

albertshau:
i think we wanted to add this as a separate repo under the hydrator org, so lets not merge it here

caskchester:
Closing this and moving to https://github.com/hydrator/whole-file-ingest per @nitinmotgi 's request

PR #694, opened 2017-07-26
test&fix: BatchXMLFileFilter to read all files when a path and regex is given

superchaoran:
disable quick fix;complete batch filter test;diable incomplete wrangler transform

fix file path not found bug; also tested by unit test

adding invalid file directory test

rm unecessory try-catch

not going to include failed test on other thread


superchaoran:
referring to https://issues.cask.co/browse/CDAP-12146

PR #695, opened 2017-07-28
Read avro parquet source

sheivin:


sheivin:
@albertshau Text format does not need to have a schema now.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -219,7 +220,7 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
    */
   @javax.ws.rs.Path("getSchema")
   public Schema getSchema(SchemaRequest request, EndpointPluginContext pluginContext) {
-    return PathTrackingInputFormat.getOutputSchema(request.pathField);
+    return PathTrackingInputFormat.getTextOutputSchema(request.pathField);
```
albertshau: this gets called when the user hits the 'get schema' button. It should not return the text schema when the source is not configured to use text.

However, now that the user is responsible for setting a schema, we should just remove this method and corresponding widget entirely, since it does not make sense anymore.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -101,19 +108,24 @@
     "URI will be used. Defaults to false.")
   public Boolean filenameOnly;
 
+  @Nullable
+  @Description("Schema for the source")
+  public String inputSchema;
```
albertshau: There is already a schema property. Don't add a new one, just use the existing one.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -137,6 +151,21 @@ protected void validate() {
       throw new IllegalArgumentException("pathField can only be used if inputFormatClass is " +
                                            CombinePathTrackingInputFormat.class.getName());
     }
+    if (format == null || !(format.equalsIgnoreCase("text") ||
+      (format.equalsIgnoreCase("avro")) || format.equalsIgnoreCase("parquet"))) {
+      throw new IllegalArgumentException("Format can only be 'text', 'avro' or 'parquet'");
+    }
+    if (!format.equalsIgnoreCase("text")) {
+      try {
+        new Schema.Parser().parse(inputSchema);
```
albertshau: use our Schema, not avro's. `Schema.parseJson()`

albertshau: there needs to be additional validation after this. If the pathField is specified, you need to validate that it is in the schema and it is either a nullable string or a string.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -137,6 +151,21 @@ protected void validate() {
...
+      } catch (Exception e){
+        throw new IllegalArgumentException("Unable to parse schema with error: " + e.getMessage(), e);
+      }
+    } else if (inputSchema != null && !inputSchema.isEmpty()) {
+      if (!(inputSchema.contains("offset") && inputSchema.contains("body"))) {
```
albertshau: this is wrong, an avro or parquet schema can have a field named 'body'. The validation here should be if format is text, the schema should be equal to the schema used by the code, taking into account whether the pathfield is set or not.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -40,18 +48,30 @@
...
       conf.set(PATH_FIELD, pathField);
     }
     conf.setBoolean(FILENAME_ONLY, filenameOnly);
+    conf.set(FORMAT, format);
+    if (schema != null) {
+      conf.set(SCHEMA, schema);
+    }
+  }
+
+  public static void setJob(Job currentJob) {
```
albertshau: what is this supposed to be doing? I don't think you need it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -129,8 +152,83 @@ public float getProgress() throws IOException, InterruptedException {
...
+    }
+  }
+
+  private static class TrackingParquetRecordReader extends TrackingRecordReader<Void, GenericRecord> {
+    private AvroToStructuredTransformer recordTransformer;
+    private Job job;
+
+    private TrackingParquetRecordReader(RecordReader<Void, GenericRecord> delegate, @Nullable String pathField,
+                                        String path, String schema, Job job) throws IOException {
+      super(delegate, pathField, path, Schema.parseJson(schema));
```
albertshau: parsing should not occur in the constructor, where it can throw an IOException. It should be parsed by the caller, with only a valid schema object passed into each different type of record reader.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -129,8 +152,83 @@ public float getProgress() throws IOException, InterruptedException {
...
     public void close() throws IOException {
       delegate.close();
     }
   }
+
+  private static class TrackingTextRecordReader extends TrackingRecordReader<LongWritable, Text> {
+
+    private TrackingTextRecordReader(RecordReader<LongWritable, Text> delegate, @Nullable String pathField,
+                                     String path) {
+      super(delegate, pathField, path, getTextOutputSchema(pathField));
```
albertshau: you should pass in the schema set in the config, not generate it here. Otherwise the schema set by the inputformat is different than the schema used here, which makes it easy to introduce bugs later on.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -129,8 +152,83 @@ public float getProgress() throws IOException, InterruptedException {
...
+    @Override
+    public void initialize(InputSplit split, TaskAttemptContext context) throws IOException,
+      InterruptedException {
+      super.initialize(split, context);
+      recordTransformer = new AvroToStructuredTransformer();
+    }
+
+    public StructuredRecord getCurrentValue(StructuredRecord.Builder recordBuilder) throws IOException,
+      InterruptedException {
+      AvroJob.setOutputKeySchema(job, new org.apache.avro.Schema.Parser().parse(schema.toString()));
```
albertshau: this does nothing at this point. It needs to be set when the job is being configured.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -129,8 +152,83 @@ public float getProgress() throws IOException, InterruptedException {
...
+    @Override
+    public void initialize(InputSplit split, TaskAttemptContext context) throws IOException, InterruptedException {
+      super.initialize(split, context);
+      recordTransformer = new AvroToStructuredTransformer();
+    }
+
+    public StructuredRecord getCurrentValue(StructuredRecord.Builder recordBuilder) throws IOException,
+      InterruptedException {
+      StructuredRecord ret = recordTransformer.transform(recordBuilder, delegate.getCurrentValue(), schema, pathField);
+      AvroReadSupport.setAvroReadSchema(job.getConfiguration(), new org.apache.avro.Schema.Parser().parse
```
albertshau: this needs to be set when configuring the job, not here.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +478,149 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+
+    return appManager;
+  }
+
+  private void workflowStartAndWait(ApplicationManager appManager) throws Exception {
+    WorkflowManager workflowManager = appManager.getWorkflowManager(SmartWorkflow.NAME);
+    workflowManager.start();
+    workflowManager.waitForRuns(ProgramRunStatus.COMPLETED, 1, 2, TimeUnit.MINUTES);
+  }
+}
```
albertshau: all files need to end with a newline

core-plugins/widgets/File-batchsource.json
```diff
@@ -106,20 +119,17 @@
       "name": "schema",
       "widget-type": "schema",
       "widget-attributes": {
-        "default-schema": {
```
albertshau: why remove the default schema?

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -463,4 +478,149 @@ public void testRecursiveTimeFilterRegex() throws Exception {
...
+
+    Assert.assertEquals(output, expected);
+  }
+
+  @Test
+  public void testFileBatchInputFormatAvro() throws Exception {
+    File fileAvro = new File(temporaryFolder.newFolder(), "test.avro");
+    String outputDatasetName = "test-filesource-avro";
+
+    ApplicationManager appManager = createSourceAndDeployApp(fileAvro, "avro", outputDatasetName, RECORD_SCHEMA);
```
albertshau: you should have avro and parquet tests that specify a schema that is different, but compatible to, write schema. This is to test the logic that is setting the input schema key in the input format. For example, I believe you should be able to remove a field from the write schema and it should read it correctly.

albertshau (commented):
the commit history is really messy. When addressing these comments, please keep changes in a single new commit so that it is easier to review just the changes.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
sheivin: yeah this is not needed because as you pointed out below, the property should be set when configuring the job. So this is useless.

core-plugins/widgets/File-batchsource.json
sheivin: i felt that this is not needed because we have added multiple formats. but it is good to have a default schema. so will put this back

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -137,6 +151,21 @@ protected void validate() {
...
                                            CombinePathTrackingInputFormat.class.getName());
     }
+    if (format == null || !(format.equalsIgnoreCase("text") ||
+      (format.equalsIgnoreCase("avro")) || format.equalsIgnoreCase("parquet"))) {
+      throw new IllegalArgumentException("Format can only be 'text', 'avro' or 'parquet'");
+    }
+    if (!format.equalsIgnoreCase("text")) {
+      try {
+        new Schema.Parser().parse(inputSchema);
+      } catch (Exception e){
```
albertshau: insert space between `){`

sheivin: yeah I had missed some of the checkstyle issues. they will be fixed in the next commit

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -168,8 +169,13 @@ public void prepareRun(BatchSourceContext context) throws Exception {
       FileInputFormat.setMaxInputSplitSize(job, config.maxSplitSize);
     }
     if (CombinePathTrackingInputFormat.class.getName().equals(config.inputFormatClass)) {
-      PathTrackingInputFormat.setJob(job);
-      PathTrackingInputFormat.configure(conf, config.pathField, config.filenameOnly, config.format, config.inputSchema);
+      if (config.format.equalsIgnoreCase("avro")) {
```
albertshau: this logic should be in the PathTrackingInputFormat's configure method

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,28 @@ protected void validate() {
...
-        throw new IllegalArgumentException("Unable to parse schema with error: " + e.getMessage(), e);
+    Schema parsedSchema;
+    try {
+      parsedSchema = Schema.parseJson(schema);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Unable to parse schema with error: " + e.getMessage(), e);
+    }
+
+    if (format.equalsIgnoreCase("text")) {
+      if (schema != null && !schema.isEmpty()) {
```
albertshau: combine this with the previous line to have  `"text".equalsIgnoreCase(format) && parsedSchema != null`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,28 @@ protected void validate() {
...
+        if (!(schema.contains("offset") && schema.contains("body"))) {
```
albertshau: you need to check the types.  check that offset is a long or nullable long and body is a string or nullable string.

```
Schema.Field offsetField = schema.getField("offset");
if (offsetField == null) {...}
Schema offsetSchema = offsetField.getSchema();
Schema.Type offsetType = offsetSchema.isNullable() ? offsetSchema.getNonNullable().getType() : offsetSchema.getType();
if (offsetType != Schema.Type.LONG) { ... }
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,28 @@ protected void validate() {
...
+        }
       }
-    } else if (inputSchema != null && !inputSchema.isEmpty()) {
-      if (!(inputSchema.contains("offset") && inputSchema.contains("body"))) {
-        throw new IllegalArgumentException("Schema must contain fields for offset and body");
+    }
+    if (pathField != null) {
+      if (schema.contains(pathField)) {
+        Schema.Field schemaPathField = parsedSchema.getField(pathField);
+        if (!(schemaPathField.getSchema().isSimpleOrNullableSimple())) {
```
albertshau: this isn't good enough, need to check string or nullable string.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,28 @@ protected void validate() {
...
+          throw new IllegalArgumentException("Value for " + pathField + " must be a string or nullable string");
+        }
+      } else {
+        throw new IllegalArgumentException("Schema must contain " + pathField);
```
albertshau: For readability, logic for handling a condition should be as close to the condition as possible. In this case, move things around to look like:
```
Schema.Field schemaPathField = parsedSchema.getField(pathField);
if (schemaPathField == null) {
  // throw condition
}
// check type
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -99,14 +91,14 @@ public static Schema getTextOutputSchema(@Nullable String pathField) {
...
+      return new TrackingAvroRecordReader(delegate, pathField, path, Schema.parseJson(schema));
     } else if ("parquet".equalsIgnoreCase(format)) {
       RecordReader<Void, GenericRecord> delegate = (new AvroParquetInputFormat<GenericRecord>())
         .createRecordReader(split, context);
-      return new TrackingParquetRecordReader(delegate, pathField, path, schema, job);
+      return new TrackingParquetRecordReader(delegate, pathField, path, Schema.parseJson(schema));
     } else {
       RecordReader<LongWritable, Text> delegate = (new TextInputFormat()).createRecordReader(split, context);
-      return new TrackingTextRecordReader(delegate, pathField, path);
+      return new TrackingTextRecordReader(delegate, pathField, path, getTextOutputSchema(pathField));
```
albertshau: Do Schema.parseJson() before this if else section and pass the schema to each record reader. As mentioned before, you shouldn't be using getTextOutputSchema() here, as it makes it easier to accidentally break the code as it evolves.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -584,6 +584,90 @@ public void testFileBatchInputFormatParquet() throws Exception {
     Assert.assertEquals(expected, output);
   }
 
+  @Test
+  public void testFileBatchInputFormatAvroMissingField() throws Exception {
+    File fileAvro = new File(temporaryFolder.newFolder(), "test.avro");
+    String outputDatasetName = "test-filesource-avro";
+
+    Schema recordSchemaWithMissingField = Schema.recordOf("record",
+                                                   Schema.Field.of("i", Schema.of(Schema.Type.INT)),
```
albertshau: indentation

core-plugins/widgets/File-batchsource.json
```diff
@@ -57,17 +57,6 @@
       "label": "Output Schema Properties",
       "properties": [
         {
-          "widget-type": "textbox",
-          "label": "Path Field",
-          "name": "pathField",
-          "plugin-function": {
```
albertshau: you shouldn't be remove the entire path field. Just remove this plugin-function section.

albertshau (commented):
just some style comments and validation comments remaining.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -149,28 +149,49 @@ protected void validate() {
       (format.equalsIgnoreCase("avro")) || format.equalsIgnoreCase("parquet"))) {
       throw new IllegalArgumentException("Format can only be 'text', 'avro' or 'parquet'");
     }
+    if (format.equalsIgnoreCase("text") && schema == null) {
```
sheivin: @albertshau This has been added for the case where the schema is not provided and format is text. Noticed this when all the other tests started failing where the schema was not specified. If schema is not specified, then we don't have to perform any of the checks that are after this. This felt like a cleaner option than adding everything else in another conditional block.

sheivin: If schema is not specified, then we don't have to perform any of the checks that are after this. This felt like a cleaner option than adding everything else in another conditional block.

albertshau: there's no need to check the format here right? If schema is null then nothing below here needs to be checked

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      return;
+    }
+    Schema parsedSchema;
+    try {
+      parsedSchema = Schema.parseJson(schema);
+    } catch (Exception e) {
+      throw new IllegalArgumentException("Unable to parse schema with error: " + e.getMessage(), e);
+    }
+
+    if (format.equalsIgnoreCase("text") && parsedSchema != null) {
```
albertshau: don't need the parsedSchema != null here anymore

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      Schema.Field offsetField = parsedSchema.getField("offset");
+      if (offsetField == null) {
+        throw new IllegalArgumentException("Schema for text format must have an offset field");
```
albertshau: must have a field named 'offset'.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      }
+      Schema offsetSchema = offsetField.getSchema();
+      Schema.Type offsetType = offsetSchema.isNullable() ? offsetSchema.getNonNullable().getType() :
+        offsetSchema.getType();
+      if (offsetType != Schema.Type.LONG) {
+        throw new IllegalArgumentException("Value of offset field in schema must be of type Long");
```
albertshau: Type of 'offset' field must be 'long', but found [whatever offsetType is]

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      }
+
+      Schema.Field bodyField = parsedSchema.getField("body");
+      if (bodyField == null) {
+        throw new IllegalArgumentException("Schema for text format must have a body field");
```
albertshau: must have a field named 'body'

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      }
+      Schema bodySchema = bodyField.getSchema();
+      Schema.Type bodyType = bodySchema.isNullable() ? bodySchema.getNonNullable().getType() : bodySchema.getType();
+      if (bodyType != Schema.Type.STRING) {
+        throw new IllegalArgumentException("Value of body field in schema must be of type String");
```
albertshau: Type of 'offset' field must be 'long', but found [whatever bodyType is]

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
+      }
+    }
+
+    if (pathField != null) {
+      Schema.Field schemaPathField = parsedSchema.getField(pathField);
+      if (schemaPathField == null) {
+        throw new IllegalArgumentException("No value specified for " + pathField);
```
albertshau, line 188: Path field [whatever pathField is] is not a field in the schema.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -155,15 +149,49 @@ protected void validate() {
...
       }
-    } else if (inputSchema != null && !inputSchema.isEmpty()) {
-      if (!(inputSchema.contains("offset") && inputSchema.contains("body"))) {
-        throw new IllegalArgumentException("Schema must contain fields for offset and body");
+      Schema pathFieldSchema = schemaPathField.getSchema();
+      Schema.Type pathFieldType = pathFieldSchema.isNullable() ? pathFieldSchema.getNonNullable().getType() :
+        pathFieldSchema.getType();
+      if (pathFieldType != Schema.Type.STRING) {
+        throw new IllegalArgumentException("value of " + pathField + " must be of type String");
```
albertshau: Path field [whatever pathField is] must be of type string, but found [whatever pathFieldType is]

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -612,7 +702,7 @@ private ApplicationManager createSourceAndDeployApp(File file, String format, St
       .build();
 
     AppRequest<ETLBatchConfig> appRequest = new AppRequest<>(BATCH_ARTIFACT, etlConfig);
-    ApplicationId appId = NamespaceId.DEFAULT.app("FileSource" + format + "Format");
+    ApplicationId appId = NamespaceId.DEFAULT.app(appName + format + "Format");
```
albertshau: there's no need to append format + "Format"

albertshau (commented):
mostly comments about clarifying the validation messages. Also, comment about keeping pathField in the widgets json still applies.

sheivin:
@albertshau addressed comments. For the pathField in the widgets, I have only removed the plugin-function section.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -159,26 +159,27 @@ protected void validate() {
...
       if (offsetField == null) {
-        throw new IllegalArgumentException("Schema for text format must have an offset field");
+        throw new IllegalArgumentException("Schema for text format must have a field named 'offset'");
       }
       Schema offsetSchema = offsetField.getSchema();
       Schema.Type offsetType = offsetSchema.isNullable() ? offsetSchema.getNonNullable().getType() :
         offsetSchema.getType();
       if (offsetType != Schema.Type.LONG) {
-        throw new IllegalArgumentException("Value of offset field in schema must be of type Long");
+        throw new IllegalArgumentException("Type of 'offset' must be 'long', but found " + offsetType.toString());
```
albertshau: 'offset' field

albertshau: you also don't need .toString(), it is done automatically.

sheivin: @albertshau oh ok. will change it.

albertshau (approved):
lgtm. please be sure to squash commits before merging.

PR #696, opened 2017-08-02
Read Avro and Parquet formats from file batch source

sheivin:


albertshau:
this is a copy of https://github.com/caskdata/hydrator-plugins/pull/695 but with commits cleaned up.

albertshau (approved):
lgtm

PR #699, opened 2017-08-04
Avro parquet filesource null schema

sheivin:


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -96,7 +96,11 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     if (!config.containsMacro("timeTable") && config.timeTable != null) {
       pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
     }
-    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    if (config.getSchema() != null) {
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    } else {
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(DEFAULT_SCHEMA);
```
albertshau: This isn't correct. For example, if schema is null and format is avro, you will set the output schema to the default schema, which is almost certainly wrong.

You should be setting the actual output schema here. The same schema you set in the input format is what should be set here.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -98,7 +98,10 @@ public static Schema getTextOutputSchema(@Nullable String pathField) {
     String format = context.getConfiguration().get(FORMAT);
     String path = filenameOnly ? fileSplit.getPath().getName() : fileSplit.getPath().toUri().toString();
     String schema = context.getConfiguration().get(SCHEMA);
-    Schema parsedSchema = Schema.parseJson(schema);
+    Schema parsedSchema = null;
```
albertshau: nit: `Schema parsedSchema = schema == null ? null : Schema.parseJson(schema)`

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -140,13 +143,19 @@ public NullWritable getCurrentKey() throws IOException, InterruptedException {
     protected abstract StructuredRecord getCurrentValue(StructuredRecord.Builder recordBuilder)
       throws IOException, InterruptedException;
 
+    protected abstract StructuredRecord getCurrentValueNull() throws IOException, InterruptedException;
```
albertshau: this is a strange contract. The fact that you return null from one of the implementations and expect it never to be used by the parent means it's not a clean abstraction.

albertshau: instead of this, reverse the order and change getCurrentValue to `StructuredRecord.Builder startCurrentValue()`. This class calls `startCurrentValue`, then sets the path field for it.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
```diff
@@ -71,4 +74,26 @@ public StructuredRecord transform(StructuredRecord.Builder builder, GenericRecor
 
     return builder.build();
   }
+
+  public StructuredRecord transform(GenericRecord genericRecord, String addFieldKey, String addFieldValue)
```
albertshau: This isn't a very natural place for this logic, as this class should be about converting records. Modify the other transform method to return a StructuredRecord.Builder and remove the builder as an argument. The logic to add a field to the schema should be in the input format class, not here.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
```diff
@@ -71,4 +74,26 @@ public StructuredRecord transform(StructuredRecord.Builder builder, GenericRecor
...
+    Schema.Field newField = Schema.Field.of(addFieldKey, Schema.of(Schema.Type.STRING));
+    List<Schema.Field> newFields = new ArrayList<>();
+
+    for (Schema.Field field : schema.getFields()) {
+      newFields.add(field);
+    }
+    newFields.add(newField);
+    Schema newSchema = Schema.recordOf(schema.getRecordName(), newFields);
+
+    GenericRecordBuilder genericRecordBuilder = new GenericRecordBuilder(new org.apache.avro.Schema.Parser().
```
albertshau: don't build a brand new record and convert it, you'll double the work. You should be converting the GenericRecord once, and setting the pathfield at the very end on the StructuredRecord.Builder.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -585,13 +591,13 @@ public void testFileBatchInputFormatParquet() throws Exception {
...
         .set("l", Long.MAX_VALUE)
-        .set("file", fileParquet.toURI().toString())
+        .set("file", fileAvro.toURI().toString())
         .build()
     );
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
     List<StructuredRecord> output = MockSink.readOutput(outputManager);
-    Assert.assertEquals(expected, output);
+    Assert.assertEquals(output, expected);
```
albertshau: expected is always first.

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -686,18 +772,33 @@ private ApplicationManager createSourceAndDeployApp(String appName, File file, S
...
-                                .put(Properties.File.FILESYSTEM, "Text")
-                                .put(Properties.File.PATH, file.getAbsolutePath())
-                                .put(Properties.File.FORMAT, format)
-                                .put(Properties.File.IGNORE_NON_EXISTING_FOLDERS, "false")
-                                .put("pathField", "file")
-                                .put(Properties.File.SCHEMA, schemaString)
-                                .build(),
-                              null));
+    ETLStage source;
+    if (schema != null) {
```
albertshau: this is unnecessary repeat code. The only thing you need to do if schema is not null is put the schema property in the ImmutableMap builder

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
sheivin: @albertshau For setting the actual output schema in the case of when the schema is not defined, we will only be able to get the schema once the record reader has been initialized and we start reading

albertshau: when you set a null output schema at configure time, it means the schema can either be variable or it is not known until runtime. If schema is null and format is not text, you should be setting output schema here to null.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -99,7 +99,11 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     if (config.getSchema() != null) {
       pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
     } else {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(DEFAULT_SCHEMA);
+      if (config.format.equalsIgnoreCase("text")) {
```
albertshau: this is still not correct. you need to set it to the output schema if you know it. If schema is set provided, you should have validated it already and you can set it as the output schema. If schema is not provided but format is text, you can determine the correct output schema depending on if pathfield is set. This is exactly what you're doing elsewhere in the code, when setting schema in the Configuration. If schema is not provided and format is not text, you set output schema to null.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -83,6 +83,16 @@ public static Schema getTextOutputSchema(@Nullable String pathField) {
     return Schema.recordOf("file.record", fields);
   }
 
+  public static Schema addFieldToSchema(Schema schema, String addFieldKey) {
+    List<Schema.Field> newFields = new ArrayList<>();
+    for (Schema.Field field : schema.getFields()) {
```
albertshau: replace this with
```
newFields = new ArrayList<>(schema.getFields().size() + 1);
newField.addAll(schema.getFields());
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
```diff
@@ -212,13 +208,13 @@ public void initialize(InputSplit split, TaskAttemptContext context) throws IOEx
...
-    }
-
-    public StructuredRecord getCurrentValueNull() throws IOException, InterruptedException {
-      return recordTransformer.transform(delegate.getCurrentKey().datum(), pathField, path);
+    public StructuredRecord.Builder startCurrentValue() throws IOException, InterruptedException {
+      if (schema != null) {
+        return recordTransformer.transform(delegate.getCurrentKey().datum(), schema, pathField);
+      }
+      Schema schema = recordTransformer.generateStructuredSchema(delegate.getCurrentKey().datum());
+      return recordTransformer.transform(delegate.getCurrentKey().datum(), addFieldToSchema(schema, pathField),
```
albertshau: this will break when path field is null, you should only add the field if path is not null.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
```diff
@@ -58,42 +43,34 @@ public StructuredRecord transform(GenericRecord genericRecord, Schema structured
       String fieldName = field.getName();
       builder.set(fieldName, convertField(genericRecord.get(fieldName), field.getSchema()));
     }
-
     return builder.build();
   }
 
-  public StructuredRecord transform(StructuredRecord.Builder builder, GenericRecord genericRecord,
-                                    Schema structuredSchema, String skipField)
+  public StructuredRecord.Builder transform(GenericRecord genericRecord, Schema structuredSchema, String skipField)
```
albertshau: should annotate skipField as `@Nullable`

core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java
```diff
@@ -58,42 +43,34 @@ public StructuredRecord transform(GenericRecord genericRecord, Schema structured
...
-    Schema schema = Schema.parseJson(genericRecord.getSchema().toString());
-    Schema.Field newField = Schema.Field.of(addFieldKey, Schema.of(Schema.Type.STRING));
-    List<Schema.Field> newFields = new ArrayList<>();
-
-    for (Schema.Field field : schema.getFields()) {
-      newFields.add(field);
-    }
-    newFields.add(newField);
-    Schema newSchema = Schema.recordOf(schema.getRecordName(), newFields);
+  public Schema generateStructuredSchema(GenericRecord genericRecord) throws IOException {
```
albertshau: take an avro Schema as the argument instead of a Generic Record and rename it convertSchema

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/PathTrackingInputFormat.java
albertshau: should do something like:
```
GenericRecord genericRecord = delegate.getCurrentKey().datum();
Schema recordSchema = schema;
if (schema == null) {
  recordSchema = recordTransformer.generateStructuredSchema(genericRecord.getSchema());
  if (pathField != null) {
    recordSchema = addFieldToSchema(recordSchema, pathField);
  }
}
return recordTransformer.transform(genericRecord, recordSchema, pathField);

```

sheivin:
@albertshau addressed comments. I had forgotten about the point related to the first comment earlier that the default schema (when schema is not set) can still have the pathField if it is not null (in the case of text). Now I am calling the function from PathTrackingInputFormat to set the schema if the format is text.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -96,7 +96,16 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     if (!config.containsMacro("timeTable") && config.timeTable != null) {
       pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
     }
-    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    if (config.getSchema() != null) {
+      pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+    } else {
+      if (config.format.equalsIgnoreCase("text")) {
```
albertshau: you can have one layer of ifs instead of having a nested if.

```
if (config.getSchema() != null) {
...
} else if ("text".equalsIgnoreCase(config.format)) {
...
} else {
...
}

File: core-plugins/src/main/java/co/cask/hydrator/plugin/common/AvroToStructuredTransformer.java [hunk 14]
```diff
@@ -55,20 +45,32 @@ public StructuredRecord transform(GenericRecord genericRecord, Schema structured
...
     }
-
     return builder.build();
   }
 
-  public StructuredRecord transform(StructuredRecord.Builder builder, GenericRecord genericRecord,
-                                    Schema structuredSchema, String skipField)
-    throws IOException {
+  public StructuredRecord.Builder transform(GenericRecord genericRecord, Schema structuredSchema, @Nullable String
+    skipField) throws IOException {
```
albertshau: variable type and name should be on the same line.

albertshau (commented):
a couple style comments, otherwise lgtm. fix and merge.

PR #700, opened 2017-08-08
CDAP-10653 add arguments to script transform contexts

albertshau:
Adding a way to get pipeline arguments from the context objects
exposed by the python and javascript transforms. Also updating
the docs to include examples of using arguments and emitting alerts.


albertshau:
https://builds.cask.co/browse/HYP-BAD346

sreevatsanraman:
LGTM

PR #701, opened 2017-08-09
CDAP-12226 add delimiter property to HDFS sink

albertshau:
Adding an optional delimiter property to the HDFS sink that allows
users to configure the delimiter used to concatenate fields.


gokulavasan:
LGTM 👍

albertshau:
https://builds.cask.co/browse/HYP-BAD349

PR #702, opened 2017-08-10
CDAP-9953 fix hive source and sink on spark

albertshau:
Changes the Hive source and sink to set the HCat schema as an
argument instead of writing it to a dataset. This allows the
plugins to be used in Spark since they no longer try to get a
dataset at runtime. This also fixes CDAP-12194, where a race
condition can occur if multiple pipelines run at the same time
and set the same key on the previously used dataset, causing
a transaction conflict exception.


albertshau:
https://builds.cask.co/browse/HYP-BAD348

rohitsinha54:
LGTM

PR #706, opened 2017-08-16
Updated pom for icon support

sheivin:


sheivin:
On hold till https://github.com/caskdata/cdap-maven-plugin is deployed to maven central

PR #707, opened 2017-08-16
making protocol lowercase

russorat:
https://issues.cask.co/browse/CDAP-10916

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
@@ -200,7 +200,7 @@ public class EmailAction extends PostAction {
       ClassLoader oldClassLoader = Thread.currentThread().getContextClassLoader();
       Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
       try {
-        Transport transport = session.getTransport(config.protocol);
+        Transport transport = session.getTransport(config.protocol.toLowerCase());
         transport.connect(config.host, config.port, config.username, config.password);
         try {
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/EmailAction.java
```diff
@@ -200,7 +200,7 @@ public PasswordAuthentication getPasswordAuthentication() {
       ClassLoader oldClassLoader = Thread.currentThread().getContextClassLoader();
       Thread.currentThread().setContextClassLoader(getClass().getClassLoader());
       try {
-        Transport transport = session.getTransport(config.protocol);
+        Transport transport = session.getTransport(config.protocol.toLowerCase());
```
anwar6953, line 203: the `Session.getTransport(String)` is case sensitive?
Then, what if there's some protocol that has an uppercase letter in it?

Edit: The [defaults](https://github.com/javaee/javamail/blob/master/mail/src/main/java/javax/mail/Session.java#L970-L988) seem to be all lowercase, but non-default protocols may have uppercase in the protocol, right?

russorat: what do you recommend? also, Session.getTransport(String) seems to expect lowercase versions of the protocols.

anwar6953: I guess its unlikely that someone will use a transport with a non-lowercase protocol, since most people will just use default protocols.
LGTM

russorat:
looks like the test isn't passing now. investigating.

PR #715, opened 2017-08-25
Add a profile to skipTests

sreevatsanraman:


chtyim:
LGTM

anwar6953:
This is needed because BUT build runs tests across all submodules, and so just doing `-DskipTests` would skip tests for all other submodules as well (such as cdap, cask-tracker, etc).

PR #718, opened 2017-09-09
Bump cdap versions for CDAP 5.0.0

anwar6953:


hdfs-plugins/pom.xml
```diff
@@ -27,11 +27,6 @@
   <artifactId>hdfs-plugins</artifactId>
   <modelVersion>4.0.0</modelVersion>
 
-  <properties>
-    <!-- properties for script build step that creates the config files for the artifacts -->
-    <app.parents>system:cdap-data-pipeline[4.3.0-SNAPSHOT,4.4.0-SNAPSHOT)</app.parents>
```
albertshau: don't remove these, they're here so that these plugins aren't available to realtime pipelines

anwar6953: Oh ok. Couldn't we restrict that just by plugin type? Realtime pipelines can exclude "batch" plugins? Or only include the "realtime" plugin?

albertshau: batchsink is used by both types of pipelines, so we can't restrict that way. It's probably worth updating the comment so that it explains what this is doing.

anwar6953: Sounds good; will do that.

albertshau (approved):
lgtm

PR #720, opened 2017-09-15
[CDAP-12526] Update tests to use Tephra's 0.13 transaction calls

anew:


sagarkapare (approved):

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLTPFSTestRun.java
```diff
@@ -310,8 +310,8 @@ public void testAvroSourceConversionToAvroSink() throws Exception {
     dataFileWriter.flush();
 
     txTpfs.commitTx();
-    txService.canCommit(tx1, txTpfs.getTxChanges());
-    txService.commit(tx1);
+    txService.canCommit(tx1.getTransactionId(), txTpfs.getTxChanges());
```
albertshau: we should remove usage of TransactionManager completely and just call fileSetManager.flush().

anew: agreed, will push another commit

albertshau (approved):
lgtm

PR #722, opened 2017-09-20
CDAP-12639-handle row id

CuriousVini:
Issue: https://issues.cask.co/browse/CDAP-12639

image

image

Here RID is the `RowId`.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
@@ -179,6 +179,9 @@ public final class DBUtils {
         type = Schema.Type.NULL;
         break;
 
+      case Types.ROWID:
+        break;
+
       case Types.BOOLEAN:
       case Types.BIT:
         type = Schema.Type.BOOLEAN;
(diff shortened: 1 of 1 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
```diff
@@ -224,7 +227,6 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
       case Types.JAVA_OBJECT:
       case Types.OTHER:
       case Types.REF:
-      case Types.ROWID:
       case Types.SQLXML:
       case Types.STRUCT:
         throw new SQLException(new UnsupportedTypeException("Unsupported SQL Type: " + sqlType));
```
bdmogal, line 232: is there any way to change this exception message to include the name of the sqlType instead of the number?

bdmogal:
minor comment. rest LGTM

database-plugins/src/main/java/co/cask/hydrator/plugin/DBUtils.java
CuriousVini: I am not sure how we can do it since we have not inferred the type.

PR #726, opened 2017-09-27
HbaseSource check Row field Type Error

xho22:
When Schema Type is UNION with null and other simple type. It will run
into an Exception


anwar6953:
@xho22 Why did you close this PR? Is this not a change that should be merged?

PR #728, opened 2017-10-10
Adds outputSchema plugin function to widget json of Splitter plugins

tonybach:
This is needed to fetch the new output schemas after the user has specified the field to split on.


tonybach:
Closed due to https://github.com/caskdata/hydrator-plugins/pull/729.

PR #732, opened 2017-10-13
Simplify ETL plugin tests with single method to deploy and run pipeline

ecfm:
https://builds.cask.co/browse/HYP-BAD371

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
@@ -22,10 +22,17 @@ import co.cask.cdap.api.artifact.ArtifactVersion;
 import co.cask.cdap.api.data.schema.Schema;
 import co.cask.cdap.api.dataset.lib.TimePartitionedFileSet;
 import co.cask.cdap.datapipeline.DataPipelineApp;
+import co.cask.cdap.datapipeline.SmartWorkflow;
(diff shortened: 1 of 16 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -178,4 +190,52 @@ protected void cleanPartitions(TimePartitionedFileSet fileSet) throws IOExceptio
...
+  }
+
+  /**
+   * Run the SmartWorkflow in the given ETL application for once and wait for the workflow's COMPLETED status
+   * with 5 minutes timeout.
+   *
+   * @param appManager the ETL application to run
+   * @param arguments the arguments to be passed when running SmartWorkflow
+   */
+  protected WorkflowManager runETLOnce(ApplicationManager appManager, Map<String, String> arguments)
```
anwar6953, line 218: this seems like a bad API - that if I want to pass in arguments, then I can't pass in a timeout.
And if I want to pass in a timeout, then I can't pass in arguments.

ecfm: I'm thinking maybe the API to set timeout is not necessary, since the default 5 min timeout should be sufficient.

anwar6953: For unit tests, I agree.

anwar6953 (approved):

PR #735, opened 2017-10-31
Update and pin the netty version to 4.1.16.Final.

anwar6953:
Update and pin the netty version to 4.1.16.Final.

### DIFF
--- a/http-plugins/pom.xml
+++ b/http-plugins/pom.xml
@@ -53,6 +53,14 @@
       <groupId>co.cask.cdap</groupId>
       <artifactId>hydrator-test</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.http</groupId>
+      <artifactId>netty-http</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/http-plugins/src/test/java/co/cask/hydrator/plugin/batch/HttpCallbackActionTest.java
+++ b/http-plugins/src/test/java/co/cask/hydrator/plugin/batch/HttpCallbackActionTest.java
@@ -96,9 +96,9 @@ public class HttpCallbackActionTest extends HydratorTestBase {
     List<HttpHandler> handlers = new ArrayList<>();
     handlers.add(new MockFeedHandler());
     httpService = NettyHttpService.builder("MockService")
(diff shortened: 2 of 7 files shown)


http-plugins/pom.xml
```diff
@@ -53,6 +53,12 @@
       <groupId>co.cask.cdap</groupId>
       <artifactId>hydrator-test</artifactId>
     </dependency>
+    <dependency>
```
chtyim, line 56: You should define this in the parent pom `dependencyManagement` section, with scope test there. Then in here, simply have:

```
<dependency>
  <groupId>io.netty</groupId>
  <artifactId>netty-all</artifactId>
</dependency>
```

http-plugins/src/test/java/co/cask/hydrator/plugin/batch/HttpCallbackActionTest.java
```diff
@@ -96,9 +96,9 @@ public static void setupTestClass() throws Exception {
     List<HttpHandler> handlers = new ArrayList<>();
     handlers.add(new MockFeedHandler());
     httpService = NettyHttpService.builder("MockService")
```
chtyim, line 98: Since `NettyHttpService` is used directly, you should have it declared as direct dependecy

chtyim: I mean a direct dependency on the cask netty-http library

anwar6953: ok; done

http-plugins/pom.xml
```diff
@@ -53,6 +53,18 @@
...
+      <groupId>io.netty</groupId>
+      <artifactId>netty-all</artifactId>
+    </dependency>
+    <dependency>
```
chtyim, line 60: Seems duplicated

chtyim (commented):
Only one comment about duplicated dependency

http-plugins/pom.xml
anwar6953: oops, yea. will remove.

chtyim:
LGTM

anwar6953:
Squashed commits. `mvn clean test` passed locally.
The bamboo build won't pass yet, because the cdap artifacts with the latest netty changes haven't been pushed yet to the snapshot repository.

PR #736, opened 2017-11-02
Fix error messages' formatting

anwar6953:
Otherwise, there are error messages like the following, which make debugging difficult:

```
co.cask.cdap.api.data.schema.UnsupportedTypeException: {} type is currently not supported in ORC
    at co.cask.hydrator.plugin.common.StructuredToOrcTransformer.convertToWritable(StructuredToOrcTransformer.java:122) ~[1509581904569-0/:na]
...
```

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
@@ -41,15 +41,15 @@ public final class CubeUtils {
       if (split.length != 2) {
         //should not happen
         throw new IllegalArgumentException(
-          String.format("For Property %s Parameter {}, name or value is empty, please check config", property, group));
+          String.format("For Property %s Parameter %s, name or value is empty, please check config", property, group));
       }
       if (Strings.isNullOrEmpty(split[0])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in %s", property));
+          String.format("Empty key/value is not allowed in property %s, group %s", property, group));
       }
       if (Strings.isNullOrEmpty(split[1])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in %s", property));
+          String.format("Empty key/value is not allowed in property %s, group %s", property, group));
       }
       result.put(keyFunction.apply(split[0]), split[1]);
     }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
@@ -41,15 +41,15 @@ public final class CubeUtils {
       if (split.length != 2) {
         //should not happen
         throw new IllegalArgumentException(
-          String.format("For Property {} Parameter {} , name or value is empty, please check config", property, group));
+          String.format("For Property %s Parameter {}, name or value is empty, please check config", property, group));
       }
       if (Strings.isNullOrEmpty(split[0])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in {}", property));
+          String.format("Empty key/value is not allowed in %s", property));
       }
       if (Strings.isNullOrEmpty(split[1])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in {}", property));
+          String.format("Empty key/value is not allowed in %s", property));
       }
       result.put(keyFunction.apply(split[0]), split[1]);
     }
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
@@ -203,7 +203,6 @@ public class FileSetUtil {
       HiveSchemaConverter.appendType(builder, schemaObj);
       return builder.toString();
     } catch (IOException e) {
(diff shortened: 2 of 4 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToOrcTransformer.java
```diff
@@ -120,7 +120,7 @@ private WritableComparable convertToWritable(Schema.Field field, StructuredRecor
           return new BytesWritable(Bytes.getBytes((ByteBuffer) fieldVal));
         }
       default:
-        throw new UnsupportedTypeException(String.format("{} type is currently not supported in ORC",
+        throw new UnsupportedTypeException(String.format("%s type is currently not supported in ORC",
```
ecfm: `throw new UnsupportedTypeException(String.format("Type '%s' of field '%s' is currently not supported in ORC",
                                                         fieldType.name(), field.getName()));` would be better

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
```diff
@@ -41,15 +41,15 @@ private CubeUtils() {
       if (split.length != 2) {
         //should not happen
         throw new IllegalArgumentException(
-          String.format("For Property {} Parameter {} , name or value is empty, please check config", property, group));
+          String.format("For Property %s Parameter {}, name or value is empty, please check config", property, group));
```
ecfm: also replace the second {}. Also I feel this message doesn't seem to make sense, `split.length != 2` may not just mean name or value is empty. Also is `group` a parameter?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
```diff
@@ -41,15 +41,15 @@ private CubeUtils() {
...
       }
       if (Strings.isNullOrEmpty(split[0])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in {}", property));
+          String.format("Empty key/value is not allowed in %s", property));
```
ecfm: Better to add `group` or `propertyValue` to the message too

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
```diff
@@ -41,15 +41,15 @@ private CubeUtils() {
...
       }
       if (Strings.isNullOrEmpty(split[1])) {
         throw new IllegalArgumentException(
-          String.format("Empty key/value is not allowed in {}", property));
+          String.format("Empty key/value is not allowed in %s", property));
```
ecfm: ditto

core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java
```diff
@@ -197,17 +197,16 @@ private static String parseHiveSchema(String schemaString, String configuredSche
...
   private static String parseOrcSchema(String configuredSchema) {
-    co.cask.cdap.api.data.schema.Schema schemaObj = null;
     try {
-      schemaObj = co.cask.cdap.api.data.schema.Schema.parseJson(configuredSchema);
+      co.cask.cdap.api.data.schema.Schema schemaObj = co.cask.cdap.api.data.schema.Schema.parseJson(configuredSchema);
       StringBuilder builder = new StringBuilder();
       HiveSchemaConverter.appendType(builder, schemaObj);
       return builder.toString();
     } catch (IOException e) {
       LOG.debug("{} is not a valid schema", configuredSchema, e);
```
ecfm: Unnecessary log?

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToOrcTransformer.java
```diff
@@ -60,7 +60,7 @@ public OrcStruct transform(StructuredRecord input, Schema schema) {
         orcRecord.setFieldValue(fields.get(i).getName(), writable);
       } catch (UnsupportedTypeException e) {
         LOG.debug("{} is not a supported type", field.getName(), e);
```
ecfm: unnecessary log?

albertshau (approved):
lgtm

core-plugins/src/main/java/co/cask/hydrator/plugin/common/StructuredToOrcTransformer.java
```diff
@@ -60,7 +60,7 @@ public OrcStruct transform(StructuredRecord input, Schema schema) {
...
-        throw new IllegalArgumentException(String.format("{} is not a supported type", field.getName()), e);
+        throw new IllegalArgumentException(String.format("%s is not a supported type", field.getName()), e);
```
ecfm: is it possible to include `input`?

anwar6953: The field type and name will already be in the exception message, so I think its fine.

core-plugins/src/main/java/co/cask/hydrator/plugin/common/CubeUtils.java
anwar6953: I'm not sure if group is a parameter, but I replaced the second `{}` now.

ecfm (approved):
LGTM

PR #739, opened 2017-11-07
Changes input field in splitter plugins' widget json to be of type input-field-selector

tonybach:
Since the field to split on in these plugins must always be a field from the input schema, I have changed the widget type from `textbox` to `input-field-selector` (which is a dropdown of fields in the input schema).


albertshau (approved):
lgtm

PR #741, opened 2017-11-08
Fix the normalize plugin

chtyim:
- After upgraded to Java 8, the behavior about generic method return type changed


chtyim:
Detail explanation in https://stackoverflow.com/questions/30521974/why-does-the-java-8-generic-type-inference-pick-this-overload

albertshau (approved):
lgtm

PR #744, opened 2017-11-10
set failOnError to false for javadoc plugin

gsps1:


gsps1:
Need to update the submodules : wrangler, condition plugins and realtime source.

chtyim:
LGTM

PR #747, opened 2017-11-16
fixes dependencies for spark and core plugins

albertshau:
remove usages of commons-lang3 from core plugins and add guava
as a direct dependency of spark-plugins.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
@@ -32,7 +32,6 @@ import co.cask.cdap.etl.api.Transform;
 import co.cask.cdap.etl.api.TransformContext;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.python.core.Py;
 import org.python.core.PyCode;
 import org.python.core.PyException;
@@ -41,6 +40,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
```diff
@@ -179,10 +180,17 @@ public void transform(StructuredRecord input, final Emitter<StructuredRecord> em
...
-      throw new IllegalArgumentException("Could not transform input.\n" + ExceptionUtils.getStackTrace(e));
+      throw new IllegalArgumentException("Could not transform input.\n" + getStackTrace(e));
     }
   }
 
+  private String getStackTrace(Throwable throwable) {
+    StringWriter sw = new StringWriter();
+    PrintWriter pw = new PrintWriter(sw, true);
+    throwable.printStackTrace(pw);
+    return sw.getBuffer().toString();
```
chtyim: Just do `sw.toString()`.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
```diff
@@ -179,10 +180,17 @@ public void transform(StructuredRecord input, final Emitter<StructuredRecord> em
     } catch (PyException e) {
       // We put the stack trace as the exception message, because otherwise the information from PyException is lost.
       // PyException only exposes the actual cause (Python stack trace) if printStackTrace() is called on it.
-      throw new IllegalArgumentException("Could not transform input.\n" + ExceptionUtils.getStackTrace(e));
+      throw new IllegalArgumentException("Could not transform input.\n" + getStackTrace(e));
```
chtyim, line 183: BTW, why the stacktrace is included as part of the message instead of the cause of the IllegalArgumentException?

albertshau: from what I remember, PyException has funny behavior where it overrides printStackTrace and does special logic. @anwar6953 may remember more details.

chtyim: I see.

chtyim:
LGTM

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/PythonEvaluator.java
anwar6953: I don't recall the details, but apparently, the comment above it mentions something similar to the reason Albert gave.

PR #750, opened 2017-11-27
CDAP-11870 remove usage of error datasets

albertshau:
Modified a unit test that used error datasets to use an error
collector instead.


gsps1:
LGTM

PR #751, opened 2017-11-27
use the new Engine api in test

gsps1:
`co.cask.cdap.etl.proto.Engine`  was removed in https://github.com/caskdata/cdap/pull/9787 

Dependent change to use the Engine class in CDAP-API.


albertshau (approved):
lgtm

PR #755, opened 2017-12-01
temporarily change group by aggregates widget

albertshau:
Changing the aggregate widget to be a csv widget in order to
support the macro use case. This is temporary until
function-dropdown-with-alias has support for macros.


sreevatsanraman:
LGTM

PR #756, opened 2017-12-06
CDAP-12942 removing macro support for aggregates

albertshau:
Removing macro support for the aggregates field until the UI has
a way to display it.


sreevatsanraman:
LGTM

PR #757, opened 2017-12-07
[CDAP-12883] Adds emit-errors and emit-alerts properties to widget json of Javascript, Python, and Scala editors

tonybach:
JIRA: https://issues.cask.co/browse/CDAP-12883

This lets the UI know that these plugins can emit alerts and errors. On-hold until ~~https://github.com/caskdata/cdap/pull/9817 is merged.~~ we know for sure this is how we want to surface this information.

### DIFF
--- a/core-plugins/widgets/JavaScript-transform.json
+++ b/core-plugins/widgets/JavaScript-transform.json
@@ -1,6 +1,6 @@
 {
   "metadata": {
-    "spec-version": "1.0"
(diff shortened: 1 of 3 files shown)


bdmogal (changes requested):
@tonybach can you please move these configs below `configuration-groups`? Don't know if this is immediately important or even if users will know what to do about it as the first thing they see

bdmogal (approved):

PR #758, opened 2017-12-08
Merge 1 8

albertshau:


pom.xml
```diff
@@ -945,6 +945,7 @@
             <configuration>
               <additionalparam>-Xdoclint:none</additionalparam>
               <failOnError>false</failOnError>
+              <additionalparam>-Xdoclint:none</additionalparam>
```
elfenheart: duplicate with line 946?

albertshau: good catch, guess it was done on both branches

bdmogal:
Everything else seems ok, but https://github.com/caskdata/hydrator-plugins/pull/758 is wrong, since it doesn't update the spec version to 1.5. @ajainarayanan ?

Not sure how we want to handle this.

albertshau:
after discussion, we'll merge this, as the spec stuff is a separate and larger issue

PR #761, opened 2017-12-11
CDAP-12942 re-enable macro for group property

albertshau:


bdmogal:
@albertshau, with https://github.com/caskdata/cdap/pull/9830, this change can be on `release/1.8` now as well.

albertshau:
ok, thought we weren't doing it on 1.8 for some reason. Will just do it there.

PR #766, opened 2018-01-02
CDAP-13026 fix parquet snapshot plugins to create the same dataset

albertshau:
This fixes a bug where snapshot parquet source and sink could not
be created in the same pipeline.


albertshau:
cherry-pick of #765

sreevatsanraman:
LGTM

PR #774, opened 2018-02-02
CDAP-12875 set parquet read schema

albertshau:
This works around an issue in CDAP versions prior to 5.0.0, where
avro classes where being exposed by the data-pipeline app. This
sets the avro read schema directly to avoid going down a code
path that would trigger the bug.


anew (approved):
LGTM.

PR #776, opened 2018-02-28
[CDAP-13139] Improve DB source schema generation

anew:
for decimal(n,0) and numeric (n,0), use int or long instead of double


albertshau (approved):
lgtm

PR #777, opened 2018-03-01
temporarily ignore test case

albertshau:
Ignore a test case for file source where a subset of the write
schema is used as the read schema. This functionality is broken
due to a workaround we had to put in place to let it read parquet
files that were created by a different library. Instead, catch
the error that happens and include an error message about modifying
the plugin schema.


sreevatsanraman (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD419

PR #786, opened 2018-04-23
CDAP-13339 add guava as direct dependency

albertshau:
HDFS sink directly uses guava, and should have it as a direct
dependency. At some point, it stopped getting bundled in the fat
jar since it was a transitive dependency.


chtyim:
LGTM

PR #788, opened 2018-05-09
Convert schema type if configured

sreevatsanraman:


core-plugins/widgets/Excel-batchsource.json
```diff
@@ -119,12 +119,6 @@
           "widget-attributes": {
             "showDelimiter": "false",
             "dropdownOptions": [
-              "boolean",
-              "bytes",
```
CuriousVini, line 123: why do we need this change?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/ExcelInputFormat.java
```diff
@@ -214,8 +230,13 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
             if (HSSFDateUtil.isCellDateFormatted(cell)) {
               sb.append(colName).append(COLUMN_SEPERATOR).append(cell.getDateCellValue()).append(CELL_SEPERATOR);
             } else {
-              sb.append(colName)
-                  .append(COLUMN_SEPERATOR).append(cell.getNumericCellValue()).append(CELL_SEPERATOR);
+              if (outputSchemaMapping.containsKey(colName)) {
+                String val = formatter.formatCellValue(cell);
```
CuriousVini, line 234: Why not use [DataFormatter](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.poi/poi/3.7/org/apache/poi/ss/usermodel/DataFormatter.java#590) for all the supported types? It uses `Format` for Numeric types which will avoid padding. That would make code simpler I believe.

PR #789, opened 2018-06-06
Support phoenix sink

anwar6953:
The changes in this PR are to support phoenix as a sink.
See JIRA for more details: https://issues.cask.co/browse/CDAP-13499
This also resolves https://issues.cask.co/browse/CDAP-13499.
This also resolves https://issues.cask.co/browse/CDAP-11096.

The changes are organized (and so easier to review) by commit.

### DIFF
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -47,9 +47,6 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
-**enableUpsert:** Whether to enable UPSERT for queries run by this sink. Defaults to false. For drivers
-such as Phoenix where INSERT is not supported, this should be set to true.
-
 **enableTransactionIsolation:** Whether to enable transaction isolation for queries run by this sink.
 Defaults to true. The Phoenix jdbc driver will throw an exception if the Phoenix database does not have
 transactions enabled and this setting is set to true. For drivers like that, this should be set to false.
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -47,6 +47,13 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
+**enableUpsert:** Whether to enable UPSERT for queries run by this sink. Defaults to false. For drivers
+such as Phoenix where INSERT is not supported, this should be set to true.
+
+**enableTransactionIsolation:** Whether to enable transaction isolation for queries run by this sink.
+Defaults to true. The Phoenix jdbc driver will throw an exception if the Phoenix database does not have
+transactions enabled and this setting is set to true. For drivers like that, this should be set to false.
+
 **schema:** The schema of records output by the source. This will be used in place of whatever schema comes 
 back from the query. However, it must match the schema that comes back from the query, 
 except it can mark fields as nullable and can contain a subset of the fields. 
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -30,6 +30,11 @@ names are the same when the case is ignored (optional).
 
 **connectionString:** JDBC connection string including database name. (Macro-enabled)
 
+**connectionArguments:** A list of arbitrary string tag/value pairs as connection arguments. This is a
+semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies the key and
+value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be given
+arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
+
 **user:** User identity for connecting to the specified database. Required for databases that need
 authentication. Optional for databases that do not require authentication. (Macro-enabled)
 
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -47,13 +47,10 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
-**enableTransactionIsolation:** Whether to enable transaction isolation for queries run by this sink.
-Defaults to true. The Phoenix jdbc driver will throw an exception if the Phoenix database does not have
-transactions enabled and this setting is set to true. For drivers like that, this should be set to false.
-
-**schema:** The schema of records output by the source. This will be used in place of whatever schema comes 
-back from the query. However, it must match the schema that comes back from the query, 
-except it can mark fields as nullable and can contain a subset of the fields. 
+**transactionIsolationLevel:** The transaction isolation level for queries run by this sink.
+Defaults to TRANSACTION_SERIALIZABLE. See java.sql.Connection#setTransactionIsolation for more details.
+The Phoenix jdbc driver will throw an exception if the Phoenix database does not have transactions enabled
+and this setting is set to true. For drivers like that, this should be set to TRANSACTION_NONE.
 
 Example
 -------
--- a/database-plugins/docs/Database-batchsink.md
+++ b/database-plugins/docs/Database-batchsink.md
@@ -30,10 +30,11 @@ names are the same when the case is ignored (optional).
 
 **connectionString:** JDBC connection string including database name. (Macro-enabled)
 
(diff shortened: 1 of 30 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/sink/DBSink.java
```diff
@@ -230,6 +255,8 @@ private void setResultSetMetadata() throws Exception {
       conf.put(DBConfiguration.OUTPUT_FIELD_NAMES_PROPERTY, dbSinkConfig.columns);
     }
 
+
```
anwar6953: I'll remove these spaces

database-plugins/docs/Database-batchsource.md
```diff
@@ -39,6 +39,11 @@ names are the same when the case is ignored (optional).
 
 **connectionString:** JDBC connection string including database name. (Macro-enabled)
 
+**connectionArguments:** A list of arbitrary string tag/value pairs as connection arguments. This is a
```
sreevatsanraman: This description does not makes sense from an end user perspective looking at the plugin. The widgets will do the translation. Can you document this as when users will need to use the connectionArguments. Same comment applies in other places.

sreevatsanraman (changes requested):
One minor comment. Rest looks good.

rohitsinha54:
Allow configuring transaction isolation level also in the database.... should have its jira id in commit

database-plugins/docs/Database-batchsink.md
```diff
@@ -47,9 +52,10 @@ Normally this setting does not matter. It only matters if you are using a jdbc d
 driver -- that will error when the commit operation is run, or a driver that will error when auto-commit is
 set to false. For drivers like those, you will need to set this to 'true'.
 
-**schema:** The schema of records output by the source. This will be used in place of whatever schema comes 
-back from the query. However, it must match the schema that comes back from the query, 
-except it can mark fields as nullable and can contain a subset of the fields. 
+**transactionIsolationLevel:** The transaction isolation level for queries run by this sink.
+Defaults to TRANSACTION_SERIALIZABLE. See java.sql.Connection#setTransactionIsolation for more details.
+The Phoenix jdbc driver will throw an exception if the Phoenix database does not have transactions enabled
```
rohitsinha54, line 58: Does it throw Exception or a particular exception?

database-plugins/src/main/java/co/cask/hydrator/plugin/ConnectionConfig.java
```diff
@@ -77,4 +93,40 @@ public ConnectionConfig() {
...
+  public static Properties getConnectionArguments(@Nullable String connectionArguments,
+                                                  @Nullable String user, @Nullable String password) {
+    KeyValueListParser kvParser = new KeyValueListParser("\\s*;\\s*", "=");
+
+    Map<String, String> connectionArgumentsMap = new HashMap<>();
+    if (!Strings.isNullOrEmpty(connectionArguments)) {
+      for (KeyValue<String, String> keyVal : kvParser.parse(connectionArguments)) {
+        String key = keyVal.getKey();
+        String val = keyVal.getValue();
+        connectionArgumentsMap.put(key, val);
```
rohitsinha54: Not need to extract as variable if its just used in in inner block

```
connectionArgumentsMap.put(keyVal.getKey(), keyVal.getValue());
```

database-plugins/src/main/java/co/cask/hydrator/plugin/ConnectionConfig.java
```diff
@@ -77,4 +93,40 @@ public ConnectionConfig() {
...
+   *
+   * @param connectionArguments See {@link ConnectionConfig#connectionArguments}.
+   * @param user See {@link ConnectionConfig#user}.
+   * @param password See {@link ConnectionConfig#password}.
+   */
+  public static Properties getConnectionArguments(@Nullable String connectionArguments,
+                                                  @Nullable String user, @Nullable String password) {
+    KeyValueListParser kvParser = new KeyValueListParser("\\s*;\\s*", "=");
+
+    Map<String, String> connectionArgumentsMap = new HashMap<>();
```
rohitsinha54, line 108: We don't name variable with theirs types connectionArguments is fine.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/TransactionIsolationLevel.java
```diff
@@ -0,0 +1,84 @@
...
+   * Translates a transaction isolation level string to the corresponding int constant.
+   * For instance, if given "TRANSACTION_NONE", will return 0, as that is the
+   * value of {@link Connection#TRANSACTION_NONE}. Defaults to {@link Connection#TRANSACTION_SERIALIZABLE}
+   * if the given input is {@code null}.
+   *
+   * @param level String version of the level.
+   * @return int corresponding to the constant in {@link Connection} for the level.
+   */
+  public static int getLevel(@Nullable String level) {
+    if (level == null) {
```
rohitsinha54, line 61: you an use ternary operator here to make code read more naturally 

```
return level == null ? Connection.TRANSACTION_SERIALIZABLE : Level.valueOf(level.toUpperCase()).level;
```

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/action/DBRun.java
```diff
@@ -17,29 +17,28 @@
...
   private final Class<? extends Driver> driverClass;
-  private static final Logger LOG = LoggerFactory.getLogger(DBRun.class);
 
   public DBRun(QueryConfig config, Class<? extends Driver> driverClass) {
     this.config = config;
     this.driverClass = driverClass;
   }
 
-  public void run() {
+  public void run() throws IOException, SQLException, InstantiationException, IllegalAccessException {
```
rohitsinha54: IOException is redundant

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/action/DBRun.java
```diff
@@ -17,29 +17,28 @@
...
   private final QueryConfig config;
   private final Class<? extends Driver> driverClass;
-  private static final Logger LOG = LoggerFactory.getLogger(DBRun.class);
 
   public DBRun(QueryConfig config, Class<? extends Driver> driverClass) {
     this.config = config;
     this.driverClass = driverClass;
   }
 
-  public void run() {
```
rohitsinha54, line 42: Not your PR but should add a java doc as this public

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/action/DBRun.java
```diff
@@ -17,29 +17,28 @@
...
+  public void run() throws IOException, SQLException, InstantiationException, IllegalAccessException {
     DBManager dbManager = new DBManager(config);
 
     try {
```
rohitsinha54: Not your PR but config can be null should check that and throw helpful message than NPE

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSourceTestRun.java
```diff
@@ -33,6 +33,7 @@
 import co.cask.hydrator.plugin.ConnectionConfig;
 import co.cask.hydrator.plugin.DBConfig;
 import co.cask.hydrator.plugin.DatabasePluginTestBase;
+import co.cask.hydrator.plugin.db.batch.TransactionIsolationLevel;
```
rohitsinha54: unused

database-plugins/src/main/java/co/cask/hydrator/plugin/ConnectionConfig.java
anwar6953: That variable name was already taken in this context.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/TransactionIsolationLevel.java
anwar6953: I think ternary operator makes the code more concise, but not read more naturally than what the code currently is.

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/action/DBRun.java
anwar6953: It won't actually be null when this code runs, because the constructor has a default, non-null value.

sreevatsanraman (approved):

database-plugins/src/main/java/co/cask/hydrator/plugin/db/batch/TransactionIsolationLevel.java
rohitsinha54: FYI for null check return statement we generally prefer ternary operator. Whole CDAP code base is full of it. But yes, since we don't have any defined set of rules it is your call!

PR #790, opened 2018-06-21
CDAP-13339 add guava as direct dependency.

anwar6953:
DBSink directly uses guava, and should have it as a direct
dependency. At some point, it stopped getting bundled in the fat
jar since it was a transitive dependency.

Similar to this PR: https://github.com/caskdata/hydrator-plugins/pull/786


albertshau (approved):
lgtm

PR #794, opened 2018-07-16
CDAP-13527 macro enable dbsink properties

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-13527
Build: https://builds.cask.co/browse/HYP-BAD436

These variables are not used in configurePipeline() method, so they can be macro-enabled


albertshau (commented):
Can you add a unit test for this?

albertshau (approved):
lgtm

PR #797, opened 2018-07-26
CDAP-13917

mattwuenschel:
Add file plugins as a submodule
https://issues.cask.co/browse/CDAP-13917

Update the branches for wrangler-transform, realtime-stream-source, and condition-plugins.


mattwuenschel:
On hold until https://github.com/data-integrations/wrangler branch release/3.1 gets cut.

PR #800, opened 2018-07-27
package file plugin

mattwuenschel:
Package the file plugin


albertshau:
closing this since we're adding File sink to core-plugins, to make sure its in the same artifact as File source. (#806)

PR #806, opened 2018-08-15
CDAP-14023 move file sink to core plugins

albertshau:
Moved hdfs sink to core-plugins as file sink, since it is able to
write to more filesystems than just hdfs. Also added the capability
to write files as avro or parquet in addition to just text.

Also created File versions of the HDFS move and delete actions
for the same reason. They are able to operate on file systems
other than just HDFS.

### DIFF
--- a/core-plugins/docs/File-batchsink.md
+++ b/core-plugins/docs/File-batchsink.md
@@ -9,6 +9,8 @@ For the text format, each record is written out as delimited text.
 Non-string fields will be converted to strings using their ``toString()`` Java method,
 so fields should be limited to the string, long, int, double, float, and boolean types.
 
+All types are supported when using the avro or parquet format.
+
 
 Properties
 ----------
@@ -25,6 +27,11 @@ If not specified, no suffix is used.
 
 **jobProperties:** Advanced feature to specify any additional properties that should be used with the sink,
 specified as a JSON object of string to string. These properties are set on the job at runtime. (Macro-enabled)
+
+**format:** The format to write in. Must be 'text', 'avro', or 'parquet'. Defaults to 'text'. (Macro-enabled)
+
+**schema:** The schema to use when writing data in 'avro' or 'parquet' format. Defaults to the input
+schema. (Macro-enabled)
 
 Example
 -------
--- a/core-plugins/docs/File-batchsink.md
+++ b/core-plugins/docs/File-batchsink.md
@@ -1,0 +1,42 @@
+# File Sink
+
+
+Description
+-----------
+Writes to a filesystem in text, avro, or parquet format.
+
+For the text format, each record is written out as delimited text.
+Non-string fields will be converted to strings using their ``toString()`` Java method,
+so fields should be limited to the string, long, int, double, float, and boolean types.
+
+
+Properties
+----------
+**referenceName:** This will be used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**path:** The directory to write to. For example, ``hdfs://mycluster.net:8020/my/desired/location``.
+
+**suffix:** Time suffix to append to the path for each run of the pipeline. For example,
+``YYYY-MM-dd-HH-mm`` will take the start time of the pipeline run, convert it into
+year-month-day-hour-minute format, and append that to the path to get the final output directory.
+If not specified, no suffix is used.
+
+**delimiter:** Delimiter used to concatenate record fields. Defaults to a comma ','. (Macro-enabled)
+
+**jobProperties:** Advanced feature to specify any additional properties that should be used with the sink,
+specified as a JSON object of string to string. These properties are set on the job at runtime. (Macro-enabled)
+
+Example
+-------
+This example writes to the Hadoop FileSystem with its namenode running on ``mycluster.net``,
+to the ``/etc/accesslogs/YYYY-MM-dd-HH-mm`` path. For example if the run was scheduled to
+run midnight on new years day 2016, the pipeline would write to ``/etc/accesslogs/2016-01-01-00-00``. 
+
+    {
+        "name": "HDFS",
+        "type": "batchsink",
+        "properties": {
+            "path": "hdfs://mycluster.net:8020/etl/accesslogs",
+            "suffix": "YYYY-MM-dd-HH-mm"
+        }
+    }
--- a/core-plugins/docs/File-batchsink.md
+++ b/core-plugins/docs/File-batchsink.md
@@ -40,7 +40,7 @@ to the ``/etc/accesslogs/YYYY-MM-dd-HH-mm`` path. For example if the run was sch
 run midnight on new years day 2016, the pipeline would write to ``/etc/accesslogs/2016-01-01-00-00``. 
 
     {
-        "name": "HDFS",
+        "name": "File",
         "type": "batchsink",
         "properties": {
             "path": "hdfs://mycluster.net:8020/etl/accesslogs",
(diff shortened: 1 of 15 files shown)


albertshau:
FileMove and FileDelete are just copies of HDFSMove and HDFSDelete, with 'hdfs' in descriptions removed. FileSink is a copy of HDFSSink, except with additional logic for avro and parquet formats.

albertshau:
https://builds.cask.co/browse/HYP-BAD447

bdmogal:
@albertshau for the file sink, can you add jump button config to the widget json per https://issues.cask.co/browse/CDAP-9048

core-plugins/docs/File-batchsink.md
```diff
@@ -0,0 +1,42 @@
+# File Sink
+
+
+Description
+-----------
+Writes to a filesystem in text, avro, or parquet format.
+
+For the text format, each record is written out as delimited text.
+Non-string fields will be converted to strings using their ``toString()`` Java method,
+so fields should be limited to the string, long, int, double, float, and boolean types.
```
gsps1, line 10: should we also document the behavior for parquet or avro format here.

core-plugins/docs/FileMove-action.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+Description
+-----------
+Moves a file or files.
+
+
+Use Case
+--------
+This action can be used when a file or files need to be moved to a new location in an HDFS cluster,
```
gsps1: should we remove `hdfs` ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/FileDeleteAction.java
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright © 2016 Cask Data, Inc.
```
gsps1: 2018

core-plugins/docs/File-batchsink.md
```diff
@@ -0,0 +1,42 @@
...
+**suffix:** Time suffix to append to the path for each run of the pipeline. For example,
+``YYYY-MM-dd-HH-mm`` will take the start time of the pipeline run, convert it into
+year-month-day-hour-minute format, and append that to the path to get the final output directory.
+If not specified, no suffix is used.
+
+**delimiter:** Delimiter used to concatenate record fields. Defaults to a comma ','. (Macro-enabled)
+
+**jobProperties:** Advanced feature to specify any additional properties that should be used with the sink,
+specified as a JSON object of string to string. These properties are set on the job at runtime. (Macro-enabled)
+
```
gsps1, line 30: there should be format and schema properties too right ?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
```diff
@@ -0,0 +1,263 @@
...
+      schema = context.getInputSchema();
+    }
+
+    if ("text".equals(config.format)) {
+      context.addOutput(Output.of(config.referenceName,
+                                  new SinkOutputFormatProvider(TextOutputFormat.class, hConf)));
+    } else if ("avro".equals(config.format)) {
+      if (schema == null) {
+        throw new IllegalStateException("Cannot write using avro format when schema is unknown. "
+                                          + "Please set the schema explicitly.");
```
gsps1, line 105: should these checks also be part of validate during config ? when its non-macro.

core-plugins/docs/FileMove-action.md
gsps1: might be good to say the source and destination path should be in the same filesystem.

albertshau: ah yes, missed this, will fix.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
albertshau: true, will do

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/action/FileDeleteAction.java
albertshau: done

core-plugins/docs/File-batchsink.md
albertshau: done

core-plugins/docs/File-batchsink.md
albertshau: done

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
```diff
@@ -241,12 +239,28 @@ private void validate() {
...
         new Gson().fromJson(fileSystemProperties, new TypeToken<Map<String, String>>() { }.getType());
       }
-      if (!"text".equalsIgnoreCase(format) && !"avro".equalsIgnoreCase(format) && !"parquet".equalsIgnoreCase(format)) {
+      if (!isText() && !isAvro() && !isParquet()) {
         throw new IllegalArgumentException(String.format("Invalid format '%s. Must be 'text', 'avro', or 'parquet'.",
                                                          format));
       }
       getFSProperties();
-      getSchema();
+      Schema schema = getSchema(inputSchema);
```
gsps1: getSchema parses the schema if its not null. however we should parse only if its not a macro right ? should we support macro for schema field ?

core-plugins/docs/File-batchsink.md
```diff
@@ -0,0 +1,49 @@
...
+schema. (Macro-enabled)
+
+Example
+-------
+This example writes to the Hadoop FileSystem with its namenode running on ``mycluster.net``,
+to the ``/etc/accesslogs/YYYY-MM-dd-HH-mm`` path. For example if the run was scheduled to
+run midnight on new years day 2016, the pipeline would write to ``/etc/accesslogs/2016-01-01-00-00``. 
+
+    {
+        "name": "HDFS",
```
gsps1: name is `File`

gsps1 (commented):
one question and one comment, lgtm otherwise.

core-plugins/docs/File-batchsink.md
albertshau: fixed

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
albertshau: ah good catch.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    private String getOutputDir(long logicalStartTime) {
+      String timeSuffix = !Strings.isNullOrEmpty(suffix) ? new SimpleDateFormat(suffix).format(logicalStartTime) : "";
+      return String.format("%s/%s", path, timeSuffix);
+    }
+
+    @Nullable
+    private Schema getSchema(@Nullable Schema inputSchema) {
+      if (containsMacro("schema")) {
```
gsps1, line 223: is this check required ? getSchema is called only if its not a macro in validate.

albertshau: it's not currently called, but thought it would be better to handle the scenario in this method too since it's pretty tricky.

gsps1: ok 👍

gsps1:
LGTM

PR #809, opened 2018-08-21
Adding date/time support in database plugins

CuriousVini:
**Dependency of PR**
Depends on https://github.com/caskdata/cdap/pull/10549, can not merge before that.

Summary:

- Adding support for date/time in database source and sink 
- Fixed a flaky test: `testNonExistentDBTable` - The test was using `waitForRun` to verify programRunStatus. However, that api only blocks until first run record of given programRunStatus. Instead of that using `waitForRuns` api which blocks until given number of run records are available.

### DIFF
--- a/database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2015 Cask Data, Inc.
+ * Copyright © 2015-2018 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -37,8 +37,10 @@ import java.sql.SQLException;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.List;
-import javax.annotation.Nullable;
 import javax.sql.rowset.serial.SerialBlob;
 
 /**
@@ -49,7 +51,6 @@ import javax.sql.rowset.serial.SerialBlob;
  * @see DBWritable DBWritable
  */
 public class DBRecord implements Writable, DBWritable, Configurable {
-
   private StructuredRecord record;
   private Configuration conf;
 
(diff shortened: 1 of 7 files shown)


database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
```diff
@@ -103,12 +105,26 @@ public void readFields(ResultSet resultSet) throws SQLException {
...
+  private void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field, int sqlType,
+                        int sqlPrecision, int sqlScale) throws SQLException {
+    Object o = DBUtils.transformValue(sqlType, sqlPrecision, sqlScale, resultSet, field.getName());
+    if (o instanceof Date) {
+      recordBuilder.setDate(field.getName(), ((Date) o).toLocalDate());
+    } else if (o instanceof Time) {
+      recordBuilder.setTime(field.getName(), ((Time) o).toLocalTime());
+    } else if (o instanceof Timestamp) {
+      Instant instant = ((Timestamp) o).toInstant();
+      recordBuilder.setTimestamp(field.getName(), ZonedDateTime.ofInstant(instant, ZoneId.systemDefault()));
```
albertshau: use `instant.atZone()`, and use UTC, not the system default.

database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
```diff
@@ -207,12 +224,11 @@ private void writeToDB(PreparedStatement stmt, Schema.Type fieldType, @Nullable
         stmt.setBoolean(sqlIndex, (Boolean) fieldValue);
         break;
       case INT:
-        // write short or int appropriately
-        writeInt(stmt, fieldIndex, sqlIndex, fieldValue);
+        writeIntOrDate(stmt, fieldName, fieldValue, fieldIndex, sqlIndex);
```
albertshau: methods like this are a bad pattern. They indicate that the condition being checked is happening in the wrong place.

instead of introducing writeIntOrDate, refactor this method to take a Schema.Field instead of a Schema.Type and field name. Then check for logical type before checking physical type. Once we get here, we should always be writing an int.

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSinkTestRun.java
```diff
@@ -86,6 +98,17 @@ public void testDBSink() throws Exception {
     }
   }
 
+  private ZonedDateTime getZonedDateTime(long ts, ZoneId zoneId, TimeUnit unit) {
```
albertshau: this doesn't look like it's called

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSourceTestRun.java
```diff
@@ -174,6 +179,17 @@ public void testDBSource() throws Exception {
     Assert.assertEquals(CLOB_DATA, row2.get("CLOB_COL"));
   }
 
+  private ZonedDateTime getZonedDateTime(long ts, ZoneId zoneId, TimeUnit unit) {
```
albertshau: this can be replaced with `Instant.ofEpochMillis(ts).atZone(zoneId)`

database-plugins/src/test/java/co/cask/hydrator/plugin/db/batch/sink/DBSourceTestRun.java
```diff
@@ -152,15 +158,14 @@ public void testDBSource() throws Exception {
...
     sdf = new SimpleDateFormat("H:mm:ss");
-    long expectedTimeTimestamp = Time.valueOf(sdf.format(date)).getTime();
-    Assert.assertEquals(expectedDateTimestamp, (long) row1.get("DATE_COL"));
-    Assert.assertEquals(expectedDateTimestamp, (long) row2.get("DATE_COL"));
-    Assert.assertEquals(expectedTimeTimestamp, (long) row1.get("TIME_COL"));
-    Assert.assertEquals(expectedTimeTimestamp, (long) row2.get("TIME_COL"));
-    Assert.assertEquals(CURRENT_TS, (long) row1.get("TIMESTAMP_COL"));
-    Assert.assertEquals(CURRENT_TS, (long) row2.get("TIMESTAMP_COL"));
+    LocalTime expectedTime = Time.valueOf(sdf.format(date)).toLocalTime();
+    ZonedDateTime expectedTs = getZonedDateTime(CURRENT_TS, ZoneId.systemDefault(), TimeUnit.MILLISECONDS);
```
albertshau: use UTC, not system default

database-plugins/src/main/java/co/cask/hydrator/plugin/DBRecord.java
CuriousVini: Thanks for the suggestion!

CuriousVini:
@albertshau Thanks for the review. I have addressed comments. Please take another pass when you get a chance.

albertshau (commented):
code lgtm but there seem to be unintentional changes to the condition-plugins, realtime-stream-source, and wrangler-transform submodules.

CuriousVini:
The are latest references to those submodules generated through `git submodule update --init --recursive --remote`. 

image

albertshau:
they still should not be changed in this PR

CuriousVini:
done

albertshau (approved):
lgtm

CuriousVini:
The dependent CDAP PR #10549 is merged. Build: https://builds.cask.co/browse/HYP-BAD450-10

CuriousVini:
The build passed. Squashed the commits, merging.

PR #810, opened 2018-08-31
CDAP-13924: Update requirement specification

rohitsinha54:
Updates with the changes in the PR: https://github.com/caskdata/cdap/pull/10610

This PR should only be merged after the above PR. 

Build: Waiting on CDAP PR to be merged to run.


albertshau (approved):
lgtm.

As an aside, even though a stream isn't a dataset, maybe we should still treat it as one for requirements.

PR #817, opened 2018-09-17
CDAP-14296 Reduce timestamp precision from millis to micros for avro source and sink

CuriousVini:
Build: https://builds.cask.co/browse/HYP-BAD459-6


albertshau:
decided not to do this

PR #820, opened 2018-09-18
Bump all submodules to the same version as parent submodule (a few were missing).

anwar6953:
In https://github.com/caskdata/hydrator-plugins/pull/816, not all of the submodule pom files were updated.


albertshau (approved):
lgtm

PR #822, opened 2018-09-21
CDAP-13274 Emit FLL from JavaScript, Python, GroupBy transform

rohitsinha54:
- Add FLL for JS and Python transform
- Add FLL in GroupBy transform

JS
image

Python
image

Group By
image

@albertshau : Showing the script in description. There is no checks on the description it is just stored as a string but the script can be really long. Should we display it?

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
@@ -25,12 +25,18 @@ import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.StageConfigurer;
 import co.cask.cdap.etl.api.batch.BatchAggregator;
+import co.cask.cdap.etl.api.batch.BatchAggregatorContext;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldTransformOperation;
+import co.cask.hydrator.common.SchemaValidator;
 import co.cask.hydrator.plugin.batch.aggregator.function.AggregateFunction;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import javax.ws.rs.Path;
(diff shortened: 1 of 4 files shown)


chtyim:
@ajainarayanan @elfenheart @tonybach Is there a way for the UI to format the code? Or at least show it as verbatim?

hydrator-common/src/main/java/co/cask/hydrator/common/SchemaValidator.java
```diff
@@ -116,6 +120,25 @@ public static void validateFieldsArePresentInSchema(Schema schema, String... req
...
   }
 
+  /**
+   * Checks that the given schema is not null
+   *
+   * @param schema the schema to be checked
+   * @param name the name of the schema
+   * @return true if the schema is not null else false
+   */
+  public static boolean validateSchemaIsNotNull(Schema schema, String name) {
```
albertshau: should be annotated as nullable

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -74,6 +81,29 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    List<String> outputFields = new ArrayList<>();
+    Schema inputSchema = context.getInputSchema();
+    if (SchemaValidator.validateSchemaIsNotNull(inputSchema, "input")) {
+      //noinspection ConstantConditions
+      inputFields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    Schema outputSchema = context.getOutputSchema();
+    if (SchemaValidator.validateSchemaIsNotNull(outputSchema, "output")) {
+      //noinspection ConstantConditions
+      outputFields = outputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
```
albertshau: group by should not be treated like javascript and python. There should be one transform operation per aggregate configured in the plugin.

For example, if we generate a new field 'avgPrice' by calculating 'avg(price)', that should be a single transform operation where the input is 'price', the output is 'avgPrice', and the operation name is probably something like 'GroupBy:avg'.

hydrator-common/src/main/java/co/cask/hydrator/common/SchemaValidator.java
albertshau: methods that return a boolean should start with 'is' or 'should' or something like that. Also, these log messages are very specific to FLL. 'canRecordLineage()' would be a better name.

core-plugins/src/main/java/co/cask/hydrator/plugin/transform/JavaScriptTransform.java
```diff
@@ -137,6 +144,26 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
...
+    if (SchemaValidator.validateSchemaIsNotNull(inputSchema, "input")) {
+      //noinspection ConstantConditions
+      inputFields = inputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    Schema outputSchema = context.getOutputSchema();
+    if (SchemaValidator.validateSchemaIsNotNull(outputSchema, "output")) {
+      //noinspection ConstantConditions
+      outputFields = outputSchema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList());
+    }
+    FieldOperation dataPrepOperation = new FieldTransformOperation("JavaScript", config.script, inputFields,
```
albertshau: the script could possibly be very long. I think it's best if we enforce a maximum length (we should have done this from wrangler as well). Maybe only include the first 30 lines or something? @bdmogal what do you think?

albertshau:
@chtyim There is a jira open to retain newlines in the UI.
@bdmogal @elfenheart what should happen if a line is longer than the cell width? Will it wrap or will there be a scrollbar? Any thoughts on what the maximum length for descriptions should be?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -84,24 +83,22 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-      " Aggregates: " + aggregates, inputFields, outputFields);
-    context.record(Collections.singletonList(dataPrepOperation));
+    LinkedList<FieldOperation> fllOperations = new LinkedList<>();
+    // in configurePipeline all the necessary checks have been performed already to set output schema
+    if (SchemaValidator.canRecordLineage(context.getOutputSchema(), "output")) {
+      Schema inputSchema = context.getInputSchema();
+      // for every function record the field level operation details
+      for (GroupByConfig.FunctionInfo functionInfo : conf.getAggregates()) {
+        Schema.Field outputSchemaField = getOutputSchemaField(functionInfo, inputSchema);
+        String operationName = String.format("Group '%s'", functionInfo.getField());
```
albertshau: don't need the enclosing single quotes.

albertshau (commented):
minor comment plus pending question about script length

rohitsinha54:
@albertshau Removed single quotes. How about for now we merge this and once we have a decision on how we want to treat the script we can change that in a separate PR.

albertshau (approved):
lgtm

PR #824, opened 2018-09-24
CDAP-14300 Fix orc sink plugin

CuriousVini:
Build: https://builds.cask.co/browse/HYP-BAD468-2


albertshau (approved):
I assume this protobuf was being used instead of the one from orc?

CuriousVini:
Yes, that right. Now its being used from right jar. 

image

CuriousVini:
The build is green. Merging.

PR #825, opened 2018-09-25
CDAP-14013: Record schema for external dataset

rohitsinha54:
### Summary
- Record Schema for External Datasets
- To record the Schema the dataset creation is done with the schema which is indexed by the metadata store as the schema of the external dataset.
- The creation is done in configurePipeline stage rather than prepareRun so that the external dataset is created when the pipeline is deployed and not when the run is started. 
- The Field Level Lineage uses the schema stored in the metadata store to return the schema of the dataset so this will now allow FLL to display the schema of external datasets.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
@@ -60,6 +60,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /**
  * Abstract file batch source
@@ -101,14 +102,19 @@ public abstract class AbstractFileBatchSource<T extends FileSourceConfig>
     if (!config.containsMacro("timeTable") && config.timeTable != null) {
       pipelineConfigurer.createDataset(config.timeTable, KeyValueTable.class, DatasetProperties.EMPTY);
     }
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(getSchema());
+  }
+
+  @Nullable
+  @Override
+  public Schema getSchema() {
     if (config.getSchema() != null) {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+      return config.getSchema();
     } else if ("text".equalsIgnoreCase(config.format)) {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(PathTrackingInputFormat.getTextOutputSchema
-        (config.pathField));
-    } else {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(null);
+      return PathTrackingInputFormat.getTextOutputSchema(config.pathField);
     }
+    //TODO: Should we return DEFAULT Schema here as that is what we emit record as
+    return null;
   }
 
   @Override
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSink.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSink.java
@@ -16,8 +16,13 @@
 
 package co.cask.hydrator.common;
 
+import co.cask.cdap.api.data.schema.Schema;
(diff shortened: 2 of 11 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/AbstractFileBatchSource.java
```diff
@@ -101,14 +102,19 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+      return config.getSchema();
     } else if ("text".equalsIgnoreCase(config.format)) {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(PathTrackingInputFormat.getTextOutputSchema
-        (config.pathField));
-    } else {
-      pipelineConfigurer.getStageConfigurer().setOutputSchema(null);
+      return PathTrackingInputFormat.getTextOutputSchema(config.pathField);
     }
+    //TODO: Should we return DEFAULT Schema here as that is what we emit record as
```
albertshau, line 116: is it even possible to get here?

hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSink.java
```diff
@@ -38,5 +43,14 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
     // Verify that reference name meets dataset id constraints
     IdUtils.validateId(config.referenceName);
+    // create the external dataset this is done in the configure pipeline stage so that the external dataset with or
+    // without the schema is created when the pipeline is deployed itself.
+    Schema externalSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
```
albertshau, line 48: this logic isn't correct, as sinks may transform their input before writing. If the config has a schema, it should be used. If it does not have a schema, then no schema should be set.

hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSource.java
```diff
@@ -38,5 +44,20 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    }
+    pipelineConfigurer.createDataset(config.referenceName, Constants.EXTERNAL_DATASET_TYPE,
+                                     DatasetProperties.of(dsProperties));
   }
+
+  /**
+   * @return The schema of the source if known else null
+   */
+  @Nullable
+  public abstract Schema getSchema();
```
albertshau, line 62: This encourages plugins to put schema parsing logic in their class instead of in their plugin config. In general, we want the config object to expose a getSchema() method so that the string value of the schema is never used. I think it makes more sense to put this in the config class and not have it as abstract, returning null by default.

hydrator-common/src/main/java/co/cask/hydrator/common/ReferenceBatchSource.java
```diff
@@ -38,5 +44,20 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
     // Verify that reference name meets dataset id constraints
     IdUtils.validateId(config.referenceName);
+    // create the external dataset this is done in the configure pipeline stage so that the external dataset with or
+    // without the schema is created when the pipeline is deployed itself.
+    Schema externalSchema = getSchema();
```
albertshau, line 49: Instead of using class inheritance for all of this logic, I think it makes more sense to encapsulate the lineage logic in some other class. It would remove code duplication in the source and sink and make it easier to use in other projects. I'm thinking of how we could use this in other projects like the gcp plugins that have their own class hierarchy, where they cannot extend this class or their configs can't extend ReferencePluginConfig. 

I think we can have a LineageRecorder class in hydrator-common that is responsible for creating this dataset if it doesn't exist, and also recording FLL. That way the plugin author doesn't need to know about external datasets and also doesn't have to remember to use the same name for the external dataset as the FLL endpoint.

rohitsinha54:
Closed in favor of https://github.com/caskdata/hydrator-plugins/pull/826

PR #826, opened 2018-09-26
CDAP-14013: Record schema for external datasets

rohitsinha54:
### Summary
- Record Schema for External Datasets
- To record the Schema the dataset creation is done with the schema which is indexed by the metadata store as the schema of the external dataset.
- The Field Level Lineage uses the schema stored in the metadata store to return the schema of the dataset so this will now allow FLL to display the schema of external datasets.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
@@ -29,6 +29,7 @@ import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
 import co.cask.cdap.etl.api.batch.BatchSink;
 import co.cask.cdap.etl.api.batch.BatchSinkContext;
+import co.cask.hydrator.common.LineageRecorder;
 import co.cask.hydrator.common.ReferenceBatchSink;
 import co.cask.hydrator.common.ReferencePluginConfig;
 import co.cask.hydrator.common.batch.JobUtils;
@@ -95,6 +96,11 @@ public class FileSink extends ReferenceBatchSink<StructuredRecord, Object, Objec
     }
 
     Schema schema = config.getSchema(context.getInputSchema());
+
+    // create the external dataset with the given schema
+    LineageRecorder lineageRecorder = new LineageRecorder(context, config.referenceName);
+    lineageRecorder.createExternalDataset(schema);
+
 
     if ("text".equals(config.format)) {
       context.addOutput(Output.of(config.referenceName,
@@ -223,11 +229,8 @@ public class FileSink extends ReferenceBatchSink<StructuredRecord, Object, Objec
       if (containsMacro("schema")) {
         return null;
       }
-      if (schema == null) {
-        return inputSchema;
-      }
       try {
-        return Schema.parseJson(schema);
+        return schema == null ? null : Schema.parseJson(schema);
       } catch (IOException e) {
         throw new IllegalArgumentException("Unable to parse schema: " + e.getMessage(), e);
       }
--- a/hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
+++ b/hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright © 2018 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package co.cask.hydrator.common;
+
+import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.dataset.DatasetManagementException;
+import co.cask.cdap.api.dataset.DatasetProperties;
+import co.cask.cdap.api.dataset.InstanceConflictException;
+import co.cask.cdap.api.lineage.field.EndPoint;
+import co.cask.cdap.api.lineage.field.ReadOperation;
+import co.cask.cdap.api.lineage.field.WriteOperation;
+import co.cask.cdap.etl.api.batch.BatchContext;
+import co.cask.cdap.etl.api.lineage.field.FieldReadOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldWriteOperation;
+import org.slf4j.Logger;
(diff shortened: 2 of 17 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
```diff
@@ -96,6 +97,11 @@ public void prepareRun(BatchSinkContext context) throws IOException {
 
     Schema schema = config.getSchema(context.getInputSchema());
 
+    if (schema != null) {
```
albertshau: don't think we need to check non-null, the dataset should be created regardless.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -0,0 +1,88 @@
...
+    this.context = context;
+    this.dataset = dataset;
+  }
+
+  /**
+   * Creates an external dataset with given schema is not null
+   *
+   * @param schema the schema of the external dataset
+   */
+  public void createDataset(@Nullable String schema) {
```
albertshau: should take a Schema object, not the string. That ensures it never records an invalid schema string.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -0,0 +1,88 @@
...
+    if (schema != null) {
+      try {
+        context.createDataset(dataset, Constants.EXTERNAL_DATASET_TYPE,
+                              DatasetProperties.of(Collections.singletonMap(DatasetProperties.SCHEMA, schema)));
+      } catch (DatasetManagementException e) {
+        Throwables.propagate(e);
```
albertshau: this is always equivalent to throw new RuntimeException(), except it has a less friendly error message. I think it's better to throw one with a nicer error message

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Creates an external dataset with given schema is not null
+   *
+   * @param schema the schema of the external dataset
+   */
+  public void createDataset(@Nullable String schema) {
+    if (schema != null) {
+      try {
+        context.createDataset(dataset, Constants.EXTERNAL_DATASET_TYPE,
```
albertshau: should check it does not exist before trying to create it.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -0,0 +1,88 @@
...
+                              DatasetProperties.of(Collections.singletonMap(DatasetProperties.SCHEMA, schema)));
+      } catch (DatasetManagementException e) {
```
albertshau: Can you tell if this failed because the dataset already exists? Should ignore those types of failures

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -0,0 +1,88 @@
...
+    this.dataset = dataset;
+  }
+
+  /**
+   * Creates an external dataset with given schema is not null
+   *
+   * @param schema the schema of the external dataset
+   */
+  public void createDataset(@Nullable String schema) {
+    if (schema != null) {
```
albertshau: why only create if the schema is not null? Is it to keep the input/output format properties? It seems inconsistent to sometimes have them and sometimes not. If we really want to keep them, this method should support passing in a properties map.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/sink/FileSink.java
rohitsinha54: The schema can be null here. The dataset will be created later on when context.addOutput is handled.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
rohitsinha54: Yes, the idea was that to keep the properties. I will change it now to create in all cases and if we want to retain the properties then we can add that in another PR later.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
rohitsinha54: Done. Handled the case of already existing.

hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/HiveConfig.java
```diff
@@ -50,6 +52,13 @@ public HiveConfig() {
     dbName = "default";
   }
 
+  @Name(Hive.SCHEMA)
+  @Description("Optional schema to use while reading (in case of Source) or writing (in case of Sink) Hive table. " +
```
albertshau: The description is user facing and should not have this type of language in it. Please move them back to where they were before. This means there is code duplication, but don't think there's anything we can do about that.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -43,18 +48,28 @@ public LineageRecorder(BatchContext context, String dataset) {
...
+    if (schema == null) {
+      datasetProperties = DatasetProperties.EMPTY;
+    } else {
+      datasetProperties = DatasetProperties.of(Collections.singletonMap(DatasetProperties.SCHEMA, schema.toString()));
+    }
+    try {
+      if (!context.datasetExists(dataset)) {
+        context.createDataset(dataset, Constants.EXTERNAL_DATASET_TYPE, datasetProperties);
+      } else {
+        LOG.info("Failed to create dataset {} with schema {}. A dataset with the given name already exists.",
```
albertshau: this is not a failure and should not be logged as one. Don't think this needs to be logged at all, it will happen whenever a pipeline is run more than once.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
```diff
@@ -43,18 +48,28 @@ public LineageRecorder(BatchContext context, String dataset) {
...
+                 dataset, schema);
       }
+    } catch (DatasetManagementException e) {
+      throw new RuntimeException(String.format("Failed to create dataset %s with schema %s.", dataset, schema), e);
```
albertshau, line 76: I don't see any handling for if this fails because the dataset already exists, which can happen when multiple pipelines are scheduled to run at the same time. Can you check what type of exception is thrown if you try to create a dataset that already exists? is it InstanceConflictException?

hive-plugins/src/main/java/co/cask/hydrator/plugin/batch/HiveConfig.java
rohitsinha54: Done.

hydrator-common/src/main/java/co/cask/hydrator/common/LineageRecorder.java
rohitsinha54: Done.

albertshau (approved):
lgtm

PR #827, opened 2018-09-27
CDAP-12852 add a property to copy headers of text files

albertshau:
This is a hack to allow wrangler's parse-as-csv to operate
correctly when a file is split into parts. Each record reader will
emit the header as its first record, which will allow an attached
wrangler transform to parse-as-csv with a header correctly.

The property is hidden, and is only expected to be set by the
dataprep service.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/CombinePathTrackingInputFormat.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/CombinePathTrackingInputFormat.java
@@ -17,8 +17,12 @@
 package co.cask.hydrator.plugin.batch.source;
 
 import co.cask.cdap.api.data.format.StructuredRecord;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.JobContext;
 import org.apache.hadoop.mapreduce.RecordReader;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.input.CombineFileInputFormat;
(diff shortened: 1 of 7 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/CombinePathTrackingInputFormat.java
```diff
@@ -17,33 +17,110 @@
...
+   */
+  @Override
+  public List<InputSplit> getSplits(JobContext job) throws IOException {
+    List<InputSplit> fileSplits = super.getSplits(job);
+    Configuration hConf = job.getConfiguration();
+
+    boolean copyHeader = hConf.getBoolean(PathTrackingInputFormat.COPY_HEADER, false);
+    List<InputSplit> splits = new ArrayList<>(fileSplits.size());
+
+    if (!copyHeader) {
```
CuriousVini: would this provide better readability? 

```java
  public List<InputSplit> getSplits(JobContext job) throws IOException {
    List<InputSplit> fileSplits = super.getSplits(job);
    Configuration hConf = job.getConfiguration();

    boolean copyHeader = hConf.getBoolean(PathTrackingInputFormat.COPY_HEADER, false);
    List<InputSplit> splits = new ArrayList<>(fileSplits.size());

    String header = null;
    for (InputSplit split : fileSplits) {
      CombineFileSplit combineFileSplit = (CombineFileSplit) split;

      if (copyHeader && header == null) {
        header = getHeader(hConf, header, combineFileSplit);
      }

      splits.add(new CombineHeaderFileSplit(combineFileSplit, header));
    }

    return splits;
  }
```

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
```diff
@@ -213,4 +219,8 @@ protected Schema getSchema() {
   }
 
   protected abstract String getPath();
+
+  boolean shouldCopyHeader() {
```
CuriousVini, line 223: why do we need this method?

CuriousVini (commented):
Are these changes also needed for other plugins like s3 and google cloud plugins?

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/CombinePathTrackingInputFormat.java
albertshau: yeah I agree, will restructure it.

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/FileSourceConfig.java
albertshau: using it to make it return a boolean instead of a Boolean to avoid warnings about possible nulls. The variable is annotated as nullable to tell cdap that it's optional even though we always set it to a value.

albertshau:
yeah this is needed in s3 and gcs plugins as well. I need to do a separate refactoring to make sure they all use common code instead of copied code

CuriousVini (approved):
LGTM 👍

PR #832, opened 2018-10-02
bump the netty http version to get race condition fix

albertshau:
This will fix flaky test problems where apps sometimes fail to
deploy due to CDAP internal dataset service calls.


rohitsinha54:
LGTM

PR #833, opened 2018-10-02
CDAP-12852 move file source validation logic

albertshau:
Moving validation logic from the config to the source. This is
to ensure that validation is performed when plugins don't
extend the AbstractFileSourceConfig and just implement the
FileSourceProperties interface.


rohitsinha54 (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD478

PR #834, opened 2018-10-02
CDAP-14355 Validate referenceName during pipeline deployment

rohitsinha54:
- Validate referenceName in configurePipeline stage.
- This is done in the validate which is being called in configurePipeline. 
- This validation is also done in ReferenceBatchSource so a plugin which extends it will have the validation. If a plugin does not extend it then the plugin must perform the validation.
- The validation here applies to all the file source/sink which uses this common format.


albertshau (approved):
lgtm

PR #835, opened 2018-10-03
CDAP-14321 add formats to file source

albertshau:
Added csv, tsv, delimited, json, and blob formats to the file
source. This is to mirror the formats that are available in
the sink, so that whatever content is written by a File sink
can also be read by a File source.

Refactored much of the input format logic to be similar to what
is done for output formats. This mostly amounted to moving the
inner static record readers from PathTrackingInputFormat into
their own classes, and creating hooks to get the record reader
directly from the FileFormat enum.

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -35,9 +35,14 @@ an inputFormatClass is given.
 **filenameOnly:** If true and a pathField is specified, only the filename will be used.
 If false, the full URI will be used. Defaults to false.
 
-**format:** Format of the file. Must be "text", "avro" or "parquet". Defaults to "text".
+**Format:** Format of the data to read.
+The format must be one of 'avro', 'blob', 'csv', 'delimited', 'json', 'parquet', 'text', or 'tsv'.
+If the format is 'blob', every input file will be read into a separate record.
+The 'blob' format also requires a schema that contains a field named 'body' of type 'bytes'.
+If the format is 'text', the schema must contain a field named 'offset' of type 'long' and a field
+named 'body' of type 'string'. (Macro-enabled)
 
-**schema:** Schema for the source.
+**schema:** Schema of the data to read.
 
 **maxSplitSize:** Maximum split-size for each mapper in the MapReduce Job. Defaults to 128MB. (Macro-enabled)
 
--- a/format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
+++ b/format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
@@ -17,6 +17,14 @@
 package co.cask.hydrator.format;
 
 import co.cask.cdap.api.data.schema.Schema;
+import co.cask.hydrator.format.input.AvroInputProvider;
+import co.cask.hydrator.format.input.BlobInputProvider;
+import co.cask.hydrator.format.input.DelimitedInputProvider;
+import co.cask.hydrator.format.input.FileInputFormatter;
+import co.cask.hydrator.format.input.FileInputFormatterProvider;
+import co.cask.hydrator.format.input.JsonInputProvider;
+import co.cask.hydrator.format.input.ParquetInputProvider;
+import co.cask.hydrator.format.input.TextInputProvider;
 import co.cask.hydrator.format.output.AvroOutputProvider;
 import co.cask.hydrator.format.output.DelimitedTextOutputProvider;
 import co.cask.hydrator.format.output.FileOutputFormatter;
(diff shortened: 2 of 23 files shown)


albertshau:
https://builds.cask.co/browse/HYP-BAD480

core-plugins/docs/File-batchsource.md
```diff
@@ -35,9 +35,14 @@ an inputFormatClass is given.
 **filenameOnly:** If true and a pathField is specified, only the filename will be used.
 If false, the full URI will be used. Defaults to false.
 
-**format:** Format of the file. Must be "text", "avro" or "parquet". Defaults to "text".
-
-**schema:** Schema for the source.
+**Format:** Format of the data to read.
+The format must be one of 'avro', 'blob', 'csv', 'delimited', 'json', 'parquet', 'text', or 'tsv'.
+If the format is 'blob', every input file will be read into a separate record.
+The schema must contain a field named 'body' of type 'bytes'.
```
rohitsinha54: This line while reading does not seems connected with the line above about blob. Can we say "Also" or something to show continuation.

format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
```diff
@@ -66,9 +77,27 @@ public boolean canRead() {
   public <K, V> FileOutputFormatter<K, V> getFileOutputFormatter(Map<String, String> properties,
                                                                  @Nullable Schema schema) {
     //noinspection unchecked
+    if (outputProvider == null) {
```
rohitsinha54, line 87: rather than null check can will it better to check the boolean flag which is being set?

format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
```diff
@@ -66,9 +77,27 @@ public boolean canRead() {
...
+  /**
+   * Create the FileInputFormatter for this format.
+   *
+   * @param properties plugin properties
+   * @param schema schema for the pipeline stage
+   * @return the FileInputFormatter for this format
+   * @throws IllegalArgumentException if the properties or schema are not valid
+   */
+  public FileInputFormatter getFileInputFormatter(Map<String, String> properties, @Nullable Schema schema) {
+    if (inputProvider == null) {
```
rohitsinha54, line 102: same here?

format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
```diff
@@ -66,9 +77,27 @@ public boolean canRead() {
   public <K, V> FileOutputFormatter<K, V> getFileOutputFormatter(Map<String, String> properties,
```
rohitsinha54, line 84: Can we add the doc like below here

format-common/src/main/java/co/cask/hydrator/format/input/BlobInputFormatter.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Map<String, String> getFormatConfig() {
+    return Collections.emptyMap();
+  }
+
+  @Override
+  public RecordReader<NullWritable, StructuredRecord.Builder> create(FileSplit split, TaskAttemptContext context) {
+    if (split.getLength() > Integer.MAX_VALUE) {
```
rohitsinha54, line 52: Is there a relation between the split length and the MAX_VALUE for file size greater than 2GB?

rohitsinha54 (approved):

format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
albertshau: doing this null check will remove a warning, which is one of the reasons it's here. If it's ever null it means somebody is using the class incorrectly.

format-common/src/main/java/co/cask/hydrator/format/input/BlobInputFormatter.java
albertshau: we can't create a byte[] using a long, which is why it's restricted by integer size. It's also a terrible idea to have a 2gb record.

core-plugins/docs/File-batchsource.md
albertshau: done

format-common/src/main/java/co/cask/hydrator/format/FileFormat.java
albertshau: done

format-common/src/main/java/co/cask/hydrator/format/input/BlobInputFormatter.java
rohitsinha54: 👍

PR #836, opened 2018-10-04
fixing potential flaky tests

albertshau:
Instead of using waitForRun(), switching to startAndWaitForRun().
The waitForRun method should not be used if the workflow can
run more than once, as it only checks for the existence of at
least one run record with that status.


rohitsinha54:
LGTM

PR #837, opened 2018-10-05
CDAP-14468 improve distinct aggregator widget

albertshau:
Makes the fields property a drop down and adds a get schema
button.

### DIFF
--- a/core-plugins/widgets/Distinct-batchaggregator.json
+++ b/core-plugins/widgets/Distinct-batchaggregator.json
@@ -1,15 +1,25 @@
 {
   "metadata": {
-    "spec-version": "1.0"
+    "spec-version": "1.5"
   },
   "configuration-groups": [
     {
       "label": "Distinct",
       "properties": [
         {
-          "widget-type": "textbox",
+          "widget-type": "csv",
           "label": "Fields",
-          "name": "fields"
+          "name": "fields",
+          "widget-attributes": {
+            "delimiter": ",",
+            "value-placeholder": "Field Name"
+          },
+          "plugin-function": {
+            "method": "POST",
+            "widget": "outputSchema",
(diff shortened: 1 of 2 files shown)


core-plugins/widgets/Distinct-batchaggregator.json
```diff
@@ -1,15 +1,25 @@
...
+          "widget-type": "csv",
           "label": "Fields",
-          "name": "fields"
+          "name": "fields",
+          "widget-attributes": {
+            "delimiter": ",",
+            "value-placeholder": "Field Name"
+          },
+          "plugin-function": {
+            "method": "POST",
```
CuriousVini, line 18: What is the default label for this widget?

core-plugins/widgets/Distinct-batchaggregator.json
```diff
@@ -1,15 +1,25 @@
...
+            "widget": "outputSchema",
+            "output-property": "schema",
+            "plugin-method": "outputSchema"
```
CuriousVini, line 21: do we need `position` and `button-class`?

CuriousVini (approved):
2 questions about widget json. Code changes lgtm 👍

core-plugins/widgets/Distinct-batchaggregator.json
albertshau: It shows up as 'Get Schema'

core-plugins/widgets/Distinct-batchaggregator.json
albertshau: don't know what those are, but it renders fine, and this is what the group by aggregator does.

CuriousVini:
LGTM 👍

PR #838, opened 2018-10-05
CDAP-14470 make offset optional when reading with text format

albertshau:
Makes the offset field optional for the file source. It will be
there by default, but users can remove it if they don't want it.

This is also required to fix issues with wrangler, where it
makes assumptions about column numbering and ends up creating
bad data if the offset field is there.

### DIFF
--- a/format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
+++ b/format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
@@ -53,12 +53,12 @@ public class TextInputProvider implements FileInputFormatterProvider {
     }
 
     if (schema == null) {
-      schema = getSchema(properties.get("pathField"));
+      schema = getDefaultSchema(properties.get("pathField"));
     }
     return new TextInputFormatter(schema);
   }
 
-  public static Schema getSchema(@Nullable String pathField) {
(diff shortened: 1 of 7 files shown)


format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
```diff
@@ -32,14 +32,13 @@
...
-        offsetSchema.getType();
-      if (offsetType != Schema.Type.LONG) {
-        throw new IllegalArgumentException("Type of 'offset' field must be 'long', but found " + offsetType);
+      if (offsetField != null) {
+        Schema offsetSchema = offsetField.getSchema();
+        Schema.Type offsetType = offsetSchema.isNullable() ? offsetSchema.getNonNullable().getType() :
+          offsetSchema.getType();
+        if (offsetType != Schema.Type.LONG) {
+          throw new IllegalArgumentException("Type of 'offset' field must be 'long', but found " + offsetType);
+        }
```
CuriousVini, line 41: Do we also need changes near line 63?

albertshau: no that just gets the default schema. Renaming that method.

CuriousVini (approved):
lgtm 👍

PR #839, opened 2018-10-05
Remove Stream Formatter transform

CuriousVini:
Remove Stream Formatter transform because it converts all the fields into text. This functionality can be achieved by `CSVFormatter`.

Build passed locally: 
image


albertshau (approved):
lgtm

PR #842, opened 2018-10-08
CDAP-14481 check for extra schema fields

albertshau:
Usability improvements to the file source to check if there are
extra fields in the schema that are not supported when using the
text or blob formats. This is to address user behavior where
the format is changed but schema is not changed, resulting in a
invalid schema.

Also adding a get schema button to help users generate the correct
schema when changing formats.

### DIFF
--- a/format-common/src/main/java/co/cask/hydrator/format/input/BlobInputProvider.java
+++ b/format-common/src/main/java/co/cask/hydrator/format/input/BlobInputProvider.java
@@ -17,7 +17,10 @@
 package co.cask.hydrator.format.input;
 
 import co.cask.cdap.api.data.schema.Schema;
+import co.cask.hydrator.format.plugin.FileSourceProperties;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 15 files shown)


format-common/src/main/java/co/cask/hydrator/format/input/BlobInputProvider.java
```diff
@@ -26,19 +29,50 @@
...
+      throw new IllegalArgumentException("The schema for the 'blob' format must have a field named 'body'");
+    }
+    Schema bodySchema = bodyField.getSchema();
+    Schema.Type bodyType = bodySchema.isNullable() ? bodySchema.getNonNullable().getType() : bodySchema.getType();
+    if (bodyType != Schema.Type.BYTES) {
+      throw new IllegalArgumentException(String.format("The 'body' field must be of type 'bytes', but found '%s'",
+                                                       bodyType.name().toLowerCase()));
+    }
+
+    int numExpectedFields = pathField == null ? 1 : 2;
```
yaojiefeng, line 63: I think it is good to add some comments to this method or the javadoc of this class to specify what is the validation logic for the blob format

format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
```diff
@@ -28,42 +29,80 @@
...
       }
     }
 
-    if (schema == null) {
-      schema = getDefaultSchema(properties.get("pathField"));
+    Schema.Field bodyField = schema.getField("body");
+    if (bodyField == null) {
+      throw new IllegalArgumentException("The schema for the 'text' format must have a field named 'body'");
     }
+    Schema bodySchema = bodyField.getSchema();
```
yaojiefeng, line 63: Same here, it is good to add some comments to this method or the javadoc of this class to specify what is the validation logic for the text format.

format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
```diff
@@ -28,42 +29,80 @@
...
     return new TextInputFormatter(schema);
   }
 
   public static Schema getDefaultSchema(@Nullable String pathField) {
     List<Schema.Field> fields = new ArrayList<>();
     fields.add(Schema.Field.of("offset", Schema.of(Schema.Type.LONG)));
-    fields.add(Schema.Field.of("body", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
-    if (pathField != null) {
-      fields.add(Schema.Field.of(pathField, Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    fields.add(Schema.Field.of("body", Schema.of(Schema.Type.STRING)));
```
yaojiefeng, line 106: Is there a reason we change this to not nullable?

yaojiefeng (approved):
Just some minor comments, lgtm

format-common/src/main/java/co/cask/hydrator/format/input/TextInputProvider.java
albertshau: it wasn't nullable in the previous release, and the widget for the file starts off with it not nullable.

It only became nullable during some recent refactoring, and reverting it back

PR #846, opened 2018-10-10
CDAP-14015 update widgets and reference for file source and sink

albertshau:


core-plugins/docs/File-batchsink.md
```diff
@@ -3,50 +3,28 @@
...
+If not specified, nothing will be appended to the path."
 
 **Format:** Format to write the records in.
-The format must be one of 'json', 'avro', 'parquet', 'csv', 'tsv', or 'delimited'. (Macro-enabled)
+The format must be one of 'json', 'avro', 'parquet', 'csv', 'tsv', or 'delimited'.
 
 **Delimiter:** Delimiter to use if the format is 'delimited'.
-The delimiter will be ignored if the format is anything other than 'delimited'.
-
-**Schema:** Schema of the data to write. (Macro-enabled)
```
yaojiefeng, line 37: Should we include some description about the schema? Since there can be some validation about the schema

core-plugins/docs/File-batchsource.md
```diff
@@ -3,86 +3,39 @@
...
 **Format:** Format of the data to read.
 The format must be one of 'avro', 'blob', 'csv', 'delimited', 'json', 'parquet', 'text', or 'tsv'.
 If the format is 'blob', every input file will be read into a separate record.
 The 'blob' format also requires a schema that contains a field named 'body' of type 'bytes'.
-If the format is 'text', the schema must contain a field named 'body' of type 'string'. (Macro-enabled)
+If the format is 'text', the schema must contain a field named 'body' of type 'string'.
+
+**Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
-**schema:** Schema of the data to read.
```
yaojiefeng, line 44: Same here

yaojiefeng (approved):
just two minor questions about the schema, otherwise lgtm

core-plugins/docs/File-batchsink.md
albertshau: the sink doesn't actually validate schema, it just uses whatever is configured. The source does validation, which is mentioned in the format section.

core-plugins/docs/File-batchsource.md
albertshau: It's mentioned in the format description, as that is what determines the schema restrictions. A description like 'Schema of the data to read.' does not give any information than knowing that schema is there.

PR #849, opened 2018-10-10
drop snapshot

rohitsinha54:
Submodules will have version bumps in their own repo and then updated here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>co.cask.hydrator</groupId>
   <artifactId>hydrator-plugins</artifactId>
-  <version>2.1.0-SNAPSHOT</version>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -128,7 +128,7 @@
     <json-path.version>2.2.0</json-path.version>
     <json.version>20160212</json.version>
     <netty.version>4.1.16.Final</netty.version>
-    <netty-http.version>1.1.0-SNAPSHOT</netty-http.version>
```
albertshau: I did this already. Is this from restoring the branch from an earlier point?

rohitsinha54: Yes. It was one commit ahead. 

image

albertshau (approved):
lgtm

PR #856, opened 2018-10-29
Record FLL for the Table source plugin.

sagarkapare:
Recording `READ` operation from the Table source plugin.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TableSource.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TableSource.java
@@ -22,18 +22,27 @@ import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.annotation.Requirements;
 import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
+import co.cask.cdap.api.dataset.DatasetManagementException;
 import co.cask.cdap.api.dataset.lib.KeyValue;
 import co.cask.cdap.api.dataset.table.Row;
 import co.cask.cdap.api.dataset.table.Table;
+import co.cask.cdap.api.lineage.field.EndPoint;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.PipelineConfigurer;
 import co.cask.cdap.etl.api.batch.BatchRuntimeContext;
+import co.cask.cdap.etl.api.batch.BatchSourceContext;
+import co.cask.cdap.etl.api.lineage.field.FieldOperation;
+import co.cask.cdap.etl.api.lineage.field.FieldReadOperation;
+import co.cask.hydrator.common.LineageRecorder;
 import co.cask.hydrator.common.RowRecordTransformer;
 import co.cask.hydrator.plugin.common.Properties;
 import co.cask.hydrator.plugin.common.TableSourceConfig;
 import com.google.common.collect.Maps;
 
+import java.io.IOException;
+import java.util.Collections;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * CDAP Table Dataset Batch Source.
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/TableSource.java
```diff
@@ -75,6 +83,27 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     }
   }
 
+  @Override
+  public void prepareRun(BatchSourceContext context) throws DatasetManagementException {
+    super.prepareRun(context);
```
rohitsinha54, line 89: We introduced LineageRecorder to make lineage recording require less code. Please consider using that.

sagarkapare: I tried using that but does not seem to help much in this case since we are not creating any external datasets here. Also I am not sure why we create external datasets in the LineageRecorder class since these are two different things.  Because of this association, the class is not usable in the transform plugins as well right?

rohitsinha54: LineageRecorder is mostly just to help in recording lineage. We wrapped the externalDataset creation and recording lineage as lineage for external dataset is recorded as that dataset and both happens with consistent referenceName. 
It is not a complete replacement for recording lineage as it just exposes helper for read and write. This is what we needed so far. Transform helper can be added if needed in future.

rohitsinha54: Also don't get the schema string and parse it here again. 
You can do something like this:

```
@Override
  public void prepareRun(BatchSourceContext context) throws DatasetManagementException {
    super.prepareRun(context);
    LineageRecorder lineageRecorder = new LineageRecorder(context, tableConfig.getName());
    Schema schema = tableConfig.getSchema();
    if (schema != null && schema.getFields() != null) {
      lineageRecorder.recordRead("Read", "Read from Table dataset",
                                 schema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList()));  
    }
  }
```

sagarkapare: Good point about the schema.. fixed it. 
However I am still not sure why we need to create `LineageRecorder` here. Based on the second parameter to the constructor of `LineageRecorder` it seems like its the name of the external dataset, however we do not create external dataset for Table dataset.  Also coupling of creation of "external" dataset with the field level lineage recording seems weird since they are independent of each other.

anew: I am not sure I see the big difference between using LineageRecorder and directly recording the read through the context. Are there examples of where the LineageRecorder significantly simplifies the code? 

Also I am surprised by the choice of naming for this utility class. LineageRecorder is also the name of the interface that is implemented by the context to record lineage directly. However, this LineageRecorder does not implement that interface.

rohitsinha54: The second parameter is the name with which the lineage needs to be recorded. The externalDataset  does not need to be created if it is not needed. They are not coupled as externalDataset does not necessarily need to be be created to record lineage or that lineage cannot be recorded without creating one. They are just two different helper method exposed by utility class.

Agreed, here the difference seems insignificant. It simplifies code in cases where external dataset needs to be created that is all external source/sink like google cloud plugins etc. It also to standardize that the external dataset name lineage record are done in same name.
Examples can found here: 
https://github.com/data-integrations/google-cloud/blob/release/0.11/src/main/java/co/cask/gcp/spanner/source/SpannerSource.java#L133
and
https://github.com/caskdata/hydrator-plugins/blob/develop/format-common/src/main/java/co/cask/hydrator/format/plugin/AbstractFileSink.java#L79

I don't have a strong preference about this but I do believe that following one consistent way will help. Yes the name does not seem to be most intuitive. It should be renamed to signify that it is a helper class.

anew: I don't mind using the utility class, just don't see how it helps in this case.

anew (approved):

PR #857, opened 2018-10-30
CDAP-14442 refactor file sink formats into plugins

albertshau:
This will fix the classloading issues we have when plugins from
different artifacts are using avro or parquet.


albertshau:
https://builds.cask.co/browse/HYP-BAD499

format-avro/src/main/java/co/cask/format/avro/AvroOutputFormatProvider.java
```diff
@@ -0,0 +1,107 @@
...
+  public Map<String, String> getOutputFormatConfiguration() {
+    Map<String, String> configuration = new HashMap<>();
+    if (conf.schema != null) {
+      configuration.put(SCHEMA_KEY, conf.schema);
+    }
+
+    if (conf.compressionCodec != null && !conf.containsMacro("compressionCodec") &&
+      !"none".equalsIgnoreCase(conf.compressionCodec)) {
+
+      configuration.put(MAPRED_OUTPUT_COMPRESS, "true");
```
chtyim: Better use the `CodecFactory` class from the avro library to parse and pickup the codec to simplify the logic.

```java
try {
  CodecFactory.fromString(conf.compressionCodec);
} catch (Exception e) {
  throw new IllegalArgumentException(.....);
}

File: format-parquet/pom.xml [hunk 2]
```diff
@@ -0,0 +1,105 @@
...
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>3.3.0</version>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <_exportcontents>
+              co.cask.format.parquet.*;
+              org.apache.parquet.avro.*;
+              org.apache.parquet.hadoop.*;
```
chtyim, line 78: Why need to expose the parquet? I thought we wrap the input/output format class?

albertshau: oh true, I had just copied these over without thinking. I'll try removing these.

albertshau: There are some classloading issues if I remove these, I think because of the fact that they haven't been removed from core-plugins yet since I haven't moved the inputformats out yet.

chtyim: I see. So I suppose these will be removed when the inputformat/source modification is done.

chtyim (approved):

PR #860, opened 2018-11-02
CDAP-14442 refactoring input formats into plugins

albertshau:
Moved file input formats into their own plugins to consolidate
logic in to a single place and to fix classloading issues.

As part of this refactoring, removed avro, parquet, and orc as
dependencies in core-plugins.


albertshau:
https://builds.cask.co/browse/HYP-BAD504

chtyim (approved):
LGTM

PR #861, opened 2018-11-05
CDAP-14560 remove lzo as a compression option

albertshau:
LZO is not packaged in the plugins due to licensing issues and
should not be listed as a compression option, as it is guaranteed
to break.


CuriousVini (approved):
Code changes lgtm 👍 

Should we also remove: https://github.com/caskdata/hydrator-plugins/blob/91ea9bae30ebfb959913455c741d374946fde7be/core-plugins/src/main/java/co/cask/hydrator/plugin/common/FileSetUtil.java#L55?

sreevatsanraman (approved):

albertshau:
@CuriousVini that was refactored and no longer exists on the release branch

PR #866, opened 2018-11-09
Depend on cdap 5.1.1

rohitsinha54:
Depend on cdap 5.1.1. This is needed to run the ITN where cdap is bumped to 5.1.1 and we want hydrator to use 5.1.1 which has the [latest fix](https://github.com/caskdata/cdap/pull/10832) rather than fetching 5.1.0 from maven central.

Drop snapshot of hydrator will come in another PR later.


CuriousVini (approved):
lgtm

PR #872, opened 2018-12-03
remove realtime stream source

albertshau:
CDAP streams are being removed in 6.0.0 so removing the stream
source as a submodule


chtyim (approved):

anwar6953:
Should we also remove other plugins, such as https://github.com/cdapio/hydrator-plugins/blob/develop/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/source/StreamBatchSource.java

albertshau:
yeah we should. I'm guessing it's being used in some integration tests so we would want to modify those first

PR #874, opened 2018-12-06
Remove reference to StreamFormatter.

anwar6953:
StreamFormatter was removed in https://github.com/cdapio/hydrator-plugins/pull/839.

Same change as https://github.com/cdapio/hydrator-plugins/pull/873, but against release/2.1.


albertshau (approved):
lgtm

PR #876, opened 2018-12-11
remove format classes from core-plugins test artifact

albertshau:
This avoids classloading conflicts in unit tests, where the same
class is defined in multiple classloaders, due to being exposed
in multiple artifacts.

### DIFF
--- a/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
+++ b/core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
@@ -99,6 +99,7 @@ import org.apache.parquet.hadoop.ParquetReader;
 import org.apache.twill.filesystem.Location;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
+import org.xerial.snappy.Snappy;
 
 import java.io.IOException;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


core-plugins/src/test/java/co/cask/hydrator/plugin/batch/ETLBatchTestBase.java
```diff
@@ -182,7 +183,8 @@ public static void setupTest() throws Exception {
     addPluginArtifact(NamespaceId.DEFAULT.artifact("formats-parquet", "4.0.0"), DATAPIPELINE_ARTIFACT_ID,
                       ImmutableSet.of(ParquetOutputFormatProvider.PLUGIN_CLASS,
                                       ParquetInputFormatProvider.PLUGIN_CLASS),
-                      ParquetOutputFormatProvider.class, ParquetInputFormatProvider.class);
+                      ParquetOutputFormatProvider.class, ParquetInputFormatProvider.class,
+                      Snappy.class);
```
anwar6953, line 187: Why need to add Snappy here? We're not removing anything from the formats-parquet otherwise, right?

albertshau: I'm not entirely sure why, but without this a parquet snappy test fails, maybe because it doesn't get pulled into the test jar through dependency tracing (in the jar built through maven, it's there)

anwar6953 (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD525

PR #885, opened 2019-02-05
Fix RunningStats#stddev

anwar6953:
https://issues.cask.co/browse/CDAP-14857

If there is only one entry, then the current implementation doesn't handle that well (divide by 0 error).
The standard deviation of a single element is defined as 0, so making the implementation reflect that.

Also adding unit tests for the RunningStats class, which fails without the fix to the class.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
@@ -22,25 +22,26 @@ package co.cask.hydrator.plugin.batch.aggregator.function;
  * http://www.johndcook.com/blog/skewness_kurtosis/
  */
 public final class RunningStats  {
-  private long entries = 0L;
+  private long numEntries = 0L;
   private double mean1, mean2, mean3, mean4 = 0d;
 
   /**
-   * Pushes a number into machinary that computes a lot of statistics.
+   * Pushes a number into machinery that computes a lot of statistics.
    * @param x number to be added to computing statistics.
    */
   public void push(double x) {
     double delta, deltaN, deltaN2, term1;
 
-    long n1 = entries;
-    entries++;
+    long n1 = numEntries;
+    numEntries++;
     delta = x - mean1;
-    deltaN = delta / entries;
+    deltaN = delta / numEntries;
     deltaN2 = deltaN * deltaN;
     term1 = delta * deltaN * n1;
     mean1 += deltaN;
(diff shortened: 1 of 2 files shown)


sreevatsanraman (approved):

core-plugins/src/main/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStats.java
```diff
@@ -55,7 +56,10 @@ public double mean() {
    * @return Variance of all numbers.
    */
   public double variance() {
-    return mean2 / (entries - 1.0d);
+    if (numEntries == 1) {
+      return 0;
+    }
+    return mean2 / (numEntries - 1);
```
albertshau: Using (numEntries - 1) means this is calculating a sample variance and not the true variance of the data being read in the pipeline. I don't think this is what we want anyway, and that this should have just been numEntries instead of (numEntires - 1) to begin with.

anwar6953: I think that the [source blog](https://www.johndcook.com/blog/skewness_kurtosis/) was using [Bessel's correction](https://en.wikipedia.org/wiki/Bessel%27s_correction) for computing the sample statistic.

I think that even if we use numEntries for the denominator, it won't give us the true variance. Instead, we'll just be accounting for the bias less in the sample statistic.

albertshau: No, the formula for stddev of a sample and of a population just differs by that denominator, whether you subtract 1. Without subtracting 1, you get the actual variance of the data that was processed. Subtracting 1 means you think the data is a sample of a larger population and you are estimating the variance of that population. 

For the purposes of a pipeline, I think it makes more sense to treat it like a population, not as a sample.

anwar6953: Ok, I pushed this diff, as well as a change to the test case accordingly:
image

albertshau (approved):
lgtm

core-plugins/src/test/java/co/cask/hydrator/plugin/batch/aggregator/function/RunningStatsTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @Test
+  public void testVariance() {
+    RunningStats runningStats = new RunningStats();
+    runningStats.push(5);
+    runningStats.push(6);
+    runningStats.push(10);
+    runningStats.push(14);
+    runningStats.push(15);
+
+    Assert.assertEquals(4.04969, runningStats.stddev(), 0.001);
```
anwar6953, line 56: verified correctness:
image

PR #889, opened 2019-02-14
Feature/cdap 14738 python improvements move out commons

aonischuk:
There are some common classes used by both Python and Javascript transform.
The scope of this jira is to:
a) move common files to hydrator-commons so they can be used by python-plugins repo.
b) remove Python transform from hydrator-plugins.


hydrator-common/src/main/java/co/cask/hydrator/common/validator/CoreValidator.java
```diff
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package co.cask.hydrator.plugin.validator;
+package co.cask.hydrator.common.validator;
```
albertshau: Unless I'm missing something, I don't think this class needs to be moved.

Would like to keep it in core-plugins to avoid adding a commons dependency to hydrator-common. This it is a module shared across many projects, we try to keep it as light as possible. It actually shouldn't even contain guava ideally.

aonischuk: @albertshau fixed that. Can you please review again.

It was used in PythonEvaluatorTest.java but actually seems like we can live without it there.
So I have removed the usage from  PythonEvaluatorTest and will undo moving to commons.

albertshau (approved):
lgtm

PR #892, opened 2019-02-28
CDAP-14844 remove plugin endpoints

albertshau:
Removed all plugin endpoints except for the one in DBSource, which
will be removed in a future commit. Output schema is now fetched
through the system service validation endpoint.

### DIFF
--- a/core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
+++ b/core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
@@ -21,17 +21,16 @@ import co.cask.cdap.api.annotation.Name;
 import co.cask.cdap.api.annotation.Plugin;
 import co.cask.cdap.api.data.format.StructuredRecord;
 import co.cask.cdap.api.data.schema.Schema;
-import co.cask.cdap.api.plugin.EndpointPluginContext;
 import co.cask.cdap.api.plugin.PluginConfig;
 import co.cask.cdap.etl.api.Emitter;
 import co.cask.cdap.etl.api.ErrorRecord;
 import co.cask.cdap.etl.api.ErrorTransform;
 import co.cask.cdap.etl.api.PipelineConfigurer;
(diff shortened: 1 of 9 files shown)


core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
```diff
@@ -50,27 +49,27 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
     Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
     if (inputSchema != null) {
       if (config.messageField != null && inputSchema.getField(config.messageField) != null) {
-        throw new IllegalArgumentException(String.format(
+        throw new InvalidConfigPropertyException(String.format(
```
CuriousVini, line 52: Are we using `InvalidConfigPropertyException ` instead of `IllegalArgumentException `?

CuriousVini (commented):
should we also remove getSchema endpoint from other plugins such as google-cloud?

core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
albertshau: yep, this will allow better UI integration in the future, as it will be able to highlight which field is invalid.

albertshau:
yes, will remove those in google-cloud plugins after this.

core-plugins/src/main/java/co/cask/hydrator/plugin/error/ErrorCollector.java
CuriousVini: I see. should we also change it for other plugins, eg: Joiner, DedupAggregator?

albertshau: Only when it's tied to a specific config property. If it's something about the input schema then it doesn't really make sense.

CuriousVini (approved):
lgtm 👍

PR #893, opened 2019-03-07
CDAP-14772 remove plugin endpoint from db source

albertshau:
Instead, the source will lookup the schema if the schema is null,
and has the UI configured to always send a null schema when using
the get schema button.


CuriousVini (approved):
lgtm 👍

elfenheart (approved):
👍 LGTM

PR #901, opened 2019-04-19
TableSink: Add deploy-time error if rowkey field is not in output schema

kmeelu:
https://issues.cask.co/browse/CDAP-11732

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
@@ -73,13 +73,15 @@ public class TableSink extends BatchWritableSink<StructuredRecord, byte[], Put> 
     Schema outputSchema =
       SchemaValidator.validateOutputSchemaAndInputSchemaIfPresent(tableSinkConfig.getSchemaStr(),
                                                                   tableSinkConfig.getRowField(), pipelineConfigurer);
-    if ((outputSchema != null) && (outputSchema.getFields().size() == 1)) {
-      String fieldName = outputSchema.getFields().get(0).getName();
-      if (fieldName.equals(tableSinkConfig.getRowField())) {
-        throw new IllegalArgumentException(
-          String.format("Output schema should have columns other than rowkey."));
+    if (outputSchema != null) {
+      if (outputSchema.getField(tableSinkConfig.getRowField()) == null) {
+        throw new IllegalArgumentException("Output schema should contain the rowkey column.");
+      }
+      if (outputSchema.getFields().size() == 1) {
+        throw new IllegalArgumentException("Output schema should have columns other than rowkey.");
       }
     }
+
     // NOTE: this is done only for testing, once CDAP-4575 is implemented, we can use this schema in initialize
     pipelineConfigurer.getStageConfigurer().setOutputSchema(outputSchema);
   }
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/sink/TableSinkTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/sink/TableSinkTest.java
@@ -88,10 +88,9 @@ public class TableSinkTest extends ETLBatchTestBase {
 
   @Test
   public void testTableSinkWithComplexTypeSkipped() {
-    // This doesn't include the rowkey in the output record. The row key inclusion in output record
-    // is optional.
     Schema outputSchema = Schema.recordOf(
       "purchase",
+      Schema.Field.of("rowkey", Schema.of(Schema.Type.STRING)),
       Schema.Field.of("user", Schema.of(Schema.Type.STRING))
     );
 
@@ -168,6 +167,7 @@ public class TableSinkTest extends ETLBatchTestBase {
   public void testTableSink() {
     Schema outputSchema = Schema.recordOf(
       "purchase",
+      Schema.Field.of("rowkey", Schema.of(Schema.Type.STRING)),
       Schema.Field.of("user", Schema.of(Schema.Type.STRING)),
       Schema.Field.of("count", Schema.of(Schema.Type.INT))
     );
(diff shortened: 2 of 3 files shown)


kmeelu:
Added Unit Test. Ran locally to confirm error being shown (https://photos.app.goo.gl/4Eccz59PsBmaUNpNA) when output schema doesn't include rowkey (https://photos.app.goo.gl/xL9LkAaASZsbBbq59).

anew:
I think the change in the wrangler submodule is unrelated... can you remove that?

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
```diff
@@ -73,13 +73,19 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     Schema outputSchema =
       SchemaValidator.validateOutputSchemaAndInputSchemaIfPresent(tableSinkConfig.getSchemaStr(),
                                                                   tableSinkConfig.getRowField(), pipelineConfigurer);
-    if ((outputSchema != null) && (outputSchema.getFields().size() == 1)) {
-      String fieldName = outputSchema.getFields().get(0).getName();
-      if (fieldName.equals(tableSinkConfig.getRowField())) {
+    if (outputSchema != null) {
+      if (outputSchema.getFields().size() == 1) {
```
anew, line 80: hmmm this code looks suspicious. I think there are three different error scenarios:
a) schema has one field and that field is the row field
b) schema has one field and that field is different from the row field
c) schema has more fields but does not contain the row field

It looks as if case b) is not covered?

anew (changes requested):

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
kmeelu: Moved fieldName assignment out of conditional and merged logic from nested conditional (lmk if it makes more sense to just remove fieldName assignment and use outputSchema.getFields().get(0).getName().equals(tableSinkConfig.getRowField()).

Also, updated the tests. PTAL.

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
```diff
@@ -73,13 +73,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
       SchemaValidator.validateOutputSchemaAndInputSchemaIfPresent(tableSinkConfig.getSchemaStr(),
                                                                   tableSinkConfig.getRowField(), pipelineConfigurer);
-    if ((outputSchema != null) && (outputSchema.getFields().size() == 1)) {
+    if (outputSchema != null) {
       String fieldName = outputSchema.getFields().get(0).getName();
-      if (fieldName.equals(tableSinkConfig.getRowField())) {
+      if (outputSchema.getFields().size() == 1 && fieldName.equals(tableSinkConfig.getRowField())) {
         throw new IllegalArgumentException(
-          String.format("Output schema should have columns other than rowkey."));
+            String.format("Output schema should have columns other than rowkey."));
```
CuriousVini: do we need `String.format`? Can't it just be `throw new IllegalArgument("Output schema should have columns other than rowkey.")`?

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
```diff
@@ -73,13 +73,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+      } else if (outputSchema.getField(tableSinkConfig.getRowField()) == null) {
+        throw new IllegalArgumentException(
+            String.format("Output schema should contain the rowkey column."));
```
CuriousVini: same here.

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
anew: It seems easier to read the code if: 
1. first check if the row key is in the schema
2. check if there are more than 1 field

Also, please add tests for each case. Thanks

kmeelu: Yep! Added the extra test, as well.

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
kmeelu: Don't think so. I assumed it was convention, but it seems like it's left out other places. Fixing this.

Thanks!

kmeelu:
Removed the wrangler submodule, as requested.

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
```diff
@@ -73,13 +73,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
                                                                   tableSinkConfig.getRowField(), pipelineConfigurer);
-    if ((outputSchema != null) && (outputSchema.getFields().size() == 1)) {
-      String fieldName = outputSchema.getFields().get(0).getName();
-      if (fieldName.equals(tableSinkConfig.getRowField())) {
-        throw new IllegalArgumentException(
-          String.format("Output schema should have columns other than rowkey."));
+    if (outputSchema != null) {
+      if (outputSchema.getField(tableSinkConfig.getRowField()) == null) {
+        throw new IllegalArgumentException("Output schema should contain the rowkey column.");
+      } else if (outputSchema.getFields().size() == 1) {
```
anew: this should not be in an else branch. And if the size is 1, then the next check is not needed, because you know that the row field is one of the fields, that is, the only field.

anew (changes requested):
See my comment - the code can be simplified. Also, did you find a way to configure the widget.json so that the fields is a dropdown from the fields of the schema?

core-plugins/src/main/java/io/cdap/plugin/batch/sink/TableSink.java
kmeelu: As in multiple if statements? Updated, let me know if this is what you mean.

Is there a Java best practices guide that the repo adheres to?

kmeelu:
> See my comment - the code can be simplified. Also, did you find a way to configure the widget.json so that the fields is a dropdown from the fields of the schema?

Looked at https://docs.cask.co/cdap/4.2.0/en/developer-manual/pipelines/developing-plugins/presentation-plugins.html#plugin-widgets for a "output-field-selector" (similar to "input-field-selector", but for output schema). None exists. Is it possible to create one? Should I reach out to a member of the team about this?

anew:
> > See my comment - the code can be simplified. Also, did you find a way to configure the widget.json so that the fields is a dropdown from the fields of the schema?
> 
> Looked at https://docs.cask.co/cdap/4.2.0/en/developer-manual/pipelines/developing-plugins/presentation-plugins.html#plugin-widgets for a "output-field-selector" (similar to "input-field-selector", but for output schema). None exists. Is it possible to create one? Should I reach out to a member of the team about this?

It does not harm to ask. If it's not there, you can always open an improvement Jira, it is useful to bring it up even if it does not get addressed right away.

Would it make sense to use the drop down for input fields? At least that is a proxy for the output fields.

anew (approved):
But please see my reply to your other comment

kmeelu:
Row field now uses input-field-selector widget (i.e. allows selection from input schema).

Screen Shot 2019-04-23 at 12 09 18 PM

anew (approved):

PR #912, opened 2019-05-22
Move HTTPPollerSource dependency under io.cdap.plugin.spark

rohitsinha54:
https://issues.cask.co/browse/CDAP-15412

These classes are not used by any other plugin and does not serve any purpose for being in commons.


rohitsinha54:
Not really HTTPCallbackAction does uses HTTPConfig.

PR #918, opened 2019-06-21
Updates Database  source to have browse widget associated with connection string

ajainarayanan:


albertshau (approved):
lgtm assuming it doesn't break earlier versions of the UI

ajainarayanan:
@albertshau Good point. Right now it does throw an exception but this will not be visible to the user nor will it be blocking the user flow. It should be fine if the plugin is used in earlier versions of CDAP (6.0). Have filed a [JIRA-CDAP-15592](https://issues.cask.co/browse/CDAP-15592)  anyways to fix this in 6.0 and 6.1

PR #924, opened 2019-08-19
CDAP-15767 add support for reading int96 fields

albertshau:
Added support for reading int96 fields by copying relevant classes
from the parquet library and slightly modifying them to handle
int96 fields. INT96 fields can be read as bytes or as a long.
When read as a long, the value is converted into a unix timestamp
in microseconds, which means the nanosecond component of the
original data is lost.

Fixed a bug with the existing parquet format so that it correctly
uses the user provided schema as the read schema, which does not
have to be the same as the write schema.

### DIFF
--- a/format-parquet/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
+++ b/format-parquet/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
@@ -1,0 +1,584 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.parquet.avro;
+
+import java.lang.reflect.Constructor;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericArray;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.generic.IndexedRecord;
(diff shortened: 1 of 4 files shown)


format-parquet/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -0,0 +1,581 @@
...
+ * This {@link Converter} class materializes records as Avro
+ * {@link IndexedRecord} instances. This is replaced by
+ * {@link AvroRecordConverter}, but is included for backward-compatibility.
+ *
+ * This is a copy of Parquet's class, except with special logic for int96.
+ *
+ * @param <T> a subclass of Avro's IndexedRecord
+ */
+class AvroIndexedRecordConverter<T extends IndexedRecord> extends GroupConverter {
+  // These constants are added
```
albertshau, line 55: these constants are modified

format-parquet/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -0,0 +1,581 @@
...
+    if (avroField == null) {
+      throw new InvalidRecordException(String.format("Parquet/Avro schema mismatch. Avro field '%s' not found.",
+          parquetFieldName));
+    }
+    return avroField;
+  }
+
+  private static Converter newConverter(Schema schema, Type type,
+      GenericData model, ParentValueContainer parent) {
+    // this is the modified section
```
albertshau, line 150: this section is modified

format-parquet/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -0,0 +1,367 @@
...
+          public Schema convertINT32(PrimitiveTypeName primitiveTypeName) {
+            return Schema.create(Schema.Type.INT);
+          }
+          @Override
+          public Schema convertINT64(PrimitiveTypeName primitiveTypeName) {
+            return Schema.create(Schema.Type.LONG);
+          }
+          @Override
+          public Schema convertINT96(PrimitiveTypeName primitiveTypeName) {
+            // this is the modified part
```
albertshau, line 254: this line is modified

albertshau:
most of the code is copied, have commented on the modified sections

format-parquet/src/main/java/org/apache/parquet/avro/AvroIndexedRecordConverter.java
```diff
@@ -0,0 +1,581 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
```
sreevatsanraman: We should retain the copyright from the class.

format-parquet/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -0,0 +1,367 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
```
sreevatsanraman: Retain copyright

sreevatsanraman (approved):
Fix the minor comments and merge.

PR #925, opened 2019-08-22
CDAP-15767 fix failure when reading subset of written parquet fields

albertshau:
Fixed a bug where the file source would fail to read parquet files
with a schema that was a subset of the actual written data. This
bug was introduced in the fix for int96, which fixed a bug where
the user provided schema was being used as the write schema instead
of the read schema.

Fixed by performing a no-op when reading fields that do not exist
in the read schema.


chtyim (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD567

PR #926, opened 2019-08-27
Add a Schemas util class to get display name for schema types

CuriousVini:
sample plugin usage:
https://github.com/data-integrations/google-cloud/pull/142/commits/c49d7795a0821c9397c3c10346ad04bb73018412#diff-32b7b66d98a6948af91713bbda871493R363

### DIFF
--- a/hydrator-common/src/main/java/io/cdap/plugin/common/Schemas.java
+++ b/hydrator-common/src/main/java/io/cdap/plugin/common/Schemas.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


hydrator-common/src/main/java/io/cdap/plugin/common/Schemas.java
```diff
@@ -0,0 +1,65 @@
...
+   * LogicalType.TIME_MILLIS -> "time"
+   * LogicalType.TIME_MACROS -> "time"
+   * LogicalType.TIMESTAMP_MILLIS -> "timestamp"
+   * LogicalType.TIMESTAMP_MACROS -> "timestamp"
+   * LogicalType.DECIMAL -> "decimal"
+   *
+   * @param schema schema to be used to get type name
+   * @return type name associated to the schema
+   */
+  public static String getSchemaTypeName(Schema schema) {
```
chtyim: Why in hydrator-plugins? Shouldn't it goes together with the `Schema` class in `cdap-api-common`?

CuriousVini: This is because in data pipeline studio we map logical types such as timestamp_millis and timestamp_micros as timestamp and these checks mostly will be performed by plugins. So this util method would help provide error messages with correct logical types.

albertshau: I think it makes sense to put it in the Schema class, this is useful in other places, such as the pipeline app.

PR #933, opened 2019-09-11
CDAP-15879 fix RegexPathFilter class loading

albertshau:
Fixed a bug where RegexPathFilter needed to be exported by any
plugins using the AbstractFileSource. It is a shared class that
should not be exported by multiple plugins, otherwise classloading
conflicts could arise.

The class gets instantiated by FileInputFormat when generating
splits. Fixed the format input formats to set the Configuration
classloader to the plugin classloader, which will be able to
create RegexPathFilter without exports since the class is packaged
in the plugin jar.

This also fixes a related bug, where the FileSystem used by
RegexPathFilter would only contain the filesystems deployed
directly on the cluster, and not the connectors packaged in the
plugins. This was causing the GCS plugins to fail
when run on clusters without the GCS connector installed.


albertshau:
https://builds.cask.co/browse/HYP-BAD574

CuriousVini (approved):

PR #936, opened 2019-09-13
CDAP-15794 use a better error message when delimited schema is wrong

albertshau:
Added an explicit error message in the delimited format when the
data has more fields than the schema. This replaces the previous
behavior, which would throw a NoSuchElementException.

Added a special check to see if the user meant to use the 'text'
format, as that seems to be a very common user error when the
file source is paired with wrangler.


albertshau:
https://builds.cask.co/browse/HYP-BAD577

CuriousVini (approved):

PR #940, opened 2019-09-17
[CDAP-15787] Add plugin validation in spark plugins and http plugin

CuriousVini:
build: https://builds.cask.co/browse/HYP-BAD578-4


http-plugins/src/main/java/io/cdap/plugin/batch/HTTPCallbackAction.java
```diff
@@ -148,18 +157,32 @@ public HttpRequestConf() {
     @SuppressWarnings("ConstantConditions")
     public void validate() {
```
albertshau: this shouldn't exist anymore right?

albertshau (approved):

http-plugins/src/main/java/io/cdap/plugin/batch/HTTPCallbackAction.java
CuriousVini: Yes, we can remove the overridden method. Done

PR #945, opened 2019-09-18
CDAP-15428 fixed MD5 hasher plugin to check for field type

albertshau:
Fixed the md5 hasher plugin to validate that the fields specified
are strings and to fail if they are not.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/Hasher.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/Hasher.java
@@ -64,7 +64,9 @@ public final class Hasher extends Transform<StructuredRecord, StructuredRecord> 
 
   @Override
   public void prepareRun(StageSubmitterContext context) throws Exception {
-    config.validate(context.getInputSchema(), context.getFailureCollector());
+    FailureCollector failureCollector = context.getFailureCollector();
+    config.validate(context.getInputSchema(), failureCollector);
+    failureCollector.getOrThrowException();
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


transform-plugins/src/main/java/io/cdap/plugin/Hasher.java
```diff
@@ -124,14 +131,37 @@ public Config(String hash, String fields) {
...
+                                    "Allowed hashers are md2, md5, sha1, sha256, sha384 and sha512");
       }
+
+      if (inputSchema == null) {
+        return;
+      }
+
+      for (String field : getFields()) {
+        Schema.Field inputField = inputSchema.getField(field);
+        if (inputField == null) {
```
CuriousVini, line 151: Isn't this error case where field to be hashed is not present in input?

transform-plugins/src/main/java/io/cdap/plugin/Hasher.java
```diff
@@ -44,28 +49,27 @@
...
+  public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
+    StageConfigurer stageConfigurer = pipelineConfigurer.getStageConfigurer();
+    config.validate(stageConfigurer.getInputSchema(), stageConfigurer.getFailureCollector());
+    stageConfigurer.getFailureCollector().getOrThrowException();
+    stageConfigurer.setOutputSchema(stageConfigurer.getInputSchema());
+  }
+
+  @Override
+  public void prepareRun(StageSubmitterContext context) throws Exception {
+    config.validate(context.getInputSchema(), context.getFailureCollector());
```
CuriousVini: should throw validation exception using `collector.getOrThrowException()` in prepareRun.

CuriousVini (commented):
just 2 comments. rest lgtm

transform-plugins/src/main/java/io/cdap/plugin/Hasher.java
albertshau: The current behavior is that it gets ignored, so I didn't want to change it.

CuriousVini (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD588

PR #951, opened 2019-09-19
[CDAP-15787] Updated ErrorCollector for new validation API

dli357:
Updated ErrorCollector.java for the new validation API.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
@@ -25,9 +25,8 @@ import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
 import io.cdap.cdap.etl.api.ErrorRecord;
 import io.cdap.cdap.etl.api.ErrorTransform;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
-import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
-
 import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
```diff
@@ -40,28 +39,37 @@
 public class ErrorCollector extends ErrorTransform<StructuredRecord, StructuredRecord> {
   private final Config config;
 
+  private static final String MESSAGE_FIELD = "messageField";
```
albertshau: put these in the config class

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
```diff
@@ -40,28 +39,37 @@
...
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
     if (inputSchema != null) {
       if (config.messageField != null && inputSchema.getField(config.messageField) != null) {
-        throw new InvalidConfigPropertyException(String.format(
-          "Input schema already contains message field '%s'. Please set message field to a different value.",
-          config.messageField), "messageField");
+        collector.addFailure(
+            String.format("Input schema already contains message field '%s'.", config.messageField),
+            "Please set message field to a different value.")
```
albertshau: as we're going through these messages, we've stopped putting 'Please', as it's weird to have it in some places but not others, and sometimes it doesn't make sense.

Set the message field to a different value

albertshau: similar comments below

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
```diff
@@ -40,28 +39,37 @@
...
+    // Get failure collector for updated validation API
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
     Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
     if (inputSchema != null) {
       if (config.messageField != null && inputSchema.getField(config.messageField) != null) {
-        throw new InvalidConfigPropertyException(String.format(
-          "Input schema already contains message field '%s'. Please set message field to a different value.",
-          config.messageField), "messageField");
+        collector.addFailure(
+            String.format("Input schema already contains message field '%s'.", config.messageField),
```
albertshau: already contains field '%s'.

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
dli357: Shouldn't we consider leaving in "Input schema" to be more clear about what already contains that field?

albertshau: yeah, was lazy and didn't type the whole thing:

Input schema already contains field '%s'.

albertshau (approved):
lgtm

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
dli357: Fixed

PR #952, opened 2019-09-19
[CDAP-15787] Updated ExcelInputReader for New Validation API

dli357:
Updated ExcelInputReader.java for the new validation API. Was unable to run the unit tests due to an InvalidArtifactException.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
@@ -29,12 +29,14 @@ import io.cdap.cdap.api.common.Bytes;
 import io.cdap.cdap.api.data.batch.Input;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.dataset.DatasetManagementException;
 import io.cdap.cdap.api.dataset.DatasetProperties;
 import io.cdap.cdap.api.dataset.lib.CloseableIterator;
 import io.cdap.cdap.api.dataset.lib.KeyValue;
 import io.cdap.cdap.api.dataset.lib.KeyValueTable;
 import io.cdap.cdap.api.dataset.table.Table;
 import io.cdap.cdap.etl.api.Emitter;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.batch.BatchRuntimeContext;
 import io.cdap.cdap.etl.api.batch.BatchSource;
@@ -75,6 +77,14 @@ public class ExcelInputReader extends BatchSource<LongWritable, Object, Structur
   private static final String KEY = "key";
   private static final String FILE = "file";
   private static final String SHEET = "sheet";
+  private static final String SHEET_VALUE = "sheetValue";
+  private static final String TABLE_EXPIRY_PERIOD = "tableExpiryPeriod";
+  private static final String ROWS_LIMIT = "rowsLimit";
+  private static final String COLUMN_LIST = "columnList";
+  private static final String OUTPUT_SCHEMA = "outputSchema";
+  private static final String COLUMN_MAPPING = "columnMapping";
+  private static final String IF_ERROR_RECORD = "ifErrorRecord";
+  private static final String ERROR_DATASET_NAME = "errorDatasetName";
   private static final String RECORD = "record";
   private static final String EXIT_ON_ERROR = "Exit on error";
   private static final String WRITE_ERROR_DATASET = "Write to error dataset";
@@ -112,13 +122,15 @@ public class ExcelInputReader extends BatchSource<LongWritable, Object, Structur
   @Override
   public void initialize(BatchRuntimeContext context) throws Exception {
     batchRuntimeContext = context;
-    init();
+    FailureCollector collector = context.getFailureCollector();
+    init(collector);
+    collector.getOrThrowException();
   }
 
   /**
    * Initialize and set maps from input config object
    */
-  private void init() {
+  private void init(FailureCollector collector) {
     if (!Strings.isNullOrEmpty(excelInputreaderConfig.columnList)) {
       String[] columnsList = excelInputreaderConfig.columnList.split(",");
       inputColumns = Arrays.asList(columnsList);
@@ -128,13 +140,18 @@ public class ExcelInputReader extends BatchSource<LongWritable, Object, Structur
       String[] mappings = excelInputreaderConfig.columnMapping.split(",");
       for (String map : mappings) {
         String[] columns = map.split(":");
-        if (CollectionUtils.isNotEmpty(inputColumns) && !inputColumns.contains(columns[0])) {
-          throw new IllegalArgumentException("Column name: " + columns[0] + " in 'Column-Label Mapping' does not " +
-                                               "match the columns in the 'Column To Be Extracted' input text box. " +
-                                               "It has to be one of the columns in 'Column To Be Extracted' " +
-                                               "input text box.");
+        if (columns.length < 2 || Strings.isNullOrEmpty(columns[1])) {
+          collector.addFailure(
+            String.format("Column: '%s' in 'Column-Label Mapping' must have an alias specified.", columns[0]), null)
+            .withConfigElement(COLUMN_MAPPING, map);
+        } else if (CollectionUtils.isNotEmpty(inputColumns) && !inputColumns.contains(columns[0])) {
+          collector.addFailure(
+            String.format("Column: '%s' in 'Column-Label Mapping' " +
+                            "must be included in 'Column To Be Extracted'", columns[0]), null)
+            .withConfigElement(COLUMN_MAPPING, map);
+        } else {
+          columnMapping.put(columns[0], columns[1]);
         }
-        columnMapping.put(columns[0], columns[1]);
       }
     }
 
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -129,12 +138,13 @@ private void init() {
       for (String map : mappings) {
         String[] columns = map.split(":");
         if (CollectionUtils.isNotEmpty(inputColumns) && !inputColumns.contains(columns[0])) {
-          throw new IllegalArgumentException("Column name: " + columns[0] + " in 'Column-Label Mapping' does not " +
-                                               "match the columns in the 'Column To Be Extracted' input text box. " +
-                                               "It has to be one of the columns in 'Column To Be Extracted' " +
-                                               "input text box.");
+          collector.addFailure(
+              String.format("Column: %s in 'Column-Label Mapping' "
```
albertshau: indentation

albertshau: single quotes around '%s'

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -143,14 +153,16 @@ private void init() {
       for (String schema : schemaList) {
         String[] columns = schema.split(":");
         if (CollectionUtils.isNotEmpty(inputColumns) && !inputColumns.contains(columns[0])) {
-          throw new IllegalArgumentException("Column name: " + columns[0] + " in 'Field Name Schema Type Mapping'" +
-                                               " does not match the columns in the 'Column To Be Extracted' input " +
-                                               "text box. It has to be one of the columns in " +
-                                               "'Column To Be Extracted' input text box.");
+          collector.addFailure(
+              String.format("Column: %s in 'Field Name Schema Type Mapping' "
```
albertshau: single quotes around '%s'

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -310,6 +325,15 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
   }
 
   private void createDatasets(@Nullable PipelineConfigurer pipelineConfigurer, @Nullable BatchSourceContext context) {
```
albertshau: FYI, this is a bad pattern. A method shouldn't take two arguments, one of which it expects to always be null and the other not.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -310,6 +325,15 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
...
+    FailureCollector collector;
+    if (pipelineConfigurer != null) {
+      collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    } else if (context != null) {
+      collector = context.getFailureCollector();
+    } else {
+      throw new IllegalArgumentException("Unable to retrieve failure collector.");
+    }
     try {
       if (!excelInputreaderConfig.containsMacro("errorDatasetName") &&
```
albertshau, line 362: woah just read how this plugin is handling errors and it is completely wrong. Can you modify the widget .json file for this plugin to set the widget type for errorDatasetName and ifErrorRecord to 'hidden' so nobody tries to use them?

In fact, I think the existence of this entire source is questionable, so I've opened CDAP-15937 to revisit this next release.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -143,14 +153,16 @@ private void init() {
...
+              String.format("Column: %s in 'Field Name Schema Type Mapping' "
+                  + "must be included in 'Column To Be Extracted'", columns[0]), null)
+              .withConfigElement(OUTPUT_SCHEMA, schema);
+        } else {
+          outputSchemaMapping.put(columns[0], columns[1]);
         }
-        outputSchemaMapping.put(columns[0], columns[1]);
       }
     }
+    collector.getOrThrowException();
```
albertshau: whether to call this or not should be decided by the caller and not placed here at the end of the method

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -489,20 +515,25 @@ public ExcelInputReaderConfig() {
...
+            .withConfigProperty(TABLE_EXPIRY_PERIOD);
       }
 
       if (!Strings.isNullOrEmpty(rowsLimit) && !StringUtils.isNumeric(rowsLimit)) {
-        throw new IllegalArgumentException(String.format("Invalid row limit: %s. " +
-                                                           "Numeric value expected.", rowsLimit));
+        collector.addFailure(String.format("Invalid row limit: '%s'.", rowsLimit), "Numeric value expected.")
+            .withConfigProperty(ROWS_LIMIT);
       }
+      collector.getOrThrowException();
```
albertshau: should not be here, the caller should decide whether to call it. In the case of configurePipeline(), it looks like more things can be validated.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -266,23 +278,26 @@ public void transform(KeyValue<LongWritable, Object> input, Emitter<StructuredRe
...
-    init();
+    init(collector);
     getOutputSchema();
     pipelineConfigurer.getStageConfigurer().setOutputSchema(outputSchema);
   }
 
   @Override
   public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
-    excelInputreaderConfig.validate();
+    excelInputreaderConfig.validate(batchSourceContext.getFailureCollector());
```
albertshau: call getOrThrowException after this method call

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -266,23 +278,26 @@ public void transform(KeyValue<LongWritable, Object> input, Emitter<StructuredRe
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
 
-    excelInputreaderConfig.validate();
+    // Get failure collector for updated validation API
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    excelInputreaderConfig.validate(collector);
 
     if (Strings.isNullOrEmpty(excelInputreaderConfig.columnList) &&
```
albertshau: this looks like it belongs in the validate() method

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Yeah I built off of the existing method which already did this, I can go ahead and create overloaded methods instead which are probably better.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Sounds good, changed to hidden for now

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
albertshau: its ok, you don't have to refactor this, was just mentioning it.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -309,7 +333,37 @@ public void prepareRun(BatchSourceContext batchSourceContext) throws Exception {
...
+        collector.addFailure("Error dataset name should not be empty if choosing write to error "
+                               + "dataset for 'On Error' input.", null)
+          .withConfigProperty(IF_ERROR_RECORD).withConfigProperty(ERROR_DATASET_NAME);
+      }
+
+      if (!excelInputreaderConfig.containsMacro("memoryTableName") &&
+        !Strings.isNullOrEmpty(excelInputreaderConfig.memoryTableName)) {
+        pipelineConfigurer.createDataset(excelInputreaderConfig.memoryTableName, KeyValueTable.class);
+      }
+    } catch (Exception e) {
```
albertshau: don't think anything throws an exception anymore. If not, you should remove this

albertshau (approved):
minor comment otherwise lgtm. fix, squash, then merge

core-plugins/widgets/Excel-batchsource.json
```diff
@@ -136,7 +136,7 @@
       "label": "Error Handling",
       "properties": [
         {
-          "widget-type": "select",
+          "widget-type": "hidden",
```
albertshau, line 139: forgot to mention, please remove mention of these from the documentation if they exist

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
dli357: Fixed

core-plugins/widgets/Excel-batchsource.json
dli357: Fixed

PR #954, opened 2019-09-20
[CDAP-15787] Updated ValueMapper for new Validation API

dli357:
Updated ValueMapper.java and ValueMapperTest.java, added additional validation to ensure that the mapped columns exist in the input and that the default mappings exist in the mapped columns. Added additional unit tests to test for these cases and fixed static variable storage problem which was causing a race condition when running unit tests.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/ValueMapper.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/ValueMapper.java
@@ -19,16 +19,18 @@ package io.cdap.plugin;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
+import io.cdap.cdap.api.annotation.Requirements;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.dataset.table.Table;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.Lookup;
 import io.cdap.cdap.etl.api.LookupTableConfig;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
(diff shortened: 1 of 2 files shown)


transform-plugins/src/main/java/io/cdap/plugin/ValueMapper.java
```diff
@@ -77,55 +82,56 @@ public Config(String mapping, String defaults) {
...
+   * It is required to use in configurePiperline as well as transform method (run in initialize).
+   */
+  private void parseConfiguration(Config config, FailureCollector collector) {
+    if (!config.defaults.isEmpty()) {
+      String[] defaultsList = config.defaults.split(",");
+      for (String defaultValue : defaultsList) {
+        String[] defaultsArray = defaultValue.split(":");
+        if (defaultsArray.length != 2) {
+          collector.addFailure(
+              String.format("Invalid default: %s.", defaultValue),
```
albertshau: indentation, here and rest of PR

transform-plugins/src/test/java/io/cdap/plugin/ValueMapperTest.java
```diff
@@ -367,6 +385,83 @@ public void testMappingValidation() throws Exception {
...
+
+    MockPipelineConfigurer configurer = new MockPipelineConfigurer(inputSchema);
+    ValueMapper mapper = new ValueMapper(config);
+    String caughtException = "";
+    try {
+      mapper.configurePipeline(configurer);
+    } catch (Exception e) {
+      caughtException = e.getMessage();
+    }
+    Assert.assertEquals(validationExceptionMessage, caughtException);
```
albertshau: shouldn't check exception message

transform-plugins/src/main/java/io/cdap/plugin/ValueMapper.java
dli357: Fixed

transform-plugins/src/test/java/io/cdap/plugin/ValueMapperTest.java
dli357: Fixed

dli357:
Overall, the validation logic in this plugin is rather complicated, but for now, due to the volume of plugins that need updating, we will simply patch it and revisit it later.

albertshau (approved):

PR #955, opened 2019-09-20
[CDAP-15787] Updated XMLParser for new Validation API

dli357:
Updated XMLParser.java and XMLParserTest.java for new validation API. Added a validation check and unit test to check for input field as well.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
@@ -25,10 +25,12 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.InvalidEntry;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.Transform;
 import io.cdap.cdap.etl.api.TransformContext;
+import io.cdap.cdap.etl.api.validation.ValidationFailure;
 import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
@@ -63,6 +65,7 @@ import javax.xml.xpath.XPathFactory;
 @Name("XMLParser")
 @Description("Parse XML events based on XPath")
 public class XMLParser extends Transform<StructuredRecord, StructuredRecord> {
+  private static final String XPATH_MAPPINGS = "xPathMappings";
   private static final String EXIT_ON_ERROR = "Exit on error";
   private static final String WRITE_ERROR_DATASET = "Write to error dataset";
   private final Config config;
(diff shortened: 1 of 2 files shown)


transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -77,66 +82,79 @@ public XMLParser(Config config) {
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
     super.configurePipeline(pipelineConfigurer);
-    outSchema = config.getOutputSchema();
+    // Get failure collector for updated validation API
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    outSchema = config.getOutputSchema(collector);
+    if (!this.config.containsMacro(INPUT)
+        && pipelineConfigurer.getStageConfigurer().getInputSchema().getField(this.config.inputField) == null) {
```
albertshau: this is difficult to read, it's better to have a separate variable for the input schema.

Also, the input schema can be null. This check should be skipped if it's null.

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -77,66 +82,79 @@ public XMLParser(Config config) {
...
+      collector.addFailure(String.format("Input field '%s' must exist in input schema.", this.config.inputField), null)
+          .withConfigProperty(INPUT);
+    }
     try {
-      validateXpathAndSchema();
+      validateXpathAndSchema(collector);
     } catch (UnsupportedEncodingException e) {
```
albertshau: whatever can throw this exception should be handled within the validate method and not out here

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -77,66 +82,79 @@ public XMLParser(Config config) {
...
 
   @Override
   public void initialize(TransformContext context) throws Exception {
     super.initialize(context);
-    outSchema = config.getOutputSchema();
-    xPathMapping = getXPathMapping();
+    // Get failure collector for updated validation API
+    FailureCollector collector = getContext().getFailureCollector();
+    outSchema = config.getOutputSchema(collector);
+    xPathMapping = getXPathMapping(collector);
```
albertshau, line 101: call collector.getOrThrowException after this to fail the pipeline run here instead of continuing on.

This should actually be validated in prepareRun(), which is one once at the start of the run. initialize() is called by each mapper at the start of their lifecycle, so if there are 100 mappers created for the run, you'll see 100 error messages in the log, all saying the same thing

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -77,66 +82,79 @@ public XMLParser(Config config) {
...
     List<Schema.Field> outFields = outSchema.getFields();
     // Checks if all the fields in the XPath mapping are present in the output schema.
     // If they are not a list of fields that are not present is included in the error message.
-    StringBuilder notOutput = new StringBuilder();
     for (Schema.Field field : outFields) {
       String fieldName = field.getName();
       if (!xPathMapping.keySet().contains(field.getName())) {
-        notOutput.append(fieldName + ";");
+        ValidationFailure failure = collector.addFailure(
+            String.format("Field '%s' must be present in XPath Mapping.", fieldName), null);
```
albertshau: indentation

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -291,26 +309,32 @@ public Config(String inputField, String encoding, String xPathFieldMapping, Stri
...
-        Schema.Field field = Schema.Field.of(fieldName, Schema.nullableOf(Schema.of(Schema.Type.valueOf(
-          params[1].trim().toUpperCase()))));
-        if (fields.contains(field)) {
-          throw new IllegalArgumentException(String.format("Field %s already has type specified. Duplicate field %s",
-                                                           fieldName, fieldName));
+          collector.addFailure(String.format("A type must be provided for field '%s'.", fieldName), null)
+              .withConfigElement(FIELD_TYPE_MAPPING, mapping);
         } else {
-          fields.add(field);
+          Schema.Field field = Schema.Field.of(fieldName, Schema.nullableOf(Schema.of(Schema.Type.valueOf(
```
albertshau: When there's so many function calls within function calls, it gets hard to read. Break this up into 2 statements.

transform-plugins/src/test/java/io/cdap/plugin/XMLParserTest.java
```diff
@@ -45,12 +52,22 @@ public void testInvalidConfig() throws Exception {
...
+    String caughtMessage = "";
     try {
       new XMLParser(config).configurePipeline(configurer);
       Assert.fail();
-    } catch (IllegalArgumentException e) {
-      Assert.assertEquals("Type cannot be null. Please specify type for category", e.getMessage());
+    } catch (Exception e) {
+      caughtMessage = e.getMessage();
     }
+    Assert.assertEquals(validationExceptionMessage, caughtMessage);
```
albertshau: shouldn't check exception message

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed, and moved the runtime validation to prepareRun instead of initialize

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed

transform-plugins/src/test/java/io/cdap/plugin/XMLParserTest.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Currently, the unit tests rely on initialize to create a mock transform context. For now I'll leave the current initialization logic as it is, but we should probably revisit it at some point.

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -77,65 +80,79 @@ public XMLParser(Config config) {
...
-      }
-
-      if (xpathmap[0] == null || xpathmap[0].trim().isEmpty()) {
-        throw new IllegalArgumentException("XPath mapping is missing a field name : '" + xpath + "'");
-      }
-
-      if (xpathmap[1] == null || xpathmap[1].trim().isEmpty()) {
-        throw new IllegalArgumentException("XPath mapping is missing xpath : '" + xpath + "'");
+        collector.addFailure(String.format("Invalid XPath mapping: '%s'.", xpath), null)
+            .withConfigElement(XPATH_MAPPINGS, xpath);
```
albertshau: nit: indentation

albertshau (approved):
lgtm

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: Fixed, also I switched to the proper CDAP IntelliJ config so hopefully these shouldn't happen anymore

PR #958, opened 2019-09-20
[CDAP-15787] Updated XMLToJSON for new Validation API

dli357:
Added two unit tests as well for the new validation checks.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
@@ -24,6 +24,7 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.Transform;
 import io.cdap.cdap.etl.api.TransformContext;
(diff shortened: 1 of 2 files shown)


transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
```diff
@@ -80,14 +83,16 @@ public void initialize(TransformContext context) throws Exception {
...
 
     Schema inputSchema = pipelineConfigurer.getStageConfigurer().getInputSchema();
     if (inputSchema != null) {
       if (inputSchema.getField(config.inputField) == null) {
-        throw new IllegalArgumentException(String.format("Field %s is not present in input schema", config.inputField));
-      }
-      if (inputSchema.getField(config.inputField).getSchema().getType() != Schema.Type.STRING) {
-        throw new IllegalArgumentException(String.format("Field %s must be of type string.", config.inputField));
+        collector.addFailure(String.format("Field '%s' must be present in input schema.", config.inputField), null)
+            .withConfigProperty(INPUT_FIELD);
```
albertshau: indentation (and rest of PR)

transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
```diff
@@ -80,14 +83,16 @@ public void initialize(TransformContext context) throws Exception {
...
+      } else if (inputSchema.getField(config.inputField).getSchema().getType() != Schema.Type.STRING) {
```
albertshau: this is overly verbose, it's easier to read if it's like

```
Schema.Field field = inputSchema.getField(config.inputField);
if (field == null) {
  ...
} else if (...
```

transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/XMLToJSON.java
dli357: Fixed

albertshau (approved):

PR #959, opened 2019-09-24
[CDAP-15787] Updated NullFieldSplitter for new Validation API

dli357:
Made an overload method for getNonNullSchema since it seems that SplitterTransform does not have a way to retrieve context.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
@@ -24,6 +24,7 @@ import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.plugin.PluginConfig;
+import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.MultiOutputEmitter;
 import io.cdap.cdap.etl.api.MultiOutputPipelineConfigurer;
 import io.cdap.cdap.etl.api.MultiOutputStageConfigurer;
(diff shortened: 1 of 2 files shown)


transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
```diff
@@ -97,10 +102,11 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
     }
   }
 
-  private static Map<String, Schema> getOutputSchemas(Schema inputSchema, Conf conf) {
+  private static Map<String, Schema> getOutputSchemas(Schema inputSchema, Conf conf, FailureCollector collector) {
     Map<String, Schema> outputs = new HashMap<>();
     if (inputSchema.getField(conf.field) == null) {
-      throw new IllegalArgumentException("Field " + conf.field + " does not exist in input schema.");
+      collector.addFailure(String.format("Field '%s' must exist in input schema.", conf.field), null)
```
albertshau: in input schema -> in the input schema

transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
```diff
@@ -139,10 +145,43 @@ static Schema getNonNullSchema(Schema nullableSchema, String fieldName) {
     return Schema.recordOf(nullableSchema.getRecordName() + ".nonnull", fields);
   }
 
+  @VisibleForTesting
+  static Schema getNonNullSchema(Schema nullableSchema, String fieldName, FailureCollector collector) {
```
albertshau: shouldn't this replace the other method?

transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/NullFieldSplitter.java
dli357: The reason I made an overload is that the SplitterTransform does not contain a getContext() method necessary to get a failure collector, and getNonNullSchema is also used in the transform method, so there's no easy way to pass a failure collector in that context.

albertshau: I see. Since these methods are basically the same, let's make the FailureCollector nullable and pass in a null instead of calling the other one.

dli357: Sounds good, fixed

albertshau (approved):

PR #960, opened 2019-09-24
[CDAP-15787] Updated UnionSplitter for new Validation API

dli357:


transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -174,42 +177,45 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
-
-    int numFields = inputSchema.getFields().size();
-    ArrayList<Schema.Field> outputFields = new ArrayList<>(numFields);
-    int i = 0;
-    int unionIndex = -1;
-    for (Schema.Field inputField : inputSchema.getFields()) {
-      if (inputField.getName().equals(unionField)) {
-        unionIndex = i;
-        outputFields.add(null);
+      collector.addFailure(String.format("Field '%s' must exist in input schema.", unionField), null)
```
albertshau: in input schema -> in the input schema

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -174,42 +177,45 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
+        .withConfigProperty(Conf.UNION_FIELD);
+    } else {
```
albertshau: if the union field is null, it doesn't make sense to go further as the output schemas returned would be invalid. This should be:

```
if (unionSchemaField == null) {
  collector.addFailure(...);
  throw collector.getOrThrowException();
}

Schema unionSchema = ...
```

this also removes a layer of nesting, making it easier to follow

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -174,42 +177,45 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
+      Schema unionSchema = unionSchemaField.getSchema();
+      if (unionSchema.getType() != Schema.Type.UNION) {
+        collector.addFailure(String.format("Field '%s' must be of type union, but is of type '%s'.",
+                                           unionField, unionSchema.getType()), null)
+          .withConfigProperty(Conf.UNION_FIELD);
```
albertshau: similar situation here, if it's not a union there's no point in going forward.

```
if (unionSchema.getType() != Schema.Type.UNION) {
  collector...
  throw collector.getOrThrowException();
}

...
```

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -174,42 +177,45 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
+        }
+
+        for (Schema schema : unionSchema.getUnionSchemas()) {
+          Schema.Type type = schema.getType();
+          switch (type) {
+            case ENUM:
+            case MAP:
+            case ARRAY:
+            case UNION:
+              collector.addFailure(String.format("Unsupported type '%s' within a union.", type), null)
```
albertshau: I'm not sure these should be separate failures for each type. It seems more fitting to use a single failure, with the message saying which types are invalid.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Added a corrective action to indicate what the unsupported types are. We can also remove the specific typing message if that's desired.

albertshau: what I meant here is that since this is in a for loop, a field with multiple invalid types in its union will result in multiple failures. This should instead be a single failure, with the message mentioning which types are invalid.

dli357: Gotcha, fixed this by using if statement coupled with a break statement upon discovering a failure.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -204,14 +204,12 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
-        case ENUM:
-        case MAP:
-        case ARRAY:
-        case UNION:
-          collector.addFailure(String.format("Unsupported type '%s' within a union.", type),
-                               "The following types are unsupported: enum, map, array, and union.")
-            .withConfigProperty(Conf.UNION_FIELD);
+      if (type.equals(Schema.Type.ENUM) || type.equals(Schema.Type.MAP)
+        || type.equals(Schema.Type.ARRAY) || type.equals(Schema.Type.UNION)) {
+        collector.addFailure(String.format("Unsupported type '%s' within a union.", type),
```
albertshau: Also include the field name

within union field '%s'.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -204,14 +204,12 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
+                             "The following types are unsupported: enum, map, array, and union.")
```
albertshau: unsupported -> not supported

albertshau (approved):
minor comments, otherwise lgtm. Fix, squash, and merge

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Fixed

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
dli357: Changed second "unsupported" to "not supported"

PR #962, opened 2019-09-24
[CDAP-15787] Updated Encoder for new Validation API

dli357:


transform-plugins/src/main/java/io/cdap/plugin/Encoder.java
```diff
@@ -68,45 +69,50 @@ public Encoder(Config config) {
...
+        }
       }
     }
   }
 
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws IllegalArgumentException {
     super.configurePipeline(pipelineConfigurer);
-    parseConfiguration(config.encode);
+    // Get failure collector for updated validation API
```
albertshau: don't need this comment

albertshau (approved):

transform-plugins/src/main/java/io/cdap/plugin/Encoder.java
dli357: Fixed

PR #965, opened 2019-09-24
[CDAP-15787] Updated Normalizer for new Validation API

dli357:
Would recommend changing the names of the fields in the Normalizer widget because it is not intuitive that the second and third components of the Normalizing field are type and value.

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/Normalize.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/Normalize.java
@@ -16,7 +16,6 @@
 
 package io.cdap.plugin;
 
-import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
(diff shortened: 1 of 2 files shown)


transform-plugins/src/main/java/io/cdap/plugin/Normalize.java
```diff
@@ -81,48 +89,85 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) throws Ille
...
-      Preconditions.checkArgument(mappings.length == 2, "Mapping field '" + mappings[0] + "' is invalid. Both input" +
-        " and output schema fields required.");
-      //Input schema cannot be null, check added for JUnit test case run.
-      if (inputSchema != null) {
-        Preconditions.checkArgument(inputSchema.getField(mappings[0]) != null, "Mapping field '" + mappings[0]
-          + "' not present in input schema.");
+      if (mappings.length != 2) {
+        collector.addFailure("Field name and mapping must be provided.", null)
+          .withConfigElement(NormalizeConfig.FIELD_MAPPING, fieldMapping);
+      } else {
```
albertshau: Avoid adding more layers of nesting when possible
```
if (mappings.length != 2) {
  ...
  continue;
}

...
```

dli357: Fixed, and in other places as well

albertshau (approved):

PR #973, opened 2019-09-25
throw IOException in AbstractFileSInk

yaojiefeng:


format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -70,7 +71,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   }
 
   @Override
-  public void prepareRun(BatchSinkContext context) throws InstantiationException {
+  public void prepareRun(BatchSinkContext context) throws InstantiationException, IOException {
```
albertshau: just change this to Exception, in case there are other sinks in the future that throw other types of exceptions

albertshau (approved):

PR #982, opened 2019-10-14
[CDAP-16039] Fix macros to use property names instead of actual values

CuriousVini:
build: https://builds.cask.co/browse/HYP-BAD623-1


albertshau (approved):

CuriousVini:
build is green. Merging.

PR #991, opened 2019-11-08
support fll for error collector and bump version

yaojiefeng:


core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
```diff
@@ -43,6 +48,40 @@ public ErrorCollector(Config config) {
...
+    // the originial fields will not change in transform
+    inputSchema.getFields().forEach(
+      field -> fieldOperations.add(new FieldTransformOperation(String.format("Error collector for field %s",
+                                                                             field.getName()),
+                                                              "Error collector identity transform",
+                                                               Collections.singletonList(field.getName()),
+                                                               field.getName())));
+
+    // emit records for generated fields
+    if (config.messageField != null) {
```
CuriousVini: What if input schema does not have this field? I think we should also check `inputSchema.getField(config.messageField) != null`
Also we should add empty string check

CuriousVini (commented):
just one comment.
Assuming we would be updating wrangler reference in next PR.

core-plugins/src/main/java/io/cdap/plugin/error/ErrorCollector.java
yaojiefeng: The input schema is guaranteed to not have this field, we did the validation check in `configurePipeline()`, this field is a generated field in the plugin, that's why the fll operation for this field does not contain the input fields.

CuriousVini (approved):

PR #995, opened 2019-11-13
Refactoring of the Joiner implementation

chtyim:
- Generate the join key schema once instead of per record
-- This should boost the transform performance 
- Don’t assign fields in get methods. Only assign in init/initialize methods
- Fix bunch of IDE warnings
-- Potential NPE, upper casing constants, access constants via class instead of instance
- Fix the join key schemas validation logic
-- Should stick to one schema to compare, otherwise a wrong config in one stage can be emitting incorrect error messages for later input stage
-- Use LinkedHashMap for stable input stages ordering


CuriousVini (approved):
lgtm 👍

PR #996, opened 2019-11-21
add regex file pattern

zhaotianai:
Before the getFilePattern() function is returning null, I added in a regex expression to do some filtering on the file path based on the path specified in the original code: 'prefix://username:password@hostname:port/path'


albertshau:
closed due to age

PR #1005, opened 2019-12-17
Fix SFTP plugin to return regex pattern matcher

theodoresiu:
Currently fileRegex returns a null and SFTP plugin does not successfully parse files based on regex. Fixing the getFilePattern method to return the Pattern

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
@@ -141,7 +141,7 @@ public class FTPBatchSource extends AbstractFileSource {
     @Nullable
     @Override
     public Pattern getFilePattern() {
-      return null;
+      return Pattern.compile(fileRegex);
     }
 
     @Override


nitinmotgi:
LGTM 👍

nitinmotgi (approved):
LGTM 👍

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -141,7 +141,7 @@ public FileFormat getFormat() {
     @Nullable
     @Override
     public Pattern getFilePattern() {
-      return null;
+      return Pattern.compile(fileRegex);
```
chtyim, line 144: If `fileRegex` is null, a NPE will be throw. However, it should return `null` in such case instead.

theodoresiu: Putting in a new PR for this: https://github.com/cdapio/hydrator-plugins/pull/1005/files#r359378170

PR #1010, opened 2019-12-18
Put  regex with null check

theodoresiu:
Move regex parsing to try catch block incase no regex is specified


core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -141,7 +141,14 @@ public FileFormat getFormat() {
     @Nullable
     @Override
     public Pattern getFilePattern() {
-      return Pattern.compile(fileRegex);
+      Pattern p;
```
chtyim: Simply do a `null` check is better.

```java
if (fileRegex == null) {
  return null;
}
return Pattern.compile(fileRegex);
```

theodoresiu: Done

chtyim (approved):

PR #1022, opened 2020-02-14
CDAP-16002 fix bug where header is emitted multiple times

albertshau:
When multiple files with headers are being read, the header will
be emitted once per file within the same RecordReader. This causes
Wrangler's parse-as-csv with header to fail when it reaches the
header for the 2nd file.

To fix this, ignore the header for everything except for the first
file in the split. This is only done when the hidden copyHeader
variable is set, as we do not actually want to filter out the first
row for files that do not have a special header.


yaojiefeng (approved):

albertshau:
https://builds.cask.co/browse/HYP-BAD650

PR #1021, opened 2020-02-14
[CDAP-16289] Added FLL for core plugins

vravish:

PR #1030, opened 2020-02-26
[CDAP-16289] Added FLL for core plugins

vravish:


yaojiefeng:
This seems same to the previous one, please address comments in #1021

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DistinctAggregator.java
```diff
@@ -106,6 +108,17 @@ public void prepareRun(BatchAggregatorContext context) throws Exception {
     validate(context.getInputSchema(), conf.getFields(), context.getFailureCollector());
     context.getFailureCollector().getOrThrowException();
 
+    if (conf.getFields() == null) {
+      context.record(TransformLineageRecorderUtils.generateOneToOnes(
+        TransformLineageRecorderUtils.getFields(context.getInputSchema()),
```
yaojiefeng: You can just create this field with `List<String> fields = conf.getFields() == null ?  ... : ...` to avoid the if else block

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DistinctAggregator.java
```diff
@@ -106,6 +108,17 @@ public void prepareRun(BatchAggregatorContext context) throws Exception {
...
+        "distinctAggregator",
+        "Removed duplicates in input records."));
+    } else {
+      context.record(TransformLineageRecorderUtils.generateOneToOnes(
+        Lists.newArrayList(conf.getFields()),
```
yaojiefeng: use `new ArrayList<>(conf.getFields())`

core-plugins/src/main/java/io/cdap/plugin/transform/ProjectionTransform.java
```diff
@@ -138,6 +146,25 @@ public void initialize(TransformContext context) throws Exception {
     collector.getOrThrowException();
   }
 
+  @Override
+  public void prepareRun(StageSubmitterContext context) {
+    init(context.getInputSchema(), context.getFailureCollector());
+
+    List<FieldOperation> output = new ArrayList<>();
+    Collection<String> fieldsToKeep = this.fieldsToKeep;
```
yaojiefeng: just create a keep list using `new ArrayList<>(fieldsToKeep)`

core-plugins/src/main/java/io/cdap/plugin/transform/ProjectionTransform.java
```diff
@@ -138,6 +146,25 @@ public void initialize(TransformContext context) throws Exception {
...
+    if (fieldsToKeep.isEmpty()) {
+      fieldsToKeep = TransformLineageRecorderUtils.getFields(context.getInputSchema());
+    }
+    output.addAll(TransformLineageRecorderUtils.generateOneToOnes(new ArrayList<>(fieldsToKeep),
+      "identity", TransformLineageRecorderUtils.IDENTITY_TRANSFORM_DESCRIPTION));
+    output.addAll(TransformLineageRecorderUtils.generateDrops(new ArrayList<>(fieldsToDrop)));
```
yaojiefeng: for drops, if the fieldsToKeep is not empty, any field that is not in the keep list is dropped. See https://github.com/cdapio/hydrator-plugins/blob/087c35db5174f754d1a5fba6abc209ee645f864e/core-plugins/src/main/java/io/cdap/plugin/transform/ProjectionTransform.java#L175-L176

yaojiefeng (commented):
Some comments, we are close to merge

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DistinctAggregator.java
vravish: That won't work since conf.getFields() returns an Iterable, and there's no ArrayList constructor that converts an Iterable to a list.

yaojiefeng (approved):
one minor nit, LGTM, please wait for the build and squash the commits before merging. Also, please document what plugin does not have FLL in the sheet

PR #1031, opened 2020-02-27
(CDAP-16324) Improve macro support in Joiner plugin

rmstar:
Output schema is now macro enabled.
Handle null input schema - this happens when the output schema of the previous node is a macro.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
@@ -81,7 +81,12 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
     FailureCollector collector = pipelineConfigurer.getMultiInputStageConfigurer().getFailureCollector();
     init(inputSchemas, collector);
     collector.getOrThrowException();
-    //validate the input schema and get the output schema for it
+    if (!conf.inputSchemasAvailable(inputSchemas) && !conf.containsMacro(conf.OUTPUT_SCHEMA) &&
+      conf.getOutputSchema(collector) == null) {
+      // If input schemas are unknown, an output schema must be provided.
+      collector.addFailure("Output schema must be specified", null).withConfigProperty(conf.OUTPUT_SCHEMA);
+    }
+
     Schema outputSchema = getOutputSchema(inputSchemas, collector);
     if (outputSchema != null) {
       // Set output schema if it's not a macro.
@@ -95,9 +100,15 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
       context.setNumPartitions(conf.getNumPartitions());
     }
     FailureCollector collector = context.getFailureCollector();
-    init(context.getInputSchemas(), collector);
+    Map<String, Schema> inputSchemas = context.getInputSchemas();
+    if (!conf.inputSchemasAvailable(inputSchemas)) {
+      // inputSchemas will be empty if the output schema of a previous node is a macro
+      return;
+    }
+
+    init(inputSchemas, collector);
     collector.getOrThrowException();
-    Collection<OutputFieldInfo> outputFieldInfos = createOutputFieldInfos(context.getInputSchemas(), collector);
+    Collection<OutputFieldInfo> outputFieldInfos = createOutputFieldInfos(inputSchemas, collector);
     collector.getOrThrowException();
     context.record(createFieldOperations(outputFieldInfos, perStageJoinKeys));
   }
@@ -186,10 +197,16 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
     List<Schema.Field> fields = new ArrayList<>();
     Schema schema = record.getSchema();
 
+
     List<String> joinKeys = perStageJoinKeys.get(stageName);
     int i = 1;
     for (String joinKey : joinKeys) {
-      Schema.Field joinField = Schema.Field.of(String.valueOf(i++), schema.getField(joinKey).getSchema());
+      Schema.Field field = schema.getField(joinKey);
+      if (field == null) {
+        throw new IllegalArgumentException(String.format("Join key field '%s' does not exist in schema from '%s'.",
+                                                         joinKey, stageName));
+      }
+      Schema.Field joinField = Schema.Field.of(String.valueOf(i++), field.getSchema());
       fields.add(joinField);
     }
     Schema keySchema = Schema.recordOf("join.key", fields);
@@ -250,8 +267,8 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
 
       Schema schema = inputSchemas.get(stageName);
       if (schema == null) {
-        collector.addFailure(String.format("Input schema for input stage '%s' cannot be null.", stageName), null);
-        collector.getOrThrowException();
+        // Input schema will be null if the output schema of the previous node is a macro
+        return;
       }
 
       for (String joinKey : entry.getValue()) {
@@ -280,7 +297,8 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
     duplicateFields = ArrayListMultimap.create();
     List<Schema.Field> outputFields = getOutputFields(createOutputFieldInfos(inputSchemas, collector));
     if (outputFields.isEmpty()) {
-      return null;
+      // Could not derive output schema from input schema. Try to get output schema from config.
+      return conf.getOutputSchema(collector);
     } else {
       return Schema.recordOf("join.output", outputFields);
     }
@@ -296,6 +314,11 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
     // Selected Field name to output field info
     Map<String, OutputFieldInfo> outputFieldInfo = new LinkedHashMap<>();
     List<String> duplicateAliases = new ArrayList<>();
+
+    if (!conf.inputSchemasAvailable(inputSchemas)) {
+      // inputSchemas will be empty if the output schema of a previous node is a macro
+      return outputFieldInfo.values();
+    }
 
     // order of fields in output schema will be same as order of selectedFields
     Set<Table.Cell<String, String, String>> rows = perStageSelectedFields.cellSet();
@@ -448,7 +471,7 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
 
   private void validateRequiredInputs(Map<String, Schema> inputSchemas, FailureCollector collector) {
     for (String requiredInput : requiredInputs) {
-      if (!inputSchemas.containsKey(requiredInput)) {
+      if (!inputSchemas.isEmpty() && !inputSchemas.containsKey(requiredInput)) {
         collector.addFailure(String.format("Provided input '%s' must be an input stage name.", requiredInput), null)
           .withConfigElement(conf.REQUIRED_INPUTS, requiredInput);
       }
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
@@ -31,6 +31,7 @@ import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.common.KeyValueListParser;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -45,6 +46,7 @@ public class JoinerConfig extends PluginConfig {
   public static final String SELECTED_FIELDS = "selectedFields";
   public static final String REQUIRED_INPUTS = "requiredInputs";
   public static final String JOIN_KEYS = "joinKeys";
+  public static final String OUTPUT_SCHEMA = "schema";
   private static final String NUM_PARTITIONS_DESC = "Number of partitions to use when joining. " +
     "If not specified, the execution framework will decide how many to use.";
   private static final String JOIN_KEY_DESC = "List of join keys to perform join operation. The list is " +
@@ -81,6 +83,12 @@ public class JoinerConfig extends PluginConfig {
   @Macro
   protected String requiredInputs;
 
+  @Nullable
+  @Macro
+  @Description(OUTPUT_SCHEMA)
+  private String schema;
+
+
   public JoinerConfig() {
     this.joinKeys = "";
     this.selectedFields = "";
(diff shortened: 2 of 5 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -95,11 +95,15 @@ public void prepareRun(BatchJoinerContext context) {
...
     }
     FailureCollector collector = context.getFailureCollector();
-    init(context.getInputSchemas(), collector);
-    collector.getOrThrowException();
-    Collection<OutputFieldInfo> outputFieldInfos = createOutputFieldInfos(context.getInputSchemas(), collector);
-    collector.getOrThrowException();
-    context.record(createFieldOperations(outputFieldInfos, perStageJoinKeys));
+    Map<String, Schema> inputSchemas = context.getInputSchemas();
+    // inputSchemas will be empty if the output schema of the previous node is a macro
+    if (!inputSchemas.isEmpty()) {
```
albertshau: nit: it's more readable in these situations to do:

```
if (inputSchemas.isEmpty()) {
  return;
}

// rest of logic
```

doesn't make much difference here, but is more true if the logic within the block is longer.

core-plugins/widgets/Joiner-batchjoiner.json
```diff
@@ -45,5 +45,24 @@
...
     }
   ],
-  "outputs": []
+  "outputs": [
+    {
+      "name": "schema",
+      "widget-type": "schema",
+      "widget-attributes": {
+        "schema-types": [
+          "boolean",
```
albertshau: are these types needed? Do the logical types (timestamp, date, etc) show up in the UI with this configuration?

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
albertshau: the correct behavior for the app would not be to give an empty map, but to have a null value as the schema. Let's check for that scenario as well (there is at least one null in the map values), so that this doesn't need to be revisited once the getInputSchemas() method is fixed.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -178,6 +182,13 @@ public void initialize(BatchJoinerRuntimeContext context) {
     init(inputSchemas, collector);
     collector.getOrThrowException();
     outputSchema = getOutputSchema(inputSchemas, collector);
+    if (outputSchema == null) {
+      // Could not derive output schema from input schema. Try to get output schema from config.
+      outputSchema = conf.getOutputSchema(collector);
+      if (outputSchema == null) {
```
albertshau: this should never be null, it should be validated in the configurePipeline() method.

albertshau: in configurePipeline(), you should validate that a valid output schema is provided if the input schemas are unknown.

albertshau (commented):
please also add a unit test for the macro use case.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java

core-plugins/widgets/Joiner-batchjoiner.json
rmstar: removed as it's not needed.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
rmstar: moved validation to configurePipeline()

rmstar:
I have added test cases for handling null input schemas. PTAL.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -361,6 +362,13 @@ Schema getOutputSchema(Map<String, Schema> inputSchemas, FailureCollector collec
 
     return outputFieldInfo.values();
   }
+  private boolean inputSchemasAvailable(Map<String, Schema> inputSchemas) {
+    // TODO: Remove isEmpty() check when CDAP-16351 is fixed
+    if (inputSchemas.isEmpty() || inputSchemas.values().stream().anyMatch(v -> v == null)) {
+      return false;
```
albertshau: can change this to a single line:

`return !inputSchemas.isEmpty() && ...`

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -217,7 +217,11 @@ private boolean isAliasPresent(Iterable<String> stageOldNameAliasPair) {
 
   void validateJoinKeySchemas(Map<String, Schema> inputSchemas, Map<String, List<String>> joinKeys,
                               FailureCollector collector) {
-    if (!containsMacro(JoinerConfig.JOIN_KEYS) && inputSchemas.size() != 0 && joinKeys.size() != inputSchemas.size()) {
+    // Skip validation if joinKeys is a macro, or if any input's output schema is a macro.
+    if (!containsMacro(JoinerConfig.JOIN_KEYS) &&
+      // TODO: Remove isEmpty() check when CDAP-16351 is fixed
+      !inputSchemas.isEmpty() && !inputSchemas.values().stream().anyMatch(v -> v == null) &&
```
albertshau: looks like duplicate code, can it be refactored so it's not in multiple places?

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -361,6 +362,13 @@ Schema getOutputSchema(Map<String, Schema> inputSchemas, FailureCollector collec
 
     return outputFieldInfo.values();
   }
+  private boolean inputSchemasAvailable(Map<String, Schema> inputSchemas) {
```
albertshau: style: should have a newline separating methods

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -77,37 +82,55 @@ public void testInnerJoin() throws Exception {
...
+      .addStage(filmActorStage)
+      .addStage(filmCategoryStage)
+      .addStage(joinStage)
+      .addStage(joinSinkStage)
+      .addConnection(filmStage.getName(), joinStage.getName())
+      .addConnection(filmActorStage.getName(), joinStage.getName())
+      .addConnection(filmCategoryStage.getName(), joinStage.getName())
+      .addConnection(joinStage.getName(), joinSinkStage.getName())
+      .build();
+    ApplicationManager appManager = deployETL(config, "joiner-test");
```
albertshau: is it an issue if the same name is used across runs?

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -77,37 +82,55 @@ public void testInnerJoin() throws Exception {
      *
      */
 
+    ETLBatchConfig config = ETLBatchConfig.builder("* * * * *")
```
albertshau: this method is deprecated and shouldn't be used, there is a builder method that doesn't take any arguments.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -216,6 +228,30 @@ public void testOuterJoin() throws Exception {
...
+
+  @Test
+  public void testOuterJoinWithUnknownInputSchemas() throws Exception {
+    String filmDatasetName = "film-outerjoin-unknown-inputschemas";
+    String filmCategoryDatasetName = "film-category-outerjoin-unknown-inputschemas";
+    String filmActorDatasetName = "film-actor-outerjoin-unknown-inputschemas";
+    String joinedDatasetName = "outerjoin-output-unknown-inputschemas";
+
+
+    ETLStage filmStage = new ETLStage("film", MockSource.getPlugin(filmDatasetName, FILM_SCHEMA));
```
albertshau: the schemas are being set here, isn't this test supposed to be for null input schemas? You should be using the method that doesn't take a schema as an argument.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -309,9 +310,32 @@ public void testOuterJoinWithoutRequiredInputs() throws Exception {
...
+                                            Properties.TimePartitionedFileSetDataset.TPFS_NAME, joinedDatasetName),
+                            null));
+
+    joinHelper(filmStage, filmActorStage, filmCategoryStage, joinStage, joinSinkStage,
+               filmDatasetName, filmActorDatasetName, filmCategoryDatasetName, joinedDatasetName,
+               outputSchema, this::verifyOuterJoinWithoutRequiredInputs);
+  }
+
+  @Test
+  public void testOuterJoinWithoutRequiredInputs() throws Exception {
```
albertshau, line 288: empty required inputs means an outer join? We don't need to run a full pipeline in order to test this, it should be handled in the JoinerConfigTest with maybe some refactoring (these pipeline tests take a long time to run and are really more like integration tests).

Also, is this testing some behavior that you changed in this PR?

albertshau (commented):
what happens when somebody gives the wrong output schema as a macro? We should make sure a good error message is logged

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: Yes, I see this error when dataset name is reused:
```
Exception in thread "WorkflowDriver" io.cdap.cdap.api.dataset.lib.PartitionAlreadyExistsException: Dataset 'innerjoin-output' already has a partition with the same key: {year=2020, month=3, day=2, hour=14, minute=17}
```

albertshau: yes, so is it a problem if this same 'joiner-test' app name is used? does this need to be unique?

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
rmstar: Refactored this into `JoinerConfig.inputSchemasAvailable()`

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
rmstar: done.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
rmstar: code moved to JoinerConfig.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: The test passes when the 'joiner-test' app name is reused. Is it incorrect reuse the app name? If yes, I can pass in app name as param to `joinHelper`

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: fixed.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: thanks for catching that! fixed.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: this PR isn't really changing outer join behavior. I've removed `testOuterJoinWithoutRequiredInputsAndInputSchemas`.

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
albertshau: it's probably safer to generate UUID.

we've had bugs in the past where a test will start an app, then wait for at least one completed run. However, since the app name is re-used, and the app was started in a different test, it immediately finds a completed run and doesn't actually wait for the current one to complete.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -198,10 +198,19 @@ public StructuredRecord joinOn(String stageName, StructuredRecord record) {
...
     Schema schema = record.getSchema();
 
+
     List<String> joinKeys = perStageJoinKeys.get(stageName);
     int i = 1;
     for (String joinKey : joinKeys) {
-      Schema.Field joinField = Schema.Field.of(String.valueOf(i++), schema.getField(joinKey).getSchema());
+      Schema.Field field = schema.getField(joinKey);
+      if (field == null) {
+        collector.addFailure(
```
albertshau: collector isn't really needed in this scenario, just throw an exception directly. The collector is useful mostly at configure time, to collect more than one failure and display it to the user at once.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -198,10 +198,19 @@ public StructuredRecord joinOn(String stageName, StructuredRecord record) {
...
+          String.format("Join key field '%s' is not present in input stage of '%s'.", joinKey, stageName), null)
```
albertshau: Join key field '[field name]' does not exist in schema from '[stage name]'

albertshau (approved):
minor comments, please fix and squash commits before merging.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
rmstar: updated to throw IllegalArgumentException

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
rmstar: thanks, fixed to use UUID.

PR #1032, opened 2020-03-04
CDAP-16367 fix fll for source and sink

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-16367
build: https://builds.cask.co/browse/HYP-BAD659-1

Currently we emit field lineage for source and sink using one operation containing all the fields, this basically means these fields are corelated, and if there are no transforms between source and sink, a lineage graph from all input to all output fields are generated. For pipelines with transform, this just happens to work since the transform is emitting the correct lineage for the output fields.

### DIFF
--- a/hydrator-common/src/main/java/io/cdap/plugin/common/LineageRecorder.java
+++ b/hydrator-common/src/main/java/io/cdap/plugin/common/LineageRecorder.java
@@ -85,10 +85,11 @@ public class LineageRecorder {
    * @param fields output fields of this read operation
    */
   public void recordRead(String operationName, String operationDescription, List<String> fields) {
-    context.record(Collections.singletonList(new FieldReadOperation(operationName,
-                                                                    operationDescription,
-                                                                    EndPoint.of(context.getNamespace(), dataset),
-                                                                    fields)));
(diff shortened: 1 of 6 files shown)


hydrator-common/src/main/java/io/cdap/plugin/common/LineageRecorder.java
```diff
@@ -85,23 +85,25 @@ public void createExternalDataset(@Nullable Schema schema) {
...
    * @param operationDescription description for the operation
-   * @param fields input fields of this read operation
+   * @param fields input fields of this write operation
    */
   public void recordWrite(String operationName, String operationDescription, List<String> fields) {
-    context.record(Collections.singletonList(new FieldWriteOperation(operationName,
-                                                                     operationDescription,
-                                                                     EndPoint.of(context.getNamespace(), dataset),
-                                                                     fields)));
+    fields.forEach(field -> context.record(
```
albertshau, line 103: It is confusing to me why one of these is 'correct' and the other is not. Conceptually, it makes sense to me that the sink is performing a single operation. Why does it need to be an operation per field?

PR #1033, opened 2020-03-07
[CDAP-16323] Added field name to DB source exception messages

vravish:


database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -196,7 +196,7 @@ private void writeToDataOut(DataOutput out, Schema.Field field) throws IOExcepti
         out.write((byte[]) fieldValue);
         break;
       default:
-        throw new IOException(String.format("Unsupported datatype: %s with value: %s.", fieldType, fieldValue));
+        throw new IOException(String.format("Unsupported datatype: %s with value: %s and name: %s.", fieldType, fieldValue, field.getName()));
```
bdmogal: I think we should fix the grammar in this sentence. How about - `Column %s with value %s has an unsupported datatype %s`.

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -259,7 +259,7 @@ private void writeToDB(PreparedStatement stmt, Schema.Field field, int fieldInde
         writeBytes(stmt, fieldIndex, sqlIndex, fieldValue);
         break;
       default:
-        throw new SQLException(String.format("Unsupported datatype: %s with value: %s.", fieldType, fieldValue));
+        throw new SQLException(String.format("Unsupported datatype: %s with value: %s and name: %s.", fieldType, fieldValue, field.getName()));
```
bdmogal: ditto

bdmogal (changes requested):
@vravish minor suggestion to improve the grammar. Are these the only instances where this needs to be fixed?

vravish:
@bdmogal Yes these are the only DB source exception messages. Please let me know if I missed any

bdmogal (approved):

PR #1034, opened 2020-03-07
(PLUGIN-114) Added COUNTDISTINCT function to count the distinct value…

bdmogal:
…s in a column

Issue: [PLUGIN-114](https://issues.cask.co/browse/PLUGIN-114)

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2016-2019 Cask Data, Inc.
+ * Copyright © 2016-2020 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -62,6 +62,7 @@ public class GroupByAggregator extends RecordAggregator {
     put("VARIANCE", "Variance");
     put("COLLECTLIST", "CollectList");
     put("COLLECTSET", "CollectSet");
+    put("COUNTDISTINCT", "CountDistinct");
   }};
 
   private List<String> groupByFields;
@@ -130,9 +131,27 @@ public class GroupByAggregator extends RecordAggregator {
                                            functionInfo.getName(), functionInfo.getField()), null)
           .withConfigElement("aggregates", collectorFieldName);
       }
+
+      // TODO: CDAP-16401 - Push down validation to individual aggregate functions
+      if (GroupByConfig.Function.COUNTDISTINCT == functionInfo.getFunction()) {
+        validateCountDistinct(inputField, collector, collectorFieldName);
+      }
     }
+  }
 
-
+  private void validateCountDistinct(Schema.Field inputField, FailureCollector collector, String validationFieldName) {
+    if (inputField != null) {
+      Schema.Type type = inputField.getSchema().isNullable() ?
+        inputField.getSchema().getNonNullable().getType() :
+        inputField.getSchema().getType();
+      if (type != Schema.Type.STRING && type != Schema.Type.INT && type != Schema.Type.LONG
+        && type != Schema.Type.BOOLEAN) {
+        collector.addFailure(
+          String.format("Distinct counting is not supported for the field %s of type %s.", inputField.getName(), type),
+          "Please specify a string, integer, long or boolean field.")
+          .withConfigElement("aggregates", validationFieldName);
+      }
+    }
   }
 
   @Override
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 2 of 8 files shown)


bdmogal:
Build succeeded - https://builds.cask.co/browse/HYP-BAD661

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Count distinct values of a specific column
+ */
+public class CountDistinct implements AggregateFunction<Integer> {
+  private final String fieldName;
+  private Set<Object> collectSet;
```
albertshau: this will only work for some field types. Those types should be validated at configure time, and again at prepare time if it couldn't be checked at configure time due to macros. More specifically, this will not "work" except on ints, longs, strings, floats, and doubles. And for floats and doubles, it doesn't really make much sense.

If it needs to work for other types, it is quite a bit more complex.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public CountDistinct(String fieldName) {
+    this.fieldName = fieldName;
+  }
+
+  @Override
+  public void beginFunction() {
+    collectSet = new HashSet<>();
```
albertshau: a little better to have the Set as final and create it in the constructor, and clear it here.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
bdmogal: @albertshau can you explain why we'd need to validate specific types? most of the use cases I can think of would be satisfied with strings, ints, longs, floats and doubles. Perhaps booleans in addition. But I'm unable to follow why it wouldn't work for others, and by others you mean the other types in the Schema object, right?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
bdmogal: fixed

bdmogal:
@albertshau addressed comments. Please take another look when you get a chance.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/CountDistinct.java
bdmogal: fixed

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -131,9 +131,29 @@ public void validate(Schema inputSchema, List<String> groupByFields,
                                            functionInfo.getName(), functionInfo.getField()), null)
           .withConfigElement("aggregates", collectorFieldName);
       }
-    }
 
+      // TODO: CDAP-16401 - Push down validation to individual aggregate functions
+      if (GroupByConfig.Function.COUNTDISTINCT == functionInfo.getFunction()) {
+        if (!conf.containsMacro("aggregates")) {
```
albertshau: is it possible for this to be false at this point? seems like it wouldn't be.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -131,9 +131,29 @@ public void validate(Schema inputSchema, List<String> groupByFields,
...
+    }
+  }
 
+  private void validateCountDistinct(Schema.Field inputField, FailureCollector collector, String validationFieldName) {
+    if (inputField != null) {
+      Schema.Type type = inputField.getSchema().isNullable() ?
+        inputField.getSchema().getNonNullable().getType() :
+        inputField.getSchema().getType();
+      if (type != Schema.Type.STRING && type != Schema.Type.INT && type != Schema.Type.LONG &&
+        type != Schema.Type.FLOAT && type != Schema.Type.DOUBLE && type != Schema.Type.BOOLEAN) {
```
albertshau: does it really make sense to allow float and double? It could potentially go out of memory easily, with 0.000001 different than 0.00000001, etc.

In Java when you check equality of floats and doubles most IDEs will flag it as a warning.

albertshau: though i guess the same could be set for CollectSet. We probably need a way to collect warnings and failures, with this falling under the warning category.

albertshau (approved):
minor question about whether a macro check is required, otherwise lgtm.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
bdmogal: ah! the macro check for aggregates is kind of in the weeds so I missed it. Removed it.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
bdmogal: Reported an improvement. https://issues.cask.co/browse/CDAP-16418. However, at this point, I don't think we need to absorb the risk of an OOM. I've removed support for FLOAT and DOUBLE right now. Can add it if someone brings up a use case explicitly.

albertshau (approved):
lgtm

bdmogal:
Latest successful build - https://builds.cask.co/browse/HYP-BAD661-8. Merging.

PR #1037, opened 2020-03-12
[CDAP-16330] Replaced config.getProperties with config.getRawProperti…

vravish:
…es to make sure validation happens on raw value before macros are evaluated

### DIFF
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/sink/FileSinkTestRun.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/sink/FileSinkTestRun.java
@@ -178,8 +178,8 @@ public class FileSinkTestRun extends ETLBatchTestBase {
     properties.put("path", outputDir.getAbsolutePath());
     properties.put("referenceName", format.name());
     properties.put("format", format.name());
-    //properties.put("schema", "${schema}");
-    properties.put("schema", SCHEMA.toString());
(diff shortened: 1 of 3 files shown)


core-plugins/src/test/java/io/cdap/plugin/batch/sink/FileSinkTestRun.java
```diff
@@ -178,8 +178,8 @@ private void runPipeline(FileFormat format, @Nullable String delimiter, FileCons
     properties.put("path", outputDir.getAbsolutePath());
     properties.put("referenceName", format.name());
     properties.put("format", format.name());
-    //properties.put("schema", "${schema}");
-    properties.put("schema", SCHEMA.toString());
+    properties.put("schema", "${schema}");
+    // properties.put("schema", SCHEMA.toString());
```
albertshau: can just remove this line

albertshau (approved):
one minor comment, otherwise lgtm.

Please wait for https://github.com/cdapio/cdap/pull/11954 to be merged before merging this.

vravish:
@albertshau I've already merged the other PR, but this is still not recognizing the getRawProperties method in PluginConfig class so the CI build is failing. Should we add that to a particular repo?

albertshau:
It's because the build has not run successfully yet, so nothing got published to maven central. I'm looking into the build failures

PR #1044, opened 2020-03-21
[CDAP-16453] Setting DBConf before creating record reader in DataDrivenETLDBInputFormat

vravish:


database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DataDrivenETLDBInputFormat.java
```diff
@@ -121,6 +121,7 @@ public Connection createConnection() {
 
   @Override
   protected RecordReader createDBRecordReader(DBInputSplit split, Configuration conf) throws IOException {
+    setConf(conf);
```
albertshau, line 124: the MapReduce framework is supposed to set this if the input format implements Configurable. It looks like the parent does implement it, so I'm wondering if this is some sort of MapReduce bug? Does it get set if you declare that this class implements Configurable?

vravish: I'm not sure if adding `implements Configurable` would help, since DataDrivenDBInputFormat already implements that interface. This might be a MapReduce bug if setConf is not being called

albertshau: Yes, that's what I'm trying to say. We need to investigate why it is not being set instead of just trying to workaround this issue. If there is a problem somewhere else, it could affect other plugins.

Can you try making it implement Configurable and see if MapReduce sets it? It's possible it is not checking the class correctly and is missing the fact that the parent is configurable. If that doesn't work, we'll have to identify where MapReduce is making the check and go from there.

Also, the jira looks weird, it says preview fails in Spark mode, but the stack trace looks like MapReduce. Does it fail in Spark too? With the same null pointer as the cause?

vravish: How do I set mode to Spark instead of MapReduce?

sagarkapare: oh yeah. it fails for Spark too.. Added stack trace for it - https://issues.cask.co/browse/CDAP-16453

albertshau: If the pipeline config menu, under the 'engine' section. If unit testing, using setEngine() on the ETLConfig builder

albertshau: Also, this only fails in preview but not when the pipeline is deployed and then run? If so it seems like something off in Preview and not something wrong with the plugin.

vravish: Update: adding `implements Configurable` doesn't change anything; the dbConf is still not set unless we call it from createDBRecordReader. And this fix works for preview in both Spark and MapReduce mode, both of which fail without the fix. This seems to be a blocker that we should merge ASAP, so can we just merge this for now? Or which method in the DBSource plugin is called when we preview this run?x

vravish: Update: adding `implements Configurable` doesn't change anything; the dbConf is still not set unless we call it from createDBRecordReader. And this fix works for preview in both Spark and MapReduce mode, both of which fail without the fix. This seems to be a blocker that we should merge ASAP, so can we just merge this for now? Or which method in the DBSource plugin is called when we preview this run?

albertshau: like I said, this is not the correct fix. If it only breaks in preview that means something is broken in preview. Just patching it over here will just lead to more bugs in some other place if the root cause is not identified and fixed.

PR #1047, opened 2020-03-24
Cache key schemas in Joiner

rmstar:
- cache key schemas in `joinOn`.
- removed unnecessary `validateJoinKeySchemas`  call in `getOutputSchema` (it's already called from `init`).

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
@@ -69,6 +69,7 @@ public class Joiner extends BatchJoiner<StructuredRecord, StructuredRecord, Stru
   private Table<String, String, String> perStageSelectedFields;
   private Set<String> requiredInputs;
   private Multimap<String, String> duplicateFields = ArrayListMultimap.create();
+  private Map<String, Schema> keySchemas = new HashMap<>();
 
   public Joiner(JoinerConfig conf) {
     this.conf = conf;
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -194,29 +195,37 @@ public void initialize(BatchJoinerRuntimeContext context) {
...
-    }
-    Schema keySchema = Schema.recordOf("join.key", fields);
-    StructuredRecord.Builder keyRecordBuilder = StructuredRecord.builder(keySchema);
-    i = 1;
-    for (String joinKey : joinKeys) {
-      keyRecordBuilder.set(String.valueOf(i++), record.get(joinKey));
+      keySchema = Schema.recordOf("join.key", fields);
+      keySchemas.put(stageName, keySchema);
     }
+      StructuredRecord.Builder keyRecordBuilder = StructuredRecord.builder(keySchema);
```
CuriousVini: nit: indentation

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -194,29 +195,37 @@ public void initialize(BatchJoinerRuntimeContext context) {
...
+      int i = 1;
+      for (String joinKey : joinKeys) {
+        keyRecordBuilder.set(String.valueOf(i++), record.get(joinKey));
+      }
+
+      StructuredRecord keyRecord = keyRecordBuilder.build();
+      return keyRecord;
```
CuriousVini: nit: `return keyRecordBuilder.build();`

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -291,7 +300,6 @@ void validateJoinKeySchemas(Map<String, Schema> inputSchemas, Map<String, List<S
   }
 
   Schema getOutputSchema(Map<String, Schema> inputSchemas, FailureCollector collector) {
-    validateJoinKeySchemas(inputSchemas, conf.getPerStageJoinKeys(), collector);
```
CuriousVini, line 294: why do we need to remove this? we still need this in case inputschemas are non-null right?

CuriousVini (commented):
minor comments

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
rmstar: `getOutputSchema()` is called from a couple of places:
1. `configurePipeline()`
2. `initialize()`
In both of these functions, `init()`, which calls `validateJoinKeySchemas()`,  is called before `getOuputSchema()`.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java

CuriousVini (approved):
lgtm

PR #1049, opened 2020-03-25
Bump version to 2.3.5

ajainarayanan:


albertshau (commented):
cdap version should also be changed to non-snapshot

albertshau:
the parent version range will also need to be updated to have 6.1.2 as the bottom, since the plugins rely on 6.1.2 they won't work on 6.1.1

albertshau (approved):

PR #1053, opened 2020-03-26
Add support in Joiner for input stages with a combination of macro an…

rmstar:
…d non-macro output schemas.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
@@ -46,8 +46,8 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
-
 /**
  * Batch joiner to join records from multiple inputs
  */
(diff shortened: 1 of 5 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -383,6 +378,12 @@ Schema getOutputSchema(Map<String, Schema> inputSchemas, FailureCollector collec
     return outputFields;
   }
 
+  private boolean inputSchemasAvailable(Map<String, Schema> inputSchemas, List<String> inputStages) {
+    return inputSchemas.keySet().containsAll(inputStages) &&
+      inputSchemas.values().stream().noneMatch(Objects::isNull);
+
```
albertshau: nit: remove empty line

pom.xml
```diff
@@ -678,7 +678,7 @@
           <version>1.1.0</version>
           <configuration>
             <cdapArtifacts>
-              <parent>system:cdap-data-pipeline[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
+              <parent>system:cdap-data-pipeline[6.2.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
               <parent>system:cdap-data-streams[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
```
albertshau: should also update the lower bound for this

albertshau (approved):
minor comments, otherwise lgtm. 

Please be sure the build (https://builds.cask.co/browse/HYP-BAD674) is green before merging. It looks like it's currently red due to some unrelated failure, so I kicked off another run

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java

pom.xml
rmstar: fixed.

PR #1056, opened 2020-03-31
CDAP-16508 add split quotes option

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-16508
build: https://builds.cask.co/browse/HYP-BAD677/latest

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -23,10 +23,9 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
 **Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
-is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a record looks like '1, "a, b, c"' and the delimeter
-is comma, the output result will be a structured record with field value '1' and 'a, b, c'.
-The newline delimeter cannot be within quotes and quotes will be trimmed. It also
-assumes the quotes are well enclosed. Default value is false.
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
+The first field will have '1' as its value and the second will have 'a, b, c' as its value.
+The newline delimiter cannot be within quotes and quotes will be trimmed. It also assumes the quotes are well enclosed.
 
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -22,6 +22,12 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a record looks like '1, "a, b, c"' and the delimeter
+is comma, the output result will be a structured record with field value '1' and 'a, b, c'.
+The newline delimeter cannot be within quotes and quotes will be trimmed. It also
+assumes the quotes are well enclosed. Default value is false.
+
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
 The default value is 128MB.
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -526,7 +526,7 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
     testReadDelimitedText(FileFormat.DELIMITED.name(), "\u0001", true);
   }
 
-  private void testReadDelimitedText(String format, String delimiter, boolean splitQuotes) throws Exception {
+  private void testReadDelimitedText(String format, String delimiter, boolean enableQuotedValues) throws Exception {
     File fileText = new File(temporaryFolder.newFolder(), "test.txt");
     String outputDatasetName = UUID.randomUUID().toString();
 
@@ -539,7 +539,7 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
 
     String appName = UUID.randomUUID().toString();
     ApplicationManager appManager = createSourceAndDeployApp(appName, fileText, format, outputDatasetName, schema,
-                                                             splitQuotes, delimiter);
+                                                             enableQuotedValues, delimiter);
 
     String join = Joiner.on(delimiter).join(new String[] {"\"a", "b", "c\""});
     String inputStr = new StringBuilder()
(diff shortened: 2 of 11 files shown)


format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -94,6 +94,12 @@
     + "is anything other than 'delimited'.")
   private String delimiter;
 
+  @Macro
+  @Nullable
+  @Description("Whether to split the content between quotes. This value will only be used if the format is 'csv', " +
+                 "'tsv' or 'delimited'. Default value is true.")
```
bdmogal: Shouldn't the default be false?

yaojiefeng: that will change the current behavior since without this change, we will split the quotes

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Common config for delimited related formats
+ */
+public class DelimitedConfig extends PathTrackingConfig {
+  public static final Map<String, PluginPropertyField> DELIMITED_FIELDS;
+  private static final String SPLIT_QUOTES_DESC =
+    "Whether to split the content between quotes. This value will only be used if the format is 'csv', 'tsv' " +
```
albertshau: at this level, it shouldn't mention 'csv', 'tsv', or 'delimited', because it is already the config for a specific format

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -72,9 +78,9 @@ public NullWritable getCurrentKey() {
         StructuredRecord.Builder builder = StructuredRecord.builder(schema);
         Iterator<Schema.Field> fields = schema.getFields().iterator();
 
-        for (String part : Splitter.on(delimiter).split(delimitedString)) {
+        for (String part : Splitter.on(pattern).split(delimitedString)) {
           if (!fields.hasNext()) {
-            int numDataFields = delimitedString.split(delimiter).length;
+            int numDataFields = delimitedString.split(pattern.pattern()).length;
```
albertshau: existing bug, but this should actually be length + 1

core-plugins/widgets/File-batchsource.json
```diff
@@ -47,6 +47,25 @@
           "widget-attributes": {
             "placeholder": "Delimiter if the format is 'delimited'"
           }
+        },
+        {
+          "widget-type": "radio-group",
```
albertshau: for booleans, we now use the 'toggle' type

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -94,6 +94,12 @@
     + "is anything other than 'delimited'.")
   private String delimiter;
 
+  @Macro
+  @Nullable
+  @Description("Whether to split the content between quotes. This value will only be used if the format is 'csv', " +
```
albertshau: I think it's more intuitive to have the opposite, a property that controls whether quotes should be treated as one value. Perhaps named 'Enable Quoted Values'

This should also be mentioned in the reference doc with an example. For example, it needs to be mentioned that the line delimiter (newline) cannot be within the quotes, and that the quotes themselves will not be included in the value. Also good to mention what happens if there is an unenclosed quote (odd number of quotes).

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
albertshau: since we're about to throw an exception and fail the pipeline, it would be good to also check if the line contains any quotes. If so, we can use a more specific message, something like:

"Found a row with 12 fields when the schema only contains 10 fields. Check if quoted values should be allowed"

albertshau (commented):
There are submodule updates that looked unintended.

Also, there should be corresponding changes to the output formats, where they can quote values if they contain the delimiter. This can be done in a separate PR though.

core-plugins/docs/File-batchsource.md
```diff
@@ -22,6 +22,12 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a record looks like '1, "a, b, c"' and the delimeter
+is comma, the output result will be a structured record with field value '1' and 'a, b, c'.
+The newline delimeter cannot be within quotes and quotes will be trimmed. It also
+assumes the quotes are well enclosed. Default value is false.
```
albertshau: don't need to mention the default here, since we default it in the UI already

core-plugins/docs/File-batchsource.md
```diff
@@ -22,6 +22,12 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a record looks like '1, "a, b, c"' and the delimeter
```
albertshau: For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields. The first field will have '1' as its value and the second will have 'a, b, c' as its value. The newline delimiter...

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -88,12 +96,19 @@ public NullWritable getCurrentKey() {
...
+              message += " Check if quoted values should be allowed.";
+            }
             throw new IOException(message + " Check that the schema contains the right number of fields.");
           }
 
           if (part.isEmpty()) {
             builder.set(fields.next().getName(), null);
           } else {
+            // if this part contains the original delimeter, remove the quotes
+            if (enableQuotesValue && part.contains(delimeter)) {
```
albertshau: somebody could choose to quote something that doesn't contain the delimiter, we still want to remove the enclosing quotes in that case.

yaojiefeng:
The sink change will happen later since that requires a change in cdap format

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -508,45 +510,56 @@ public void testReadJson() throws Exception {
...
+                                    Schema.Field.of("val2", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                                    Schema.Field.of("val3", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
                                     Schema.Field.of("file", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
 
     String appName = UUID.randomUUID().toString();
     ApplicationManager appManager = createSourceAndDeployApp(appName, fileText, format, outputDatasetName, schema,
-                                                             delimiter);
+                                                             enableQuotedValues, delimiter);
 
+    String join = Joiner.on(delimiter).join(new String[] {"\"a", "b", "c\""});
```
albertshau: can we also test multiple quoted values on the same line, something like:

"a, b", "c, d"

to make sure it grabs "a, b" as one value and "c, d" as another. We shouldn't have to run a pipeline to test this type of logic, let's do some refactoring so that these types of unit tests can be done efficiently.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -46,7 +48,13 @@
                                                                                     @Nullable Schema schema) {
 
     RecordReader<LongWritable, Text> delegate = (new TextInputFormat()).createRecordReader(split, context);
-    String delimiter = context.getConfiguration().get(DELIMITER);
+    String delimeter = context.getConfiguration().get(DELIMITER);
+    String regex = delimeter;
+    boolean enableQuotesValue = context.getConfiguration().getBoolean(ENABLE_QUOTES_VALUE, false);
+    if (enableQuotesValue) {
+      regex = String.format("%s(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", regex);
```
albertshau, line 55: I don't really understand this regex and can't tell that it behaves correctly, can you add some comments about it? We also need more unit tests, like testing what happens with multiple quoted values on the same line, odd number of quotes, etc.

PR #1058, opened 2020-04-01
CDAP-16517 support skip header

yaojiefeng:
Create  against the quote branch to avoid conflicts

### DIFF
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -69,7 +69,6 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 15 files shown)


core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -374,6 +373,62 @@ public void testPathGlobbing() throws Exception {
...
+                             Schema.Field.of("val3", Schema.of(Schema.Type.INT)),
+                             Schema.Field.of("file", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+    testSkipHeader("csv", ",", schema);
+    testSkipHeader("tsv", "\t", schema);
+    testSkipHeader("delimited", " ", schema);
+
+  }
+
+  private void testSkipHeader(String format, String delimeter, Schema schema) throws Exception {
+    File inputFile = temporaryFolder.newFile();
```
albertshau, line 396: better to write this once and pass it in as an argument.

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -374,6 +373,62 @@ public void testPathGlobbing() throws Exception {
...
+    String uri = inputFile.toURI().toString();
+    ApplicationManager appManager = createSourceAndDeployApp(appName, inputFile, format, outputDatasetName, schema,
+                                                             false, delimeter, true);
+
+    appManager.getWorkflowManager(SmartWorkflow.NAME).startAndWaitForRun(ProgramRunStatus.COMPLETED,
+                                                                         5, TimeUnit.MINUTES);
+
+    Set<StructuredRecord> expected = null;
+
+    if (format.equals("text")) {
```
albertshau, line 416: changing behavior based on some input argument is usually something to avoid. I think it's better to have this method return the output and have the caller assert equality.

albertshau (commented):
minor comments on code.

This change isn't related to quote support and should be merged as it's own commit.

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
yaojiefeng: the content in the file is different for each delimeter, cannot be created once for all tests

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
yaojiefeng: Fixed

PR #1059, opened 2020-04-01
CDAP-16517 skip header

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-16517
build: https://builds.cask.co/browse/HYP-BAD681-1

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -22,6 +22,8 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
+**Skip Header** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
+
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
 The default value is 128MB.
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.batch.source;
 
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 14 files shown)


core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -372,6 +373,51 @@ public void testPathGlobbing() throws Exception {
...
+      StructuredRecord.builder(schema).set("val1", 4).set("val2", 5).set("val3", 6).build());
+    Assert.assertEquals(expected, testSkipHeader("csv", ",", schema));
+    Assert.assertEquals(expected, testSkipHeader("tsv", "\t", schema));
+    Assert.assertEquals(expected, testSkipHeader("delimited", " ", schema));
+  }
+
+  private Set<StructuredRecord> testSkipHeader(String format, String delimeter, Schema schema) throws Exception {
+    File inputFile = temporaryFolder.newFile();
+
+    try (Writer writer = new FileWriter(inputFile)) {
```
albertshau, line 401: better to write just one file in the caller and pass in the File as an argument, especially since the caller is now checking the output.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -94,6 +93,12 @@
     + "is anything other than 'delimited'.")
   private String delimiter;
 
+  @Macro
+  @Nullable
+  @Description("Whether to skip header for the files. Supported formats are 'text', 'csv', 'tsv', 'delimited'. " +
```
albertshau: Whether to skip the first line of each file.

core-plugins/docs/File-batchsource.md
```diff
@@ -22,6 +22,8 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Delimiter:** Delimiter to use when the format is 'delimited'. This will be ignored for other formats.
 
+**Skip Header** Whether to skip header for the files. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
```
albertshau: Whether to skip the first line of each file

albertshau (approved):
minor comments, otherwise lgtm

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
yaojiefeng: the content in the file is different for each delimiter, so we cannot use one single file for testing

PR #1060, opened 2020-04-01
CDAP-16508 add quote option

yaojiefeng:
https://github.com/cdapio/hydrator-plugins/pull/1059 still needs a gcp changes to pass the build, this pr can rebase once we update that branch

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -24,6 +24,11 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Skip Header** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
+The first field will have '1' as its value and the second will have 'a, b, c' as its value.
+The newline delimiter cannot be within quotes and quotes will be trimmed. It also assumes the quotes are well enclosed.
+
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
 The default value is 128MB.
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.format.delimited.input;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
@@ -30,7 +31,10 @@ import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
+import java.util.TreeMap;
 import javax.annotation.Nullable;
 
 /**
@@ -39,6 +43,7 @@ import javax.annotation.Nullable;
 public class PathTrackingDelimitedInputFormat extends PathTrackingInputFormat {
   static final String DELIMITER = "delimiter";
   static final String SKIP_HEADER = "skip_header";
+  static final String ENABLE_QUOTES_VALUE = "enable_quotes_value";
 
   @Override
   protected RecordReader<NullWritable, StructuredRecord.Builder> createRecordReader(FileSplit split,
@@ -47,8 +52,10 @@ public class PathTrackingDelimitedInputFormat extends PathTrackingInputFormat {
                                                                                     @Nullable Schema schema) {
 
     RecordReader<LongWritable, Text> delegate = (new TextInputFormat()).createRecordReader(split, context);
-    String delimiter = context.getConfiguration().get(DELIMITER);
     boolean skipHeader = context.getConfiguration().getBoolean(SKIP_HEADER, false);
+
+    String delimiter = context.getConfiguration().get(DELIMITER);
+    boolean enableQuotesValue = context.getConfiguration().getBoolean(ENABLE_QUOTES_VALUE, false);
 
     return new RecordReader<NullWritable, StructuredRecord.Builder>() {
 
@@ -80,10 +87,17 @@ public class PathTrackingDelimitedInputFormat extends PathTrackingInputFormat {
 
         StructuredRecord.Builder builder = StructuredRecord.builder(schema);
         Iterator<Schema.Field> fields = schema.getFields().iterator();
+        List<String> splits = new ArrayList<>();
+        if (!enableQuotesValue) {
+          Iterable<String> iter = Splitter.on(delimiter).split(delimitedString);
+          iter.forEach(splits::add);
+        } else {
+          splits = splitQuotesString(delimitedString, delimiter);
+        }
 
-        for (String part : Splitter.on(delimiter).split(delimitedString)) {
+        for (String part : splits) {
           if (!fields.hasNext()) {
-            int numDataFields = delimitedString.split(delimiter).length;
+            int numDataFields = splits.size() + 1;
             int numSchemaFields = schema.getFields().size();
             String message = String.format("Found a row with %d fields when the schema only contains %d field%s.",
                                            numDataFields, numSchemaFields, numSchemaFields == 1 ? "" : "s");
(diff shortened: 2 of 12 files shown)


core-plugins/docs/File-batchsource.md
```diff
@@ -24,6 +24,11 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Skip Header** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
+The first field will have '1' as its value and the second will have 'a, b, c' as its value.
+The newline delimiter cannot be within quotes and quotes will be trimmed. It also assumes the quotes are well enclosed.
```
albertshau, line 30: nit:

The newline delimiter cannot be within quotes. The quote characters will be trimmed.

albertshau: It also assumes the quotes are well enclosed -> If there is an unenclosed quote, ...

what happens when there are unenclosed quotes? Does the pipeline fail? Is everything after the quote treated as one value? Should provide an example of this as well.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
     };
   }
+
+  /**
+   * Split the delimited string based on the delimiter. The delimiter should not contain any quotes. The method will
+   * behave like this:
+   * 1. if there is no quote, it will behave same as {@link String#split(String)}
+   * 2. if there are quotes in the string, the method will find pairs of quotes, content within each pair of quotes
+   * will not get splitted even if there is delimiter in that.
+   * For example, if string is "a.\"b.c\".\"d.e.f\"" and delimiter is '.', it will get split into [a, b.c, d.e.f].
```
albertshau, line 147: I think it would be more clear to just write a."b.c"."d.e.f"

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+   * if string is "val1\".\"val2\"", then it will not get splitted since the '.' is within pair of quotes.
```
albertshau, line 148: can you also document what happens if there are an odd number of quotes

core-plugins/docs/File-batchsource.md
albertshau: After reading through the test cases, we need more documentation and examples on what happens when there are multiple quotes.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static List<String> splitQuotesString(String delimitedString, String delimiter) {
+    // use a tree map so we can easily look up if a delimiter is within a pair of quotes
+    // the map has key of index and value to the index of paired quote.
+    TreeMap<Integer, Integer> quotesPos = new TreeMap<>();
+    int firstQuotePos = -1;
+    for (int i = 0; i < delimitedString.length(); i++) {
+      if (delimitedString.charAt(i) == '\"') {
```
albertshau, line 161: more readable to have a constant for the quote character and to remove a layer of nesting and do:

```
if (delimitedString.charAt(i) != QUOTE) {
  continue;
}
...
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+   * if string is "val1\".\"val2\"", then it will not get splitted since the '.' is within pair of quotes.
+   *
+   * @param delimitedString the string to split
+   * @param delimiter the separtor
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static List<String> splitQuotesString(String delimitedString, String delimiter) {
+    // use a tree map so we can easily look up if a delimiter is within a pair of quotes
+    // the map has key of index and value to the index of paired quote.
```
albertshau, line 157: nit: the map key is the index of the starting quote and the value is the index of the enclosing quote. An un-enclosed quote will not be represented in the map.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+      }
+    }
+
+    List<String> result = new ArrayList<>();
+    StringBuilder split = new StringBuilder();
+    for (int i = 0; i < delimitedString.length(); i++) {
+      // if the length is not enough for the delimiter, just add it to split
+      if (i + delimiter.length() > delimitedString.length()) {
+        split.append(delimitedString.charAt(i));
+      } else {
```
albertshau, line 178: readability: can remove a layer of nesting by removing the else and modifying the above to:

```
if (...) {
  split.append(...);
  continue;
}
...
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+          // if the they exist and they form a pair, we will not split, just add the current character to split
+          if (prevQuotePos != null && nextQuotePos != null && quotesPos.get(prevQuotePos).equals(nextQuotePos)) {
+            split.append(delimitedString.charAt(i));
+          } else {
+            // else we need to add the finding split to result, and increment the index, reset the split
+            result.add(trimQuotes(split.toString()));
+            i = i + delimiter.length() - 1;
+            split = new StringBuilder();
+          }
+        } else {
```
albertshau, line 193: similar comment here, it would be more readable to handle this above the larger chunk of code:

```
String possibleDelimiter = delimitedString.substring(i, i + delimiter.length());
if (!possibleDelimiter.equals(delimiter)) {
  split.append(...);
  continue;
}
...
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+        split.append(delimitedString.charAt(i));
+      } else {
+        // find a delimiter
+        if (delimitedString.substring(i, i + delimiter.length()).equals(delimiter)) {
+          // find the prev quote and next quote index
+          Integer prevQuotePos = quotesPos.floorKey(i);
+          Integer nextQuotePos = quotesPos.ceilingKey(i);
+          // if the they exist and they form a pair, we will not split, just add the current character to split
+          if (prevQuotePos != null && nextQuotePos != null && quotesPos.get(prevQuotePos).equals(nextQuotePos)) {
+            split.append(delimitedString.charAt(i));
```
albertshau, line 186: at this point it's better to append everything up to the nextQuote and update i accordingly, otherwise there is a lot of repeat work

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -120,4 +137,77 @@ public void close() throws IOException {
...
+      if (i + delimiter.length() > delimitedString.length()) {
+        split.append(delimitedString.charAt(i));
+      } else {
+        // find a delimiter
+        if (delimitedString.substring(i, i + delimiter.length()).equals(delimiter)) {
+          // find the prev quote and next quote index
+          Integer prevQuotePos = quotesPos.floorKey(i);
+          Integer nextQuotePos = quotesPos.ceilingKey(i);
+          // if the they exist and they form a pair, we will not split, just add the current character to split
+          if (prevQuotePos != null && nextQuotePos != null && quotesPos.get(prevQuotePos).equals(nextQuotePos)) {
```
albertshau, line 185: I'm a bit confused here. Say the line is:
```
pos: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
     " | a | , | b | " | , | " | c | , | d | "  |
```

that means quotesPos is:

0 -> 4
6 -> 10

So doesn't this mean it mean `quotesPos.get(prevQuotePos)` will never be the same as `nextQuotePos`?

albertshau: ok just realized the map has both left -> right as well as right -> left, so it would be:

```
0 -> 4
4 -> 0
6 -> 10
10 -> 6
```

Can you add more comments with an example?

yaojiefeng:
Putting this on hold, we might want to have a different strategy for quotes

PR #1064, opened 2020-04-02
CDAP-16525 add config to filter pattern in field name

yaojiefeng:
JIRA: https://issues.cask.co/browse/CDAP-16525
https://builds.cask.co/browse/HYP-BAD688-1

Sometimes a db can contain some characters, for exmaple "\", which is not valid in the avro schema, we need to add a way to replace these patterns with something valid

### DIFF
--- a/database-plugins/widgets/Database-batchsource.json
+++ b/database-plugins/widgets/Database-batchsource.json
@@ -161,6 +161,21 @@
           }
         }
       ]
+    },
+    {
+      "label": "Column Names Normalization",
+      "properties": [
+        {
+          "widget-type": "textbox",
+          "label": "Pattern To Replace",
+          "name": "patternToReplace"
+        },
+        {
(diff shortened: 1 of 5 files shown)


database-plugins/widgets/Database-batchsource.json
```diff
@@ -159,6 +159,16 @@
             ],
             "default": "TRANSACTION_SERIALIZABLE"
           }
+        },
+        {
```
bdmogal, line 168: Can we put this in a separate Group? So that users don't confuse this as something to replace the pattern in the data itself. Something like Column Names Normalization?

chtyim (approved):
One comment about the modules in pom.xml

They shouldn't be commented out.

yaojiefeng:
I made a mistake to commit the pom change, already removed that

PR #1069, opened 2020-04-04
[CDAP-16290] Added recordLineage method to ReferenceBatchSink and ReferenceBatchSource so all source and sink plugins can use it

vravish:


vravish:
The code is now in the CDAP repo; only PR needed is https://github.com/cdapio/cdap/pull/12036

yaojiefeng:
Let's have the method to record lineage in hydrator, don't put that in cdap, let's reopen this one

PR #1070, opened 2020-04-05
[CDAP-16532] multi-level input schema serialization as csv, tsv or delimited validation errors

nitinmotgi:
This PR adds a validation check to make sure that when user has multi-level or nested input structure and are serializing it as CSV, TSV or DELIMITED there is a appropriate error message as well as limiting that combination.

### DIFF
--- a/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
@@ -84,13 +84,32 @@ public abstract class AbstractFileSinkConfig extends PluginConfig implements Fil
           .withConfigProperty(NAME_SUFFIX).withStacktrace(e.getStackTrace());
       }
     }
+
+    FileFormat format = FileFormat.CSV;
     try {
-      getFormat();
+      format = getFormat();
     } catch (IllegalArgumentException e) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_FORMAT).withStacktrace(e.getStackTrace());
     }
     try {
-      getSchema();
+      Schema schema = getSchema();
+      // Checks if the output schema has fields that are not simple type.
+      // If there are non-simple field types, then serializing as CSV might not be appropriate.
+      // This restriction is on Csv, Delimited and Tab separated files.
+      if (format == FileFormat.CSV || format == FileFormat.DELIMITED || format == FileFormat.TSV) {
+        boolean simpleFields = true;
+        for (Schema.Field field : schema.getFields()) {
+          if (!field.getSchema().isSimpleOrNullableSimple()) {
+            simpleFields = false;
+            break;
+          }
+        }
+        collector.addFailure(
+          String.format("Input has multi-level structure that cannot be represented appropriately in %s.",
+                        format.name()),
+          "Consider using json, avro or parquet to write data."
+        ).withConfigProperty(NAME_FORMAT);
+      }
     } catch (IllegalArgumentException e) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_SCHEMA).withStacktrace(e.getStackTrace());
     }
--- a/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
@@ -85,7 +85,7 @@ public abstract class AbstractFileSinkConfig extends PluginConfig implements Fil
       }
     }
 
-    FileFormat format = FileFormat.CSV;
+    FileFormat format = FileFormat.JSON;
     try {
       format = getFormat();
     } catch (IllegalArgumentException e) {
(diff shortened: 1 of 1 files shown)


format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
```diff
@@ -84,13 +84,32 @@ public void validate(FailureCollector collector) {
           .withConfigProperty(NAME_SUFFIX).withStacktrace(e.getStackTrace());
       }
     }
+
+    FileFormat format = FileFormat.CSV;
```
chtyim: Was the default CSV? Or are we changing it to CSV?

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
```diff
@@ -84,13 +84,32 @@ public void validate(FailureCollector collector) {
...
     }
     try {
-      getSchema();
+      Schema schema = getSchema();
+      // Checks if the output schema has fields that are not simple type.
+      // If there are non-simple field types, then serializing as CSV might not be appropriate.
+      // This restriction is on Csv, Delimited and Tab separated files.
+      if (format == FileFormat.CSV || format == FileFormat.DELIMITED || format == FileFormat.TSV) {
+        boolean simpleFields = true;
+        for (Schema.Field field : schema.getFields()) {
```
chtyim: Can be just

```
boolean allSimple = schema.getFields().stream().map(Schema.Field::getSchema).allMatch(Schema::isSimpleOrNullableSimple)
```

nitinmotgi: this will logic will not short circuit right and go through all the elements.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
nitinmotgi: We are not changing from what it is. But, in any case changed it to JSON (which is the default).

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
nitinmotgi: Found that it does. changing it.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
```diff
@@ -84,13 +84,32 @@ public void validate(FailureCollector collector) {
...
+          if (!field.getSchema().isSimpleOrNullableSimple()) {
+            simpleFields = false;
+            break;
+          }
+        }
+        collector.addFailure(
+          String.format("Input has multi-level structure that cannot be represented appropriately in %s.",
```
bdmogal: should we use the word "nested" in the message? Also, it would help to add the name of the field that has the problem?

nitinmotgi:
I thought about that. But if there are multiple fields then we need to list
them all. But ultimately, if you think from user perspective we are not
suggesting them to remove the fields, the intent here is to suggest user to
use the right format for the data that they have. Suggesting to remove the
field can be conceived as limitation. So had to carefully word it.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
nitinmotgi: I thought about that. But if there are multiple fields then we need to list
them all. But ultimately, if you think from user perspective we are not
suggesting them to remove the fields, the intent here is to suggest user to
use the right format for the data that they have. Suggesting to remove the
field can be conceived as limitation. So had to carefully word it.

nitinmotgi:
I have tested it again and made sure everything works as expected. Tried with Nested structures and non-nested structures. @chtyim PTAL.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSinkConfig.java
```diff
@@ -84,13 +84,30 @@ public void validate(FailureCollector collector) {
...
     } catch (IllegalArgumentException e) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_FORMAT).withStacktrace(e.getStackTrace());
     }
     try {
-      getSchema();
+      Schema schema = getSchema();
+      // Checks if the output schema has fields that are not simple type.
+      // If there are non-simple field types, then serializing as CSV might not be appropriate.
+      // This restriction is on Csv, Delimited and Tab separated files.
+      if (format == FileFormat.CSV || format == FileFormat.DELIMITED || format == FileFormat.TSV) {
```
albertshau, line 99: this validation should be done in the format (DelimitedOutputFormatProvider's validate method), not in the file sink.

albertshau: discussed offline, the problem is there isn't a way to pass down the name of the format field, since we want to highlight it. So we can keep it for now and open a jira to fix this in a cleaner way.

albertshau (approved):

nitinmotgi:
@albertshau have open a JIRA - https://issues.cask.co/browse/CDAP-16537 to address the problem. Merging the change.

PR #1072, opened 2020-04-07
Macro support for format field in AbstractFileSource, AbstractFileSink

rmstar:
JIRA: https://issues.cask.co/browse/CDAP-16572

### DIFF
--- a/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
@@ -20,6 +20,8 @@ import io.cdap.cdap.api.data.batch.Output;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.dataset.lib.KeyValue;
+import io.cdap.cdap.api.plugin.InvalidPluginConfigException;
+import io.cdap.cdap.api.plugin.InvalidPluginProperty;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
 import io.cdap.cdap.etl.api.FailureCollector;
@@ -33,12 +35,16 @@ import io.cdap.plugin.common.batch.sink.SinkOutputFormatProvider;
 import io.cdap.plugin.format.FileFormat;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.text.SimpleDateFormat;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
@@ -49,7 +55,8 @@ import javax.annotation.Nullable;
  */
 public abstract class AbstractFileSink<T extends PluginConfig & FileSinkProperties>
   extends BatchSink<StructuredRecord, NullWritable, StructuredRecord> {
-  private static final String FORMAT_PLUGIN_ID = "format";
+  private static final Logger LOG = LoggerFactory.getLogger(AbstractFileSink.class);
+  private static final String NAME_FORMAT = "format";
   private final T config;
 
   protected AbstractFileSink(T config) {
@@ -64,10 +71,26 @@ public abstract class AbstractFileSink<T extends PluginConfig & FileSinkProperti
     // invalid
     collector.getOrThrowException();
 
+    if (config.containsMacro(NAME_FORMAT)) {
+      // Deploy all format plugins. This ensures that the required plugin is available when
+      // the format macro is evaluated in prepareRun.
+      for (FileFormat f: FileFormat.values()) {
+        try {
+          pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, f.name().toLowerCase(),
+                                       f.name().toLowerCase(), config.getRawProperties());
+        } catch (InvalidPluginConfigException e) {
+          LOG.warn("Failed to register format '{}', which means it cannot be used when the pipeline is run. " +
+                     "Missing properties: {}, invalid properties: {}", f.name(), e.getMissingProperties(),
+                   e.getInvalidProperties().stream().map(InvalidPluginProperty::getName).collect(Collectors.toList()));
+        }
+      }
+      return;
+    }
+
     FileFormat format = config.getFormat();
     ValidatingOutputFormat validatingOutputFormat =
       pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, format.name().toLowerCase(),
-                                   FORMAT_PLUGIN_ID, config.getRawProperties());
+                                   format.name().toLowerCase(), config.getRawProperties());
     FormatContext context = new FormatContext(collector, pipelineConfigurer.getStageConfigurer().getInputSchema());
     validateOutputFormatProvider(context, format, validatingOutputFormat);
   }
@@ -76,8 +99,21 @@ public abstract class AbstractFileSink<T extends PluginConfig & FileSinkProperti
   public void prepareRun(BatchSinkContext context) throws Exception {
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
-    ValidatingOutputFormat validatingOutputFormat = context.newPluginInstance(FORMAT_PLUGIN_ID);
     FileFormat fileFormat = config.getFormat();
+    ValidatingOutputFormat validatingOutputFormat;
+    try {
+      validatingOutputFormat = context.newPluginInstance(fileFormat.name().toLowerCase());
+    } catch (InvalidPluginConfigException e) {
+      Set<String> properties = new HashSet<>(e.getMissingProperties());
+      for (InvalidPluginProperty invalidProperty: e.getInvalidProperties()) {
+        properties.add(invalidProperty.getName());
+      }
+      String errorMessage = String.format("Format '%s' cannot be used because properties %s were not provided or " +
+                                            "were invalid when the pipeline was deployed. Set the format to a " +
+                                            "different value, or re-create the pipeline with all required properties.",
+                                          fileFormat.name(), properties);
+      throw new IllegalArgumentException(errorMessage, e);
+    }
     FormatContext formatContext = new FormatContext(collector, context.getInputSchema());
     validateOutputFormatProvider(formatContext, fileFormat, validatingOutputFormat);
     collector.getOrThrowException();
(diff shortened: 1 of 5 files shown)


rmstar:
Failure in io.cdap.plugin.HBaseTest looks unrelated to this PR.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -64,10 +64,20 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     // invalid
     collector.getOrThrowException();
 
+    if (config.containsMacro(NAME_FORMAT)) {
+      // Deploy all format plugins. This ensures that the required plugin is available when
+      // the format macro is evaluated in prepareRun,
+      for (FileFormat f: FileFormat.values()) {
+        pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, f.name().toLowerCase(), f.name().toLowerCase(),
```
albertshau: are there any format properties that are required by some formats and optional for others? If so, this can run into errors

albertshau (commented):
one question

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
rmstar: Thanks for pointing this out. Handled InvalidPluginConfigException as discussed offline.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -64,10 +68,25 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
 
+    if (config.containsMacro(NAME_FORMAT)) {
+      // Deploy all format plugins. This ensures that the required plugin is available when
+      // the format macro is evaluated in prepareRun,
+      for (FileFormat f: FileFormat.values()) {
+        try {
+          pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, f.name().toLowerCase(),
+                                       f.name().toLowerCase(), config.getRawProperties());
+        } catch (InvalidPluginConfigException e) {
+          LOG.warn("Failed to configure {}, missing properties: {}, invalid properties: {}",
```
albertshau: error messages should inform the reader about the implications. Something like:

Failed to register format '{}', which means it cannot be used when the pipeline is run. Missing properties ...

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -76,8 +95,15 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
   public void prepareRun(BatchSinkContext context) throws Exception {
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
-    ValidatingOutputFormat validatingOutputFormat = context.newPluginInstance(FORMAT_PLUGIN_ID);
     FileFormat fileFormat = config.getFormat();
+    ValidatingOutputFormat validatingOutputFormat;
+    try {
+      validatingOutputFormat = context.newPluginInstance(fileFormat.name().toLowerCase());
+    } catch (InvalidPluginConfigException e) {
+      LOG.error("Failed to create config for {} plugin missing properties: {} invalid properties: {}",
```
albertshau: In general, avoid this type of pattern where an exception is caught, logged, and then re-thrown. What ends up happening is that the error gets logged here and then later on by the caller, generating extra logs and confusing the user.

Instead, you can wrap the exception with a more specific error message and propagate it up. Something like:

```
throw new IllegalArgumentException("...", e);
```

The message should also give more information to the user about what they should do to fix the problem. This particular case can only happen when the format was a macro, otherwise it would have caused an error in configurePipeline. So a better message would be:

Format '{}' cannot be used because properties {} were not provided or were invalid when the pipeline was deployed. Set the format to a different value, or re-create the pipeline with all required properties.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -76,8 +96,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-    ValidatingOutputFormat validatingOutputFormat = context.newPluginInstance(FORMAT_PLUGIN_ID);
     FileFormat fileFormat = config.getFormat();
+    ValidatingOutputFormat validatingOutputFormat;
+    try {
+      validatingOutputFormat = context.newPluginInstance(fileFormat.name().toLowerCase());
+    } catch (InvalidPluginConfigException e) {
+      String errorMessage = String.format("Format '%s' cannot be used because properties %s were not provided or " +
+                                          "were invalid when the pipeline was deployed. Set the format to a " +
+                                          "different value, or re-create the pipeline with all required properties.",
+                                          fileFormat.name(), e.getMissingProperties());
```
albertshau: properties in the message should include both the missing and the invalid. Otherwise if it was from just invalid, the message will look weird.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
```diff
@@ -79,11 +79,21 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     // invalid
     collector.getOrThrowException();
 
+    if (config.containsMacro(NAME_FORMAT)) {
+      // Deploy all format plugins. This ensures that the required plugin is available when
+      // the format macro is evaluated in prepareRun,
+      for (FileFormat f: FileFormat.values()) {
+        pipelineConfigurer.usePlugin(ValidatingInputFormat.PLUGIN_TYPE, f.name().toLowerCase(), f.name().toLowerCase(),
+          config.getRawProperties());
+      }
```
albertshau, line 101: we should have similar error handling here in the source if there is an error at configure or prepare time.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -60,28 +61,33 @@
...
 
   @Name(PATH_FIELD)
+  @Macro
   @Nullable
   @Description("Output field to place the path of the file that the record was read from. "
     + "If not specified, the file path will not be included in output records. "
     + "If specified, the field must exist in the output schema as a string.")
   private String pathField;
 
+  @Macro
```
albertshau, line 84: have we tested a pipeline where all of these are macros? Have seen bugs introduced in the past where the property is assumed to not be a macro and a null pointer exception is generated at configure time.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
rmstar: Added invalid properties in error msg.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
rmstar: fixed

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
rmstar: yes, I tested a File source -> File sink pipeline where all these fields are macros.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -76,8 +96,17 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
-    ValidatingOutputFormat validatingOutputFormat = context.newPluginInstance(FORMAT_PLUGIN_ID);
     FileFormat fileFormat = config.getFormat();
+    ValidatingOutputFormat validatingOutputFormat;
+    try {
+      validatingOutputFormat = context.newPluginInstance(fileFormat.name().toLowerCase());
+    } catch (InvalidPluginConfigException e) {
+      String errorMessage = String.format("Format '%s' cannot be used because properties %s were not provided or " +
+                                          "were invalid %s when the pipeline was deployed. Set the format to a " +
```
albertshau: e.getInvalidProperties() is a set of objects, not property names. It will not display nicely. The location of that third '%s' is also awkward, it won't be a good sentence. 

Let's remove that last %s

invalid %s -> invalid

and create a union of the property names from missing properties and invalid properties.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
```diff
@@ -79,11 +83,27 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
+    if (config.containsMacro(NAME_FORMAT)) {
+      // Deploy all format plugins. This ensures that the required plugin is available when
+      // the format macro is evaluated in prepareRun,
+      for (FileFormat f: FileFormat.values()) {
+        try {
+          pipelineConfigurer.usePlugin(ValidatingInputFormat.PLUGIN_TYPE, f.name().toLowerCase(),
+                                       f.name().toLowerCase(), config.getRawProperties());
+        } catch (InvalidPluginConfigException e) {
+          LOG.warn("Failed to register format '{}', which means it cannot be used when the pipeline is run." +
+              "Missing properties: {}, invalid properties: {}",
```
albertshau: indentation should be 2 spaces not 4

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
```diff
@@ -99,8 +119,18 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
     FailureCollector collector = context.getFailureCollector();
     config.validate(collector);
-    ValidatingInputFormat validatingInputFormat = context.newPluginInstance(FORMAT_PLUGIN_ID);
     FileFormat fileFormat = config.getFormat();
+    ValidatingInputFormat validatingInputFormat;
+    try {
+      validatingInputFormat = context.newPluginInstance(fileFormat.name().toLowerCase());
+    } catch (InvalidPluginConfigException e) {
+      String errorMessage = String.format("Format '%s' cannot be used because properties %s were not provided or " +
+          "were invalid %s when the pipeline was deployed. Set the format to a " +
```
albertshau: indentation

albertshau: same comment as above about the error message

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
albertshau: also, indentation here is inconsistent. Continuing a statement on the next line should be a 2 space indent, not aligned with the previous line. alignment is used when there are multiple arguments for a single method call spread out across multiple lines.

```
format("message '%s' " + 
         "wraps ",
       arg);
```

please see rest of the project for examples, and update your IDE accordingly.

rmstar: fixed indentation and error msg.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
rmstar: fixed

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
rmstar: fixed indent and error msg

albertshau (approved):

PR #1106, opened 2020-06-03
CDAP-16875 changed joiner plugin to use new auto-join API

albertshau:
Changed the joiner to use the new AutoJoiner API to take
advantage of the performance improvements it allows in Spark
pipelines.

Introduced two new optional properties. The first determines
whether the join will use null safe equality, and the second
provides hint to the execution engine about which input datasets
should be broadcast to perform an in-memory join.

Removed much of the existing join logic because it has been moved
into the application code. Updated the documentation to use
the property names that show up in the UI, to display the properties
in the same order that they appear in the UI, and to include
descriptions of the two new properties.

### DIFF
--- a/core-plugins/docs/Joiner-batchjoiner.md
+++ b/core-plugins/docs/Joiner-batchjoiner.md
@@ -1,49 +1,90 @@
 # Joiner
 
-
 Description
 -----------
-Joins records from one or more input based on join keys. Supports `inner` and `outer` joins, selection and renaming of output fields.  
-
-Use Case
---------
-The transform is used when you want to combine fields from one or more input, similar to the joins in SQL.
+Joins records from one or more input based on join key equality.
+Supports `inner` and `outer` joins, selection and renaming of output fields.
+The plugin is used when you want to combine fields from one or more inputs, similar to joins in SQL.
 
 Properties
 ----------
-**joinKeys:** List of keys to perform the join operation. The list is separated by `&`. 
+**Fields:** List of fields from each input that should be included in the output. 
+Output field names must be unique. If the same field name exists in more than one input,
+each field must be aliased (renamed) to a unique output name.
+
+**Join Type:** Type of join to perform.
+A join between two required input is an inner join. A join between a required input and an optional
+input is a left outer join. A join between two optional inputs is an outer join.
+
+A join of more than two inputs is logically equivalent to performing inner joins over all the
+required inputs, followed by left outer joins on the optional inputs.
+
+**Join Condition:** List of keys to perform the join operation. The list is separated by `&`. 
 Join key from each input stage will be prefixed with `<stageName>.` and the relation among join keys from different inputs is represented by `=`. 
 For example: customers.customer_id=items.c_id&customers.customer_name=items.c_name means the join key is a composite key
 of customer id and customer name from customers and items input stages and join will be performed on equality 
 of the join keys. This transform only supports equality for joins.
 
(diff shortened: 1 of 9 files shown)


core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -1,49 +1,91 @@
...
+| 3           | bob    |
+
+The join is a left outer join on purchases.customer_name = customers.name.
+If this property is set to true, the joined output would be:
+
+| purchase_id | customer_name | item   | customer_id | name  |
+| ----------- | ------------- | ------ | ----------- | ----- |
+| 1           | alice         | donut  | 1           | alice |
+| 2           |               | coffee | 2           |       |
+| 2           |               | coffee | 3           |       |
```
yaojiefeng: Should this be 2 and next line be 3?

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -1,49 +1,91 @@
...
+| 3           | bob           | water  |             |       |
+
+Note that the rows with a null customer name were joined together.
+If this property is set to false, the joined output would be:
+
+| purchase_id | customer_name | item   | customer_id | name  |
+| ----------- | ------------- | ------ | ----------- | ----- |
+| 1           | alice         | donut  | 1           | alice |
+| 2           |               | coffee |             |       |
```
yaojiefeng, line 66: Here the customer_id should be 2 right?

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -58,40 +60,83 @@
...
+        .select(conf.getSelectedFields(collector))
+        .from(inputs)
+        .on(JoinCondition.onKeys()
+              .setKeys(conf.getJoinKeys(collector))
+              .setNullSafe(conf.isNullSafe())
+              .build());
+      if (useOutputSchema) {
+        joinBuilder.setOutputSchema(conf.getOutputSchema(collector));
+      } else {
+        joinBuilder.setOutputSchemaName("join.output");
```
yaojiefeng, line 114: Just curious - is the schema name useful for the output schema? I believe the builder will use its own schema name if this is not provided, and because different plugins seem to give different schema name, and the name basically is not used anywhere. Is it important and worth to pay attention when developing a plugin in the future?

pom.xml
```diff
@@ -837,7 +837,7 @@
         <module>google-cloud</module>
         <module>kafka-plugins</module>
         <module>amazon-s3-plugins</module>
-        <module>azure</module>
+        <!--module>azure</module-->
```
yaojiefeng: nit - remove this line

yaojiefeng (approved):
Minor comments, LGTM

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
albertshau: it's important with certain types of schemas that have recursive structures, because the schema name is used instead of the object schema to prevent infinite recursion. In most situations it doesn't matter too much, setting this here since this is the schema name used before.

core-plugins/docs/Joiner-batchjoiner.md
albertshau: yeah, thanks for catching, this row should be deleted. I changed the example input data and didn't change this.

PR #1107, opened 2020-06-07
CDAP-16895 use new aggregator api

yaojiefeng:
This is taking much longer than expected because of big changes in all the aggregation functions and unit tests in groupBy plugin, still cannot find a good way to compute the population variance based on variance from two groups, so change it to sample variance.


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -87,6 +84,13 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     Schema outputSchema = getOutputSchema(inputSchema);
     FailureCollector collector = stageConfigurer.getFailureCollector();
     validateSchema(outputSchema, uniqueFields, functionInfo, collector);
+
+    if (functionInfo != null) {
```
albertshau, line 88: if the input schema is null, moving this check down will cause the validation to be skipped.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -174,26 +179,50 @@ public void groupBy(StructuredRecord record, Emitter<StructuredRecord> emitter)
   }
 
   @Override
-  public void aggregate(StructuredRecord groupKey, Iterator<StructuredRecord> iterator,
-                        Emitter<StructuredRecord> emitter) throws Exception {
-    if (!iterator.hasNext()) {
-      return;
+  public Table<String, String, Object> initializeAggregateValue(StructuredRecord record) {
+    Table<String, String, Object> table = HashBasedTable.create();
+    table.put(SCHEMA, "", record.getSchema());
```
albertshau: I'm confused about what is being put in this Table. Whatever object this is needs to be Serializable right? Does this work properly?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/AggregateFunction.java
```diff
@@ -16,27 +16,46 @@
...
+public interface AggregateFunction<T> {
+
+  /**
+   * Merge the given record to the aggregated collection. This function is guaranteed to get called at least once
+   * before {@link #mergeAggregates(Table, Table)} is called.
+   *
+   * @param agg the aggregated collection, the merged result should be added to it
+   * @param record the record to merge
+   */
+  void mergeValue(Table<String, String, Object> agg, StructuredRecord record);
```
albertshau: I'm confused about what is in this Table.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
albertshau: after reading some more, it seems like a proper object here would be better, would make it more clear what is happening instead of using special keys.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Avg.java
```diff
@@ -38,32 +43,45 @@ public Avg(String fieldName, Schema fieldSchema) {
...
+    computeAvg(agg1, agg2Count, (Number) val);
   }
 
-  @Override
-  public Double getAggregate() {
-    if (count == 0) {
-      // only happens if the field value was always null
-      return null;
+  private void computeAvg(Table<String, String, Object> table, long deltaCount, Number oldAvg) {
+    if (!table.contains(fieldName, AGG_KEY)) {
```
albertshau: ok so it looks like Table is used to prevent conflicts between fields in the group key and intermediate data?

It seems like it would be better to have a separate class that contains the group key as a field along with any additional fields that are required.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/AggregateFunction.java
albertshau: After reading some more it seems like it's used to carry intermediate data? Using Table as the intermediate data makes things harder to understand and maintain because it's not clear what is in the Table. 

It seems like it would be better for each individual function to be able to decide what reduced data it needs to keep track of, using an actual object instead of a Table or a Map with special keys.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Stddev.java
```diff
@@ -16,51 +16,50 @@
...
 
   public Stddev(String fieldName, Schema fieldSchema) {
-    this.fieldName = fieldName;
-    boolean isNullable = fieldSchema.isNullable();
-    Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
+    Schema.Type fieldType = fieldSchema.isNullable() ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
     if (!AggregationUtils.isNumericType(fieldType)) {
       throw new IllegalArgumentException(String.format(
-        "Cannot compute standard deviation on field %s because its type %s is not numeric", fieldName, fieldType));
+        "Cannot compute standard devisation on field %s because its type %s is not numeric", fieldName, fieldType));
```
albertshau: typo

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Variance.java
```diff
@@ -16,17 +16,28 @@
...
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.plugin.batch.aggregator.AggregationUtils;
 
+import javax.annotation.Nullable;
+
 /**
- * Calculates Variance
+ * Calculates Sample Variance
+ * TODO: CDAP-16911 figure out a way to compute population variance
```
albertshau: I thought the difference between population and sample variance is just the denominator, where one divides by n, and the other by n+1? If you're keeping track of the squared mean and count, it seems like it should be straightforward division?

yaojiefeng: Variance doesn't work in this way. The squared mean only makes sense for its own group since the mean and count are accumulated across the data. 
If you consider following code:
```
    count++;
    double delta = value.doubleValue() - mean;
    mean += delta / count;  // here the count only makes sense for individual group
    double delta2 = value.doubleValue() - mean;
    squareMean += delta * delta2;
```
For entire group, we cannot simply add up square mean as the new square mean because the actual square mean should be computed using the actual mean and count across the entire data.

albertshau: Isn't the population variance just the sample variance * (N-1)/N, where N is the count?

yaojiefeng: Ah, you are right, I thought it is complicated than this, will fix.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
yaojiefeng: If input schema is null, then this check cannot be performed, since the function will require a field schema to start with, passing a null to it just doesn't make sense.

yaojiefeng:
https://builds.cask.co/browse/HYP-BAD720

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Min.java
```diff
@@ -16,80 +16,54 @@
...
+    combine(otherAgg.getAggregate());
   }
 
-  @Override
-  protected void startDouble() {
-    minDouble = null;
-  }
-
-  @Override
-  protected void updateInt(int val) {
```
albertshau, line 57: I think these were originally done so that we don't do an extra comparison on type for every record. I don't think it actually matters much from a performance perspective, don't imagine this was a bottleneck in any way. Are we just changing it for readability?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
```diff
@@ -19,172 +19,57 @@
...
  */
 public abstract class NumberSelection implements SelectionFunction {
-  private final SelectionFunction delegate;
   private final String fieldName;
+  private final Schema.Type fieldType;
 
-  public NumberSelection(final String fieldName, @Nullable Schema fieldSchema) {
+  public NumberSelection(String fieldName, Schema fieldSchema) {
     this.fieldName = fieldName;
-    // if schema is not known before we start getting records, just use doubles.
```
albertshau, line 37: there used to be null handling before, how come it is being removed? Was that actually never supported?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
```diff
@@ -19,172 +19,57 @@
...
-    final boolean isNullable = fieldSchema.isNullable();
-    Schema.Type fieldType = isNullable ? fieldSchema.getNonNullable().getType() : fieldSchema.getType();
-    switch (fieldType) {
-      case INT:
-        delegate = new SelectionFunction() {
-          @Override
-          public List<StructuredRecord> getSelectedRecords() {
-            return getRecords();
-          }
+  protected abstract float combineFloat(float val1, float val2);
```
albertshau: these methods should be called 'compare' instead of 'combine', and they should all return an int.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Min.java
yaojiefeng: Yes, the original class is really hard to understand and it creates delegates to make sure the correct function is called. Since this is just a simple comparison, I want to remove it for better readability.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
yaojiefeng: By default it uses double to handle null, but a null schema scenario actually doesn't exist. If a field does not exist, the validation should fail. If a field does exist, then it has to have a field schema. Previously this was mainly used to validate and unit test that a null schema can fetch the result, so I removed this.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
yaojiefeng: good catch, the implementation is based on compare, I forgot to change this.

albertshau (approved):

PR #1108, opened 2020-06-08
CDAP-16875 changed joiner plugin to use new auto-join API

albertshau:
Changed the joiner to use the new AutoJoiner API to take
advantage of the performance improvements it allows in Spark
pipelines.

Introduced two new optional properties. The first determines
whether the join will use null safe equality, and the second
provides hint to the execution engine about which input datasets
should be broadcast to perform an in-memory join.

Removed much of the existing join logic because it has been moved
into the application code. Updated the documentation to use
the property names that show up in the UI, to display the properties
in the same order that they appear in the UI, and to include
descriptions of the two new properties.

### DIFF
--- a/core-plugins/docs/Joiner-batchjoiner.md
+++ b/core-plugins/docs/Joiner-batchjoiner.md
@@ -1,49 +1,90 @@
 # Joiner
 
-
 Description
 -----------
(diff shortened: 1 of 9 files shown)


albertshau:
cherry-pick of #1106

yaojiefeng (approved):

PR #1113, opened 2020-06-09
Feature/cdap cherry pick aggregator

yaojiefeng:
https://builds.cask.co/browse/HYP-BAD726-1
cherry pick #1107 
with differences:
1. bump the cdap version to 6.3.0-SNAPSHOT to match the cdap repo, 6.2.1-SNAPSHOT will not compile
2. support the newly added count distinct function


albertshau:
I shouldn't have merged my change without 6.2.1-snapshot getting published first. It's ok to bump to 6.3.0-snapshot for now, but we should change it back once the artifact is published. We should always be using the lowest needed api dependency version to allow plugins to be compatible with as many cdap versions as possible.

albertshau (approved):

elfenheart:
cdap 6.2.1-SNAPSHOT version bump is merged

yaojiefeng:
This is develop branch, which it should relate to the cdap develop branch right? The cdap develop branch now has 6.3.0-SNAPSHOT as the version

yaojiefeng:
The local build still failed for me. I will merge this pr with 6.3.0-SNAPSHOT. Also the develop branch of cdap is on 6.3.0-SNAPSHOT, should this match that?

PR #1117, opened 2020-06-10
Feature/update submodules

hsiaoching:


yaojiefeng:
I think you need to open the branch against release/2.3 instead of develop, the submodule pr should look like #983

PR #1122, opened 2020-06-15
Upgraded Hive dependency to 2.4.0

bdmogal:
Note: AvroGenericRecordReader is copied from https://github.com/apache/hive/blob/rel/release-2.3.7/ql/src/java/org/apache/hadoop/hive/ql/io/avro/AvroGenericRecordReader.java

At this point, not including the plugin with CDAP, but just making it work with Hive 2.x.

Tested with Hive Metastore:

image

### DIFF
--- a/hive-plugins/pom.xml
+++ b/hive-plugins/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  Copyright © 2015-2019 Cask Data, Inc.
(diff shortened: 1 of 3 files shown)


albertshau:
does this break when used with older hive instances? I'm wondering if it should be a different plugin for Hive2

bdmogal:
hmmm, just checked - all CDH 5.x versions (appear to) use Hive 1.x You're right then, perhaps we should make a separate plugin. I'll close this PR, and keep the branch, perhaps someone can pick it up later.

PR #1125, opened 2020-06-17
(PLUGIN-165) Added fetchSize property to DBSource

chtyim:
JDBC supports Statement.setFetchSize to provide JDBC driver hint on pre-fetching/buffering rows. Exposing this through the DB Source configuration would allow users fine tune the reading performance.


albertshau (approved):
lgtm. The db specific plugins should also be updated to support this, can you make sure that's tracked in a jira somewhere

database-plugins/widgets/Database-batchsource.json
```diff
@@ -80,6 +80,14 @@
           "widget-attributes" : {
              "default": "1"
            }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Fetch Size",
+          "name": "fetchSize",
+          "widget-attributes" : {
+            "default": "1000"
```
wyzhang, line 89: Do we know what's the behavior when each row is gigantic? OOM ? or JDBC is intelligent to stream data in piecemeal (i.e. smaller than fetch size)

chtyim: No, JDBC won't do that. The process can go OOM if that happen. That's why the fetch size and the executor resources need to be set accordingly, which is mentioned in the description of this attribute

chtyim:
For vendor specific DB source, this fetch size may not apply, since it's based on the exact mechanism being used by those vendor specific source.

PR #1130, opened 2020-06-29
Additional aggregate functions

ailegion:
added: GroupByAggregator - Concat function
added: GroupByAggregator - ConcatDistinct function
added: GroupByAggregator - CountNulls function
added: GroupByAggregator - LogicalAnd function
added: GroupByAggregator - LogicalOr function
added: GroupByAggregator - LongestString function
added: GroupByAggregator - ShortestString function
added: GroupByAggregator - CorrectedSumOfSquares function
added: GroupByAggregator - SumOfSquares function

Jira Task: https://issues.cask.co/browse/PLUGIN-166


ailegion:
@googlebot I signed it!

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -61,6 +61,17 @@
     put("VARIANCE", "Variance");
     put("COLLECTLIST", "CollectList");
     put("COLLECTSET", "CollectSet");
+    put("LATESTDATE", "LatestDate");
```
albertshau: why are latestdate and earliestdate needed? These seems like more specific versions of max and min, do max and min not work with dates for some reason?

ailegion: this is how it is specified in the requirements doc.

bdmogal: Will max and min work for dates as well? If that's the case we don't need to add more functions. But please test the scenario.

ailegion: @bdmogal @albertshau tested with dates and both min and max functions worked. Removed earliest and latest date from PR.

bdmogal: Yup that's fine.

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/function/SumOfSquaresTest.java
```diff
@@ -0,0 +1,36 @@
...
+ * the License.
+ */
+
+package io.cdap.plugin.batch.aggregator.function;
+
+import io.cdap.cdap.api.data.schema.Schema;
+import org.junit.Test;
+
+/**
+ *
```
CuriousVini: Please add javadoc

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Concat.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public void initialize() {
+    this.concatString = "";
+  }
+
+  @Override
+  public void mergeValue(StructuredRecord record) {
+    if (record.get(fieldName) != null) {
+      String value = Objects
+          .requireNonNull(record.get(fieldName)).toString();
```
CuriousVini: nit: this line can move to above line

CuriousVini: why do we check for `requireNonNull ` when we already have if condition?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Concat.java
```diff
@@ -0,0 +1,88 @@
...
+      if (firstString) {
```
CuriousVini, line 55: why only first string is being merged? This will happen in each task?

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/function/SumOfSquaresTest.java
ailegion: updated

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Concat.java
ailegion: updated

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Concat.java
ailegion: @CuriousVini this condition is only for the first string in order to avoid starting the string with the leading comma ex: "test1" instead of ", test1". The first statement after the if condition is used to merge the rest of the strings ```concatString = String.format("%s, %s", concatString, value);``` - line 62

CuriousVini: Will this method get executed by all the tasks in the framework? If so this will be applied to all the splits

CuriousVini:
Also please open the PR against develop branch

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Concat.java
ailegion: Correct, and split are also joined the same way - with String format - in which case also this check prevents situations like ``test1 ,, test2``.

ailegion:
> Also please open the PR against develop branch

new PR against develop branch: https://github.com/cdapio/hydrator-plugins/pull/1186

PR #1132, opened 2020-07-16
(CDAP-17080) Add Data Cacher Plugin

MEseifan:
https://issues.cask.co/browse/CDAP-17080


spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright © 2016-2019 Cask Data, Inc.
```
chtyim: Wrong year

spark-plugins/docs/DataCacher-sparkcompute.md
```diff
@@ -0,0 +1,27 @@
+# Data Cacher
```
chtyim, line 17: Missing copyright comment

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,76 @@
...
+public class DataCacher extends SparkCompute<StructuredRecord, StructuredRecord> {
+
+  private final DataCacherConfig config;
+
+  public DataCacher(DataCacherConfig config) {
+    this.config = config;
+  }
+
+  @Override
+  public JavaRDD<StructuredRecord> transform(
```
chtyim: A better line breaking:

```
public JavaRDD<StructuredRecord> transform(SparkExecutionPluginContext pluginContext,
                                           JavaRDD<StructuredRecord> javaRDD) throws Exception {
```

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,76 @@
...
+    SparkExecutionPluginContext sparkExecutionPluginContext, JavaRDD<StructuredRecord> javaRDD)
+    throws Exception {
+
+    StorageLevel storageLevel = config.getStorageLevel();
+    if (storageLevel == StorageLevel.NONE()) {
```
chtyim, line 67: This should be a validation error that should be checked at deployment time.

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public static class DataCacherConfig extends PluginConfig {
+
+    @Name("StorageLevel")
+    @Description("Spark storage level used to cache the data")
+    @Macro
+    private String storageLevel;
+
+    StorageLevel getStorageLevel() {
+      if (Strings.isNullOrEmpty(storageLevel)) {
```
chtyim: You can just check for empty string since the config is not optional (i.e. doesn't have `@Nullable` annotation)

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: Fixed

spark-plugins/docs/DataCacher-sparkcompute.md
MEseifan: Do we add copyright comments to markdown files? I dont see the copyright comment in any of the other plugin docs

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: done

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: We cant check at deployment time because the spark classes are not available during deployment/validation. I dont think this is a big problem because the plugin uses a dropdown to specify the storage level so the only way the user can pass an incorrect value is using a macro which we cant check during deployment anyway.

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: Fixed

spark-plugins/docs/DataCacher-sparkcompute.md
chtyim: I think we should have copyright in all files in the source repo

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
chtyim: Maybe check that the string value is in one of the allowed value.

MEseifan:
I manually tested this plugin using a pipeline (source -> data cacher->sink) with autocaching disabled. I was able to see that the RDDs were cached using the SparkUI

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: Sure, done.

spark-plugins/docs/DataCacher-sparkcompute.md
MEseifan: ok, added copyright to this doc and the rest of the docs

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,104 @@
...
+    @Description("Spark storage level used to cache the data")
+    @Macro
+    private String storageLevel;
+
+    public void validate(FailureCollector collector) {
+      if (containsMacro(STORAGE_LEVEL)) {
+        return;
+      }
+
+      String[] allowedStrings = {"DISK_ONLY", "DISK_ONLY_2", "MEMORY_ONLY", "MEMORY_ONLY_2", "MEMORY_ONLY_SER",
```
chtyim: Use a Set

```
Set<String> allowed = new HashSet<>(Arrays.asList(
  "DISK_ONLY",
  "DISK_ONLY_2",
  ...
));
```

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,104 @@
...
+        "MEMORY_ONLY_SER_2", "MEMORY_AND_DISK", "MEMORY_AND_DISK_2", "MEMORY_AND_DISK_SER", "MEMORY_AND_DISK_SER_2"};
+
+      if (storageLevel.isEmpty()) {
```
chtyim: I don't think this check is needed

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
```diff
@@ -0,0 +1,104 @@
...
+        collector.addFailure("Storage level is a required value. ", "Please select the desired storage level.")
+                 .withConfigProperty(STORAGE_LEVEL);
+      } else if (!Arrays.asList(allowedStrings).contains(storageLevel.toUpperCase())) {
+        collector
+          .addFailure("Invalid value for Storage Level ", "Please select one of the valid values from the dropdown.")
```
chtyim: Instead of saying dropdown, it would be better to have the list of allowed values in the suggestive action, since this API can be used in REST API (imagine when someone export and import a pipeline, and due to future version change, some of the value might no longer be valid).

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: done

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: removed

spark-plugins/src/main/java/io/cdap/plugin/spark/DataCacher.java
MEseifan: done

chtyim (approved):

spark-plugins/docs/DataCacher-sparkcompute.md
```diff
@@ -0,0 +1,42 @@
...
+ * the License.
+ -->
+ 
+# Data Cacher
+
+Description
+-----------
+Data cacher plugin will cache any record that is passed through it. This is 
+useful for pipelines that have auto-caching disabled but would still like to cache 
+records are certain points in the pipeline. Spark caching prevents unnecessary
```
albertshau: are -> at

spark-plugins/docs/File-streamingsource.md
```diff
@@ -1,3 +1,18 @@
+<!---
```
albertshau, line 1: do these not show up in the markdown doc?

I guess we haven't been putting copyrights in these doc files, don't know that they're really needed here though.

spark-plugins/docs/DataCacher-sparkcompute.md
MEseifan: whoops..good catch

spark-plugins/docs/File-streamingsource.md
MEseifan: I've confirmed that they dont show up. 

I'm not sure if its needed but it was requested in one of the previous review comments

PR #1144, opened 2020-07-23
(PLUGIN-303) Add distribution to Joiner Plugin

MEseifan:


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -144,25 +162,30 @@ public JoinDefinition define(AutoJoinerContext context) {
   public void prepareRun(BatchJoinerContext context) {
     if (conf.getNumPartitions() != null) {
       context.setNumPartitions(conf.getNumPartitions());
+      if (conf.getDistributionFactor() != null && conf.getDistributionFactor() < conf.getNumPartitions()) {
+        LOG.warn(String.format(
+          "Number of partitions (currently %d) should be greater than or equal to distribution factor (currently %d) "
```
albertshau: nit: remove the 'currently' in parentheses to shorten the message a little.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -67,6 +68,15 @@
     " Required input stages decide the type of the join. If all the input stages are present in required inputs, " +
     "inner join will be performed. Otherwise, outer join will be performed considering non-required inputs as " +
     "optional.";
+  private static final String DISTRIBUTION_STAGE_DESC =
+    "Name of the skewed input stage. The skewed input stage is the one that contains many rows that join "
+      + "to the same row in the non-skewed stage. Ex. If stage A has 10 rows that join on the same row in stage B, then"
+      + " stage A is the skewed input stage";
+  private static final String DISTRIBUTION_FACTOR_DESC =
+    "This controls the size of the salt that will be generated for distribution, the number of partitions "
```
albertshau: distribution. The number

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -104,6 +114,22 @@
...
+  private Integer distributionFactor;
+
+  @Nullable
+  @Macro
+  @Description(DISTRIBUTION_STAGE_DESC)
+  private String distributionStageName;
+
+  @Macro
+  @Nullable
+  @Description("Distribution is useful when the input data is skewed, enabling distribution will allow you to "
```
albertshau: skewed. Enabling will increase the parallelism

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -113,6 +116,21 @@ public JoinDefinition define(AutoJoinerContext context) {
...
+              .withConfigProperty("distributionFactor");
+          }
+          if (Strings.isNullOrEmpty(conf.getDistributionStageName())) {
+            collector.addFailure("Skewed Stage name is a required value if distribution is enabled.", "")
+              .withConfigProperty("distributionStageName");
+          }
+        }
+      }
       return joinBuilder.build();
     } catch (InvalidJoinException e) {
```
albertshau, line 123: if there is some error with the distribution values (ex: not an inner or left outer join), they should be handled here, with the proper property highlighted. This will require some more specific error in your other CDAP PR.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -113,6 +116,21 @@ public JoinDefinition define(AutoJoinerContext context) {
       } else {
         joinBuilder.setOutputSchemaName("join.output");
       }
+
+      if (conf.isDistributionEnabled()) {
```
albertshau: there's a lot of nesting here which makes it hard to read. Can you pull this out into a separate method?

albertshau (commented):
should update the markdown doc as well

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: Noted

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: Added new error types and handling for error types

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -74,6 +74,23 @@ Setting it to true can cause a large drop in performance if there are a lot of n
 **Number of Partitions:** Number of partitions to use when grouping fields. If not specified, the execution
 framework will decide on the number to use.
 
+**Distribution:** Enabling distribution will increase the level of parallelism when joining skewed data. A 
+skewed join is when multiple records from one stage (the skewed stage) join to a single record in
+another stage. Distribution is possible when the following conditions are met:
+1. There are exactly two input stages. 
+1. Broadcast is not enabled for either stage.
```
albertshau, line 81: not sure why the numbers here are a different color. Are they part of the file or being added by github? Should make sure it doesn't render as a bunch of 1.s

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -74,6 +74,23 @@ Setting it to true can cause a large drop in performance if there are a lot of n
...
+1. The skewed input is marked as `required`.
+
+
+Distribution requires two parameters:
+1. **Distribution Size:** This controls the size of the salt that will be generated for distribution. The **Number of Partitions** 
+property should be greater than or equal to this number for optimal results. A larger value will lead to more 
+parallelism but it will also grow the size of the non-skewed dataset by this factor.
+1. **Skewed Input Stage:**  Name of the skewed input stage. The skewed input stage is the one that contains many rows that join 
+to the same row in the non-skewed stage. Ex. If stage A has 10 rows that join on the same row in stage B, then
+stage A is the skewed input stage.
```
albertshau, line 91: Seeing how many issues we see with skew in user pipelines, I think it's worth having a separate section after the example section that talks about what skew is and how to handle it. 

We should start by defining what skew is, use an example to illustrate, and explain why it results in slowness. The first approach to deal with it is to load the smaller side of the join in memory. If that isn't possible, use the distribution properties.

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -74,6 +74,23 @@ Setting it to true can cause a large drop in performance if there are a lot of n
 **Number of Partitions:** Number of partitions to use when grouping fields. If not specified, the execution
 framework will decide on the number to use.
 
+**Distribution:** Enabling distribution will increase the level of parallelism when joining skewed data. A 
+skewed join is when multiple records from one stage (the skewed stage) join to a single record in
```
albertshau: skew is a little different than this. It is possible to have multiple records joining to a single record and still be unskewed. I think it's better to say something like:

A skewed join happens when a significant percentage of input records have the same key.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -238,4 +266,47 @@ boolean isNullSafe() {
...
+
+    //If distribution is disabled then no need to validate further
+    if (!containsMacro("distributionEnabled") && !distributionEnabled) {
+      return false;
+    }
+
+    if (!containsMacro(DISTRIBUTION_FACTOR) && distributionFactor == null) {
+      collector.addFailure("Distribution Size is a required value if distribution is enabled.", "")
+        .withConfigProperty(DISTRIBUTION_FACTOR);
+
```
albertshau: nit: extra newline

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -238,4 +266,47 @@ boolean isNullSafe() {
...
+        .withConfigProperty(DISTRIBUTION_STAGE);
+    }
+
+    // If there are still macro values then this config is not valid
+    if (distributionContainsMacro()) {
+      return false;
+    }
+
+    return startFailures == collector.getValidationFailures().size();
+
```
albertshau: nit: extra newline

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -134,6 +140,13 @@ public JoinDefinition define(AutoJoinerContext context) {
           case OUTPUT_SCHEMA:
             OutputSchemaError schemaError = (OutputSchemaError) error;
             failure.withOutputSchemaField(schemaError.getField());
+            break;
+          case DISTRIBUTION_SIZE:
+            failure.withConfigProperty(JoinerConfig.DISTRIBUTION_FACTOR);
+            break;
+          case DISTRIBUTION_STAGE:
```
albertshau, line 147: add one case for the broadcast error

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -134,6 +140,13 @@ public JoinDefinition define(AutoJoinerContext context) {
...
+            failure.withConfigProperty(JoinerConfig.DISTRIBUTION_STAGE);
+            break;
```
albertshau, line 149: this is an existing problem, but we should have a default case to make sure other types of errors (like the generic join error) are collected.

core-plugins/docs/Joiner-batchjoiner.md
MEseifan: I've confirmed it looks correct in the UI, probably just Github highlighting:

image

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
MEseifan: done

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: whoops, forgot to push my changes..its there now

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: The first line before the switch-case handles all the errors:
`ValidationFailure failure = collector.addFailure(error.getMessage(), error.getCorrectiveAction());`
Its just that the errors with types get modified to target a particular property. So the generic errors are captured, they just appear at the top of the plugin instead of underneath a given property

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -144,25 +160,30 @@ public JoinDefinition define(AutoJoinerContext context) {
   public void prepareRun(BatchJoinerContext context) {
     if (conf.getNumPartitions() != null) {
       context.setNumPartitions(conf.getNumPartitions());
+      if (conf.getDistributionFactor() != null && conf.getDistributionFactor() < conf.getNumPartitions()) {
+        LOG.warn(String.format(
+          "Number of partitions (%d) should be greater than or equal to distribution factor (currently %d) "
```
chtyim: Use please holder `{}` instead of `String.format` when using Logger.

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
albertshau: ah ok, makes sense

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
MEseifan: done

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -74,6 +74,95 @@ Setting it to true can cause a large drop in performance if there are a lot of n
...
+
+Skew
+----------
+### Problem
+Data skew is an important characteristic to consider when implementing joins. A skewed join happens
+when a significant percentage of input records in one dataset have the same key and therefore join to
+the same record in the second dataset. This is problematic due to the way the execution framework handles
+joins. At a high level, all records with matching keys are grouped into a partition, these partitions 
+are distributed across the nodes in a cluster to perform the join operation. In a skewed join, one 
+or more of these partitions will be significantly larger than the rest which will result in a majority
```
albertshau: nit: rest, which

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -74,6 +74,95 @@ Setting it to true can cause a large drop in performance if there are a lot of n
...
+does not support In-Memory joins.*
+
+The first approach for increasing performance of skewed joins is using an In-Memory join. An in-memory
+join is a performance improvement when a large dataset is being joined with a small dataset. In this
+approach, the small dataset is loaded into memory and broadcast to workers and loaded into 
+workers memory. Once it is in memory, a join is performed by iterating through the elements of the
+large dataset. With this approach, data from the large dataset is NEVER shuffled. Data with the 
+same key can be joined in parallel across the cluster instead of handled only by a single worker 
+providing optimal performance. Data sets that have a small size can be used for in-memory joins. 
+Make sure the total size of broadcast across and join does not exceed **2GB.**
```
albertshau: total size of broadcast data does not exceed

albertshau (approved):

core-plugins/docs/Joiner-batchjoiner.md
MEseifan: done

core-plugins/docs/Joiner-batchjoiner.md
MEseifan: done

core-plugins/docs/Joiner-batchjoiner.md
MEseifan: done

core-plugins/docs/Joiner-batchjoiner.md
MEseifan: done

PR #1145, opened 2020-07-25
Exclude logbook from Cassandra dependency

chtyim:


chtyim:
Tested in Bamboo that all tests passed https://builds.cask.co/browse/CDAP-BUT889-RCU-7

dli357 (approved):

PR #1154, opened 2020-07-31
[PLUGIN-367] Blobs should not be splitable

dli357:
See https://issues.cask.co/browse/PLUGIN-367.

Should we also consider adding a logging info message to AbstractFileSource which tells the user that the file will not be split?


albertshau (commented):
fix lgtm, though can we also add a unit test for this case?

albertshau (approved):

PR #1157, opened 2020-08-04
Enable subclasses to override OutputFormat, output path implementation

greeshmaswaminathan:
Fix for https://issues.cask.co/browse/CDAP-17123
- The change in AbstractFileSink is required to override OutputFormat implementation for GCSSink to emit metric. PR for reference - https://github.com/data-integrations/google-cloud/pull/267 
- Fixed a bug in which the outputdir was formatting as gs://\<bucket>/path//\<timeSuffix> instead of  gs://\<bucket>/path/\<timeSuffix>

### DIFF
--- a/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
@@ -88,11 +88,15 @@ public abstract class AbstractFileSink<T extends PluginConfig & FileSinkProperti
     }
 
     FileFormat format = config.getFormat();
-    ValidatingOutputFormat validatingOutputFormat =
-      pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, format.name().toLowerCase(),
-                                   format.name().toLowerCase(), config.getRawProperties());
+    ValidatingOutputFormat validatingOutputFormat = getValidatingOutputFormat(pipelineConfigurer);
     FormatContext context = new FormatContext(collector, pipelineConfigurer.getStageConfigurer().getInputSchema());
     validateOutputFormatProvider(context, format, validatingOutputFormat);
+  }
+
+  protected ValidatingOutputFormat getValidatingOutputFormat(PipelineConfigurer pipelineConfigurer) {
+    FileFormat format = config.getFormat();
+    return pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, format.name().toLowerCase(),
+                                        format.name().toLowerCase(), config.getRawProperties());
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -88,32 +88,24 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-                                   format.name().toLowerCase(), config.getRawProperties());
+    ValidatingOutputFormat validatingOutputFormat = getValidatingOutputFormat(pipelineConfigurer);
     FormatContext context = new FormatContext(collector, pipelineConfigurer.getStageConfigurer().getInputSchema());
     validateOutputFormatProvider(context, format, validatingOutputFormat);
   }
 
+  public ValidatingOutputFormat getValidatingOutputFormat(PipelineConfigurer pipelineConfigurer) {
+    FileFormat format = config.getFormat();
+    return
+        pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, format.name().toLowerCase(),
```
albertshau: style: should have something on the previous line besides just return. Also, indentation for a new line should be 2 spaces and not 4.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -135,12 +127,28 @@ public void prepareRun(BatchSinkContext context) throws Exception {
     Map<String, String> outputProperties = new HashMap<>(validatingOutputFormat.getOutputFormatConfiguration());
     outputProperties.putAll(getFileSystemProperties(context));
     outputProperties.put(FileOutputFormat.OUTDIR, getOutputDir(context.getLogicalStartTime()));
-
     context.addOutput(Output.of(config.getReferenceName(),
                                 new SinkOutputFormatProvider(validatingOutputFormat.getOutputFormatClassName(),
                                                              outputProperties)));
   }
 
+  public ValidatingOutputFormat getOutputFormatForRun(BatchSinkContext context) throws InstantiationException {
```
albertshau: I assume this is pulled into its own method so subclasses can use it? If that's the case, can this be protected instead of public?

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -88,32 +88,24 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
     FileFormat format = config.getFormat();
-    ValidatingOutputFormat validatingOutputFormat =
-      pipelineConfigurer.usePlugin(ValidatingOutputFormat.PLUGIN_TYPE, format.name().toLowerCase(),
-                                   format.name().toLowerCase(), config.getRawProperties());
+    ValidatingOutputFormat validatingOutputFormat = getValidatingOutputFormat(pipelineConfigurer);
     FormatContext context = new FormatContext(collector, pipelineConfigurer.getStageConfigurer().getInputSchema());
     validateOutputFormatProvider(context, format, validatingOutputFormat);
   }
 
+  public ValidatingOutputFormat getValidatingOutputFormat(PipelineConfigurer pipelineConfigurer) {
```
albertshau: can this be protected instead of public?

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
greeshmaswaminathan: Fixed.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
greeshmaswaminathan: Yes, protected is enough. Fixed.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
greeshmaswaminathan: Done.

albertshau (approved):

PR #1178, opened 2020-08-17
Added: Logical types time, timestamp, decimal and date.

vjosadika:
Jira ticket: https://issues.cask.co/browse/PLUGIN-246

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
@@ -31,11 +31,10 @@ import io.cdap.cdap.etl.api.MultiOutputStageConfigurer;
 import io.cdap.cdap.etl.api.SplitterTransform;
 
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.text.DecimalFormat;
+import java.util.*;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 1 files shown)


transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -31,11 +31,10 @@
...
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.text.DecimalFormat;
+import java.util.*;
```
chtyim, line 37: Don't use wildcard import. Import the needed classes one by one. You can adjust your IDE settings to do it automatically.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -104,28 +103,36 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
       valSchema = Schema.of(Schema.Type.DOUBLE);
     } else if (val instanceof String) {
       valSchema = Schema.of(Schema.Type.STRING);
-    } else if (val instanceof StructuredRecord) {
+    }else if (val instanceof Time) {
```
chtyim, line 106: missing a space after `}`

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -104,28 +103,36 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
+      valSchema = Schema.of(Schema.LogicalType.TIMESTAMP_MILLIS);
+    }else if (val instanceof Date) {
+      valSchema = Schema.of(Schema.LogicalType.DATE);
+    }else if (val instanceof StructuredRecord) {
       valSchema = ((StructuredRecord) val).getSchema();
     } else if (val.getClass().isEnum()) {
       emitter.emitError(
-        new InvalidEntry<>(300, String.format("Field '%s' is an Enum, which is not supported.", conf.unionField),
-                           record));
+              new InvalidEntry<>(300, String.format("Field '%s' is an Enum, which is not supported.", conf.unionField),
```
chtyim, line 118: No need to change the indentation. Please keep the old one.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
chtyim: Checking the instance of the `val` won't work. This is because calling `Object val = record.get(conf.unionField);` would return the numeric value, instead of the logical type class. You should use the schema type of the union field in the incoming `record` as the schema of the output value.

vjosadika:
Created new PR against develop branch: https://github.com/cdapio/hydrator-plugins/pull/1206

PR #1187, opened 2020-08-19
Feature/flatten transform plugin

ardian4:
Added: FlattenTransform plugin

JIRA Task: https://issues.cask.co/browse/PLUGIN-378

### DIFF
--- a/transform-plugins/docs/Flatten-transform.md
+++ b/transform-plugins/docs/Flatten-transform.md
@@ -21,17 +21,20 @@ For example the input record is as follows:
 
 ```
 Address
-    - Street 
+    - Street  
     - City 
+        - Name
+        - Code
     - State
     - Zip 
 ```
 
 The output schema will be specified as:
 
-| Field            |
-| ---------------- | 
-| Address_Street   | 
-| Address_City     | 
-| Address_State    | 
-| Address_Zip      | 
+| Field             |
+| ----------------  |  
+| Address_Street    | 
+| Address_City_Name |
+| Address_City_Code |
+| Address_State     | 
+| Address_Zip       | 
--- a/transform-plugins/docs/Flatten-transform.md
+++ b/transform-plugins/docs/Flatten-transform.md
@@ -1,0 +1,37 @@
+# Flatten Transform
+
+## Description
+Flatten is a transform plugin that flattens nested data structures.
+
+## Use Case
+
+The transform is used  to convert a nested data structure  into a single flattened record where each key in the record 
+is a ```_``` name path to the node in the nested structure. 
+
+Configuration
+-------------
+**Fields to flatten:** Specifies the list of fields in the input schema to be flattened.
+
+**Level to limit flattening:** Limit flattening to a certain level in nested structures.
+
+**Prefix:** An optional prefix to be used while generating names of the flattened fields. This can be used to fix conflicts that can occur if fields have the same name after flattening. The prefix is added as <prefix>_<flattened_name >.
+
+## Example
+For example the input record is as follows:
+
+```
+Address
+    - Street 
+    - City 
+    - State
+    - Zip 
+```
+
+The output schema will be specified as:
+
+| Field            |
+| ---------------- | 
+| Address_Street   | 
+| Address_City     | 
+| Address_State    | 
+| Address_Zip      | 
--- a/transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
@@ -54,6 +54,10 @@ public final class Flatten extends Transform<StructuredRecord, StructuredRecord>
   private Schema outputSchema;
   private Map<String, OutputFieldInfo> inputOutputMapping = Maps.newHashMap();
 
+  public Flatten(Config config) {
+    this.config = config;
+  }
+
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
@@ -100,8 +104,7 @@ public final class Flatten extends Transform<StructuredRecord, StructuredRecord>
 
     List<String> fieldsToFlatten = config.getFieldsToFlatten();
     if (fieldsToFlatten == null || fieldsToFlatten.isEmpty()) {
-      failureCollector.addFailure("No field(s) selected for flattening.",
-                                  "Choose field(s) to flatten.");
+      failureCollector.addFailure("Atleast one field should be selected.", "");
       failureCollector.getOrThrowException();
     }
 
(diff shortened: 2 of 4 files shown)


transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+    Map<String, OutputFieldInfo> outputFieldMap = handleDuplicateFieldName(inputOutputMapping, config.prefix);
+    Schema outputSchema = generateOutputSchema(inputSchema, outputFieldMap.values());
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(outputSchema);
+  }
+
+  private void checkSchemaType(Schema inputSchema, List<String> fieldsToFlatten, FailureCollector collector) {
+    for (String field : fieldsToFlatten) {
+      Schema schemaField = inputSchema.getField(field).getSchema();
+      if (!isRecord(schemaField)) {
+        collector.addFailure(String.format("'%s' cannot be flattened.", field),
```
CuriousVini, line 89: @bdmogal Is this an acceptable constraint from product standpoint?

transform-plugins/docs/Flatten-transform.md
```diff
@@ -0,0 +1,37 @@
...
+## Use Case
+
+The transform is used  to convert a nested data structure  into a single flattened record where each key in the record 
+is a ```_``` name path to the node in the nested structure. 
+
+Configuration
+-------------
+**Fields to flatten:** Specifies the list of fields in the input schema to be flattened.
+
+**Level to limit flattening:** Limit flattening to a certain level in nested structures.
```
CuriousVini, line 15: why is this needed? Cant we just rely on structured record to figure out nesting levels?

transform-plugins/docs/Flatten-transform.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Prefix:** An optional prefix to be used while generating names of the flattened fields. This can be used to fix conflicts that can occur if fields have the same name after flattening. The prefix is added as <prefix>_<flattened_name >.
```
CuriousVini, line 17: What happens if there are records of record? What prefix will be used?

ardian4: @CuriousVini The prefix is used only if after flattening there are fields with same name, the prefix will be added to duplicate.

example:

` {
   "Address_Number": 100,
   "Address":{
   		"Number":100
   }
 }`

The output fields will be:
Address_Number, Prefix_Address_Number

transform-plugins/docs/Flatten-transform.md
ardian4: @CuriousVini 
As for the requirments we can restricts nesting to certain level, where the default is one level.

CuriousVini: I see

transform-plugins/docs/Flatten-transform.md
CuriousVini: What would happen if there are 3 record fields in a schema with `Number` field?

CuriousVini (commented):
What is the maximum nesting supported by the plugin. How does it perform when there is deeply nested schemas?

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+  @Override
+  public void initialize(TransformContext context) throws Exception {
+    super.initialize(context);
+
+    FailureCollector failureCollector = context.getFailureCollector();
+
+    List<String> fieldsToFlatten = config.getFieldsToFlatten();
+    if (fieldsToFlatten == null || fieldsToFlatten.isEmpty()) {
+      failureCollector.addFailure("No field(s) selected for flattening.",
+                                  "Choose field(s) to flatten.");
```
CuriousVini: "Atleast one field should be selected"

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+      .collect(Collectors.toList());
+    return Schema.recordOf(inputSchema.getRecordName() + ".flatten", fields);
+  }
+
+  private boolean isRecord(Schema schema) {
+    if (schema == null) {
+      return false;
+    }
+    Schema.Type type = schema.getType();
+    if (type == Schema.Type.RECORD) {
```
CuriousVini: we can refactor this if and the one below into more compact code by checking nonnullable using ternary operator

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+
+    @Macro
+    @Name(PROPERTY_NAME_FIELDS_TO_MAP)
+    @Description("Specifies the list of fields in the input schema to be flattened.")
+    private String fieldsToFlatten;
+
+    @Macro
+    @Nullable
+    @Name(PROPERTY_NAME_LEVEL_TO_LIMIT)
+    @Description("Limit flattening to a certain level in nested structures.")
```
CuriousVini: We should add what is the default value for this

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+    private String levelToLimitFlattening;
+
+    @Macro
+    @Nullable
+    @Name(PROPERTY_NAME_PREFIX)
+    @Description("An optional prefix to be used while generating names of the flattened fields. " +
```
CuriousVini, line 323: what would happen if there are duplicates and this optional property is not provided?

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+
+  private List<OutputFieldInfo> flattenField(Schema.Field field, int levelToLimitFlattening) {
+    Schema schema = field.getSchema();
+
+    if (isRecord(schema) && levelToLimitFlattening != 0) {
+      return generateOutputFieldInfoForRecord(field, levelToLimitFlattening);
+    }
+
+    ArrayList<OutputFieldInfo> maps = new ArrayList<>();
+    maps.add(generateOutputFieldInfo(field));
```
CuriousVini, line 168: Is this logic correct? we would still add the field to the map, if schema is not a record?

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+    return outputFieldInfo;
+  }
+
+  private List<OutputFieldInfo> generateOutputFieldInfoForRecord(Schema.Field field, int levelToLimitFlattening) {
+
+    List<OutputFieldInfo> outputFieldInfos = new ArrayList<>();
+    Schema schema = field.getSchema();
+    List<Schema.Field> fields;
+
+    if (schema.isNullable()) {
```
CuriousVini: This can be single line using ternary operator

transform-plugins/docs/Flatten-transform.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Example
+For example the input record is as follows:
+
+```
+Address
+    - Street
```
CuriousVini: we should add nested example here with 3 level nesting

CuriousVini (commented):
Is there any specific reason for adding this transform to hydrator plugins and not in data integrations repository?

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
```diff
@@ -0,0 +1,393 @@
...
+    node.fieldName = field.getName();
+    node.fieldSchema = fieldSchema;
+
+    OutputFieldInfo outputFieldInfo = new OutputFieldInfo();
+    outputFieldInfo.fieldName = field.getName();
+    outputFieldInfo.node = node;
+    return outputFieldInfo;
+  }
+
+  private List<OutputFieldInfo> generateOutputFieldInfoForRecord(Schema.Field field, int levelToLimitFlattening) {
```
CuriousVini, line 249: The method is using DFS to flatten the schema. could you add a javadoc with a nested schema example on how it will be flatten

CuriousVini (commented):
Also could you please add unit tests for the plugin and the config?

transform-plugins/docs/Flatten-transform.md
ardian4: If there are more than one duplicate the value of last duplicate will be mapped to last prefix.
 In order to prevent this scenario we can do one of the following:
      1. Raise Exception during schema generation 
      2. Add counter to prefix: prefix_1_field, prefix_2_field ....

ardian4:
> What is the maximum nesting supported by the plugin. How does it perform when there is deeply nested schemas?

Currently we have not set any limition here. Regarding how will it react with many nesting levels - the schema generation process will run only once, so it should not have any heavy impact. However we haven't colleted any metrics regarding performance for multiple depth levels. Please let us know if there is a need to test with different nesting levels.

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
ardian4: The logic is correct, since regardles if the schema is record or not it would still be mapped.

transform-plugins/src/main/java/io/cdap/plugin/Flatten.java
ardian4: The value of duplicate will be taken, and the value of the first field will be ignored.
We can raise exception in this scenario.

ardian4:
> Is there any specific reason for adding this transform to hydrator plugins and not in data integrations repository?

Since other transform plugins were in hydrator plugins we added this one to hydrator as well. If there is a need to move it to another repo please let us know to which repo to move it.

CuriousVini:
> > What is the maximum nesting supported by the plugin. How does it perform when there is deeply nested schemas?
> 
> Currently we have not set any limition here. Regarding how will it react with many nesting levels - the schema generation process will run only once, so it should not have any heavy impact. However we haven't colleted any metrics regarding performance for multiple depth levels. Please let us know if there is a need to test with different nesting levels.

We should test this for nested schema. It is one of the requirements in the PRD:
```
Performance (includes concurrency and scalability)
There should be performance bounds defined for validate action after plugin configuration is defined. 
The delivery should include bounds for the maximum level that the plugin can effectively support.

CuriousVini:
Also please close this PR open a new one against a repository under `data-integrations`

transform-plugins/docs/Flatten-transform.md
CuriousVini: For duplicate columns, This is the behavior we should have:
1. You get a validation error when validate button is pressed.
2. The error has a clear call to action. "Rename x, y, or z fields before using the flatten transform"

ardian4:
> Also please close this PR open a new one against a repository under `data-integrations`
@CuriousVini 
new PR: https://github.com/data-integrations/hierarchy-plugins/pull/2

ardian4:
> > > What is the maximum nesting supported by the plugin. How does it perform when there is deeply nested schemas?
> > 
> > 
> > Currently we have not set any limition here. Regarding how will it react with many nesting levels - the schema generation process will run only once, so it should not have any heavy impact. However we haven't colleted any metrics regarding performance for multiple depth levels. Please let us know if there is a need to test with different nesting levels.
> 
> We should test this for nested schema. It is one of the requirements in the PRD:
> 
> ```
> Performance (includes concurrency and scalability)
> There should be performance bounds defined for validate action after plugin configuration is defined. 
> The delivery should include bounds for the maximum level that the plugin can effectively support.
> ```

While testing with deeply nested Json we got those results:

Nested Levels  |  Nested Limit in config  | results
40                    50                        Works
50               35                        Works
50               50 					   Works
200              100 					   Works
1000             300 					   Works
1000             500   				       Works
1000             900   				       Works
1000             1050  				       Works
2000             2000   				   exception Stackoverflow
3000             1000  				       exception Stackoverflow
3000             2000                      exception Stackoverflow 
3000             500 					   Works
3000             800                       Works
3000             900 					   exception Stackoverflow

also we added restriction max level of nested json to be 100.

ardian4:
new PR:  data-integrations/hierarchy-plugins#2

ardian4:
new PR: data-integrations/hierarchy-plugins#2

PR #1206, opened 2020-08-24
Supported Logical types

vjosadika:
JIRA ticket: https://issues.cask.co/browse/PLUGIN-246

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
@@ -94,19 +94,19 @@ public class UnionSplitter extends SplitterTransform<StructuredRecord, Structure
     } else if (val instanceof Boolean) {
       valSchema = Schema.of(Schema.Type.BOOLEAN);
     } else if (val instanceof ByteBuffer || val instanceof byte[] || val instanceof Byte[]) {
-      if(logicalType!=null){
+      if (logicalType!=null) {
         valSchema = Schema.of(logicalType);
       } else {
         valSchema = Schema.of(Schema.Type.BYTES);
       }
     } else if (val instanceof Integer) {
-      if(logicalType!=null){
+      if (logicalType!=null) {
         valSchema = Schema.of(logicalType);
       } else {
         valSchema = Schema.of(Schema.Type.INT);
       }
     } else if (val instanceof Long) {
-      if(logicalType!=null){
+      if (logicalType!=null) {
         valSchema = Schema.of(logicalType);
       } else {
         valSchema = Schema.of(Schema.Type.LONG);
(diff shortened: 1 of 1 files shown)


transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -87,17 +87,30 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
     Object val = record.get(conf.unionField);
+    Schema.LogicalType logicalType = fieldSchema.getNonNullable().getLogicalType();
     Schema valSchema;
     if (val == null) {
       valSchema = Schema.of(Schema.Type.NULL);
     } else if (val instanceof Boolean) {
       valSchema = Schema.of(Schema.Type.BOOLEAN);
     } else if (val instanceof ByteBuffer || val instanceof byte[] || val instanceof Byte[]) {
-      valSchema = Schema.of(Schema.Type.BYTES);
+      if(logicalType!=null){
```
chtyim: Please fix al the spacing. If statement should be spaced like this:

```
if (condition) {
  statement;
}
```

vjosadika: Fixed

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -87,17 +87,30 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
     }
 
     Object val = record.get(conf.unionField);
+    Schema.LogicalType logicalType = fieldSchema.getNonNullable().getLogicalType();
```
chtyim: You should check for `fieldSchema.isNullable()` before calling `getNonNullable()`. Otherwise you'll result in NPE.

transform-plugins/src/main/java/io/cdap/plugin/UnionSplitter.java
```diff
@@ -87,17 +87,30 @@ public void transform(StructuredRecord record, MultiOutputEmitter<StructuredReco
...
     Schema valSchema;
     if (val == null) {
       valSchema = Schema.of(Schema.Type.NULL);
     } else if (val instanceof Boolean) {
       valSchema = Schema.of(Schema.Type.BOOLEAN);
     } else if (val instanceof ByteBuffer || val instanceof byte[] || val instanceof Byte[]) {
-      valSchema = Schema.of(Schema.Type.BYTES);
+      if (logicalType!=null) {
```
chtyim: Please fix all spacing. It should be `logicalType != null`

vjosadika: Fixed

CuriousVini (commented):
Has this changes been tested?  Do we need to make changes to line 230? 
`outputFields.set(unionIndex, Schema.Field.of(unionField, modifySchema ? schema : unionSchema));`

PR #1208, opened 2020-08-28
Fixed: decimal type handled correctly

vjosadika:
JIRA ticket: https://issues.cask.co/browse/PLUGIN-180

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
@@ -232,12 +232,13 @@ public final class DBUtils {
         break;
 
       case Types.NUMERIC:
-      case Types.DECIMAL:
-        // if there are no digits after the point, use integer types
+//         if there are no digits after the point, use integer types
         type = scale != 0 ? Schema.Type.DOUBLE :
           // with 10 digits we can represent 2^32 and LONG is required
           precision > 9 ? Schema.Type.LONG : Schema.Type.INT;
         break;
+      case Types.DECIMAL:
+        return Schema.decimalOf(precision, scale);
 
       case Types.DOUBLE:
         type = Schema.Type.DOUBLE;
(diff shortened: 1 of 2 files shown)


database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -233,11 +233,7 @@ private static Schema getSchema(String typeName, int sqlType, int precision, int
 
       case Types.NUMERIC:
       case Types.DECIMAL:
-        // if there are no digits after the point, use integer types
-        type = scale != 0 ? Schema.Type.DOUBLE :
-          // with 10 digits we can represent 2^32 and LONG is required
-          precision > 9 ? Schema.Type.LONG : Schema.Type.INT;
-        break;
+        return  Schema.decimalOf(2,scale);
```
CuriousVini: why the precision is 2?

CuriousVini: Has this been tested with any database? Please add a unit test

CuriousVini (changes requested):

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -284,15 +280,7 @@ public static Object transformValue(int sqlType, int precision, int scale,
...
-          if (scale != 0) {
-            // if there are digits after the point, use double types
-            return decimal.doubleValue();
-          } else if (precision > 9) {
-            // with 10 digits we can represent 2^32 and LONG is required
-            return decimal.longValue();
-          } else {
-            return decimal.intValue();
-          }
+          return decimal.unscaledValue().toByteArray();
```
CuriousVini: Please move this logic here: 
https://github.com/cdapio/hydrator-plugins/blob/develop/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java#L137

CuriousVini (approved):
fix and squash commits

PR #1216, opened 2020-09-16
(Plugin-388) Fail Parquet output format validation if schema definition is null or empty.

supriyapatil-google:


CuriousVini (commented):
Now that I took a closer look at the issue, it happens in CDF environment but not in sandbox. The reason being in CDF environment UI is sending schema as empty string.
I think what we need in the fix is to check if schema is an empty or null string then throw an error. Does this code change fixes issue in cdf environment?

image

image

format-parquet/pom.xml
```diff
@@ -79,6 +79,10 @@
       <artifactId>jackson-mapper-asl</artifactId>
       <version>1.9.13</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
```
CuriousVini, line 83: Dont think there is need to add this dependency explicitly.

CuriousVini (approved):
one minor comment. LGTM 👍

supriyapatil-google:
Sample validation works in snapshot environment with my changes.

format-parquet/pom.xml
supriyapatil-google: I do see this dependency explicitly specified in other plugin packages like (database-plugins, http-plugins, core-plugins) so its best to follow the same pattern.

CuriousVini: sounds good

CuriousVini (approved):
lgtm 👍  Please squash all the commits into one before merging.

supriyapatil-google:
1. Added unit tests in FileSinkTestRun to validate empty and null value output schema values.
2. Overridden ParquetOutputFormatProvider#validate(FormatContext) to validate schemas as part of plugin validation.
3. Verified parquet plugin validation in Sandbox CDAP environment.

Please take another look. Thanks!

CuriousVini (approved):
lgtm 👍

PR #1226, opened 2020-10-07
feature/PLUGIN-353-Schema-autodetection

ailegion:
added: Schema auto-detection support for file source plugins

JIRA: https://issues.cask.co/browse/PLUGIN-353

Related to PR: https://github.com/cdapio/hydrator-plugins/pull/1203

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -20,6 +20,12 @@ If the format is 'blob', every input file will be read into a separate record.
 The 'blob' format also requires a schema that contains a field named 'body' of type 'bytes'.
 If the format is 'text', the schema must contain a field named 'body' of type 'string'.
 
+**Get Schema:** Auto-detects schema from file. Supported formats are: avro, parquet, csv, delimited, tsv, blob 
+and text.
+Blob - is set by default as field named 'body' of type bytes.
+Text - is set by default as two fields: 'body' of type bytes and 'offset' of type 'long'.
+JSON - is not supported, user has to manually provide the output schema.
+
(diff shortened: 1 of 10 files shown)


core-plugins/docs/File-batchsource.md
```diff
@@ -20,6 +20,11 @@ If the format is 'blob', every input file will be read into a separate record.
 The 'blob' format also requires a schema that contains a field named 'body' of type 'bytes'.
 If the format is 'text', the schema must contain a field named 'body' of type 'string'.
 
+**Get Schema:** Auto-detects schema from file. Supported formats are: avro, parquet, csv, delimited,tsv.
```
CuriousVini: `Supported formats are: avro, parquet, csv, delimited,tsv.` - It also supports blob and text right?

CuriousVini (commented):
One comment on doc. Also why was the file test failing?

core-plugins/docs/File-batchsource.md
ailegion: fixed

ailegion:
> One comment on doc. Also why was the file test failing?

@CuriousVini the test was failing because it was designed to test null schema scenario and with this PR when schema is null we try to generate it from file. However the test was designed to provide only the file name for validation but not actually generate the file before the final assert. This caused 'File not found exception' during schema detection. We moved the file generation before the validation in order to make file available for schema detection.

CuriousVini (approved):

CuriousVini:
Please squash the commits

ailegion:
> Please squash the commits

squashed

PR #1250, opened 2020-11-18
Bug fix for plugin 400 

flakrimjusufi:
FTP Batch Source - Error supporting complex Passwords in Path field

JIRA TICKET: https://cdap.atlassian.net/browse/PLUGIN-400

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
@@ -30,11 +30,14 @@ import io.cdap.cdap.etl.api.batch.BatchSourceContext;
 import io.cdap.plugin.format.FileFormat;
 import io.cdap.plugin.format.plugin.AbstractFileSource;
 import io.cdap.plugin.format.plugin.FileSourceProperties;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 
 import java.lang.reflect.Type;
+import java.net.URLEncoder;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
@@ -49,6 +52,10 @@ public class FTPBatchSource extends AbstractFileSource {
   private static final String NAME_FILE_SYSTEM_PROPERTIES = "fileSystemProperties";
   private static final String FS_SFTP_IMPL = "fs.sftp.impl";
   private static final String SFTP_FS_CLASS = "org.apache.hadoop.fs.sftp.SFTPFileSystem";
+  private static final String FTP_PROTOCOL = "ftp";
+  private static final String SFTP_PROTOCOL = "sftp";
+  private static final int DEFAULT_FTP_PORT = 21;
+  private static final int DEFAULT_SFTP_PORT = 22;
 
   public static final Schema SCHEMA = Schema.recordOf("text",
                                                       Schema.Field.of("offset", Schema.of(Schema.Type.LONG)),
@@ -78,7 +85,8 @@ public class FTPBatchSource extends AbstractFileSource {
   @SuppressWarnings("unused")
   public static class FTPBatchSourceConfig extends PluginConfig implements FileSourceProperties {
     private static final Gson GSON = new Gson();
-    private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() { }.getType();
+    private static final Type MAP_STRING_STRING_TYPE = new TypeToken<Map<String, String>>() {
+    }.getType();
 
     @Macro
     @Description("Name be used to uniquely identify this source for lineage, annotating metadata, etc.")
@@ -130,6 +138,29 @@ public class FTPBatchSource extends AbstractFileSource {
 
     @Override
     public String getPath() {
+      if (authContainsSpecialCharacters()) {
+        Path urlInfo;
+        String extractedPassword = extractPasswordFromUrl();
+        String encodedPassword = URLEncoder.encode(extractedPassword);
+        String validatePath = path.replace(extractedPassword, encodedPassword);
+        try {
+          urlInfo = new Path(validatePath);
+        } catch (Exception e) {
+          throw new IllegalArgumentException(String.format("Unable to parse url: %s %s", e.getMessage(), e));
+        }
+        String host = urlInfo.toUri().getAuthority().substring(urlInfo.toUri().getAuthority().lastIndexOf("@") + 1);
+        String user = urlInfo.toUri().getAuthority().split(":")[0];
+        String protocol = urlInfo.toUri().getScheme();
+        int port = urlInfo.toUri().getPort();
+        if (port == -1 && protocol.equals(FTP_PROTOCOL)) {
+          port = DEFAULT_FTP_PORT;
+        }
+        if (port == -1 && protocol.equals(SFTP_PROTOCOL)) {
+          port = DEFAULT_SFTP_PORT;
+        }
+        String cleanHost = host.replace(":" + port, "");
+        return urlInfo.toUri().getScheme() + "://" + cleanHost + urlInfo.toUri().getPath();
+      }
       return path;
     }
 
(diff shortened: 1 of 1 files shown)


flakrimjusufi:
@googlebot I signed it!

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+
+    public boolean authContainsSpecialCharacters() {
+      Pattern regularPasswordWithoutSpecialCharacters = Pattern.compile("[^A-Za-z0-9]");
+      Matcher regularPassword = regularPasswordWithoutSpecialCharacters.matcher(extractPasswordFromUrl());
+      return !regularPassword.matches();
+    }
+
     Map<String, String> getFileSystemProperties() {
-      if (fileSystemProperties == null) {
+      if (fileSystemProperties == null && !authContainsSpecialCharacters()) {
```
rmstar: No need for this if condition. if fileSystemProperties == null && !authContainsSpecialCharacters(), we'll return empty hashmap at the end of the function.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+          Path urlInfo;
+          String extractedPassword = extractPasswordFromUrl();
+          String encodedPassword = URLEncoder.encode(extractedPassword);
+          String validatePath = path.replace(extractedPassword, encodedPassword);
+          try {
+            urlInfo = new Path(validatePath);
+          } catch (Exception e) {
+            throw new IllegalArgumentException(String.format("Unable to parse url: %s %s", e.getMessage(), e));
+          }
+          int port = urlInfo.toUri().getPort();
```
rmstar, line 246: add comment indicating what the uri looks like

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+          String host = urlInfo.toUri().getAuthority().substring(urlInfo.toUri().getAuthority().lastIndexOf("@") + 1);
+          String user = urlInfo.toUri().getAuthority().split(":")[0];
+          if (urlInfo.toUri().getScheme().equals(FTP_PROTOCOL)) {
+            port = (port == -1) ? DEFAULT_FTP_PORT : port;
+            String cleanHostFTP = host.replace(":" + port, "");
```
rmstar, line 251: Does this work as expected when  urlInfo.toUri().getPort() returns -1, i.e. port is undefined ?

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -130,6 +138,29 @@ public String getReferenceName() {
...
+        String user = urlInfo.toUri().getAuthority().split(":")[0];
+        String protocol = urlInfo.toUri().getScheme();
+        int port = urlInfo.toUri().getPort();
+        if (port == -1 && protocol.equals(FTP_PROTOCOL)) {
+          port = DEFAULT_FTP_PORT;
+        }
+        if (port == -1 && protocol.equals(SFTP_PROTOCOL)) {
+          port = DEFAULT_SFTP_PORT;
+        }
+        String cleanHost = host.replace(":" + port, "");
```
rmstar, line 161: Why are we removing the port from the path here? Looks like we don't do that if the password doesn't contain special characters?

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+          }
+          int port = urlInfo.toUri().getPort();
+          String host = urlInfo.toUri().getAuthority().substring(urlInfo.toUri().getAuthority().lastIndexOf("@") + 1);
+          String user = urlInfo.toUri().getAuthority().split(":")[0];
+          if (urlInfo.toUri().getScheme().equals(FTP_PROTOCOL)) {
+            port = (port == -1) ? DEFAULT_FTP_PORT : port;
+            String cleanHostFTP = host.replace(":" + port, "");
+            fileSystemPropertiesMap.put("fs.ftp.host", cleanHostFTP);
+            fileSystemPropertiesMap.put(String.format("fs.ftp.user.%s", cleanHostFTP), user);
+            fileSystemPropertiesMap.put(String.format("fs.ftp.password.%s", cleanHostFTP), extractPasswordFromUrl());
```
rmstar: we can use extractedPassword here.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+            fileSystemPropertiesMap.put("fs.ftp.host.port", String.valueOf(port));
+          } else {
+            port = (port == -1) ? DEFAULT_SFTP_PORT : port;
+            String cleanHostSFTP = host.replace(":" + port, "");
+            fileSystemPropertiesMap.put(String.format("fs.sftp.user.%s", cleanHostSFTP), user);
+            fileSystemPropertiesMap.put(String.format("fs.sftp.password.%s.%s", cleanHostSFTP, user),
+                                        extractPasswordFromUrl());
```
rmstar: we can use extractedPassword here.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -181,15 +212,63 @@ public Schema getSchema() {
...
+      if (fileSystemProperties != null) {
+        fileSystemPropertiesMap.putAll(GSON.fromJson(fileSystemProperties, MAP_STRING_STRING_TYPE));
+      }
+
       try {
-        return GSON.fromJson(fileSystemProperties, MAP_STRING_STRING_TYPE);
+        if (authContainsSpecialCharacters()) {
+          Path urlInfo;
+          String extractedPassword = extractPasswordFromUrl();
+          String encodedPassword = URLEncoder.encode(extractedPassword);
```
rmstar, line 237: why are we url encoding the password? seems like we just set the un-encoded password in filesystem property fs.(s)ftp.password.<host>.<user>

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: We are removing the port from path here because of the format that is required in FTP File System when you add the necessary parameters as _File System Properties_. 

For example this will fail in File System Properties if we don't remove the port because it doesn't match the required format; 

`ftp://kimi:42@67]gfuss@192.168.0.179:21/kimi-look-here.txt`

"fs.ftp.host" --> "fs.ftp.host.192.168.0.179:21" 
"fs.ftp.user." + host  --> "fs.ftp.user.192.168.0.179:21.kimi" 
"fs.ftp.password." + host --> "fs.ftp.password.192.168.0.179:21.47@79]gfuss" 

The FTP or SFTP File system contains a **"fs.ftp.host.port"** which defaults to 21 in _FTP_ or 22 in _SFTP_ if the user doesn't specify the port in neither url or in File system properties, but if apparently we don't remove it from `host`, the job will fail because it is expecting a host like this:

"fs.ftp.host" --> "192.168.0.179" 
"fs.ftp.user." + host ---> "fs.ftp.user.192.168.0.179"  --> "kimi" 
"fs.ftp.password." + host --> "fs.ftp.password.192.168.0.179" --> "47@79]gfuss" 
"fs.ftp.host.port" -> "21" 

I've tested this specific scenario in: 
- CDAP 6.3.0 
- CDAP 6.1.4 
- CDAP 6.1.2 
- CDAP 6.2.0 

...and removing the port from host and adding it as standalone **"fs.ftp.host.port"** in File System Properties was the only way to get the job through.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: I think we need this condition because when authentication in url contains special characters but File System Properties are empty, we'll return always an empty hashmap at the end of the function but this doesn't solve our problem. 

Instead, we need to extract the necessary parameters from **url** (host, port, username, password) when this scenario happens and add these parameters as File System Properties in order to solve the issue.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: We are encoding the password in order to avoid _IllegalArgumentExceptions_. 

If we don't encode the password, the Path (org.apache.hadoop.fs.Path) will throw an Exception and we won't be able to extract the necessary parameters from Path (host, port, scheme, username, password). 

In this way, we are sure that the Path will always be valid and we will be able to extract whatever we need to put the parameters in File System Properties.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: In case when urlInfo.toUri().getPort() returns -1 or undefined we add the default FTP/SFTP port. This is how it's required in FTP File System and we've followed the same pattern. 

This works as expected. After setting up the default port we are able to process what follows.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: Nice catch.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
flakrimjusufi: For Example: 

Before encoding: 
`ftp://kimi:42@67]gfuss@192.168.0.179:21/kimi-look-here.txt`
**This will throw an IllegalArgumentException because of invalid characters in password.** 

After encoding: 
`ftp://kimi:42%4067%5Dgfuss@192.168.0.179:21/kimi-look-here.txt`
**The Path (_org.apache.hadoop.fs.Path_) will be able to parse the url.**

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: I see, thanks for the explanation!

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: If  authentication in url contains special characters but File System Properties are empty, we will go into the if condition in line 237 and update fileSystemPropertiesMap.

flakrimjusufi: You're right. Now I got your point. We don't need that if condition. I followed your suggestion and tested the scenarios after removing the condition and everything is working okay.

rmstar (approved):
lgtm. Please do the following:
- squash commits
- update PR description to reference JIRA ticket
- add comment on the JIRA ticket that links to this PR

flakrimjusufi:
> lgtm. Please do the following:
> 
> * squash commits
> * update PR description to reference JIRA ticket
> * add comment on the JIRA ticket that links to this PR

Done.

PR #1255, opened 2020-12-03
[CDAP-17248] Added support for decoding files encoded in fixed-length charsets when reading from a file source.

fernst:
As it stands, UTF-32, ISO-8859-1 and Windows-1252 formats are supported.

I've identified an edge case when some lines get skipped in the partition edges. I'm close to isolating the cause of the issue.

### DIFF
--- a/format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
@@ -1,0 +1,227 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.charset;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.cdap.plugin.format.charset.fixedlength.FixedLengthCharset;
+import io.cdap.plugin.format.charset.fixedlength.FixedLengthCharsetTransformingCodec;
+import io.cdap.plugin.format.charset.fixedlength.FixedLengthCharsetTransformingDecompressorStream;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.Seekable;
+import org.apache.hadoop.io.LongWritable;
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.compress.Decompressor;
+import org.apache.hadoop.io.compress.SplitCompressionInputStream;
+import org.apache.hadoop.io.compress.SplittableCompressionCodec;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.CompressedSplitLineReader;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.hadoop.mapreduce.lib.input.SplitLineReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * Copy of Hadoop's LineRecordReader (Hadoop Version 2.3.0). The reason we copy this class is to create a
+ * FixedLengthCharsetTransformingCodec which works for every input file with no regards to the file extension.
+ * <p>
+ * This class has been simplified since we can assume that the input needs to be decompressed.
+ * <p>
+ * This class uses a fixed Codec and Decompressor to parse records.
+ */
+public class CharsetTransformingLineRecordReader extends RecordReader<LongWritable, Text> {
+  private static final Logger LOG = LoggerFactory.getLogger(CharsetTransformingLineRecordReader.class);
+  public static final String MAX_LINE_LENGTH =
+    "mapreduce.input.linerecordreader.line.maxlength";
+
+  private final FixedLengthCharset fixedLengthCharset;
+  private final byte[] recordDelimiterBytes;
+  private long start;
+  private long pos;
+  private long end;
+  private SplitLineReader in;
+  private Seekable filePosition;
+  private int maxLineLength;
+  private LongWritable key;
+  private Text value;
+  private Decompressor decompressor;
+
+  public CharsetTransformingLineRecordReader(FixedLengthCharset fixedLengthCharset, byte[] recordDelimiter) {
+    this.fixedLengthCharset = fixedLengthCharset;
+    this.recordDelimiterBytes = recordDelimiter;
+  }
+
+  @VisibleForTesting
+  protected CharsetTransformingLineRecordReader(FixedLengthCharset fixedLengthCharset,
+                                                byte[] recordDelimiter,
+                                                SplitLineReader in,
+                                                long start,
+                                                long pos,
+                                                long end,
+                                                int maxLineLength) {
+    this(fixedLengthCharset, recordDelimiter);
+    this.in = in;
+    this.start = start;
+    this.pos = pos;
+    this.end = end;
+    this.maxLineLength = maxLineLength;
+  }
+
+  /**
+   * Initialize method from parent class, simplified for this our use case from the base class.
+   *
+   * @param genericSplit File Split
+   * @param context      Execution context
+   * @throws IOException if the underlying file or decompression operations fail.
+   */
+  public void initialize(InputSplit genericSplit,
+                         TaskAttemptContext context) throws IOException {
+    FileSplit split = (FileSplit) genericSplit;
+    Configuration job = context.getConfiguration();
+    this.maxLineLength = job.getInt(MAX_LINE_LENGTH, Integer.MAX_VALUE);
+    start = split.getStart();
+    end = start + split.getLength();
+    final Path file = split.getPath();
+
+    // open the file and seek to the start of the split
+    final FileSystem fs = file.getFileSystem(job);
+    FSDataInputStream fileIn = fs.open(file);
+
+    SplittableCompressionCodec codec = new FixedLengthCharsetTransformingCodec(fixedLengthCharset);
+    decompressor = codec.createDecompressor();
+
+    final SplitCompressionInputStream cIn =
+      codec.createInputStream(
+        fileIn, decompressor, start, end,
+        SplittableCompressionCodec.READ_MODE.CONTINUOUS);
+    in = new CompressedSplitLineReader(cIn, job,
+                                       this.recordDelimiterBytes);
+    start = cIn.getAdjustedStart();
+    end = cIn.getAdjustedEnd();
+    filePosition = cIn;
+
+    // If this is not the first split, we always throw away first record
+    // because we always (except the last split) read one extra line in
+    // next() method.
+    if (start != 0) {
+      Text t = new Text();
+      start += in.readLine(t, 4096, Integer.MAX_VALUE);
+      LOG.info("Discarded line: " + t.toString());
+    }
+    this.pos = start;
+  }
+
+  /**
+   * Returns the File position in the underlying stream.
+   * <p>
+   * Note that, as the file is read in chunks to decompress, this number will only update in batches and usually be
+   * ahead of the actual decompressed position in the underlying file.
+   *
+   * See {@link FixedLengthCharsetTransformingDecompressorStream#getCompressedData}
+   *
+   * @return File position
+   * @throws IOException if an exception is thrown from the underlying strem operation.
+   */
+  @VisibleForTesting
+  protected long getFilePosition() throws IOException {
+    return filePosition.getPos();
+  }
+
+  /**
+   * Read the next Key Value from the decompressor.
+   * <p>
+   * Note that, as we read from the original file in chunks in order to decode, as we approach the partition boundary
+   * defined by `end`, the read operation
+   *
+   * @return
+   * @throws IOException
+   */
+  public boolean nextKeyValue() throws IOException {
+    if (key == null) {
+      key = new LongWritable();
+    }
+    key.set(pos);
+    if (value == null) {
+      value = new Text();
+    }
+    int newSize = 0;
+    // We always read one extra line, which lies outside the upper
+    // split limit i.e. (end - 1)
+    while (getFilePosition() <= end || in.needAdditionalRecordAfterSplit()) {
+      newSize = in.readLine(value, maxLineLength, Integer.MAX_VALUE);
+      pos += newSize;
+      if (newSize < maxLineLength) {
+        break;
+      }
+
+      // line too long. try again
+      LOG.info("Skipped line of size " + newSize + " at pos " +
+                 (pos - newSize));
+    }
+    if (newSize == 0) {
+      key = null;
+      value = null;
+      return false;
+    } else {
+      return true;
+    }
+  }
+
+  @Override
+  public LongWritable getCurrentKey() {
+    return key;
+  }
+
+  @Override
+  public Text getCurrentValue() {
+    return value;
+  }
+
+  /**
+   * Get the progress within the split
+   */
+  public float getProgress() throws IOException {
+    if (start == end) {
+      return 0.0f;
+    } else {
+      return Math.min(1.0f, (getFilePosition() - start) / (float) (end - start));
+    }
+  }
+
+  /**
+   * Close this input stream and clean up the decompressor.
+   *
+   * @throws IOException
+   */
+  public synchronized void close() throws IOException {
+    try {
+      if (in != null) {
+        in.close();
+      }
+    } finally {
+      if (decompressor != null) {
+        decompressor.end();
+      }
+    }
+  }
+}
--- a/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.charset.fixedlength;
+
+import com.google.common.collect.ImmutableMap;
+
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.UnsupportedCharsetException;
+import java.util.Map;
+
+/**
+ * Enumeration containing all currently supported Fixed Length charsets.
+ * <p>
+ * This currently includes:
+ * - UTF-32
+ * - Any single-byte charset encoding supported by Java, including the ISO-8859, Windows-125x and EBCDIC
+ */
+public class FixedLengthCharset {
+  public static final FixedLengthCharset UTF_32 = new FixedLengthCharset("UTF-32", Charset.forName("UTF-32"), 4);
+  private static final Map<String, FixedLengthCharset> ALLOWED_MULTIBYTE_CHARSETS = ImmutableMap.of(
+    UTF_32.getName(), UTF_32
+  );
+
+  private final String name;
+  private final Charset charset;
+  private final int numBytesPerCharacter;
+
+  FixedLengthCharset(String name, Charset charset, int numBytesPerCharacter) {
+    this.name = name;
+    this.charset = charset;
+    this.numBytesPerCharacter = numBytesPerCharacter;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public Charset getCharset() {
+    return charset;
+  }
+
+  public int getNumBytesPerCharacter() {
+    return numBytesPerCharacter;
+  }
+
+  /**
+   * Find a FixedLengthCharset for a given encoding name. Throws a runtime exception if not found.
+   *
+   * @param name Charset name
+   * @return FixedLengthCharset for the desired charset.
+   */
+  public static FixedLengthCharset forName(String name) {
+    if (!isValidEncoding(name)) {
+      throw new IllegalArgumentException("Charset not supported: " + name);
+    }
+
+    FixedLengthCharset fixedLengthCharset = ALLOWED_MULTIBYTE_CHARSETS.get(name.toUpperCase());
+    if (fixedLengthCharset != null) {
+      return fixedLengthCharset;
+    }
+    Charset charset = Charset.forName(name.toUpperCase());
+    return new FixedLengthCharset(name, charset, 1);
+  }
+
+  /**
+   * Check if this file encoding is valid.
+   *
+   * @return boolean value specifying if this is a valid encoding or not.
+   */
+  public static boolean isValidEncoding(String name) {
+    if (ALLOWED_MULTIBYTE_CHARSETS.containsKey(name.toUpperCase())) {
+      return true;
+    }
+    try {
+      Charset charset = Charset.forName(name.toUpperCase());
+      CharsetEncoder encoder = charset.newEncoder();
+      return encoder.maxBytesPerChar() == 1 && encoder.averageBytesPerChar() == 1;
+    } catch (UnsupportedOperationException | UnsupportedCharsetException e) {
+      return false;
+    }
+  }
+
+  /**
+   * Takes the first word of the file encoding string, as any further words are just charset descriptions.
+   *
+   * @param fileEncoding The file encoding parameter supplied by the user
+   * @return the cleaned up file encoding name
+   */
+  public static String cleanFileEncodingName(String fileEncoding) {
+    fileEncoding = fileEncoding.trim();
+
+    return fileEncoding.split(" ")[0].toUpperCase();
+  }
+}
--- a/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.charset.fixedlength;
+
+import org.apache.hadoop.conf.Configurable;
+import org.apache.hadoop.fs.Seekable;
+import org.apache.hadoop.io.compress.CompressionOutputStream;
+import org.apache.hadoop.io.compress.Compressor;
+import org.apache.hadoop.io.compress.Decompressor;
+import org.apache.hadoop.io.compress.DefaultCodec;
+import org.apache.hadoop.io.compress.DirectDecompressor;
+import org.apache.hadoop.io.compress.SplitCompressionInputStream;
+import org.apache.hadoop.io.compress.SplittableCompressionCodec;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * Codec implementation that returns a decompressor for Fixed Length character encodings.
+ */
+public class FixedLengthCharsetTransformingCodec extends DefaultCodec
+  implements Configurable, SplittableCompressionCodec {
+  private static final Logger LOG = LoggerFactory.getLogger(FixedLengthCharsetTransformingCodec.class);
+
+  private final FixedLengthCharset sourceEncoding;
+
+  public FixedLengthCharsetTransformingCodec(FixedLengthCharset sourceEncoding) {
+    this.sourceEncoding = sourceEncoding;
+  }
+
+  @Override
+  public CompressionOutputStream createOutputStream(OutputStream out) throws IOException {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public Class<? extends Compressor> getCompressorType() {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public Compressor createCompressor() {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public Class<? extends Decompressor> getDecompressorType() {
+    return FixedLengthCharsetTransformingDecompressor.class;
+  }
+
+  @Override
+  public DirectDecompressor createDirectDecompressor() {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public String getDefaultExtension() {
+    return ".*";
+  }
+
+  @Override
+  public Decompressor createDecompressor() {
+    return new FixedLengthCharsetTransformingDecompressor(sourceEncoding);
+  }
+
+  public SplitCompressionInputStream createInputStream(InputStream seekableIn,
+                                                       Decompressor decompressor,
+                                                       long start,
+                                                       long end,
+                                                       READ_MODE readMode) throws IOException {
+    if (!(seekableIn instanceof Seekable)) {
+      throw new IOException("seekableIn must be an instance of " +
+                              Seekable.class.getName());
+    }
+
+    //Adjust start to align to the next character boundary.
+    if (start % sourceEncoding.getNumBytesPerCharacter() != 0) {
+      long adjustment = sourceEncoding.getNumBytesPerCharacter() - (start % sourceEncoding.getNumBytesPerCharacter());
+      LOG.trace("Adjusted partition start positioon from {} to {} by {} bytes",
+                start, start + adjustment, adjustment);
+      start += adjustment;
+    }
+
+    //Adjust end to align to the next character boundary.
+    if (end % sourceEncoding.getNumBytesPerCharacter() != 0) {
+      long adjustment = sourceEncoding.getNumBytesPerCharacter() - (end % sourceEncoding.getNumBytesPerCharacter());
+      LOG.trace("Adjusted partition end position from {} to {} by {} bytes",
+                end, end + adjustment, adjustment);
+      end += adjustment;
+    }
+
+    FixedLengthCharsetTransformingDecompressorStream decompressorStream =
+      new FixedLengthCharsetTransformingDecompressorStream(seekableIn, sourceEncoding, start, end);
+
+    return new TransformingCompressionInputStream(decompressorStream, start, end);
+  }
+
+}
--- a/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
@@ -1,0 +1,245 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.charset.fixedlength;
+
+import org.apache.hadoop.io.compress.Decompressor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Decompressor that can be used to convert byte streams in fixed-length character encodings to a stream of UTF-8 bytes.
+ */
+public class FixedLengthCharsetTransformingDecompressor implements Decompressor {
+
+  private static final Logger LOG = LoggerFactory.getLogger(FixedLengthCharsetTransformingDecompressor.class);
+
+  private final FixedLengthCharset sourceEncoding;
+  private final CharsetDecoder decoder;
+  private final CharsetEncoder encoder;
+  private final Charset targetCharset = StandardCharsets.UTF_8;
+
+  //Initializing all buffers.
+  private ByteBuffer inputByteBuffer = ByteBuffer.allocate(0);
+  private CharBuffer decodedCharBuffer = CharBuffer.allocate(0);
+  private ByteBuffer partialOutputByteBuffer = ByteBuffer.allocate(0);
+
+  public FixedLengthCharsetTransformingDecompressor(FixedLengthCharset sourceEncoding) {
+    this.sourceEncoding = sourceEncoding;
+    this.decoder = sourceEncoding.getCharset().newDecoder();
+    this.encoder = targetCharset.newEncoder();
+  }
+
+  @Override
+  public void setInput(byte[] b, int off, int len) {
+    //Set up incoming buffer for writes.
+    inputByteBuffer.compact();
+
+    //Expand incoming buffer if needed.
+    if (inputByteBuffer.remaining() < len) {
+      //Allocate new buffer that can fill the existing input + newly received bytes
+      ByteBuffer newIncomingBuffer = ByteBuffer.allocate(len + inputByteBuffer.capacity());
+
+      //Set up incoming buffer for reads and copy contents into new buffer.
+      inputByteBuffer.flip();
+      newIncomingBuffer.put(inputByteBuffer);
+
+      inputByteBuffer = newIncomingBuffer;
+    }
+
+    //Copy incoming payload into Input Byte Buffer
+    inputByteBuffer.put(b, off, len);
+    inputByteBuffer.flip();
+
+    //Set up char buffer for writes
+    decodedCharBuffer.compact();
+
+    //Expand the char buffer if needed.
+    if (decodedCharBuffer.capacity() < inputByteBuffer.limit() / sourceEncoding.getNumBytesPerCharacter()) {
+      decodedCharBuffer = CharBuffer.allocate(inputByteBuffer.limit() / sourceEncoding.getNumBytesPerCharacter());
+    }
+
+    //Decode bytes from the input buffer into the Decoded Char Buffer
+    decodeByteBufferIntoCharBuffer(inputByteBuffer);
+
+    //Set up decoded char buffer for reads.
+    decodedCharBuffer.flip();
+
+  }
+
+  /**
+   * Note that we only ask for additional input once we have completely depleted out outgoing buffer.
+   */
+  @Override
+  public boolean needsInput() {
+    return decodedCharBuffer.remaining() == 0;
+  }
+
+  @Override
+  public void setDictionary(byte[] b, int off, int len) {
+    //no-op
+  }
+
+  @Override
+  public boolean needsDictionary() {
+    return false;
+  }
+
+  @Override
+  public boolean finished() {
+    return decodedCharBuffer.remaining() == 0 && partialOutputByteBuffer.remaining() == 0;
+  }
+
+  @Override
+  public int decompress(byte[] b, int off, int len) throws IOException {
+
+    //Allocate new outgoing buffer
+    ByteBuffer encodedBuffer = ByteBuffer.wrap(b, off, len);
+
+    //Consume any remaining bytes from a previous decompress invocation.
+    while (partialOutputByteBuffer != null && partialOutputByteBuffer.hasRemaining() && encodedBuffer.hasRemaining()) {
+      encodedBuffer.put(partialOutputByteBuffer.get());
+    }
+
+    //Encode as many characters as possible into the Encoded Buffer.
+    encodeCharBufferIntoByteBuffer(encodedBuffer);
+
+    // Handle the case where we need to add a partial decompressed character to the output buffer.
+    // This means we need to encode one extra character and add as many bytes as possible into the output buffer.
+    // This is an expensive operation that is executed as a last resort, meaning we only do this when we were
+    // not able to add any bytes to the output payload before.
+    if (decodedCharBuffer.remaining() > 0 && encodedBuffer.remaining() > 0 && encodedBuffer.position() - off == 0) {
+      encodePartialCharIntoByteBuffer(encodedBuffer);
+    }
+
+    //Return the number of bytes copied,
+    // This is matches the actual position in the decoded buffer minus the initial offset
+    return encodedBuffer.position() - off;
+  }
+
+  /**
+   * Decode butes from the specified byteBuffer into the DecodedCharBuffer.
+   *
+   * @param buffer The target ByteBuffer
+   */
+  protected void decodeByteBufferIntoCharBuffer(ByteBuffer buffer) {
+    int initialCharBufferPos = decodedCharBuffer.position();
+
+    //Decode input buffer as characters.
+    CoderResult decodeResult = decoder.decode(buffer, decodedCharBuffer, false);
+
+    if (decodeResult.isError()) {
+      LOG.error("Unable to decode payload from file as {}", sourceEncoding.getCharset().name());
+      throw new CharacterDecodingException(decoder);
+    }
+
+    int finalCharBufferPos = decodedCharBuffer.position();
+  }
+
+  /**
+   * Encode bytes from the decodedCharBuffer into the specified ByteBuffer.
+   *
+   * @param buffer The target ByteBuffer
+   */
+  protected void encodeCharBufferIntoByteBuffer(ByteBuffer buffer) {
+    int initialCharBufferPos = decodedCharBuffer.position();
+
+    //Decode as many chars as possible into the outgoing buffer.
+    CoderResult encodeResult = encoder.encode(decodedCharBuffer, buffer, true);
+
+    if (encodeResult.isError()) {
+      LOG.error("Unable to encode decoded payload as UTF-8");
+      throw new CharacterEncodingException(encoder);
+    }
+
+    int finalCharBufferPos = decodedCharBuffer.position();
+  }
+
+  /**
+   * Encode a minimum of 1 additional character from the decoded char buffer into the output buffer.
+   * <p>
+   * Since the output is UTF-8, there can be up to 4 additional characters encoded in this temporary buffer.
+   * <p>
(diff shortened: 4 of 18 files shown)


format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,167 @@
...
+                                                       long end,
+                                                       READ_MODE readMode) throws IOException {
+    if (!(seekableIn instanceof Seekable)) {
+      throw new IOException("seekableIn must be an instance of " +
+                              Seekable.class.getName());
+    }
+
+    //Adjust start to align to the next character boundary.
+    if (start % fixedLengthCharset.getCharLength() != 0) {
+      LOG.info("Adjusted Start from {} to {} by {} bytes",
```
fernst: Ignore these log statements for now. I've been using them for debugging.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.hadoop.mapreduce.lib.input.LineRecordReader;
+import org.apache.hadoop.mapreduce.lib.input.SplitLineReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * Copy of Hadoop's Line Record Reader (Hadoop Version 2.3.0)
```
albertshau: Can you add comments about which lines are different than the copy

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
```diff
@@ -0,0 +1,82 @@
...
+   * @return FixedLengthCharset for the desired charset.
+   */
+  public static FixedLengthCharset forName(String name) {
+    for (FixedLengthCharset c : FixedLengthCharset.values()) {
+      if (name.equalsIgnoreCase(c.getName())) {
+        return c;
+      }
+    }
+
+    throw new RuntimeException("Cannot find FixedLengthCharset with name: " + name);
```
albertshau: nit: try to avoid throwing Exception or RuntimeException. This method is similar to Enum's valueOf() method, which will throw IllegalArgumentException in a similar situation.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.OutputStream;
+
+/**
+ * Codec implementation that returns a decompressor for Fixed Length character encodings.
+ */
+public class FixedLengthCharsetTransformingCodec extends DefaultCodec
+  implements Configurable, SplittableCompressionCodec {
+  private static final Logger LOG = LoggerFactory.getLogger(FixedLengthCharsetTransformingCodec.class);
+
+  FixedLengthCharset fixedLengthCharset;
```
albertshau: privatre final

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,167 @@
...
+
+  public FixedLengthCharsetTransformingCodec(FixedLengthCharset fixedLengthCharset) {
+    this.fixedLengthCharset = fixedLengthCharset;
+  }
+
+  @Override
+  public CompressionOutputStream createOutputStream(OutputStream out) throws IOException {
+    throw new RuntimeException("Not supported");
```
albertshau: UnsupportedOperationException (same with others)

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
```diff
@@ -0,0 +1,119 @@
...
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Decompressor that can be used to convert byte streams in fixed-length character encodings to a stream of UTF-8 bytes.
+ */
+public class FixedLengthCharsetTransformingDecompressor implements Decompressor {
+
+  final FixedLengthCharset origin;
```
albertshau: private final

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,77 @@
...
+  long start;
+  long end;
+
+  protected FixedLengthCharsetTransformingDecompressorStream(InputStream in,
+                                                             FixedLengthCharset fixedLengthCharset,
+                                                             long start,
+                                                             long end)
+    throws IOException {
+    super(in, new FixedLengthCharsetTransformingDecompressor(fixedLengthCharset));
+    long skippedBytes = in.skip(start);
```
albertshau: unused?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,77 @@
...
+    //If we're working with a Charset Transforming decompressor, we can calculate the current position on the input file
+    // By adding the starting position in the file with the number of bytes we have read so far.
+    if (this.decompressor instanceof FixedLengthCharsetTransformingDecompressor) {
+      FixedLengthCharsetTransformingDecompressor flcDecompressor =
+        (FixedLengthCharsetTransformingDecompressor) this.decompressor;
+
+      //Actual position is starting possition + the number of bytes we have consumed - the remaining bytes
+      return start + flcDecompressor.getNumConsumedBytes();
+    } else {
+      return super.getPos();
```
albertshau: seems like this would never happen?

albertshau (commented):
the overall approach looks good so far.

left minor comments about things that you might have been going to clean up later anyway.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: I will.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Enumeration containing all supported Fixed Length charsets.
+ */
+public enum FixedLengthCharset {
+  UTF_32("UTF-32", Charset.forName("UTF-32"), 4),
+  ISO_8859_1("ISO-8859-1", StandardCharsets.ISO_8859_1, 1),
+  WINDOWS_1252("Windows-1252", Charset.forName("windows-1252"), 1);
```
tivv: Can we make it extendable / overridable and/or add UTF-16 and EBCDIC as popular encodings

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  @Override
+  public CompressionInputStream createInputStream(InputStream in) throws IOException {
+    return super.createInputStream(in, new FixedLengthCharsetTransformingDecompressor(this.fixedLengthCharset));
+  }
+
+  @Override
+  public CompressionInputStream createInputStream(InputStream in, Decompressor decompressor) throws IOException {
```
tivv: I don't think you need this override

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
```diff
@@ -0,0 +1,119 @@
...
+
+  ByteArrayOutputStream incomingBuffer = new ByteArrayOutputStream();
+  ByteArrayInputStream outgoingBuffer = new ByteArrayInputStream(new byte[]{});
+
+  public FixedLengthCharsetTransformingDecompressor(FixedLengthCharset origin) {
+    this.origin = origin;
+  }
+
+  @Override
+  public void setInput(byte[] b, int off, int len) {
```
tivv, line 56: You are doing a lot of copying here. Please consider using java.nio.charset.Charset along with `ByteBuffer`/`CharBuffer`.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
```diff
@@ -0,0 +1,106 @@
...
+  ISO_8859_15("ISO-8859-15", Charset.forName("ISO-8859-15"), 1),
+  WINDOWS_1250("Windows-1250", Charset.forName("windows-1250"), 1),
+  WINDOWS_1251("Windows-1251", Charset.forName("windows-1251"), 1),
+  WINDOWS_1252("Windows-1252", Charset.forName("windows-1252"), 1),
+  WINDOWS_1253("Windows-1253", Charset.forName("windows-1253"), 1),
+  WINDOWS_1254("Windows-1254", Charset.forName("windows-1254"), 1),
+  WINDOWS_1255("Windows-1255", Charset.forName("windows-1255"), 1),
+  WINDOWS_1256("Windows-1256", Charset.forName("windows-1256"), 1),
+  WINDOWS_1257("Windows-1257", Charset.forName("windows-1257"), 1),
+  WINDOWS_1258("Windows-1258", Charset.forName("windows-1258"), 1);
```
tivv: Can we have EBCDIC here. Also I am not sure if it would make sense to have some kind of override to be able to use it for arbitrary charset.

tivv: You can also try to use `java.nio.charset.CharsetDecoder` maxCharsPerByte/averageCharsPerByte and skip this table altogether

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
fernst: Hmm, good point. I can probably get away with using a byte buffer for input and transform as needed when decompression is needed.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
```diff
@@ -0,0 +1,222 @@
...
+import org.apache.hadoop.mapreduce.lib.input.LineRecordReader;
+import org.apache.hadoop.mapreduce.lib.input.SplitLineReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+
+/**
+ * Copy of Hadoop's LineRecordReader (Hadoop Version 2.3.0). The reason we copy this class is to modify some behaviors
+ * related to the creation of the decompressor. This also allows us to implement
```
chtyim: What is modified and why is needed?

chtyim (commented):
Why is it limited to fixed length encoding only? Using compressor codec should be able to handle dynamic length as well?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: Good point. Fixed it.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: Indeed, we're not really using this value. Maybe I should throw an exception if it's not possible to follow this instruction.

tivv: I believe getting less than asked means we've got EOF earlier.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: @chtyim The problem with variable with encoding is that the file partitions in Spark are based on file position in bytes, meaning we might end up in a scenario where the partition starts in the middle of a character.

If this is the case, we would be in a position where the input stream cannot be decoded correctly. Essentially, trying to read from this input stream as a String will return an invalid string.

For Fixed Length charsets, my implementations makes sure to align the partition boundaries with full characters: https://github.com/cdapio/hydrator-plugins/blob/83427f7002b016044f5f4606d998b92511c92ceb/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java#L110-L124

The only flexible solution I see for variable-length encodings is to run the decoding from the start of the file in order to make sure we are always aligned to a character boundary.

fernst:
> Why is it limited to fixed length encoding only? Using compressor codec should be able to handle dynamic length as well?

@chtyim The problem with variable with encoding is that the file partitions in Spark are based on file position in bytes, meaning we might end up in a scenario where the partition starts in the middle of a character.

If this is the case, we would be in a position where the input stream cannot be decoded correctly. Essentially, trying to read from this input stream as a String will return an invalid string.

For Fixed Length charsets, my implementations makes sure to align the partition boundaries with full characters: https://github.com/cdapio/hydrator-plugins/blob/83427f7002b016044f5f4606d998b92511c92ceb/format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java#L110-L124

The only flexible solution I see for variable-length encodings is to run the decoding from the start of the file in order to make sure we are always aligned to a character boundary.

tivv:
@fernst Are there examples of variable-width chasets but UTF-* ? I believe if delimiter is line break, UTF-* are done so that it should not matter. Line break can't be a non-first character in UTF encoding

tivv:
I see ISO-2022 family, but IMHO it makes sense to have a exclude list or dynamically detect by comparing maximum character length and average character length as this is available in runtime

fernst:
@tivv I agree with you in that this will not be a problem for UTF-16, **IF** the delimiter is `\n`, `\r` or a combination of these two. We can simply align the partition boundaries to the end of the first `\n` or `\r` character we find after a partition start/end.

If the delimiter is an arbitrary sequence of bytes, we would need to scan for this sequence in the input stream. However, this does not give us a 100% safe way of making sure we are aligned to a full character. We might be in the odd scenario where the ending or one char and the beggining of another match our sequence of bytes, and this causes our charset transformation to break.

Other variable-length encodings might not be as easy to figure out. However, for our initial requirements, we only need to support ISO-8859-1 for now so we can leave variable-length charset support for a later PR.

tivv:
Actually in UTF any delimiter would work as UTF series of encodings separate leading bytes(words for UTF-16) and trailing bytes, so you are always safe. You can't end up in the middle if you are scanning for a valid character encoding

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: @chtyim The original implementation of Hadoop's LineRecordReader uses the file name to determine which codec to implement:

https://github.com/facebookarchive/hadoop-20/blob/master/src/mapred/org/apache/hadoop/mapred/LineRecordReader.java#L92

https://github.com/facebookarchive/hadoop-20/blob/master/src/core/org/apache/hadoop/io/compress/CompressionCodecFactory.java#L160-L175

This is not particularly useful in our use case as the file extension could be anything.

In order to go around this limitation, I copied this class and simplified the code as we can assume the input is "compressed" and we need to "decompress" (apply the charset conversion) as we read the file.

chtyim: It is controlled by config 

https://github.com/apache/hadoop/blob/branch-2.9.2/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/CompressionCodecFactory.java#L124

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,119 @@
...
+    // input file.
+    // By adding the starting position in the file with the number of bytes we have read so far.
+    FixedLengthCharsetTransformingDecompressor flcDecompressor =
+      (FixedLengthCharsetTransformingDecompressor) this.decompressor;
+
+    //Actual position is starting possition + the number of bytes we have consumed.
+    return start + flcDecompressor.getNumConsumedBytes();
+  }
+
+  /**
```
fernst, line 61: @chtyim @albertshau @tivv I want to bring your attention to this line function:

The problem I'm facing at this time is that, in order to ensure that we respect the partition boundaries, and process each record exactly once, I need to make sure that I only read the last line that begins in a partition. This means that if a line is "split" by a partition, the worker that was reading the previous partition is the one that reads this line.

The way Hadoop does this for normal text files is to check against the file position in order to make sure the last line has not been read. However, this poses a problem for me, because the compressor reads part of the file in chunks, and the file position does not necessarily align with the position of the last line that was read and decoded.

The block of code I'm describing can be found here: https://github.com/cdapio/hydrator-plugins/blob/834s7f7002b016044f5f4606d998b92511c92ceb/format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java#L165-L171

This poses a challenge for me. I see 2 ways to fix this:

1) As you can see in my implementation above, once I approach the partition boundary, I make sure to only read one character at a time, until the function that reads a line is able to complete successfully, and break out of the loop. This means that, as I approach the partition boundary, the execution speed slows down.

2) The other solution I saw is to calculate the file position based on reading character-by-character from the input in the decompressor. The problem I see with that is that, for every character I read, I would need to create a String instance, convert to bytes, and fill the output buffer in this way. This would give me the ability to accurately determine my position in the underlying file, but doing the Bytes -> 1 Character long String -> Bytes in UTF-8 charset conversion for every character in my input file is definitely not a performant solution, and this will impact performance everywhere in my partition, not just the end of the partition.

Let me know if you can think of a better way to approach this partition boundary issue. The hybrid approach would be to add this "character by character" logic on the decompressor, but this does not strike me as the place where this logic should live.

tivv: Did you try switching to ByteBuffer/CharBuffer? I believe those should remove most of the overhead

tivv: Also, can't you just scan without decoding until you see a line delimiter byte sequence or EOF ? This way you would know exact length of the last line and decode it fast.

tivv: The latter may have memory problems if lines are very long, so using buffers instead of strings can be a better solution.

chtyim: The way I see it is, you read into `ByteBuffer`, using a `CharsetDecoder` to decode the bytes into the source charset, and then encode it to UTF-8 `ByteBuffer`. Then you can feed the UTF-8 bytes to the regular Hadoop reader.

fernst: Thanks @tivv. I was still tying up some things. I'll take a look at both of your suggestions and commit my changes 👀

fernst:
The only remaining change that I need to make is to add the UI dropdown to select all available char encodings.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
chtyim: Can we do it without copying this class?

fernst: I'll work on extending the LineRecordReader class and override this functionality instead of copying the whole class.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: @tivv @chtyim I've reimplemented the logic using ByteBuffer and CharBuffer and it works significantly faster than before.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: @chtyim I don't think the codec-by-configuration approach will work for us because the default implementation of the LineRecordReader class always invokes this function:

https://github.com/apache/hadoop/blob/branch-2.9.2/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/io/compress/CompressionCodecFactory.java#L197-L213

This means that the codec is matched based on file extension, which is not a desirable behavior in my case.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FileBatchSource.java
```diff
@@ -48,6 +48,9 @@ public FileBatchSource(FileSourceConfig config) {
     if (config.shouldCopyHeader()) {
       properties.put(PathTrackingInputFormat.COPY_HEADER, "true");
     }
+    if (config.getFileEncoding() != null) {
```
tivv: How does UTF-8 become null here? I am missing it.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
tivv: Well, I believe if we have a single codec specified in `io.compression.codecs` configuration and it returns an empty   string as a suffix it would be picked up, is not it?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
tivv: So, can we automatically allow any single-byte encoding here?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,78 @@
...
+    in.skip(start);
+    this.fixedLengthCharset = fixedLengthCharset;
+    this.start = start;
+    this.end = end;
+  }
+
+  @Override
+  protected int decompress(byte[] b, int off, int len) throws IOException {
+    //Set input for decompression if it's needed for execution.
+    if (this.decompressor.needsInput()) {
```
tivv: Why do we need this? I see that super already handles similar case

format-common/src/main/java/io/cdap/plugin/format/input/CharsetTransformingPathTrackingInputFormat.java
```diff
@@ -0,0 +1,52 @@
...
+  public CharsetTransformingPathTrackingInputFormat(String charsetName) {
+    this.fixedLengthCharset = FixedLengthCharset.forName(charsetName);
+  }
+
+  @Override
+  public RecordReader<LongWritable, Text> createRecordReader(InputSplit split, TaskAttemptContext context) {
+    String delimiter = context.getConfiguration().get("textinputformat.record.delimiter");
+    byte[] recordDelimiterBytes = null;
+    if (null != delimiter) {
+      recordDelimiterBytes = delimiter.getBytes(fixedLengthCharset.getCharset());
```
tivv: I believe record delimiting is done after we recoded everything to UTF-8, so  this value should be in UTF-8. Could you please add a test with record delimiting?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
```diff
@@ -0,0 +1,266 @@
...
+  @Override
+  public boolean finished() {
+    return decodedCharBuffer.remaining() == 0 && partialOutputByteBuffer.remaining() == 0;
+  }
+
+  @Override
+  public int decompress(byte[] b, int off, int len) throws IOException {
+
+    //Allocate new outgoing buffer
+    ByteBuffer encodedBuffer = ByteBuffer.allocate(len - off);
```
tivv: Can we just do `ByteBuffer.wrap(b, off, len)` and save one more copy?

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
```diff
@@ -0,0 +1,266 @@
...
+      encodedBuffer.put(partialOutputByteBuffer.get());
+    }
+
+    //Encode as many characters as possible into the Encoded Buffer.
+    encodeCharBufferIntoByteBuffer(encodedBuffer);
+
+    // Handle the case where the outgoing buffer can still fit additional space.
+    // This means we need to encode one extra character and add as many bytes as possible into the output buffer.
+    if (decodedCharBuffer.remaining() > 0 && encodedBuffer.remaining() > 0) {
+      encodePartialCharacter(encodedBuffer);
```
tivv: This is a pretty heavy operation. I think we need it only when we could not encode anything. It's totally fine to fil less than `len` as soon as we fill at least something. We are not required to fill each block exactly to the length.

tivv:
Actually it would be great to reuse `org.apache.hadoop.io.compress.TestCompressorDecompressor` for good testing, but we would need to implement UTF8->Original charset as well. Can be worth it though to ensure easy testing and it should be pretty trivial as it would mean only to set source charset to UTF-8 and destination to X

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: @tivv You might be right. Let me give it a try. I had not considered that `headMap` using an empty string as key will just match every record in the map.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -104,4 +119,17 @@ protected int getCompressedData() throws IOException {
...
+  /**
+   * Method that invokes the parent method to get decompressed data, without partition boundary awareness.
+   * <p>
+   * This method is visible for the purposes of testing the partition boundaries.
+   *
+   * @return number of bytes read from source input stream
+   * @throws IOException If there was a problem reading from the underlying input stream.
+   */
+  @VisibleForTesting
+  protected int getCompressedDataSuper() throws IOException {
```
tivv: I see you committed this for test support, but I can't seem to find the test that uses it.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * DecompressorStream implementation for the FixedLengthCharsetTransformingDecompressor.
+ * <p>
+ * This can be used to convert input streams containing bytes for fixed length charsets into UTF-8 bytes.
+ */
+public class FixedLengthCharsetTransformingDecompressorStream extends DecompressorStream {
```
tivv, line 31: What does this inheritance give you? You are only using it to wrap into `TransformingCompressionInputStream`. Would it make sense to move logic from here to `TransformingCompressionInputStream` and use `DecompressorStream` from `TransformingCompressionInputStream`. Having one less layer can make position handling easier.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(FixedLengthCharsetTransformingCodec.class);
+
+  private final FixedLengthCharset sourceEncoding;
+
+  public FixedLengthCharsetTransformingCodec(FixedLengthCharset sourceEncoding) {
+    this.sourceEncoding = sourceEncoding;
+  }
+
+  @Override
+  public CompressionOutputStream createOutputStream(OutputStream out) throws IOException {
```
tivv, line 49: I believe this is not needed. Same logic would be done in `DefaultCodec`

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,131 @@
...
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public CompressionOutputStream createOutputStream(OutputStream out, Compressor compressor) throws IOException {
```
tivv: I believe this is not needed. Given compressor is passed it's fine to allow it.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
```diff
@@ -0,0 +1,131 @@
...
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public Compressor createCompressor() {
+    throw new UnsupportedOperationException("Not supported");
+  }
+
+  @Override
+  public CompressionInputStream createInputStream(InputStream in) throws IOException {
```
tivv: I believe this is not needed. Same logic would be done in DefaultCodec

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
tivv: Generally I am not sure if using DecompressorStream simplifies things or makes it harder. BZip2CompressionInputStream does not use it and it looks very clean and small.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: @tivv @chtyim After going through this attempt, I discovered a potential issue with using the default configuration:

If you look at the `setCodecClasses` method, the CompressionCodecFactory will also load CompressionCodecs in the classpath using the ServiceLoader: https://hadoop.apache.org/docs/r2.7.3/api/org/apache/hadoop/io/compress/CompressionCodecFactory.html#setCodecClasses(org.apache.hadoop.conf.Configuration,%20java.util.List).

This means the `headMap` method call will always return multiple potential candidates. The default implementation then, picks the last result in this new map:

image

I don't think we can get around this limitation (unless we override every single possible file extension with our codec) but this does not seem like a good idea.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: The DecompressorStream gives me a place where I can check for partition boundaries and align the reads with the partition boundary. This ensures the Line Record Reader is able to read all the lines in a partition: https://github.com/cdapio/hydrator-plugins/blob/125413aeb91761443be156c954426dc91869793a/format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java#L159

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
fernst: I wanted to make it clear that this codec is supposed to be used only for decoding into UTF-8, not viceversa.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
fernst: Good point. Removed.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: I've added the missing class. I forgot to stage it for commit.

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
tivv: Yep, it would be very hacky to work around ServiceLoader. We would need to change the context classloader. Not worth it in my opinion.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
tivv: I am worrying that `DefaultCodec` is still marked as `@InterfaceStability.Evolving`, so it's subject to change and less we override less problems we would have. Also we would still have two methods throwing unsupported exception that should explain the intent

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
tivv: You are using `SplitCompressionInputStream` in there. In it you are building upon `FixedLengthCharsetTransformingDecompressorStream` that extends `DecompressorStream` My question is if it simpler to implement SplitCompressionInputStream directly without additional layers, especially extending `DecompressorStream` in a non-trivial fragile way. The one example implementation of `SplitCompressionInputStream` i hadoop is `BZip2CompressionInputStream` that do not use DecompressorStream underneath, but works with BZIP algorithm directly.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FileBatchSource.java
fernst: Fixed it.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
fernst: Good catch. I'll fix this.

format-common/src/main/java/io/cdap/plugin/format/input/CharsetTransformingPathTrackingInputFormat.java
fernst: You are indeed correct. Let me fix this.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
fernst: Good catch. I've added additional logic to only execute this operation when there is no other option (Meaning, we were not able to add a single byte to the output.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingCodec.java
fernst: Good point. I've removed this method so the superclass method gets called instead.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressor.java
tivv: I actually think this can be our problem and not partition alignment. If we read something from input and then don't return it to user we can't give proper position information and records are lost

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: The DecompressorStream gives us an avenue to test the partition boundaries in https://github.com/cdapio/hydrator-plugins/blob/29aaaf55010acbc6fb19c775a3f64fbfb2c7e3df/format-common/src/test/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStreamTest.java.

I do not think the encapsulation/additional layers really pose a problem for us. I tried to keep the implementation as similar as possible to the original LineRecordReader class I copied for simplicity/maintainability in the future.

tivv: Changes for that test is actually one of the biggest indicators of structure issues. Having a method like `getCompressedDataSuper` is a big smell of structural problems. Let's sync up on this partition boundary synchronization, with a good integration test that shows the problem we are trying to solve.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: You are right. I'm removing this method and just let the call to the super method happen as usual.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: The purpose of the test is to demonstrate that the super method exceeds the partition boundary, which poses a problem when reading lines.

I will come up with a scenario to demonstrate this issue tomorrow.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
```diff
@@ -0,0 +1,109 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.charset.fixedlength;
+
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
```
tivv: Some imports are not used. Please optimize imports across files.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,120 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(FixedLengthCharsetTransformingDecompressorStream.class);
```
tivv: I don't think this one is used

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,120 @@
...
+
+  //Starting and ending position in the file.
+  private final long start;
+  private final long end;
+  private final FixedLengthCharset fixedLengthCharset;
```
tivv: I don't think you use this value outside of constructor, no need for a field

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  /**
+   * Returns the maximum of bytes to consume from the input stream
+   * Since the input is compressed, there is no way to accurately determine how many bytes we need to consume.
+   *
+   * @param pos Current file position
+   * @return Number of bytes to consume.
+   */
+  private int maxBytesToConsume(long pos) {
```
tivv: nit: pos is not used. It may make sense even to remove the method

tivv (commented):
Some style comments. Also as discussed, please add EBCDIC and remove the call super method

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
fernst: Done.

fernst:
The File Encoding selector looks like this: 
image

image

format-common/src/main/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReader.java
fernst: Removed.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: Good point. Removed.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
fernst: You're right. Removed.

core-plugins/widgets/File-batchsource.json
```diff
@@ -154,6 +154,82 @@
           "label": "File System Properties",
           "name": "fileSystemProperties"
         },
+        {
+          "widget-type": "select",
+          "label": "File encoding",
+          "name": "fileEncoding",
+          "widget-attributes": {
+            "values": [
+              "UTF-8 (default)",
```
albertshau: Should just name this "UTF-8", as the default is already controlled by the "default" widget-attribute.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -133,6 +140,17 @@ public void validate(FailureCollector collector) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_SCHEMA).withStacktrace(e.getStackTrace());
     }
 
+    if (fileEncoding != null && !fileEncoding.isEmpty()) {
```
albertshau: there was another change that was recently merged adding a shouldGetSchema() method (see #1288). That new method also needs to be updated to also check that the fileEncoding is not macro enabled.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -198,6 +216,15 @@ public boolean skipHeader() {
     return skipHeader == null ? false : skipHeader;
   }
 
+  @Nullable
+  public String getFileEncoding() {
+    return fileEncoding;
+  }
+
+  public String getDefaultFileEncoding() {
```
albertshau, line 225: is this needed? Seems simpler to just have getFileEncoding() with `return fileEncoding == null || fileEncoding.isEmpty() ? DEFAULT_FILE_ENCODING : fileEncoding`

format-common/src/test/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+package io.cdap.plugin.format.charset.fixedlength;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+import static io.cdap.plugin.format.plugin.AbstractFileSourceConfig.cleanFileEncodingName;
```
albertshau: style: we only import the class, not the static method.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -133,6 +140,17 @@ public void validate(FailureCollector collector) {
       collector.addFailure(e.getMessage(), null).withConfigProperty(NAME_SCHEMA).withStacktrace(e.getStackTrace());
     }
 
+    if (fileEncoding != null && !fileEncoding.isEmpty()) {
+      fileEncoding = cleanFileEncodingName(fileEncoding);
```
albertshau: it's better not to mutate any fields, as it could lead to weird bugs if getFileEncoding() is called before validate().

Instead, it seems like we should change the getter to clean the encoding name before returning it. This incurs the extra cost of cleaning per get call, but it shouldn't be a big deal.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -210,4 +237,16 @@ public Schema getSchema() {
...
     return copyHeader;
   }
+
+  /**
+   * Takes the first word of the file encoding string, as any further words are just charset descriptions.
+   *
+   * @param fileEncoding The file encoding parameter supplied by the user
+   * @return the cleaned up file encoding name
+   */
+  public static String cleanFileEncodingName(String fileEncoding) {
```
albertshau: ideally we would just configure the UI to use a different display name than the actual value instead of performing some cleanup in the backend. However, I don't think there's a way to do this for the select widget.  @elfenheart can you confirm?

albertshau (approved):
I didn't take a close look at the encoding logic, I'm assuming the other reviews have that covered.

Otherwise, just some small comments.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
albertshau: Looks like there is a way, using:

```
options: [ { label: 'this is the label', value: 'value' }  ]
```

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
```diff
@@ -0,0 +1,97 @@
...
+import java.nio.charset.CharsetEncoder;
+import java.util.Map;
+
+/**
+ * Enumeration containing all currently supported Fixed Length charsets.
+ * <p>
+ * This currently includes:
+ * - UTF-32,
+ * - ISO-8859 variants supported by Java
+ * - Windows single-byte code pages supported by Java.
```
tivv: Please update the javadoc

tivv (approved):
LGTM with a minor javadoc comment, but please also address @albertshau comments

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
fernst: I didn't know the UI supported this configuration format. I'll fix it.

tivv:
> The File Encoding selector looks like this:
> image
> 
> image

Is it a combo box with a fixed set of values or does it allow user to enter his own?

tivv:
FYI: this is full list of single-byte encodings that has an encoder in java 8: 
```
IBM-Thai
IBM00858
IBM01140
IBM01141
IBM01142
IBM01143
IBM01144
IBM01145
IBM01146
IBM01147
IBM01148
IBM01149
IBM037
IBM1026
IBM1047
IBM273
IBM277
IBM278
IBM280
IBM284
IBM285
IBM290
IBM297
IBM420
IBM424
IBM437
IBM500
IBM775
IBM850
IBM852
IBM855
IBM857
IBM860
IBM861
IBM862
IBM863
IBM864
IBM865
IBM866
IBM868
IBM869
IBM870
IBM871
IBM918
ISO-8859-1
ISO-8859-13
ISO-8859-15
ISO-8859-2
ISO-8859-3
ISO-8859-4
ISO-8859-5
ISO-8859-6
ISO-8859-7
ISO-8859-8
ISO-8859-9
JIS_X0201
KOI8-R
KOI8-U
TIS-620
US-ASCII
windows-1250
windows-1251
windows-1252
windows-1253
windows-1254
windows-1255
windows-1256
windows-1257
windows-1258
x-IBM1006
x-IBM1025
x-IBM1046
x-IBM1097
x-IBM1098
x-IBM1112
x-IBM1122
x-IBM1123
x-IBM1124
x-IBM1166
x-IBM737
x-IBM833
x-IBM856
x-IBM874
x-IBM875
x-IBM921
x-IBM922
x-iso-8859-11
x-MacArabic
x-MacCentralEurope
x-MacCroatian
x-MacCyrillic
x-MacDingbat
x-MacGreek
x-MacHebrew
x-MacIceland
x-MacRoman
x-MacRomania
x-MacSymbol
x-MacThai
x-MacTurkish
x-MacUkraine
x-windows-874
```
Groovy code to generate (easy to run in Intellij groovy console):
```
import java.nio.charset.Charset

Charset.availableCharsets().values().each { c ->
    if (c.canEncode() && c.newEncoder().maxBytesPerChar() == 1) {
        System.out.println(c.name())
    }
}
```

fernst:
@tivv Right now, I have a predefined list of encodings as I have tested those. This is actually a good question for @chtyim, should we simply add every single encoding that java supports to the list we display to our customers?

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
fernst: We still need this to check wether the configured encoding is not UTF-8 (which is the default), as in this case, there is no need for us to enable this decompressor.

tivv:
Well, it's not every encoding java supports, it's every single-byte encoding java supports. Since we do support those with the latest changes

format-common/src/test/java/io/cdap/plugin/format/charset/CharsetTransformingLineRecordReaderTest.java
```diff
@@ -0,0 +1,154 @@
...
+    decompressorStream =
+      new FixedLengthCharsetTransformingDecompressorStream(inputStream, FixedLengthCharset.UTF_32, 0, 32);
+    compressionInputStream = new TransformingCompressionInputStream(decompressorStream, 0, 32);
+
+    //Set up test
+    setUpRecordReaderForTest(compressionInputStream);
+
+    //Ensure the first line is read
+    Assert.assertTrue(recordReader.nextKeyValue());
+    Assert.assertEquals(recordReader.getCurrentKey().get(), 0);
```
tivv: Almost forgot. In this version of JUnit you do `assertEquals(expected, actual)`. Please fix, otherwise errors are looking weird.

fernst: Good catch. I'll fix that.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharset.java
fernst: Updated.

format-common/src/main/java/io/cdap/plugin/format/charset/fixedlength/FixedLengthCharsetTransformingDecompressorStream.java
```diff
@@ -0,0 +1,102 @@
...
+  private final long end;
+  private long totalReadBytes = 0;
+
+  public FixedLengthCharsetTransformingDecompressorStream(InputStream in,
+                                                          FixedLengthCharset fixedLengthCharset,
+                                                          long start,
+                                                          long end)
+    throws IOException {
+    super(in, new FixedLengthCharsetTransformingDecompressor(fixedLengthCharset), 4096);
+    in.skip(start);
```
tivv: Please cast to seekable and use `seek` instead. E.g. on S3 filesystem FSDataInputStream is constructed on top of S3InputStream that do not override default skip, so skip is done in old way - by reading and skipping all the data.

fernst: @tivv I've amended my commit with your suggestion: https://github.com/cdapio/hydrator-plugins/pull/1255/files#diff-805003eb518a2ad27b1fb8bec65c7fffdbb0bbc9cce75df9edd1d7b2edd161d1R56

tivv:
🎈

PR #1261, opened 2020-12-05
[CDAP-17447] set socketfactory for Cloud SQL sinks

rmstar:


database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
```diff
@@ -123,16 +138,38 @@ private Connection getConnection(Configuration conf) {
     Connection connection;
     try {
       String url = conf.get(DBConfiguration.URL_PROPERTY);
+      if (url.contains("com.google.cloud.sql.mysql.SocketFactory")) {
+        delegateClass = "com.google.cloud.sql.mysql.SocketFactory";
+        socketFactory = "io.cdap.socketfactory.mysql.SocketFactory";
+        url = url.replace("com.google.cloud.sql.mysql.SocketFactory",
```
nitinmotgi: Can use delegate class instead of string or they can all be turned into private statics no ?

database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
```diff
@@ -123,16 +138,38 @@ private Connection getConnection(Configuration conf) {
     Connection connection;
     try {
       String url = conf.get(DBConfiguration.URL_PROPERTY);
+      if (url.contains("com.google.cloud.sql.mysql.SocketFactory")) {
```
chtyim: Please parse the URI instead of doing string replacement.

chtyim: Also, we shouldn't be matching specifically for google socketfactory. the support should be generic.

rmstar: fixed.

database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
rmstar: Changed to private static.

albertshau:
closed due to age

PR #1263, opened 2020-12-07
Cherrypick broken release2.3- plugin 180

vjosadika:
Original PR:https://github.com/cdapio/hydrator-plugins/pull/1246

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
@@ -107,7 +107,7 @@ public final class DBUtils {
    * Given the result set, get the metadata of the result set and return
    * list of {@link io.cdap.cdap.api.data.schema.Schema.Field},
    * where name of the field is same as column name and type of the field is obtained using
-   * {@link DBUtils#getSchema(String, int, int, int, String)}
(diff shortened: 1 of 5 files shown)


database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -185,7 +190,25 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
       int columnSqlPrecision = metadata.getPrecision(i); // total number of digits
       int columnSqlScale = metadata.getScale(i); // digits after the decimal point
       String columnTypeName = metadata.getColumnTypeName(i);
+<<<<<<< HEAD
```
CuriousVini, line 193: These conflicts are due to missing commits from https://github.com/cdapio/hydrator-plugins/pull/1033

@vravish will open the cherrypicks for that PR. After that we should fix and merge this PR

CuriousVini: cherrypick PR is merged https://github.com/cdapio/hydrator-plugins/pull/1275

bajram-adapt:
fix is applied in different Cherrypick PR after resolving conflicts with this one: #1276

PR #1267, opened 2020-12-11
[CDAP-17447] Emit bytes written metric for database sinks

rmstar:


database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -72,6 +73,7 @@
   public DBRecord(StructuredRecord record, int [] columnTypes) {
     this.record = record;
     this.columnTypes = columnTypes;
+    this.bytesWritten = 0;
```
chtyim: No need. Default value of long field is already 0L.

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -240,14 +249,20 @@ private void writeToDB(PreparedStatement stmt, Schema.Field field, int fieldInde
       switch (fieldLogicalType) {
         case DATE:
           stmt.setDate(sqlIndex, Date.valueOf(record.getDate(fieldName)));
+          bytesWritten += ((Long) record.get(fieldName)).BYTES;
```
chtyim: Long.BYTES is a static constant, shouldn't be accessing it via instance.

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
rmstar: removed.

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
rmstar: fixed

chtyim (approved):
LGTM

PR #1269, opened 2020-12-12
[CDAP-17447] Emit bytes written metric for database sinks

rmstar:
Cherry-pick #1267.

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
@@ -55,9 +55,10 @@ import javax.sql.rowset.serial.SerialBlob;
  * @see org.apache.hadoop.mapreduce.lib.db.DBOutputFormat DBOutputFormat
  * @see DBWritable DBWritable
  */
-public class DBRecord implements Writable, DBWritable, Configurable {
+public class DBRecord implements Writable, DBWritable, Configurable, DataSizeReporter {
   private StructuredRecord record;
   private Configuration conf;
+  private long bytesWritten;
 
   /**
    * Need to cache {@link ResultSetMetaData} of the record for use during writing to a table.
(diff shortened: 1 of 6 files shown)


database-plugins/src/main/java/io/cdap/plugin/DataSizeReporter.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Reports data size.
+ *
+ */
+public interface DataSizeReporter {
+  /**
+   * Get the total size in bytes of data written.
+   * @returns number of bytes written.
+   */
+  long getBytesWritten();
```
MEseifan, line 29: Given this new approach, couldnt we do the same for bytes read? 

We probably shouldnt add it now but it might make sense to add the method to the interface with a TODO to add the counting later. What do you think?

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -262,14 +270,20 @@ private void writeToDB(PreparedStatement stmt, Schema.Field field, int fieldInde
       switch (fieldLogicalType) {
         case DATE:
           stmt.setDate(sqlIndex, Date.valueOf(record.getDate(fieldName)));
+          bytesWritten += Long.BYTES;
```
MEseifan, line 273: Is it possible for this `writeToDB` method to be called multiple times for the same DBRecord? If so the count would be incorrect right?

rmstar: `writeToDB` is called once per DBRecord. Call stack looks like ETLDBOutputFormat.DBRecordWriter.write -> DBRecord.write -> DBRecord.writeToDB

database-plugins/src/main/java/io/cdap/plugin/DataSizeReporter.java
rmstar: Makes sense, I'll do that in a new PR because this is release/6.3 cherry-pick PR for #1267.

MEseifan (approved):

PR #1274, opened 2020-12-15
PLUGIN-102 remove format hardcoding

albertshau:
Removed hardcoded format checks and instead rely on the plugin
framework to raise an error if the plugin doesn't exist.

Also changed the widget to pull the formats from the plugin API
instead of using a hardcoded list. This allows proper support
for users who deploy their own custom formats.


albertshau:
cherry-pick of #1272

supriyapatil-google (approved):
Thanks!

PR #1278, opened 2020-12-15
PLUGIN-102 fix backwards incompatibility from format fix

albertshau:
Keep the same method signature for protected method.


albertshau:
cherry-pick of #1277

MEseifan (approved):

PR #1279, opened 2020-12-16
PLUGIN-16608 convert sets properly to GenericRecord

albertshau:
Fixed a bug where Sets in an array field would cause problems
when used with avro sinks.


CuriousVini (approved):
lgtm

supriyapatil-google (approved):
Thanks!

PR #1293, opened 2020-12-21
CDAP-17486 fix gcs sink deploy

yaojiefeng:
The newly added method should not just return null since the gcs sink config does not implement this method so the format will always be null, so usePlugin will pass a null, which results in deploy failure.


CuriousVini (approved):
lgtm 👍🏼

PR #1295, opened 2020-12-21
PLUGIN-279 default join on null keys to false

yaojiefeng:


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -254,7 +254,7 @@ boolean requiredPropertiesContainMacros() {
   }
 
   boolean isNullSafe() {
-    return joinNullKeys == null ? true : joinNullKeys;
+    return joinNullKeys == null ? false : joinNullKeys;
```
CuriousVini, line 257: This change is backwards incompatible right?

yaojiefeng: Yes, this is intended, look at https://cdap.atlassian.net/browse/PLUGIN-279 for more details

CuriousVini: As a best practice, we always make backwards incompatible changes in major versions. 
Why do we need exception in this case?

sreevatsanraman: +1

yaojiefeng:
After discussion, decided to move this out of 6.3.0, so closing this one

core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
nitinmotgi: Backward incompatible is for all major features. Not for all tiny things that affect users. If this issue was priortized then we need to figure out how to make it without affecting the users. It's not a blanket rule for only making backward incompatible changes in major version. They are mostly for wire-protocol, deprecation and such. @CuriousVini @yaojiefeng @sreevatsanraman

PR #1296, opened 2020-12-23
Handle CDAP Decimal for DBSink

vjosadika:
JIRA Ticket: https://cdap.atlassian.net/browse/PLUGIN-180

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
@@ -285,6 +285,11 @@ public class DBRecord implements Writable, DBWritable, Configurable, DataSizeRep
           stmt.setTimestamp(sqlIndex, Timestamp.from(record.getTimestamp(fieldName).toInstant()));
           bytesWritten += Long.BYTES;
           break;
+        case DECIMAL:
+          BigDecimal value = record.getDecimal(fieldName);
+          stmt.setBigDecimal(sqlIndex, value);
+          bytesWritten += value.unscaledValue().bitLength() / Byte.SIZE + Integer.BYTES;
+          break;
       }
       return;
     }
--- a/database-plugins/src/test/java/io/cdap/plugin/DatabasePluginTestBase.java
+++ b/database-plugins/src/test/java/io/cdap/plugin/DatabasePluginTestBase.java
@@ -194,7 +194,9 @@ public class DatabasePluginTestBase extends HydratorTestBase {
                      "DECIMAL_INT DECIMAL(8, 0)," +
                      "DECIMAL_LONG DECIMAL(11, 0)," +
                      "NUMERIC_INT NUMERIC(8, 0)," +
-                     "NUMERIC_LONG NUMERIC(11, 0)" +
+                     "NUMERIC_LONG NUMERIC(11, 0)," +
+                     "DECIMAL_DOUBLE DECIMAL(10, 2)," +
+                     "NUMERIC_DOUBLE DECIMAL(10, 2)" +
                      ")");
       stmt.execute("CREATE TABLE \"MY_DEST_TABLE\" AS (" +
                      "SELECT * FROM \"my_table\") WITH DATA");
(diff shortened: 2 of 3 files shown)


database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -285,6 +285,11 @@ private void writeToDB(PreparedStatement stmt, Schema.Field field, int fieldInde
           stmt.setTimestamp(sqlIndex, Timestamp.from(record.getTimestamp(fieldName).toInstant()));
           bytesWritten += Long.BYTES;
           break;
+        case DECIMAL:
```
CuriousVini, line 288: `writeToDataOut` this method should also be modified

CuriousVini: what is the existing behavior when database sink writes to DB? Is this change backwards compatible?

ardian4: **What is the existing behavior when database sink writes to DB? Is this change backwards compatible?**

- Yes this change is backwards compatible. There are unit tests to confirm this, also we have created pipelines to test this. It all depends on the field type of the input structure record schema. 

**`writeToDataOut` this method should also be modified**

- `wirteToDataOut` parameter `(DataOutput out)` accepts only primitive types. For logical type Decimal the type will be BYTES. Please let us know if there is a way on how to proceed to modify the `writeToDataOut` method?

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -285,6 +285,11 @@ private void writeToDB(PreparedStatement stmt, Schema.Field field, int fieldInde
...
+          BigDecimal value = record.getDecimal(fieldName);
+          stmt.setBigDecimal(sqlIndex, value);
+          bytesWritten += value.unscaledValue().bitLength() / Byte.SIZE;
```
rmstar: Should we also add the size of the scale (int)?

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
CuriousVini: 1. From the code it looks like existing plugin writes decimal types as bytes. After this change, the db sink will write decimal type as `decimal`. Is it correct? 
https://github.com/cdapio/hydrator-plugins/blob/develop/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java#L323

2. I see

ardian4: Yes

CuriousVini (approved):

CuriousVini:
@vjosadika Please provide previous behavior to confirm this is not a backwards incompatible change.

ardian4:
> @vjosadika Please provide previous behavior to confirm this is not a backwards incompatible change.

@CuriousVini The previous behavior of DB Sink:
1. Was throwing Exception when we were trying to insert a Decimal value to a non-nullable field in DB.
```
PostgreSQL - old sink using decimal:
java.io.IOException: org.postgresql.util.PSQLException: No value specified for parameter 4.

MySQL - old sink using decimal:
Caused by: java.sql.SQLException: Field 'decimal' doesn't have a default value
```
2. Was storing null value for Decimal values when the field was nullable in DB.

The above two scenarios were happening because, for logical types, there was no case for DECIMAL type and the value for decimal column in the statement was always null (L294).

https://github.com/cdapio/hydrator-plugins/blob/9073013b7d41aee8fbe57df9d70e10efc6bf4209/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java#L269-L295

CuriousVini:
I see.. In that case it seems like existing code is not handling the decimal data correctly.

CuriousVini (approved):
lgtm, please squash the commits

bajram-adapt:
squashed

CuriousVini (approved):

PR #1297, opened 2020-12-28
[CDAP-17248] Added UI dropdown for the File Source.

fernst:
Connected encoding selection int he UI with the pipeline properties.

This change depends on https://github.com/cdapio/hydrator-plugins/pull/1255


yaojiefeng (approved):
Can we add an end-to-end unit test which reads from a file source with some encoding after we have all the changes? If so, can you open a JIRA for it?

PR #1298, opened 2020-12-29
CDAP-17522 fix npe for file sink

yaojiefeng:
The issue is we introduced a shouldGetSchema in CDAP-17473, the NPE for this pipeline is because there is a macro field in the source, so shouldGetSchema will return false, so the schema is set as null, but the output format is validating the schema, so causing the NPE.

https://cdap.atlassian.net/browse/CDAP-17522


fernst (commented):
LGTM.

fernst (approved):

PR #1299, opened 2020-12-31
Implement CSV Schema Detection Plugin

dardanxh:
- Engineering doc: https://docs.google.com/document/d/1vj_3Av_c-QEZ9Fx7dxqXHwrSr8ZPh44X/edit#heading=h.gjdgxs
- Jira: https://cdap.atlassian.net/browse/PLUGIN-594


format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -122,47 +179,101 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+   * If the header is kept, the first line of the dataset is taken to specify the column names.
+   * @param override
+   * @param line
+   * @param columnNames
+   * @param columnValues
+   */
+  private String[] setColumnNames(HashMap<String, Schema> override, String line, String[] columnNames,
+                                  String[] columnValues) {
+    if (!getSkipHeader()) {
+      columnNames = line.split(getDefaultDelimiter());
```
rmstar: Why are  we using limit = 0 here vs limit = -1 for columnValues?
```
columnValues = line.split(getDefaultDelimiter(), -1);
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,206 @@
...
+   * e.g. {@code TypeInference.isDouble("3.4 ")} returns {@code false}.<br>
+   * 5. "." is not obligatory.<br>
+   * e.g. {@code TypeInference.isDouble("100")} returns {@code true}.
+   * <P>
+   *
+   * @param value the value to be detected.
+   * @return true if the value is a double type, false otherwise.
+   */
+  public static boolean isDouble(String value) {
+    if (!isEmpty(value) && patternDouble.matcher(value).matches()) {
```
rmstar: Can't we use Double.parseDouble() and check for NumberFormatException instead of doing a complex regex match?
Similar question for other types.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/SchemaDetector.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param columnName
+   * @param dataType
+   */
+
+  public void addDataType(String columnName, DataType dataType) {
+    if (this.state.containsKey(columnName)) {
+      this.state.get(columnName).add(dataType);
+    } else {
+      this.state.put(columnName, EnumSet.noneOf(DataType.class));
```
rmstar: why are we not adding dataType to the EnumSet here?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
dardanxh: - The data type detection solution was adapted from the existing one in wrangler plugin. 

- The `Double.parseDouble()` would parse stringified integers to double values. Then you'll need further functionality to distinct between double values and integer values. 

- `Boolean.parseBoolean(rawVal)`: will return `true` only when `rawVal="true"`, otherwise it will return `false` for any other value as for example: `"false", "foo", "1"`. Thus, you will again need to check whether `rawVal=="false"` primarily.

- Date and time could come up in various formats. The only way to parse the raw date/time value correctly is to match it with a certain date/time format already provided as regex (as it is already done).

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
dardanxh: I refactored this piece of code, and it should be clearer.

To explain the reason behind, consider that you have a CSV dataset ready to infer the schema from. Then:

1. If `keepHeader=True` we split the first row with `limit=0`, and every other record with `limit=-1`. This means:
       - The raw header:`"name,surname,age,address"`  - resolves to - `name`, `surname`,`age`,`address` fields.
       - The value records:`"john,doe,,"` - resolves to - `john`, `doe`, `(empty)`, `(empty)`. Because of `limit=-1`, the empty  values at the end of the line are also considered when detecting the column data type.
      
2. If `keepHeader=False`, every record is split with `limit=-1`. This means:
       -  If the first record is `"john,doe,,"`, the output schema will have `body_1`, `body_2`, `body_3`, `body_4` fields.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/SchemaDetector.java
dardanxh: Seems that I missed it. This is fixed now.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,206 @@
...
+   * Detect if the given value is a boolean type.
+   *
+   * @param value the value to be detected.
+   * @return true if the value is a boolean type, false otherwise.
+   */
+  public static boolean isBoolean(String value) {
+    if (isEmpty(value)) {
+      return false;
+    }
+    if ((value.trim().length() == 4 || value.trim().length() == 5)
```
rmstar: is the length check required?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,206 @@
...
+   * 1. This method support only English locale.<br>
+   * e.g. {@code TypeInference.isDouble("3.4")} returns {@code true}.<br>
+   * e.g. {@code TypeInference.isDouble("3,4")} returns {@code false}.<br>
+   * 2. Exponential notation can be detected as a valid double.<br>
+   * e.g. {@code TypeInference.isDouble("1.0E+4")} returns {@code true}.<br>
+   * e.g. {@code TypeInference.isDouble("1.0e-4")} returns {@code true}.<br>
+   * e.g. {@code TypeInference.isDouble("1.0e-04")} returns {@code true}.<br>
+   * 3. Numbers marked with a type is invalid.<br>
+   * e.g. {@code TypeInference.isDouble("3.4d")} returns {@code false}.<br>
+   * e.g. {@code TypeInference.isDouble("123L")} returns {@code false}.<br>
```
rmstar: 123L is not a double. It also doesn't match patternInteger. So would it be treated parsed as a string? Is that expected?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,206 @@
...
+
+import java.math.BigInteger;
+import java.util.regex.Pattern;
+/**
+ * Type Interface provides utility functions that allow you to detect the types of data.
+ */
+public class TypeInference {
+  private static final Pattern patternInteger = Pattern.compile("^(\\+|-)?\\d+$");
+
+  private static final Pattern patternDouble = Pattern.compile(
```
rmstar: Please add unit tests for the functions in this class.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
rmstar: I see, thanks for the explanation!

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,206 @@
...
+
+  }
+
+  /**
+   * Get big integer from a string.
+   *
+   * @param value
+   * @return big integer instance , or null if numer format exception occurrs.
+   */
+  public static BigInteger getBigInteger(String value) {
```
rmstar: Where is this used?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
dardanxh: Unit tests added

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
dardanxh: The functionality to handle long numbers is added. Please check @rmstar.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
dardanxh: Deleted.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
dardanxh: Nope. Removed this.

rmstar:
Please add a link to the JIRA in the PR description

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/DataType.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright © 2017-2019 Cask Data, Inc.
```
rmstar: fix year

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/DateTimePattern.java
```diff
@@ -0,0 +1,177 @@
+/*
+ * Copyright © 2017-2019 Cask Data, Inc.
```
rmstar: fix year

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,203 @@
+/*
+ * Copyright © 2017-2019 Cask Data, Inc.
```
rmstar: fix year

format-delimited/src/main/java/io/cdap/plugin/format/delimited/dq/TypeInference.java
```diff
@@ -0,0 +1,203 @@
...
+   * e.g. {@code TypeInference.isDouble("1.0e-4")} returns {@code true}.<br>
+   * e.g. {@code TypeInference.isDouble("1.0e-04")} returns {@code true}.<br>
+   * 3. Numbers marked with a type is invalid.<br>
+   * e.g. {@code TypeInference.isDouble("3.4d")} returns {@code false}.<br>
+   * e.g. {@code TypeInference.isDouble("123L")} returns {@code false}.<br>
+   * 4. White space is invalid.<br>
+   * e.g. {@code TypeInference.isDouble(" 3.4")} returns {@code false}.<br>
+   * e.g. {@code TypeInference.isDouble("3.4 ")} returns {@code false}.<br>
+   * 5. "." is not obligatory.<br>
+   * e.g. {@code TypeInference.isDouble("100")} returns {@code true}.
```
rmstar: We are checking for double before long in https://github.com/cdapio/hydrator-plugins/pull/1299/files#diff-2c6f9c194aea0b181a6f3755660fe4387e152ae783802eff6bf9cdc8602c590bR192. 
Would the type of a  large number like Integer.MAX_VALUE + 1 be inferred as double instead of long?

dardanxh: `Integer.MAX_VALUE+1` would be inferred as long. 
`Long.MAX_VALUE+1` would be inferred as String.

dardanxh: @rmstar could you check the latest changes I've made to this plugin.

CuriousVini:
Followup:
1.) For CSV schema, Are we honoring skip header in case there is header in the file?
2.) For CSV schema, Are we honoring regex file path filter to generate schema?

dardanxh:
> Followup:
> 1.) For CSV schema, Are we honoring skip header in case there is header in the file?
> 2.) For CSV schema, Are we honoring regex file path filter to generate schema?

@CuriousVini 
In the current plugin development state, we:

1.) Always honor the `skipHeader` decision, even when we have the header data. 

As I'm thinkging, it's impossible comming with a reliable solution that correctly identifies and distincts header data from value data. Of course a solution that checks a number of rules as: 
- Check if all values are not null, 
- Check if all values are unique
- Check if all values are of string data

could be achieved. Are we interested on this?

2.) Don't honor regex file path when generating the schema. Should I implement this?

core-plugins/src/main/java/io/cdap/plugin/batch/source/FileSourceConfig.java
```diff
@@ -50,6 +50,20 @@
     + "This is an advanced feature that requires knowledge of the properties supported by the underlying filesystem.")
   private String fileSystemProperties;
 
+  @Macro
+  @Nullable
+  @Description("Specifies a list of columns for whom the user would like to specify the data types explicitly, " +
+    "instead of relying upon Data Fusion to detect the type automatically. For these columns," +
+    " Data Fusion should not detect the type automatically.")
```
nitinmotgi: Don’t put product names in configuration and also shorten the description

core-plugins/src/main/java/io/cdap/plugin/batch/source/FileSourceConfig.java
```diff
@@ -50,6 +50,20 @@
...
+  private String override;
+
+  @Macro
+  @Nullable
+  @Description("Specifies the number of records that should be pulled in into the sample for automatically detecting " +
+    " data types. The first N records should be used as the sample.")
```
nitinmotgi: Make description user friendly not using technical terms like first N

nitinmotgi (changes requested):
Minor changes request for making descriptions more user friendly and not to use technical terms.

dardanxh:
@CuriousVini I answered all the open questions for the CSV Schema Detection plugin in the document you find in the link below. 
Let me know on how to proceed further with some of the points I've stated there.

- https://docs.google.com/document/d/1Op3ilhbQio2FkkWQ-ysn-1ObCNpTi5-EuluDhFw20hQ

CuriousVini:
1. could you please file a jira to honor regex? we should honor that. 
2. Also, If the directory being pointed at has multiple files with same/different schema, what file will be picked for schema detection? If its a union of schema fields, are all the fields nullable?  
3. For skip header, if we always honor it, what would happen if files does not have any header? 

Please include following cases in integration test for this functionality:
1. Single file path
2. Path containing multiple files with different schema
3. Path containing multiple files with different schema + regex expression provided
4. Path containing multiple files with different schema without Header

CuriousVini:
From the PRD, has following been tested as part of performance tests:
```
Should produce performance bounds for the following scenarios:

Directory containing 1TB data on GCS in a single level (50 files)
Directory containing subdirectories amounting to 1TB data in total
Single 1TB file

**Comment by dardanxh** [comment]
> 1. could you please file a jira to honor regex? we should honor that.
> 2. Also, If the directory being pointed at has multiple files with same/different schema, what file will be picked for schema detection? If its a union of schema fields, are all the fields nullable?
> 3. For skip header, if we always honor it, what would happen if files does not have any header?
> 
> Please include following cases in integration test for this functionality:
> 
> 1. Single file path
> 2. Path containing multiple files with different schema
> 3. Path containing multiple files with different schema + regex expression provided
> 4. Path containing multiple files with different schema without Header

@CuriousVini 

1. Jira ticket created: https://cdap.atlassian.net/browse/PLUGIN-592. I will include this functionality.

<hr>
2. I'll answer this in two parts:

**Part I**: In a given directory, **every single file** is investigated for `sampleSize` amount of lines. Do we need to assume that all files in a directory have the same schema? My two suggestions are:

**Suggestion 1**: We assume that a directory only consist of files with the same schema. In this case we pick up a single file from the directory and infer the schema out of it. If another file with different schema is met, we throw an error.

**Suggestion 2**: We assume that a directory might consist of files with different schemas. In this case we ouptut the schema after we have investigated all the fiels in the given directory.

**Part II**: If files with different schemas are residing in the same directory, the CSV schema detection plugin takes the union of all the files' schemas. Each field could either be null or the appropriate inferred data type.

I've tested this by detecting the union schema of two files (with different schemas) and run a pipeline `File Source -> File Sink`. The plugin that reads the data after, is considering that `length(schema.fields) == length(file.fields)` and that `schema.fields` are 1-to-1 mapped (with preserving order) to columns in the data file. Since in our case `length(union_schema.fields) >= length(single_file.fields)` the plugin that reads the data after is throwing errors.

https://github.com/cdapio/hydrator-plugins/blob/34022cfd953e36822d19ec6f3e010463ca8a443c/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java#L77

How should we continue with this? My two suggestions are:

**Suggestion 1**: We assume that a directory only consist of files with the the same schema. In this case, If another file with different schema is met, we throw an error.

**Suggestion 2**: We should include the functionality to make the plugin read the data without assuming that `length(schema.fileds) == length(file.columns)`. 
<hr>
3. This is the behavior of the pluing in those cases:

**Case 1**: Data file with field names (ie. first row: `name;surname;age`)
- If `skipHeader == false` -> field names are taken from the file -> `name;surname;age`
- If `skipHeader == true` -> field names are generated automatically -> `body_1;body_2;body_3`

**Case 2**: Data file without field names (ie. first row `John;Doe;25`)
- If `skipHeader == false` -> field names are taken from the file. The plugins tries to set `John;Doe;25` as schema field names and fails to do so since 25 cannot be a schema file. This will throw an `invalid field name: 25` error. In case that the first row consist of all strings (eg. `John;Doe;Denver`) that are validated correctly as schema field names, the field names will be `John;Doe;Denver`.
- If `skipHeader == true` -> column names are generated automatically -> `body_1;body_2;body_3`

**Comment by CuriousVini** [comment]
1. Lets add this capability. 
2. We should not scan whole directory with every single file for schema detection. It will be a very expensive operation if there is large number of files involved. In that case, we can assume that all the files in a directory has same schema. If not, regex should be used. And we always pick one file to detect the schema. 
3. Case 1 behavior looks correct. Case 2 - if `skipHeader == false` then does it throw validation error? Also the error message should be more action oriented. `invalid field name: 25.` does not provide information to the user on what they should do.

**Comment by dardanxh** [comment]
> 1. Lets add this capability.
> 2. We should not scan whole directory with every single file for schema detection. It will be a very expensive operation if there is large number of files involved. In that case, we can assume that all the files in a directory has same schema. If not, regex should be used. And we always pick one file to detect the schema.
> 3. Case 1 behavior looks correct. Case 2 - if `skipHeader == false` then does it throw validation error? Also the error message should be more action oriented. `invalid field name: 25.` does not provide information to the user on what they should do.

@CuriousVini changes addressed. Please check.

File: format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java [hunk 14]
```diff
@@ -99,7 +99,7 @@ public Schema getDefaultSchema(FormatContext context) throws IOException {
       for (Map.Entry<String, String> entry : conf.getFileSystemProperties().entrySet()) {
         hconf.set(entry.getKey(), entry.getValue());
       }
-      final Path file = conf.getFilePathForSchemaGeneration(filePath, "parquet", hconf);
+      final Path file = conf.getFilePathForSchemaGeneration(filePath, ".+\\.parquet", hconf);
```
CuriousVini, line 102: why is this change needed?

format-delimited/pom.xml
```diff
@@ -29,6 +29,13 @@
   <name>Delimited format plugins</name>
   <packaging>jar</packaging>
 
+  <properties>
+    <google-cloud-storage.version>1.112.0</google-cloud-storage.version>
```
CuriousVini: why are all these dependencies added?

CuriousVini (commented):
Please add unit tests for all the strings that can be parsed to derive types.

CuriousVini:
Also please make appropriate plugin docs changes

format-delimited/pom.xml
dardanxh: It was a misunderstanding from my part on this. Already clarified with the team. Will refactor the solution and push the changes.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
dardanxh: In the previous version, method `getFilePathForSchemaGeneration` could list one file from a given file system path. It provided the functionality to filter paths by exstension (parquet and avro in that case).

In the current version, since we already use this function to list dir and pick up one file from the file system, I extended a bit the functionality to fitler files by regex rather than a pure string. Doing so, I could now pass the "file filter regex" value in this method, and get only the file that matches the regex. 

Since this method was used in the `ParquetInputFormatProvider` and `AvroInputFormatProvider` classes, I replaced the previous pure strings to a regex expression (ie. `parquet` -> `.+\\.parquet` and `avro` -> `.+\\.avro`).

dardanxh:
@CuriousVini I refactored the solution and dropped off the uneccessary functionality. Please check.

CuriousVini:
@dardanxh Please let me know once this has been reviewed internally to avoid backend forth, thanks!

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetector.java
```diff
@@ -0,0 +1,117 @@
...
+  public EnumSet<DataType> getColumnDataTypes(String columnName) {
+    return this.dataTypeInvestigationState.get(columnName);
+  }
+
+  /**
+   * Gets all column names from the data type investigation state.
+   *
+   * @return Column names of the investigated dataset.
+   */
+  public List<String> getAllColumns() {
```
ardian4: We can remove this method if its not used

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -103,66 +124,88 @@ private String getDefaultDelimiter() {
...
-   * @return {@link Schema}
-   * @throws IOException raised when error occurs during schema extraction
+   * @return The detected schema.
+   * @throws IOException If the data can't be read from the datasource.
    */
   public Schema getDefaultSchema(@Nullable FormatContext context) throws IOException {
-    final String format = getProperties().getProperties().getOrDefault(FORMAT, "delimited");
-    String delimiter = getProperties().getProperties().get(DELIMITER);
+    final String format = getProperties().getProperties().getOrDefault(NAME_FORMAT, "delimited");
+    String delimiter = getProperties().getProperties().get(NAME_DELIMITER);
```
ardian4: Here we need to use getDefaultDelimiter() method for delimiter

dardanxh:
> @dardanxh Please let me know once this has been reviewed internally to avoid backend forth, thanks!

@CuriousVini the PR is reviewed internally, and the identified issues are fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetector.java
```diff
@@ -0,0 +1,137 @@
...
+  public static Schema detectColumnDataType(EnumSet<DataType> dataTypes) {
+    DataType dataType = dataTypes.stream().max(Comparator.comparing(v -> v.priority)).orElse(DataType.STRING);
+    boolean isNull = dataTypes.stream().anyMatch(v -> v.name.equalsIgnoreCase(DataType.EMPTY.name));
+
+    switch (dataType) {
+      case STRING:
+        return getSchemaOfType(Schema.Type.STRING, isNull);
+      case INTEGER:
+        return getSchemaOfType(Schema.Type.INT, isNull);
+      case DOUBLE:
```
CuriousVini: How are we distinguishing between BigDecimal and Double?

CuriousVini (commented):
To begin with could we just support ISO 8601 for date time and timestamp? 
Also recently we have added DateTime type. we should include that as well. However, we can add it in the followup PR.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetector.java
dardanxh: - First, every stringified number with a decimal point is positioned under double values. 

Then I perform another check where I see if `Double.parseDouble(rawValue).toString().equals(rawValue)`. If yes, this confirms that the rawValue can be correctly represented in a Double data type.

-  If not, the value is considered to be a BigDecimal. BigDecimal's digits gets shorten when parsed to Double. Hence, I output STRING as a data type.

dardanxh:
> To begin with could we just support ISO 8601 for date time and timestamp?
> Also recently we have added DateTime type. we should include that as well. However, we can add it in the followup PR.

@CuriousVini ISO 8601 is supported now. Please check.

dardanxh:
@rmstar Date and Time Logical types are translated to String now. Please check.

core-plugins/docs/File-batchsource.md
```diff
@@ -28,9 +28,14 @@ Blob - is set by default as field named 'body' of type bytes.
 Text - is set by default as two fields: 'body' of type bytes and 'offset' of type 'long'.
 JSON - is not supported, user has to manually provide the output schema.
 
+**Override:**: A list of columns with the corresponding data types for whom the automatic data type detection gets
```
rmstar: remove extra ':'

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Gets a nullable or non-nullable schema type.
+   *
+   * @param type The given schema type.
+   * @param isNull True if the column investigated for data type detection has null values. Otherwise, false.
+   *
+   * @return The {@link Schema.Type} determined
+   */
+  private static Schema getSchemaOfType(Schema.Type type, boolean isNull) {
+    return !isNull ? Schema.of(type) : Schema.unionOf(Schema.of(Schema.Type.NULL), Schema.of(type));
```
rmstar: Schema.unionOf(...) -> Schema.nullableOf()

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Gets a nullable or non-nullable schema logical type
+   *
+   * @param logicalType The given schema logical type.
+   * @param isNull True if the column investigated for data type detection has null values. Otherwise, false.
+   * @return The {@link Schema.Type} determined.
+   */
+  private static Schema getSchemaOfLogicalType(Schema.LogicalType logicalType, boolean isNull) {
+    return !isNull ? Schema.of(logicalType) : Schema.unionOf(Schema.of(Schema.Type.NULL), Schema.of(logicalType));
```
rmstar: Schema.unionOf(...) -> Schema.nullableOf()

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Gets a nullable or non-nullable schema logical type
+   *
+   * @param logicalType The given schema logical type.
+   * @param isNull True if the column investigated for data type detection has null values. Otherwise, false.
+   * @return The {@link Schema.Type} determined.
+   */
+  private static Schema getSchemaOfLogicalType(Schema.LogicalType logicalType, boolean isNull) {
```
rmstar: where is this called? seems unused..

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -0,0 +1,122 @@
...
+      }
+    }
+
+    for (String column : columnNames) {
+      if (override.containsKey(column)) {
+        fields.add(Schema.Field.of(column, override.get(column)));
+      } else {
+        Schema inferredSchema = DataTypeDetectorStatusKeeper
+                .detectColumnDataType(dataTypeDetectorStatusKeeper.getColumnDataTypes(column));
+        validateSchemaFieldName(column);
```
rmstar: why don't we validate schema field names if the field type has override configured? Maybe move this check into setColumnNames() ?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DateTimePattern.java
```diff
@@ -0,0 +1,130 @@
...
+   *
+   * @param value The given raw string value.
+   * @return True if the given value is of type "DATE", false otherwise.
+   */
+  public static boolean isDate(String value) {
+    if (StringUtils.isEmpty(value)) {
+      return false;
+    }
+
+    if (value.length() < 6 || value.length() > 64) {
```
rmstar: what are 6 and 64? Please add a comment.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DateTimePattern.java
```diff
@@ -0,0 +1,130 @@
...
+   * Check if the value passed is a time or not.
+   *
+   * @param value The given raw string value.
+   * @return true If the given value is of type "TIME", false otherwise.
+   */
+  public static boolean isTime(String value) {
+    if (StringUtils.isEmpty(value)) {
+      return false;
+    }
+    if (value.length() < 4 || value.length() > 24) {
```
rmstar: Why 4 and 24? Please add a comment.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private Map<String, EnumSet<DataType>> dataTypeDetectionStatus = new LinkedHashMap<>();
+
+  /**
+   * Detects the datatype of a single value.
+   *
+   * @param value Value whom the data type would be detected from.
+   * @return The {@link DataType} of the given value.
+   */
+  public static DataType detectsValueDataType(String value) {
```
rmstar: detectsValueDataType -> detectValueDataType

core-plugins/widgets/File-batchsource.json
```diff
@@ -28,6 +28,44 @@
...
+            "value-placeholder": "Data Type",
+            "dropdownOptions": [
+              "boolean",
+              "bytes",
+              "double",
+              "float",
+              "int",
+              "long",
+              "string",
+              "date",
```
rmstar: Are date and time types supported?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
```diff
@@ -0,0 +1,131 @@
...
+
+    switch (dataType) {
+      case INTEGER:
+        return getSchemaOfType(Schema.Type.INT, isNull);
+      case DOUBLE:
+        return getSchemaOfType(Schema.Type.DOUBLE, isNull);
+      case LONG:
+        return getSchemaOfType(Schema.Type.LONG, isNull);
+      case DATE:
+        return getSchemaOfType(Schema.Type.STRING, isNull);
```
rmstar, line 70: Let's add a comment indicating why date and time types are detected as string (PLUGIN-637).

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/TypeInference.java
```diff
@@ -0,0 +1,209 @@
...
+   * @return The inferred {@link DataType} of the given value.
+   */
+  public static DataType getDataType(String value) {
+    if (TypeInference.isEmpty(value)) {
+      return DataType.EMPTY;
+    } else if (TypeInference.isBoolean(value)) {
+      return DataType.BOOLEAN;
+    } else if (TypeInference.isInteger(value)) {
+      return deepInvestigateInt(value);
+    } else if (TypeInference.isDouble(value)) {
```
rmstar: shouldn't we check isLong before isDouble (long has higher priority than double)?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorStatusKeeper.java
dardanxh: Removed.

core-plugins/widgets/File-batchsource.json
dardanxh: `Date` (eg., `2021-12T12:34`, `2021-05-19`, ... ) and `Time` (eg., `22:30`, `23:59:59Z`,...) values of ISO-806 standard **can get detected correctly** as of `Date` and `Time` data types respectively. 

Even though the output schema shows `Date` data type, when the source plugin (eg., File Source) reads such columns in the background, it will consider them as Strings.

@rmstar Should I remove `Date` and `Time` then from the data here?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/TypeInference.java
dardanxh: Ordering here doesn't matter. For the purpose of clarity I've moved `Long` before `Double` detection

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
dardanxh: You're right. Included this functionality at `setColumnNames()` method.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DateTimePattern.java
dardanxh: This check isn't necessary. Since I based the solution a bit from wrangler plugin data type detection, I've forgot removing this.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DateTimePattern.java
dardanxh: Removed!

dardanxh:
@rmstar Comments addressed. Pleace check!

core-plugins/widgets/File-batchsource.json
rmstar: Yes, let's remove date and time if the data is just as strings.

rmstar (approved):

rmstar:
@nitinmotgi is this ok to merge? Looks like you have a pending change requested.

dardanxh:
> Minor changes request for making descriptions more user friendly and not to use technical terms.

@nitinmotgi your change requests are already handled. Could you check and approve them.

@rmstar Changes requested from @nitinmotgi are minor ones (update the description of config params) which are already handled. Should we continue with merging?

nitinmotgi (approved):
lgtm 👍

PR #1313, opened 2021-01-19
Fixes XMLParser to not execute external entities.

ajainarayanan:


albertshau (commented):
not familiar with external entities, are there use cases where they should be enabled? Seems like this should be configurable rather than hardcoded in.

dli357:
Do we need to support document type definitions (DTDs) in the XMLRecordReader? Otherwise we should consider turning it off by setting SUPPORT_DTD to false:
```java
factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
```
External entities seem to only really be useful for validating external XML schemas; do we see a use case for this? The primary risk of XXE would be somebody who has control over the input data files may be able to read local files or make network calls on behalf of cluster on which this plugin is executing.

ajainarayanan:
> not familiar with external entities, are there use cases where they should be enabled? Seems like this should be configurable rather than hardcoded in.

@albertshau Have added necessary configuration for XMLReaderBatchSource and XMLParser. Let me know if I captured everything right. Will do the same for XMLMultiParser.

ajainarayanan:
> Do we need to support document type definitions (DTDs) in the XMLRecordReader? Otherwise we should consider turning it off by setting SUPPORT_DTD to false:
> 
> ```java
> factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
> ```
> 
> External entities seem to only really be useful for validating external XML schemas; do we see a use case for this? The primary risk of XXE would be somebody who has control over the input data files may be able to read local files or make network calls on behalf of cluster on which this plugin is executing.

@dli357 Have added a dynamic filter to only show general, parameter and external doc dtd configurations when  disallow doctype dtd property is set to false.

The logic is, if disallow-doctype-dtd is set to true then it doesn't matter what other property values are. 
If it is set to false then the user can partially enable allowing general, parameter or external dtds based on use case.

By default disallow-doctype-dtd is set to true which prevents xxe for any regular usage.

core-plugins/src/main/java/io/cdap/plugin/batch/source/XMLReaderBatchSource.java
```diff
@@ -296,6 +301,11 @@ public void onRunFinish(boolean succeeded, BatchSourceContext context) {
     @Macro
     private final String temporaryFolder;
 
+    @Description("Enable external entities while processing xml files")
+    @Nullable
+    @Macro
+    private final String enableExternalEntities;
```
albertshau: String -> Boolean

core-plugins/src/main/java/io/cdap/plugin/batch/source/XMLReaderBatchSource.java
```diff
@@ -137,6 +137,10 @@ public void prepareRun(BatchSourceContext context) throws Exception {
       conf.set(XMLInputFormat.XML_INPUTFORMAT_TARGET_FOLDER, config.targetFolder);
     }
 
+    if (!config.containsMacro("enableExternalEntities") && !Strings.isNullOrEmpty(config.enableExternalEntities)) {
```
albertshau: No need to check for macro here, it will always be evaluated at this point.

Should also remove the null or empty check and have a getter method in the config class that checks for null like:

```
public boolean shouldEnableExternalEntities() { 
  return enableExternalEntities == null ? false : true
}
```

This removes the possibility of unintended null pointer exceptions. Then you can simplify this to one line:

```
conf.set(..., String.valueOf(config.shouldEnableExternalEntities()));
```

core-plugins/src/main/java/io/cdap/plugin/batch/source/XMLReaderBatchSource.java
```diff
@@ -387,6 +416,12 @@ void validate(FailureCollector collector) {
             .withConfigProperty(PATH).withConfigProperty(PATTERN);
         }
       }
+
+      if (!containsMacro(ENABLE_EXTERNAL_ENTITIES) && Strings.isNullOrEmpty(enableExternalEntities)) {
```
albertshau: shouldn't have this check, as it should be allowed to not specify a value here for backwards compatibility.

transform-plugins/src/main/java/io/cdap/plugin/XMLMultiParser.java
```diff
@@ -100,6 +100,12 @@ public void initialize(TransformContext context) throws Exception {
     schema = config.getSchema(collector);
     xPathExpression = config.getXPathExpression(collector);
     fieldNames = new HashSet<>();
+    builderFactory.setFeature("http://xml.org/sax/features/external-general-entities", false);
```
albertshau: should these be set based on configuration options?

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -316,6 +344,25 @@ public Config(String inputField, String encoding, String xPathFieldMapping, Stri
       this.fieldTypeMapping = fieldTypeMapping;
       this.processOnError = processOnError;
       this.failOnArray = false;
+      this.enableExternalGeneralEntities = false;
+      this.enableExternalParameterEntities = false;
+      this.loadExternalDTD = false;
+      this.disallowDocTypeDTD = true;
+    }
```
albertshau, line 342: nit: add a newline between constructors

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -316,6 +344,25 @@ public Config(String inputField, String encoding, String xPathFieldMapping, Stri
       this.fieldTypeMapping = fieldTypeMapping;
       this.processOnError = processOnError;
       this.failOnArray = false;
+      this.enableExternalGeneralEntities = false;
```
albertshau: change this method to `this(inputField, encoding, ...., false, false, false, false, false)`

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
```diff
@@ -316,6 +344,25 @@ public Config(String inputField, String encoding, String xPathFieldMapping, Stri
...
+      this.enableExternalParameterEntities = false;
+      this.loadExternalDTD = false;
+      this.disallowDocTypeDTD = true;
```
albertshau: the default in the backend should be false for backwards compatibility. 

The default in the widget can be true if we think that's better a better default moving forward.

core-plugins/src/main/java/io/cdap/plugin/batch/source/XMLRecordReader.java
```diff
@@ -106,6 +107,8 @@ public void initialize(InputSplit split, TaskAttemptContext context) throws IOEx
     Configuration conf = context.getConfiguration();
     fs = file.getFileSystem(conf);
     XMLInputFactory factory = XMLInputFactory.newInstance();
+    enableExternalEntities = Boolean.valueOf(conf.get(XMLInputFormat.XML_INPUTFORMAT_ENABLE_EXTERNAL_ENTITIES));
```
dli357: Can we also add supportDtd as a configuration option? I know you responded to my comment earlier about DTD support in XMLParser and XMLMultiParser, but from what I understand XMLInputFactory used in XMLRecordReader is a separate XML parser (StAX) from XMLParser and XMLMultiParser's DocumentBuilder (JAXP) and thus we will need a separate switch for both external entities and supporting DTD:
```
factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
```
[More Info here](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#xmlinputfactory-a-stax-parser).

transform-plugins/src/main/java/io/cdap/plugin/XMLParser.java
dli357: +1, would prefer the default in the widget be external entities and DTDs disabled. It should be fine to have an empty plugin leave external entities enabled, but we should consider that copying pipeline JSONs will persist this "vulnerability".

transform-plugins/src/main/java/io/cdap/plugin/XMLMultiParser.java
ajainarayanan: Yeah. Wanted to check if I got everything for XMLParser before modifying XMLMultiParser.

ajainarayanan:
@albertshau @dli357 Have addressed comments. Could you give it another review.

core-plugins/docs/XMLReader-batchsource.md
```diff
@@ -37,6 +37,9 @@ will not be processed again on future runs of the pipeline.
 
 **Table Data Expiry Period (Days):** The amount of time (in days) to wait before clearing the table used to track processed filed. If omitted, data will not expire in the tracking table. Example: for `tableExpiryPeriod = 30`, data before 30 days is deleted from the table.
 
+**Enable external entities:** This enables processing external entities while reading xml file. Defaults to `false`.
```
dli357: nit: Can we add a warning here that external entities and DTD support should be disabled unless absolutely necessary for security risks? That way users can be aware of the risk of using these options. Same comment with the configuration options for XMLParser and XMLMultiParser. Thanks!

albertshau (approved):

dli357 (approved):

gurshafriri:
@ajainarayanan is there a plan release to release a version with this fix in the near future? 
As the details for this are public we (at [Snyk](https://snyk.io)) want to add this to our open-source vulnerability DB, better with a fixed version at hand.

ajainarayanan:
> @ajainarayanan is there a plan release to release a version with this fix in the near future?
> As the details for this are public we (at [Snyk](https://snyk.io)) want to add this to our open-source vulnerability DB, better with a fixed version at hand.

@gurshafriri The changes are slated for release 2.5 of Hydrator plugins (https://github.com/cdapio/hydrator-plugins/pull/1318) which gets shipped in 6.4 version of CDAP. Version 6.4 of CDAP should generally available by beginning of April 2021.

PR #1319, opened 2021-02-05
Refactored Avro Output Format to allow writing records with no predefined schema

fernst:


format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,87 @@
...
+                                           dataModel,
+                                           codecFactory,
+                                           outputStream,
+                                           syncInterval);
+        } catch (IOException ioe) {
+          throw new RuntimeException("Unable to initialize Avro Record Writer", ioe);
+        }
+      });
+
+    delegate.write(new AvroKey<>(key.datum().getGenericRecord()), value);
```
fernst: Note that the trick here is that, even when my class expects that I will write the entire GenericRecordWrapper, i just write the inner GenericRecord.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/GenericRecordWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * TODO: add
+ */
+public class GenericRecordWrapper {
+  private final GenericRecord genericRecord;
+  private final Schema schema;
+  private final int hash;
+
+  public GenericRecordWrapper(GenericRecord genericRecord, Schema schema, int hash) {
```
fernst: I use the Hash for uniqueness (only one writer per distinct schema) and the Schema entity is used to lazily initialize the writer.

format-avro/src/main/java/io/cdap/plugin/format/avro/UnstructuredToAvroTransformer.java
```diff
@@ -0,0 +1,85 @@
+
+/*
+ * Copyright © 2018-2019 Cask Data, Inc.
```
albertshau: 2021

format-avro/src/main/java/io/cdap/plugin/format/avro/UnstructuredToAvroTransformer.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Creates GenericRecords from StructuredRecords without knowing the schema beforehand.
+ */
+public class UnstructuredToAvroTransformer extends RecordConverter<StructuredRecord, GenericRecordWrapper> {
```
albertshau: the StructuredToAvro class allows passing in a null output schema, is this class different?

format-avro/src/main/java/io/cdap/plugin/format/avro/output/AvroOutputFormatProvider.java
```diff
@@ -83,22 +85,32 @@ public String getOutputFormatClassName() {
...
 
     @Macro
     @Nullable
     @Description(CODEC_DESC)
     private String compressionCodec;
+
+    @Macro
+    @Nullable
+    @Description(ALLOW_FLEXIBLE_SCHEMA)
+    private String allowFlexibleSchema;
```
albertshau: should be enough to make the schema optional, don't think we need this extra setting.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyOutputFormat.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.avro.mapreduce.AvroKeyOutputFormat;
+import org.apache.avro.mapreduce.AvroOutputFormatBase;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+import java.io.IOException;
+
+/**
+ * TODO: add
```
albertshau: javadoc

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * TODO: add
```
albertshau: javadoc

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+      });
+
+    delegate.write(new AvroKey<>(key.datum().getGenericRecord()), value);
+  }
+
+  @Override
+  public void close(TaskAttemptContext context) throws IOException, InterruptedException {
+    for (RecordWriter<AvroKey<GenericRecord>, NullWritable> delegate : delegateMap.values()) {
+      delegate.close(context);
```
albertshau: if one of these throws IOException, should still try to close the others before eventually throwing an exception.

If more than one throws an exception, can use .addSuppressed() to make sure all exceptions are captured in whatever is eventually thrown.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/GenericRecordWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+ * the License.
+ */
+
+package io.cdap.plugin.format.avro.output;
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+
+/**
+ * TODO: add
```
albertshau: javadoc

format-avro/src/main/java/io/cdap/plugin/format/avro/output/GenericRecordWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class GenericRecordWrapper {
+  private final GenericRecord genericRecord;
+  private final Schema schema;
```
albertshau: doesn't GenericRecord have a getSchema() method? Does this need to be a separate field?

albertshau (commented):
looks like there's also an unintended change to the google-cloud submodule

albertshau:
Will also need to update the docs that say avro format requires a schema.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/GenericRecordWrapper.java
fernst: You are correct, I hadn't noticed the GenericRecord contained it's own copy of the schema.

format-avro/src/main/java/io/cdap/plugin/format/avro/UnstructuredToAvroTransformer.java
fernst: I've removed this class.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/GenericRecordWrapper.java
fernst: I was able to simplify the code thanks to this realization.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/AvroOutputFormatProvider.java
fernst: I would argue that we should add this as a way to ensure that a missing macro somewhere doesn't cause a customer's existing pipeline to start outputting records in a different format (they expect a failure but suddenly they get records with no control over how fields appear).

Let me know what you think.

format-avro/src/main/java/io/cdap/plugin/format/avro/UnstructuredToAvroTransformer.java
fernst: I removed this class.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyOutputFormat.java
fernst: Added.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
fernst: Added.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
fernst: I have implemented this logic.

fernst:
@albertshau I will update the docs next week. For now, I want to focus on 6.1.4 regression for this changes in order to deliver the artifacts to the customer today.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/AvroOutputFormatProvider.java
albertshau: if the schema was a macro, it had to have been filled out at runtime, otherwise the run fails right at the start. We don't allow specifying a macro and then not providing a value for it. The macro check here is only done so that it doesn't throw a NPE at validation time.

format-avro/src/main/java/io/cdap/plugin/format/avro/StructuredToAvroTransformer.java
```diff
@@ -49,7 +49,7 @@ public GenericRecord transform(StructuredRecord structuredRecord) throws IOExcep
 
   @Override
   public GenericRecord transform(StructuredRecord structuredRecord,
-                                 io.cdap.cdap.api.data.schema.Schema schema) throws IOException {
+                                        io.cdap.cdap.api.data.schema.Schema schema) throws IOException {
```
albertshau: unintended change?

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyOutputFormat.java
```diff
@@ -0,0 +1,52 @@
...
+public class DelegatingAvroKeyOutputFormat extends AvroOutputFormatBase<AvroKey<GenericRecord>, NullWritable> {
+
+  @Override
+  public RecordWriter<AvroKey<GenericRecord>, NullWritable> getRecordWriter(TaskAttemptContext context)
+    throws IOException, InterruptedException {
+    Function<TaskAttemptContext, OutputStream> outputStreamSupplier = (ctx) -> {
+      try {
+        return getAvroFileOutputStream(ctx);
+      } catch (IOException e) {
+        throw new RuntimeException("Unable to get Outpus Stream", e);
```
albertshau: typo: Outpus

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,95 @@
...
+          throw new RuntimeException("Unable to initialize Avro Record Writer", ioe);
+        }
+      });
+
+    delegate.write(key, value);
+  }
+
+  @Override
+  public void close(TaskAttemptContext context)  {
+    RuntimeException ex = new RuntimeException("Unable to close delegate.");
```
albertshau: It's a little uglier to code, but it's better to throw the original as-is if it's the only exception that was thrown. If more exceptions are thrown, can addSuppressed to the first one. It's also better to still throw an IOException and not a RuntimeException. For example, if we end up wrapping this class again, the RuntimeException would mistakenly propagate up if the caller only catches the declared exceptions.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,95 @@
...
+
+    for (RecordWriter<AvroKey<GenericRecord>, NullWritable> delegate : delegateMap.values()) {
+      try {
+        delegate.close(context);
+      } catch (IOException | InterruptedException e) {
```
albertshau: don't need to catch the InterruptedException, better to stop doing all work and let it propagate up immediately

format-avro/src/main/java/io/cdap/plugin/format/avro/StructuredToAvroTransformer.java
fernst: Oops, I'll fix it.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyOutputFormat.java
fernst: I'll fix it.

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
fernst: I have refactored the code to throw the original IOException and add all additional exceptions (if any) as suppressed exceptions.

albertshau (approved):

format-avro/src/main/java/io/cdap/plugin/format/avro/output/DelegatingAvroKeyRecordWriter.java
```diff
@@ -0,0 +1,101 @@
...
+          Configuration conf = context.getConfiguration();
+          GenericData dataModel = AvroSerialization.createDataModel(conf);
+
+          return new AvroKeyRecordWriter<>(key.datum().getSchema(),
+                                           dataModel,
+                                           codecFactory,
+                                           outputStreamSupplier.apply(context),
+                                           syncInterval);
+        } catch (IOException ioe) {
+          throw new RuntimeException("Unable to initialize Avro Record Writer", ioe);
```
chtyim, line 74: Not a good idea to wrap IOException like this. You can avoid it by not using computeIfAbsent, but rather a simple if containsKey check.

fernst: I'll refactor this code in order to remove the additional lambda.

PR #1321, opened 2021-02-08
Removed schema requirement from Avro output format.

fernst:
Added some improvements based on Terence's suggestions.


tivv (commented):
Why are you changing from `computeIfAbsent` to plain old contains / get / put? Some exception handling issues?

albertshau (approved):
lgtm. Yeah, the change is to preserve the IOException instead of wrapping in RuntimeException.

PR #1322, opened 2021-02-11
[PLUGIN-601] Emit bytes read metric for database source

rmstar:
JIRA: https://cdap.atlassian.net/browse/PLUGIN-601

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
@@ -59,6 +59,7 @@ public class DBRecord implements Writable, DBWritable, Configurable, DataSizeRep
   private StructuredRecord record;
   private Configuration conf;
   private long bytesWritten;
+  private long bytesRead;
 
   /**
    * Need to cache {@link ResultSetMetaData} of the record for use during writing to a table.
@@ -100,6 +101,13 @@ public class DBRecord implements Writable, DBWritable, Configurable, DataSizeRep
    */
   public long getBytesWritten() {
     return bytesWritten;
+  }
+
+  /**
+   * @return the size of data read.
+   */
+  public long getBytesRead() {
+    return bytesRead;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -162,16 +170,39 @@ private void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilde
...
     } else if (o instanceof BigDecimal) {
       BigDecimal decimal = ((BigDecimal) o);
+      bytesRead += decimal.unscaledValue().bitLength() / Byte.SIZE + Integer.BYTES;
       recordBuilder.setDecimal(field.getName(), decimal);
     } else {
+      Schema schema = field.getSchema();
+      if (schema.isNullable()) {
+        schema = schema.getNonNullable();
+      }
+      if (schema.getType() == Schema.Type.INT) {
```
MEseifan: nit: use a switch-case instead of else-ifs ?

MEseifan (approved):

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
rmstar: fixed

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -184,32 +184,32 @@ private void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilde
...
-        case STRING:
-          bytesRead += ((String) o).length();
-          break;
-        case BYTES:
-          bytesRead += ((byte[]) o).length;
-          break;
-        case BOOLEAN:
-          bytesRead += Integer.BYTES;
-          break;
+      if (o != null) {
```
MEseifan, line 187: nit: combine this with the else on the line above so it would be `else if (o != null){ ...`? Might make it easier to read

MEseifan (approved):

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
rmstar: Can't combine it because we still need to execute `recordBuilder.set(field.getName(), o);` in [line 214](https://github.com/cdapio/hydrator-plugins/pull/1322/files#diff-ba8230c85bf223ca454f424bba027c5712f77202d54945939d6f01de49a0a7faR214)

PR #1329, opened 2021-02-20
CDAP-17607 add advanced join conditions

albertshau:
Added a condition type property that toggles between basic and
advanced modes. Basic is the currently behavior that joins on
key equality, while advanced is an arbitrary SQL expression.

Most of the implementation is simply to add configuration
properties and pass them down to the underlying framework.
Also did some cleanup on the JoinerConfig to make all the
properties consistent in style.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
@@ -22,10 +22,12 @@ import com.google.common.base.Strings;
 import com.google.common.collect.Iterables;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.data.schema.Schema;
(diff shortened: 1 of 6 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/JoinerConfig.java
```diff
@@ -163,8 +203,55 @@ public Schema getOutputSchema(FailureCollector collector) {
...
+  }
+
+  private JoinCondition.Op getConditionType(FailureCollector failureCollector) {
+    if (conditionType == null || conditionType.isEmpty() || BASIC.equals(conditionType)) {
+      return JoinCondition.Op.KEY_EQUALITY;
+    }
+    if (ADVANCED.equalsIgnoreCase(conditionType)) {
+      return JoinCondition.Op.EXPRESSION;
+    }
+    failureCollector.addFailure("Invalid condition type " + conditionType, "Set it to basic or advanced.");
```
fernst: nit: add quotes around `basic` and `advanced` to make it clear that these are the expected values.

fernst (approved):
LGTM

PR #1330, opened 2021-02-25
Support datetime for generic db source and sink. 

greeshmaswaminathan:
- Support datetime for generic db source and sink. 
- String columns can be mapped to datetime in source plugin
- datetime fields can be mapped to string columns in sink plugin.
- Automatic mapping of schema to datetime field type is not supported in generic db source , sink.


fernst (approved):
LGTM, very similar to the BigTable plugin changes.

PR #1331, opened 2021-02-26
fix the parent version lower bound

albertshau:
since the code relies on CDAP 6.4.0, the lower bound should also
be 6.4.0.


fernst (approved):
LGTM

PR #1334, opened 2021-03-04
CDAP-17718 force advanced outer join to be a broadcast

albertshau:
Since Spark executes advanced outer joins as
BroadcastNestedLoopJoins, force the user to specify a side to
broadcast so that the behavior is clear and they don't wonder
why the pipeline fails with memory errors.


fernst (approved):
LGTM

PR #1341, opened 2021-03-11
CDAP-17739 - Datetime validation - Aggregate functions should not allow incompatible types

greeshmaswaminathan:
- Added validations for functions to ensure incompatible types are not allowed


albertshau (approved):
lgtm, though please also open a jira to refactor this plugin to use the FailureCollector API instead of throwing exceptions. I don't think it needs to be done immediately, but it would provide a better experience that would allow the UI to display all the issues at once, and also to highlight the problematic configs.

greeshmaswaminathan:
> would allow the UI to display all the issues at once

Created https://cdap.atlassian.net/browse/PLUGIN-636 .

PR #1343, opened 2021-03-12
CDAP-17656 add option to write headers with file based sinks

albertshau:


format-common/src/main/java/io/cdap/plugin/format/output/DelegatingRecordWriter.java
```diff
@@ -34,14 +35,28 @@
...
+    this.header = header;
+    this.isHeader = true;
   }
 
   @Override
   public void write(NullWritable key, StructuredRecord value) throws IOException, InterruptedException {
+    if (isHeader && header != null) {
+      KeyValue<K, V> headerVal = header.apply(value);
+      delegate.write(headerVal.getKey(), headerVal.getValue());
+      isHeader = false;
```
yaojiefeng, line 60: Should we return after this line since L61 is writing this record again?

yaojiefeng: Also, will this work if there are multiple splits?

yaojiefeng (commented):
One question about the output format

format-common/src/main/java/io/cdap/plugin/format/output/DelegatingRecordWriter.java
albertshau: can't return because it needs to write both the header and the data row.

There is one RecordWriter per output file -- it isn't re-used across partitions so it won't skip a header.

yaojiefeng: Just confirmed offline - it writes the header based on the record schema to the file, then it writes the actual data in the record. So no need to return.

Better to put a comment to avoid confusion in the future

yaojiefeng (approved):
LGTM

albertshau:
updated to add a comment.

PR #1346, opened 2021-03-15
Deprecate FTP plugin

rmstar:


core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -45,9 +45,11 @@
 
 /**
  * {@link BatchSource} that reads from an FTP or SFTP server.
+ * This plugin is deprecated. Use https://github.com/data-integrations/ftp-plugins instead.
  */
+@Deprecated
 @Plugin(type = "batchsource")
-@Name("FTP")
+@Name("FTPDeprecated")
```
CuriousVini: If we change the name of the plugin, then it wont be backwards compatible for the case where user has a pipeline with existing plugins in system scope. Upgrading that pipeline will not find plugin with this name. Is my understanding correct?

nitinmotgi:
This should be a UI indicator not a name change.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
bdmogal: +1. Should only update display name to ``FTP (Deprecated)`` in the widget JSON.

core-plugins/docs/FTPDeprecated-batchsource.md
```diff
@@ -1,5 +1,6 @@
 # FTP Batch Source
 
+**NOTE**: This plugin is deprecated. Please use https://github.com/data-integrations/ftp-plugins instead.
```
bdmogal: ... Please use https://github.com/data-integrations/ftp-plugins instead, which is available as **FTP plugins** (> v3.0.0) from the Hub.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: Yeah this is not good; I'll revert it.
@ajainarayanan said that widget json spec doesn't allow us to set a custom name. The display name is obtained from the plugin name (camel case name is split by space). @bdmogal how should we proceed?

CuriousVini: @rmstar 
I think we can use `"display-name": "",` for the plugin widget.

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
```diff
@@ -45,9 +45,11 @@
...
 @Description("Batch source for an FTP or SFTP source. Prefix of the path ('ftp://...' or 'sftp://...') determines " +
```
CuriousVini: We should change description to deprecated, use hub plugin instead.

core-plugins/docs/FTPDeprecated-batchsource.md
```diff
@@ -1,5 +1,6 @@
 # FTP Batch Source
```
CuriousVini: we should also make this deprecated in case font color is changed in future

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
CuriousVini: https://cdap.atlassian.net/wiki/spaces/DOCS/pages/480379311/Plugin+Presentation

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: Thanks! I'll try that.

core-plugins/docs/FTPDeprecated-batchsource.md
rmstar: fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: fixed

core-plugins/docs/FTPDeprecated-batchsource.md
rmstar: fixed

core-plugins/src/main/java/io/cdap/plugin/batch/source/FTPBatchSource.java
rmstar: done, thanks

CuriousVini (approved):
LGTM 👍🏼 
Also please open a jira to remove these plugins in hydrator and mark fix version as 7.0.0

rmstar:
> Also please open a jira to remove these plugins in hydrator and mark fix version as 7.0.0

We already have a JIRA to track this task - https://cdap.atlassian.net/browse/PLUGIN-625

PR #1364, opened 2021-04-12
[CDAP-17693] Introduce spark 3 for tests, drop spark 1, create spark …

tivv:
…3 runtime for cdap sandbox and standalone

### DIFF
--- a/cassandra-plugins/pom.xml
+++ b/cassandra-plugins/pom.xml
@@ -39,7 +39,7 @@
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-data-pipeline</artifactId>
+      <artifactId>cdap-data-pipeline2_2.11</artifactId>
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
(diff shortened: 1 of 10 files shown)


format-avro/pom.xml
```diff
@@ -112,8 +112,8 @@
         <artifactId>cdap-maven-plugin</artifactId>
         <configuration combine.self="override">
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline2_2.11[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
```
albertshau: these don't need to change. The jars get renamed to cdap-data-pipeline in the corresponding spark folder in the artifacts directory, so they are always in the system as 'cdap-data-pipeline'

tivv: Got you, reverted

albertshau (approved):

PR #1374, opened 2021-04-26
Fix HBaseTest from https://builds.cask.co/browse/CDAP-BUT

tivv:


rmstar:
Can you please add a link to a failed build that would be fixed by this PR?

tivv:
@rmstar sure, https://builds.cask.co/browse/CDAP-BUT-2822

rmstar (approved):

PR #1393, opened 2021-05-17
Fix a bug in bytes read metric

rmstar:
Cherrypick #1392

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
@@ -119,6 +119,7 @@ public class DBRecord implements Writable, DBWritable, Configurable, DataSizeRep
    * @param resultSet the {@link ResultSet} to build the {@link StructuredRecord} from
    */
   public void readFields(ResultSet resultSet) throws SQLException {
+    bytesRead = 0;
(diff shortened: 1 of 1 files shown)


MEseifan (approved):

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -119,6 +119,7 @@ public long getBytesRead() {
    * @param resultSet the {@link ResultSet} to build the {@link StructuredRecord} from
    */
   public void readFields(ResultSet resultSet) throws SQLException {
+    bytesRead = 0;
```
chtyim, line 122: Why it is reset on every readFields call? Are we assuming the caller always call `getBytesRead` after calling this method? If that's the case, I don't think it's a good assumption.

rmstar: readFields is called by `RecordReader.nextKeyValue.`  
`getBytesRead` is called from RecordReader.getCurrentKey, RecordReader.getCurrentValue.
My understanding is that getCurrentKey, getCurrentValue are called when nextKeyValue returns true => getBytesRead will be called after readFields. Is that not correct?

PR #1396, opened 2021-05-18
Exclude some dependencies from hive-exec

chtyim:
- Some of the transitive dependencies are no longer available


CuriousVini (approved):
lgtm 👍🏼

PR #1398, opened 2021-05-19
Exclude some dependencies from hive-exec

rmstar:
Cherry-pick #1396


rmstar:
hive-exec dependency doesn't exist in release/2.6 branch, so I just cherrypicked the pentaho exclusion.

tivv (approved):

PR #1412, opened 2021-06-05
CDAP-17914 migrate DBConnector

seanzhougoogle:
Tested DB:
Oracle,
MySql,
SQL Server,
Postgres


albertshau (commented):
looks like there are some unintended submodule updates

database-plugins/src/main/java/io/cdap/plugin/DBManager.java
```diff
@@ -37,37 +37,29 @@
...
+  }
+
   @Nullable
   public Class<? extends Driver> validateJDBCPluginPipeline(PipelineConfigurer pipelineConfigurer,
                                                             String jdbcPluginId, FailureCollector collector) {
     if (!config.containsMacro(DBConfig.USER) && !config.containsMacro(DBConfig.PASSWORD)
-      && config.user == null && config.password != null) {
+      && config.getUser() == null && config.getPassword() != null) {
       collector.addFailure("Username and password should be provided together.", "Please provide both " +
         "user name and password, if database supports it or remove password.")
```
albertshau: not from your PR, but let's fix this message:

Username and password should be provided together. Please provide both a username and password. Or, if the database supports it, remove the password.

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -222,9 +233,46 @@ private static boolean isCompatible(Schema resultSetSchema, Schema mappedSchema)
...
+   * @param collector      the failure collector
+   * @return the loaded JDBC driver class
+   */
+  public static Class<? extends Driver> loadJDBCDriverClass(PluginConfigurer configurer, String jdbcPluginName,
+    String jdbcPluginId, FailureCollector collector) {
+    Class<? extends Driver> jdbcDriverClass =
+      configurer.usePluginClass(PLUGIN_TYPE_JDBC, jdbcPluginName, jdbcPluginId, PluginProperties.builder().build());
+    if (jdbcDriverClass == null) {
+      String error = String.format("Unable to load JDBC Driver class for plugin name '%s'.", jdbcPluginName);
+      String action = String
```
albertshau: style nit: don't have String on a line by itself
```
String.format(
  "...")
```

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -222,9 +233,46 @@ private static boolean isCompatible(Schema resultSetSchema, Schema mappedSchema)
...
+        .format("Ensure that plugin '%s' of type '%s' containing the driver has been deployed.", jdbcPluginName,
+          PLUGIN_TYPE_JDBC);
+      if (collector != null) {
```
albertshau, line 243: Annotate the collector variable as `@Nullable`

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -419,14 +466,14 @@ private static void unregisterOracleMBean(ClassLoader classLoader) {
...
-                  "Not attempting to cleanup Oracle MBean.");
+        "Not attempting to cleanup Oracle MBean.");
       return;
     }
     MBeanServer mbs = ManagementFactory.getPlatformMBeanServer();
     Hashtable<String, String> keys = new Hashtable<>();
     keys.put("type", "diagnosability");
-    keys.put("name",
-             classLoader.getClass().getName() + "@" + Integer.toHexString(classLoader.hashCode()).toLowerCase());
+    keys
```
albertshau: similar nit as above, avoid have a variable on a line by itself

```
keys.put(
  "name", ...)
```

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -456,4 +503,47 @@ private static void unregisterOracleMBean(ClassLoader classLoader) {
   private DBUtils() {
     throw new AssertionError("Should not instantiate static utility class.");
   }
+
+  public static List<StructuredRecord> parseResultSet(ResultSet resultSet) throws SQLException {
```
albertshau: having this in a utility class is dangerous, as it could easily be misused in an actual source and have memory issues. The connector should be the thing that loads an entire ResultSet into memory, better to put this method in the connector.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,316 @@
...
+      throw new IllegalArgumentException(String.format("Cannot connect to database via connection string : %s and " +
+        "arguments: %s. Make sure you have correct connection properties.", connectionString, arguments.toString()), e);
+    }
+  }
+
+  private BrowseDetail listDatabases(Connection connection, @Nullable Integer limit) throws SQLException {
+    BrowseDetail.Builder browseDetailBuilder = BrowseDetail.builder();
+    int count = 0;
+    ResultSet databaseResultSet;
+    databaseResultSet = DBDifferenceUtils.getDatabases(connection);
```
albertshau: combine with previous line

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,316 @@
...
+    } catch (SQLException e) {
+      throw new IllegalArgumentException(String.format("Cannot connect to database via connection string : %s and " +
+        "arguments: %s. Make sure you have correct connection properties.", connectionString, arguments.toString()), e);
+    }
+  }
+
+  private BrowseDetail listDatabases(Connection connection, @Nullable Integer limit) throws SQLException {
+    BrowseDetail.Builder browseDetailBuilder = BrowseDetail.builder();
+    int count = 0;
+    ResultSet databaseResultSet;
```
albertshau: does this need to get closed?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+    return browseDetailBuilder.setTotalCount(count).build();
+  }
+
+  private BrowseDetail listSchemas(Connection connection, String database, @Nullable Integer limit)
+    throws SQLException {
+    BrowseDetail.Builder browseDetailBuilder = BrowseDetail.builder();
+    int count = 0;
+    DatabaseMetaData metaData = connection.getMetaData();
+    ResultSet resultSet = metaData.getSchemas();
```
albertshau: does this need to get closed?

albertshau: similar comment everywhere ResultSet is used

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * Replace the database name in the JDBC connection string with specified database name
+   *
+   * @param connectionString the JDBC connection string
+   * @param database         the database name
+   * @return the replaced JDBC connection string
+   */
+  public static String replaceDatabase(String connectionString, @Nullable String database) {
```
albertshau: As discussed offline, let's avoid this database replacing and make the user pass in a correct connection string.

yaojiefeng:
As we discussed, let's make following changes:
1. Make sure the connection string is same as db source, so it must contain a database in it.
2. When browsing, let's start with schema/table
3. For the import query, let's make it correct with schema/table.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,316 @@
...
+      String table = path.getTable();
+      if (table == null) {
+        return listTables(connection, database, schema, browseRequest.getLimit());
+      }
+      return getTableDetail(connection, database, schema, table);
+    });
+  }
+
+  @Override
+  public List<StructuredRecord> sample(SampleRequest sampleRequest) throws IOException {
```
yaojiefeng: I feel here it is better to have this class implement `BatchConnector`, where you only need to provide the input format provider, https://github.com/cdapio/hydrator-plugins/blob/develop/database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java#L174, then you can just reuse the util class like this for the record https://github.com/cdapio/hydrator-plugins/blob/develop/database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java#L187

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/DBManager.java
seanzhougoogle: done:)

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: good point, I was thinking I can depend on connection.close to close resultset, but it looks like not a good idea.

fixed.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: Thanks for the suggestion. Since this will be a big change to my existing code. based on the fact that my existing code has been well tested and the 6.5 timeline is tight (especially when I am primary oncall this week and will be on vacation next week). I would suggest we keep it in this PR and have another JIRA to refactor it if we still have time. 

created https://cdap.atlassian.net/browse/CDAP-18049 to track it.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
seanzhougoogle: done :)

database-plugins/pom.xml
```diff
@@ -60,6 +60,30 @@
       <artifactId>hsqldb</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
```
yaojiefeng, line 65: remove these dependencies, we are not allowed to put drivers as dependency

yaojiefeng: Also please remove all the submodules change in this pr

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
```diff
@@ -52,11 +56,13 @@
  */
 public final class DBUtils {
   private static final Logger LOG = LoggerFactory.getLogger(DBUtils.class);
+  public static final String PLUGIN_TYPE_JDBC = "jdbc";
   public static final String OVERRIDE_SCHEMA = "io.cdap.hydrator.db.override.schema";
   public static final String PATTERN_TO_REPLACE = "io.cdap.plugin.db.pattern.replace";
   public static final String REPLACE_WITH = "io.cdap.plugin.db.replace.with";
   public static final String CONNECTION_ARGUMENTS = "io.cdap.hydrator.db.connection.arguments";
   public static final String FETCH_SIZE = "io.cdap.hydrator.db.fetch.size";
+  private static final String CLASS_NAME_ROWID = "oracle.sql.ROWID";
```
yaojiefeng: this seems unused, remove?

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
```diff
@@ -0,0 +1,69 @@
...
+package io.cdap.plugin.db.common;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import javax.annotation.Nullable;
+
+/**
+ * A DB Util that handles the difference of different DB
+ */
+public final class DBDifferenceUtils {
```
yaojiefeng: Do we still need this class? In our discussion, we are not supposed to have special handling in the generic db source

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Generates SQL query on the specified table with limit of number of returned rows for specific database
+   * @param productName the product name of the database
+   * @param schema the schema of the table
+   * @param table the name of the table
+   * @param limit the limit of number returned rows
+   * @return generated SQL query
+   */
+  public static String getTableQueryWithLimit(String productName, @Nullable String schema, String table, int limit) {
```
yaojiefeng: This is not very good as we don't know if any other database type does not support `LIMIT`, there is a `setFetchSize()` in the `ResultSet` you can use to limit the number of records returned, once you get the number of record you want, you can just close the result set to prevent further reading the data

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,328 @@
...
+@Description("This connector creates connections to a Database.")
+public class DBConnector implements DirectConnector {
+  public static final String NAME = "Database";
+  public static final String ENTITY_TYPE_DATABASE = "DATABASE";
+  public static final String ENTITY_TYPE_SCHEMA = "SCHEMA";
+  public static final String[] SUPPORTED_TYPES = new String[]{"TABLE", "VIEW", "MATERIALIZED VIEW"};
+
+  private static final String RESULTSET_COLUMN_TABLE_NAME = "TABLE_NAME";
+  private static final String RESULTSET_COLUMN_TABLE_TYPE = "TABLE_TYPE";
+  private static final String RESULTSET_COLUMN_TABLE_SCHEM = "TABLE_SCHEM";
```
yaojiefeng, line 82: nit - should be schema?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
```diff
@@ -0,0 +1,143 @@
...
+  public DBConnectorConfig(String user, String password, String jdbcPluginName, String connectionString,
+    String connectionArguments) {
+    this.user = user;
+    this.password = password;
+    this.jdbcPluginName = jdbcPluginName;
+    this.connectionString = connectionString;
+    this.connectionArguments = connectionArguments;
+  }
+
+
```
yaojiefeng: nit - extra line

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,328 @@
...
+      }
+    }
+    return browseDetailBuilder.setTotalCount(count).build();
+  }
+
+  private List<StructuredRecord> getTableData(Connection connection, @Nullable String schema,
+    String table, int limit) throws SQLException {
+    String productName = connection.getMetaData().getDatabaseProductName();
+    String query = DBDifferenceUtils.getTableQueryWithLimit(productName, schema, table, limit);
+    return parseResultSet(connection.createStatement().executeQuery(query));
```
yaojiefeng: here when you create a statement, you can use `statement.setFetchSize(limit)`, this way the result set will only retrieve that limit number of data in a round trip even if the query is `SELECT * FROM TABLE`, after you getting the limit, you can then close the resultset and statement.

yaojiefeng: This way you can get rid of the limit hack for specific data types

yaojiefeng: Also this resultset and statement needs to be closed

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
```diff
@@ -0,0 +1,143 @@
...
+  @Nullable
+  public String getPassword() {
+    return password;
+  }
+
+  public String getJdbcPluginName() {
+    return jdbcPluginName;
+  }
+
```
yaojiefeng: nit - extra line

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBPath.java
```diff
@@ -0,0 +1,112 @@
...
+ * the License.
+ */
+
+package io.cdap.plugin.db.connector;
+
+import javax.annotation.Nullable;
+
+/**
+ * BigQuery Path that parses the path in the request of connection service
+ * A valid path can start with/without a slash, e.g. "/" , "", "/database/schema/table", "database/schema/table" are
```
yaojiefeng: Please update javadoc here, database should no longer exist in the path

database-plugins/src/test/java/io/cdap/plugin/db/common/DBDifferenceUtilsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+package io.cdap.plugin.db.common;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class DBDifferenceUtilsTest {
+
+  @Test
+  public void testGetTableQueryWithLimit() {
```
yaojiefeng: can remove this if we can get rid of the limit hack

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
```diff
@@ -0,0 +1,69 @@
...
+        // SQL Server doesn't support LIMIT, have to use TOP
+        return String.format("SELECT TOP(%d) * FROM %s.%s", limit, schema, table);
+      default:
+        return schema == null ? String.format("SELECT * FROM %s LIMIT %d", table, limit) :
+          String.format("SELECT * FROM %s.%s LIMIT %d", schema, table, limit);
+    }
+  }
+
+  @VisibleForTesting
+  static String getPostgreDatabasesQuery() {
```
yaojiefeng: I am assuming we don't need this since we do not need to list databases?

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
```diff
@@ -0,0 +1,69 @@
...
+    return String
+      .format("SELECT datname AS %s FROM pg_database WHERE datistemplate = false;", RESULTSET_COLUMN_TABLE_CAT);
+  }
+
+  @VisibleForTesting
+  static String getOracleDatabasesQuery() {
```
yaojiefeng: same here

yaojiefeng (commented):
Some comments about limit on sampling, this is again less a problem if we use `BatchConnector`, but it is fine we deal with it in post 6.5

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
seanzhougoogle: done :)

database-plugins/pom.xml
seanzhougoogle: the scope is "test" for those driver dependencies, they are only for testing and won't be packaged in the build.
removed submodules.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: this is the column name in the resultset returned by `DatabaseMetadata.getSchemas()` which is a standard one.

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
seanzhougoogle: removed

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
seanzhougoogle: removed.

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/common/DBDifferenceUtils.java
seanzhougoogle: removed

database-plugins/src/test/java/io/cdap/plugin/db/common/DBDifferenceUtilsTest.java
seanzhougoogle: removed

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBPath.java
seanzhougoogle: done:)

database-plugins/src/main/java/io/cdap/plugin/DBUtils.java
seanzhougoogle: removed

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
seanzhougoogle: removed

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
seanzhougoogle: removed

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,339 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * A Generic Database Connector that connects to database via JDBC.
+ */
+@Plugin(type = Connector.PLUGIN_TYPE)
+@Name(DBConnector.NAME)
+@Description("This connector creates connections to a Database.")
+public class DBConnector implements DirectConnector {
```
yaojiefeng, line 74: Can you add `@Category("Database")` here?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,339 @@
...
+
+  private Connection getConnection() {
+    try {
+      return DriverManager.getConnection(config.getConnectionString(), config.getAllConnectionArguments());
+    } catch (SQLException e) {
+      throw new IllegalArgumentException(String.format("Cannot connect to database via connection string : %s and " +
+                                                         "arguments: %s. Make sure you have correct connection " +
+                                                         "properties.",
+                                                       config.getConnectionString(), config.getConnectionArguments()),
+                                         e);
```
yaojiefeng: nit, this can be in previous line

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -0,0 +1,339 @@
...
+    return browseDetailBuilder.setTotalCount(count).build();
+  }
+
+  private List<StructuredRecord> getTableData(Connection connection, @Nullable String schema,
+                                              String table, int limit) throws SQLException {
+    String query = schema == null ? String.format("SELECT * FROM %s", table) :
+      String.format("SELECT * FROM %s.%s", schema, table);
+    try (Statement statement = connection.createStatement()) {
+      statement.setFetchSize(limit);
+      return parseResultSet(statement.executeQuery(query), limit);
```
yaojiefeng: here you need to close the result set

pom.xml
```diff
@@ -113,7 +113,7 @@
     <hsql.version>2.2.4</hsql.version>
     <cdh.hive.version>1.1.0-cdh5.5.1</cdh.hive.version>
     <javamail.version>1.4.1</javamail.version>
-    <junit.version>4.11</junit.version>
+    <junit.version>4.13.1</junit.version>
```
yaojiefeng, line 116: any reason we are bumping this?

yaojiefeng (approved):
Minor comments, please fix and merge

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: done:)

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: good catch :)

pom.xml
seanzhougoogle: because we want to use assertThrow which is not available in 4.11

PR #1414, opened 2021-06-11
fix NPE

seanzhougoogle:


database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -69,7 +69,7 @@
   private DBConnectorConfig connection;
 
   public Boolean getEnableAutoCommit() {
-    return enableAutoCommit;
+    return enableAutoCommit == null ? Boolean.TRUE : enableAutoCommit;
```
yaojiefeng: let's add a constructor in this class to set the default settings for this incase this method is overriden in some classes

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -69,7 +69,7 @@
   private DBConnectorConfig connection;
 
   public Boolean getEnableAutoCommit() {
```
yaojiefeng: let's make this return `boolean`

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
yaojiefeng: it should be `false` here

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
seanzhougoogle: done:)

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
seanzhougoogle: done :)

seanzhougoogle: making the method return boolean should avoid issues for overriding.

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -68,8 +68,8 @@
   @Description("The existing connection to use.")
   private DBConnectorConfig connection;
 
-  public Boolean getEnableAutoCommit() {
-    return enableAutoCommit;
+  public boolean getEnableAutoCommit() {
```
yaojiefeng, line 71: don't see the constructor is added, can you add it like this https://github.com/cdapio/hydrator-plugins/blob/50bb1436e040b27ab36f50b44e4d78299ba4fc20/database-plugins/src/main/java/io/cdap/plugin/ConnectionConfig.java#L92-L95

yaojiefeng (approved):

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
seanzhougoogle: as discussed offline, since we already change the return value to `boolean` , no need to add a constructor here.

PR #1415, opened 2021-06-13
add file connector

yaojiefeng:


core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+    // sort the files by names
+    Arrays.sort(files);
+    int limit = request.getLimit() == null ? files.length : Math.min(request.getLimit(), files.length);
+    BrowseDetail.Builder builder = BrowseDetail.builder().setTotalCount(files.length);
+    for (int i = 0; i < limit; i++) {
+      // do not browse hidden files
+      if (files[i].isHidden()) {
+        continue;
+      }
+      builder.addEntity(generateBrowseEntity(files[i]));
```
seanzhougoogle: set count ?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+      throw new IOException(String.format("Unable to browse the path %s", path));
+    }
+
+    // sort the files by names
+    Arrays.sort(files);
+    int limit = request.getLimit() == null ? files.length : Math.min(request.getLimit(), files.length);
+    BrowseDetail.Builder builder = BrowseDetail.builder().setTotalCount(files.length);
+    for (int i = 0; i < limit; i++) {
+      // do not browse hidden files
+      if (files[i].isHidden()) {
```
seanzhougoogle: so the hidden file is not returned but counted for limit ?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+@Category("Flat File")
+public class FileConnector extends AbstractFileConnector<FileConnector.FileConnectorConfig> {
+  public static final String NAME = "File";
+  static final String FILE_TYPE_KEY = "File Type";
+  static final String SIZE_KEY = "Size";
+  static final String LAST_MODIFIED_KEY = "Last Modified";
+  static final String OWNER_KEY = "Owner";
+  static final String GROUP_KEY = "Group";
+  static final String PERMISSION_KEY = "Permission";
+  private static final String DIRECTORY = "directory";
```
seanzhougoogle: nit: ENTITY_TYPE_DIRECTORY

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+  private static final String FILE = "file";
```
seanzhougoogle: nit: ENTITY_TYPE_FILE

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  private BrowseEntity generateBrowseEntity(File file) throws IOException {
+    String path = file.getCanonicalPath();
+    boolean isDirectory = file.isDirectory();
+    BrowseEntity.Builder builder = BrowseEntity.builder(file.getName(), path, isDirectory ? DIRECTORY : FILE);
+    Set<BrowseEntityProperty> properties = new HashSet<>();
+
+    if (isDirectory) {
+      builder.canBrowse(true).canSample(true);
```
seanzhougoogle, line 124: why can directory be sampled ?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,144 @@
...
+      builder.canSample(FileTypeDetector.isSampleable(fileType));
+      properties.add(BrowseEntityProperty.builder(FILE_TYPE_KEY, fileType,
+                                                  BrowseEntityProperty.PropertyType.STRING).build());
+      properties.add(BrowseEntityProperty.builder(SIZE_KEY, String.valueOf(file.length()),
+                                                  BrowseEntityProperty.PropertyType.SIZE_BYTES).build());
+    }
+
+    properties.add(BrowseEntityProperty.builder(LAST_MODIFIED_KEY, String.valueOf(file.lastModified()),
+                                                BrowseEntityProperty.PropertyType.TIMESTAMP_MILLIS).build());
+    properties.add(BrowseEntityProperty.builder(OWNER_KEY, Files.getOwner(file.toPath()).getName(),
```
seanzhougoogle: is it possible that Files.getOwner return null ?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,154 @@
...
+
+    if (isDirectory) {
+      builder.canBrowse(true).canSample(true);
+    } else {
+      String fileType = FileTypeDetector.detectFileType(path);
+      builder.canSample(FileTypeDetector.isSampleable(fileType));
+      properties.add(BrowseEntityProperty.builder(FILE_TYPE_KEY, fileType,
+                                                  BrowseEntityProperty.PropertyType.STRING).build());
+      properties.add(BrowseEntityProperty.builder(SIZE_KEY, String.valueOf(file.length()),
+                                                  BrowseEntityProperty.PropertyType.SIZE_BYTES).build());
```
seanzhougoogle: this `SIZE_BYTES` Looks wired, other types are indicating the type of the value, but this one is kind of indicating the semantics of the value. `SIZE_KEY` is already indicating the semantics of this value, should we use PropertyType.NUMBER here?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    properties.add(BrowseEntityProperty.builder(LAST_MODIFIED_KEY, String.valueOf(file.lastModified()),
+                                                BrowseEntityProperty.PropertyType.TIMESTAMP_MILLIS).build());
+    properties.add(BrowseEntityProperty.builder(OWNER_KEY, Files.getOwner(file.toPath()).getName(),
+                                                BrowseEntityProperty.PropertyType.STRING).build());
+    properties.add(BrowseEntityProperty.builder(
+      GROUP_KEY, Files.readAttributes(file.toPath(), PosixFileAttributes.class).group().getName(),
```
seanzhougoogle: for non POSIX operating system, what will the value be here for group ?  will `group()` return null ?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * File connector to browse flat file on local system
+ */
+@Plugin(type = Connector.PLUGIN_TYPE)
+@Name(FileConnector.NAME)
+@Description("Flat file connector for local file system")
+@Category("Flat File")
```
albertshau: nit: Was "Flat File" a specific ask from somebody? Otherwise I think "File" is more clear.

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
albertshau: looks like hidden files do not count against the limit (which seems like the correct behavior)

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
albertshau: I assume this type is this so that the UI knows it can do some handling around kb,mb,gb, etc?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
albertshau: File source is frequently setup to read from a directory, in which case it reads all files in the directory.

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
albertshau: doesn't hurt to protect against possible NPE if we're unsure

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -0,0 +1,186 @@
...
+   * @param builder the builder of the spec
+   */
+  protected void setConnectorSpec(ConnectorSpecRequest request, ConnectorSpec.Builder builder) {
+    // no-op
+  }
+
+  private ValidatingInputFormat getValidatingInputFormat(ConnectorContext context, String path) throws IOException {
+    String fileType = FileTypeDetector.detectFileType(path);
+    if (!FileTypeDetector.isSampleable(fileType)) {
+      throw new IllegalArgumentException(String.format("The given path %s is not isSamplable.", path));
```
albertshau: The given path %s cannot be sampled

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -0,0 +1,186 @@
...
+    PluginProperties.Builder builder = PluginProperties.builder();
+    builder.add("path", path);
+    builder.addAll(config.getProperties().getProperties());
+    builder.addAll(getFileSystemProperties(path));
+    ValidatingInputFormat inputFormat = context.getPluginConfigurer().usePlugin(
+      ValidatingInputFormat.PLUGIN_TYPE, format, UUID.randomUUID().toString(), builder.build());
+
+    if (inputFormat == null) {
+      throw new IOException(
+        String.format("Unsupported file format %s on path %s, the supported types are avro, blob, csv, delimited, " +
```
albertshau: this is technically not correct, since the formats are pluggable. I think it's ok to leave out the supported types in this case.

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @param request the spec generation request
+   * @param builder the builder of the spec
+   */
+  protected void setConnectorSpec(ConnectorSpecRequest request, ConnectorSpec.Builder builder) {
+    // no-op
+  }
+
+  private ValidatingInputFormat getValidatingInputFormat(ConnectorContext context, String path) throws IOException {
+    String fileType = FileTypeDetector.detectFileType(path);
```
albertshau, line 166: we said directories are sample-able earlier, what does this do for directories? It seems like it returns that it's text?

albertshau (commented):
is there supposed to be a widget and doc file? Or do those come later?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
yaojiefeng: Yea, I updated yesterday to not count that towards count

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
yaojiefeng: The method is not marked as Nullable, the javadoc also doesn't mention it, so I assume it won't return null?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
yaojiefeng: Yes, this is UI ask to have this so they can better handle the timstamp and size

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
yaojiefeng: This plugin is supposed to only work for local file system

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
yaojiefeng: Yes this will return text for it. UI now won't lead users to sample a directory since it is browsable. This will work much better once we let user input the format and UI honor users to sample a directory

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
yaojiefeng: I remember I saw it in the mocks. Changed it to File.

yaojiefeng:
This FileConnector doesn't really have a config, so I don't add a widget for it. I added a doc file for it

core-plugins/docs/File-connector.md
```diff
@@ -0,0 +1,6 @@
+# File Connector
+
+
+Description
+-----------
+File connector can be used to browse and sample on the local file system
```
albertshau: on the -> the

albertshau (approved):

PR #1416, opened 2021-06-21
move db connector to common

yaojiefeng:
This has most changes except the listDatabases, still WIP on that but sent out for early review


database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -71,69 +64,25 @@
...
-  private static final String RESULTSET_COLUMN_COLUMN_NAME = "COLUMN_NAME";
-  private static final String RESULTSET_COLUMN_IS_NULLABLE = "IS_NULLABLE";
-
-  private DBConnectorConfig config;
-  private DriverCleanup driverCleanup;
+  private final DBConnectorConfig config;
 
   DBConnector(DBConnectorConfig config) {
+    super(config);
     this.config = config;
```
seanzhougoogle, line 74: can we have `config` in parent class be protected so that we don't need to keep to references to `config` in both parent class and this class ?

hydrator-common/src/main/java/io/cdap/plugin/common/db/DBConnectorPath.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * Database connector path which contains common method about the path
+ */
+public interface DBConnectorPath {
+
+  /**
+   * Return if the path supports browsing databases
```
seanzhougoogle: nit: `Return if the path can contain database`

Path itself doesn't support browsing. Path can be used for browsing/sampling/generateSpec

seanzhougoogle: same comments for schema

hydrator-common/src/main/java/io/cdap/plugin/common/db/DBConnectorPath.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  boolean supportDatabase();
```
seanzhougoogle: nit: canContainDatabase or containDatabase ?

seanzhougoogle: same comments for schema

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -146,130 +95,23 @@ public BrowseDetail browse(ConnectorContext context, BrowseRequest browseRequest
...
-        specBuilder.setSchema(Schema.recordOf("output", fields));
-        properties.put(DBSource.DBSourceConfig.IMPORT_QUERY,
-                       schema == null ? String.format("SELECT * FROM %s;", table) :
-                         String.format("SELECT * FROM %s.%s;", schema, table));
-        properties.put(DBSource.DBSourceConfig.NUM_SPLITS, "1");
-      }
-      return specBuilder.addRelatedPlugin(new PluginSpec(DBSource.NAME, BatchSource.PLUGIN_TYPE, properties)).build();
-    });
-  }
+  protected void setConnectorSpec(ConnectorSpecRequest request, DBConnectorPath path,
```
seanzhougoogle, line 98: nit: rename it to addRelatedPlugins ?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
```diff
@@ -32,7 +33,7 @@
 /**
  * Defines a base {@link PluginConfig} that Database source, sink, and action can all re-use.
  */
-public class DBConnectorConfig extends PluginConfig {
+public class DBConnectorConfig extends PluginConfig implements DBConnectorProperties {
```
seanzhougoogle, line 36: what's the purpose of "DBConnectorProperties" interface ? don't see the usage of this interface.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
yaojiefeng: We normally prefer private field in all classes, the parent class here doesn't actually know what the exact type the config is, it only knows it is a `DatasetProperties`, so good to have the config in each class to be cleaner

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
yaojiefeng: You can use it to set whatever you want, properties, even the schema, so I name it like this

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
yaojiefeng: This is to tell the abstract connector what it can expect from the config, without this, from the config field, the connector is not able to get any information, with this, the connector is able to get the jdbc url or user name or password, the individual connector config can override this interface to provide them. For example, generic db will directly return the jdbc string, specific db will construct the jdbc url by host, port, etc.

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
seanzhougoogle: hmm.. don't quite understand the point here. You are right, parent class just need to know it's a subclass of DatasetProperties. But subclass know its exact class when `extends AbstractDBConnector<DBConnectorConfig>`
parent class can only call methods of DatasetProperties while subclass can call methods of DBConnectorConfig, it doesn't prevent removing the redundancy here.

seanzhougoogle (approved):
LGTM with some minor comments

PR #1419, opened 2021-06-22
PLUGIN-706 add jdbc plugin selector in Connector plugin UI

seanzhougoogle:


elfenheart (approved):

database-plugins/widgets/Database-connector.json
```diff
@@ -7,6 +7,14 @@
     {
       "label": "Basic",
       "properties": [
+        {
+          "widget-type": "plugin-list",
+          "label": "Plugin Name",
```
bdmogal, line 12: can we call this JDBC Driver name? It is unclear what `Plugin Name` means. This is a problem with the current Database plugins as well.

PR #1424, opened 2021-06-28
CDAP-18114 minor refactor for DB specific plugin migration

seanzhougoogle:


database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
```diff
@@ -136,10 +137,7 @@ private void testSample(DBConnector connector) throws IOException {
 
   private void testBrowse(DBConnector connector) throws IOException {
     // browse DB server
-    BrowseDetail detail = connector.browse(new MockConnectorContext(new MockConnectorConfigurer()),
-                                           BrowseRequest.builder("/").build());
-    Assert.assertTrue(detail.getTotalCount() > 0);
-    Assert.assertTrue(detail.getEntities().size() > 0);
+    BrowseDetail detail = null;
```
albertshau, line 140: I can't comment on the existing code below, but there are several System.out.println() statements that should be removed.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
```diff
@@ -122,7 +123,7 @@ private void testSample(DBConnector connector) throws IOException {
     //invalid path
     Assert.assertThrows(IllegalArgumentException.class,
                         () -> connector.sample(new MockConnectorContext(new MockConnectorConfigurer()),
-                                               SampleRequest.builder(1).setPath(schema == null ? "a/b/c" : "a/b/c/d")
+                                               SampleRequest.builder(1).setPath(schema == null ? "a/b" : "a/b/c")
```
albertshau, line 126: I don't see a change of behavior in the sample() method, did something change there?

Not sure how this test could be passing before

database-plugins/widgets/Database-batchsource.json
```diff
@@ -75,22 +83,6 @@
...
-        {
-          "widget-type": "plugin-list",
-          "label": "JDBC Driver name",
-          "name": "jdbcPluginName",
-          "widget-attributes": {
-            "plugin-type": "jdbc"
-          }
-        },
-        {
-          "widget-type": "textbox",
```
albertshau, line 87: removing this means the property shows up in the UI without any formatting. If we no longer want to show this (which makes sense to me), need to set the widget type to 'hidden'.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
seanzhougoogle: since for generic DB plugin, we require the user to put database in connection string, that means the path at most contain two parts : schema and table.  

this is to test invalid path. the old test case was for the case when database is not required in the connection string and had 3 parts for db that doesn't support schema and 4 parts for db that supports schema. so it still passed the test cases.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
seanzhougoogle: removed.

database-plugins/widgets/Database-batchsource.json
seanzhougoogle: We don't have this property any more at backend. It's now hardcoded to jdbc.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
albertshau: but I don't see any change in this PR that handles the path differently. Unless I'm missing something?

database-plugins/widgets/Database-batchsource.json
albertshau: removing a property is a backwards incompatible change. We don't want this plugin type in the new connector, but the source still needs to support it for compatibility. It's possible, though unlikely, that somebody uploaded a jdbc driver as some other plugin type (they can set it to whatever they want using the API). 

If it was removed earlier, please add it back, just for the source. Also set the widget-type to 'hidden' so that new pipelines never set it to a value.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
seanzhougoogle: it's in previous PR. But I didn' t update the test case accordingly in that PR. However old test case still pass for new code. Here , for example

`a/b/c/d` is invalid when path contains database and it's still invalid when path doesn't contain database (that's why the test case didn't break in previous PR)
I updated it to `a/b/c` in this PR so that this path is only invalid when path doesn't contain database (as you can see if the path can contain database, this path is valid)

database-plugins/widgets/Database-batchsource.json
seanzhougoogle: added back.

database-plugins/src/test/java/io/cdap/plugin/db/connector/DBConnectorTest.java
albertshau: ah I see, understood.

database-plugins/src/main/java/io/cdap/plugin/DBManager.java
```diff
@@ -40,15 +40,17 @@
...
   private final DBConnectorConfig config;
-  private DriverCleanup driverCleanup;
+  private final String jdbcPluginType;
 
-  public DBManager(DBConnectorConfig config) {
+  public DBManager(DBConnectorConfig config, String jdbcPluginType) {
     this.config = config;
+    this.jdbcPluginType = jdbcPluginType;
   }
+  private DriverCleanup driverCleanup;
```
albertshau: nit: move this above the constructor, with the other variables.

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -68,10 +70,19 @@
...
+  @Nullable
+  public String jdbcPluginType;
+
   public boolean getEnableAutoCommit() {
     return enableAutoCommit == null ? false : enableAutoCommit;
   }
 
+  public String getJdbcPluginType() {
+    return jdbcPluginType == null ? DBUtils.PLUGIN_TYPE_JDBC : jdbcPluginType;
+  }
```
albertshau, line 85: nit: add a newline between this method and the next

albertshau (approved):

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -253,15 +270,51 @@ protected BrowseDetail listTables(Connection connection, @Nullable String schema
...
+    // NOTE Oracle schema name is case sensitive here
+    try (ResultSet schemas = connection.getMetaData()
+      .getSchemas(database == null ? connection.getCatalog() : database, schema)) {
+      if (!schemas.next()) {
+        throw new IllegalArgumentException(String.format("Schema '%s' does not exist.", schema));
+      }
+    }
+  }
+
+  protected Connection getConnection(@Nullable DBConnectorPath path) {
```
yaojiefeng: better to not allow this to be nullable, just have a method getConnection() which returns the config connection string. Also the path is not used, is the specific plugin supposed to create a connection string from the path?

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -253,15 +270,51 @@ protected BrowseDetail listTables(Connection connection, @Nullable String schema
...
+  }
+
+  protected void validateSchema(String database, String schema, Connection connection)
+    throws SQLException {
+    if (schema == null) {
+      return;
+    }
+    // NOTE Oracle schema name is case sensitive here
+    try (ResultSet schemas = connection.getMetaData()
+      .getSchemas(database == null ? connection.getCatalog() : database, schema)) {
```
yaojiefeng, line 298: nit - we normally split the line on parameter

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -224,20 +245,16 @@ protected BrowseDetail getTableDetail(Connection connection, @Nullable String sc
...
-      try (ResultSet schemas = metaData.getSchemas(connection.getCatalog(), schema)) {
-        if (!schemas.next()) {
-          throw new IllegalArgumentException(String.format("Schema '%s' does not exist.", schema));
-        }
-      }
-    }
-    try (ResultSet resultSet = metaData.getTables(connection.getCatalog(), schema, null, null)) {
+    validateSchema(database, schema, connection);
+    try (ResultSet resultSet = connection.getMetaData()
+      .getTables(database == null ? connection.getCatalog() : database, schema, null, null)) {
```
yaojiefeng, line 257: nit - we normally split the line on parameter

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -253,15 +270,51 @@ protected BrowseDetail listTables(Connection connection, @Nullable String schema
...
+        throw new IllegalArgumentException(String.format("Schema '%s' does not exist.", schema));
+      }
+    }
+  }
+
+  protected Connection getConnection(@Nullable DBConnectorPath path) {
+    return getConnection(config.getConnectionString(), config.getConnectionArgumentsProperties());
+  }
+
+  protected Connection getConnection(String connectionString, Properties connectionArguments) {
```
yaojiefeng, line 313: Do we still need this method if we have `getConnection(DBConnectorPath)`?

database-plugins/src/main/java/io/cdap/plugin/DBManager.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
seanzhougoogle: done :)

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
seanzhougoogle: yes, I make it Nullable on purpose so that I can have same code for both generic and specific DB plugins. 
Generic DB plugin don't need path to create a connection while specific DB plugins need

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
seanzhougoogle: this helper method can be reused in DB Specific plugin

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
seanzhougoogle: added a empty parameter getConnection()

PR #1428, opened 2021-07-02
PLUGIN-716 add database to table path

seanzhougoogle:


hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -139,7 +139,11 @@ public ConnectorSpec generateSpec(ConnectorContext connectorContext,
         return specBuilder.build();
       }
       String database = dbConnectorPath.getDatabase();
-      validateDatabase(database, connection);
+      if (database == null) {
+        database = connection.getCatalog();
```
yaojiefeng, line 143: Just curious, what does catalog mean here? Why database can get assigned to it if it is null.

yaojiefeng (approved):
LGTM, just one question

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
seanzhougoogle: catalog means current connected database, database is now only for generic database case, bc generic database plugin have database in connection string, not in the path

PR #1435, opened 2021-07-15
not using connection for DBSource

seanzhougoogle:


database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -144,4 +144,9 @@ public String getReferenceName() {
   public String getColumnNameCase() {
     return columnNameCase;
   }
+
+  @Nullable
```
albertshau: instead of possibly returning null, would be better to check for null in the getter and return a boolean. Makes it less error-prone for the caller.

database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
```diff
@@ -352,6 +352,14 @@ private String getBoundingQuery() {
 
     @SuppressWarnings("checkstyle:WhitespaceAround")
     private void validate(FailureCollector collector) {
+      if (Boolean.TRUE.equals(getUseConnection())) {
```
albertshau: more natural if the getter returns a boolean, then you don't need to use equals()

albertshau (approved):
minor comment

database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
seanzhougoogle: done :)

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
seanzhougoogle: done :)

PR #1444, opened 2021-07-21
PLUGIN-750 connection validation fix

seanzhougoogle:


yaojiefeng (approved):
LGTM, please make sure it is tested in studio with/without connection, and actually run with/without connection

PR #1446, opened 2021-07-22
make getSchema able to be overriden by subclass to handle some special data type

seanzhougoogle:


yaojiefeng (approved):
LGTM, please revert the submodule change

PR #1451, opened 2021-07-27
make entity type lower case

seanzhougoogle:


snehalakshmisha:
Can you add some info on why this change is required?

yaojiefeng (approved):
LGTM

seanzhougoogle:
> Can you add some info on why this change is required?


It's requested by Albert in this morning's demo meeting to be consistent with other connectors.

PR #1464, opened 2021-08-06
support BIGINT properly

seanzhougoogle:


hydrator-common/src/main/java/io/cdap/plugin/common/db/DBUtils.java
```diff
@@ -297,9 +322,21 @@ public static Schema getSchema(String typeName, int sqlType, int precision, int
...
+            String.format("Column %s has unsupported SQL Type: %s with precision 0.", columnName, typeName)));
+        }
         if (handleAsDecimal) {
           return Schema.decimalOf(precision, scale);
         } else {
+          //SQL DECIMAL can be 5 - 17 bytes, not all can be held in a long
+          if (precision >= 19) {
+            throw new SQLException(
+              new UnsupportedTypeException(
+                String.format("Column %s has unsupported SQL Type: %s with precision %d while trying to store decimal" +
```
nitinmotgi: Unsupported SQL Type. Unable to store column %s as decimal. Current precision is %d, it has to be less than 19.

seanzhougoogle: done :)

yaojiefeng (approved):

PR #1466, opened 2021-08-07
CDAP-18328 construct error message properly without realing password

seanzhougoogle:


hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -340,10 +339,11 @@ protected Connection getConnection(String connectionString, Properties connectio
     try {
       return DriverManager.getConnection(connectionString, connectionArguments);
     } catch (SQLException e) {
-      throw new IllegalArgumentException(String.format("Cannot connect to database via connection string : %s and " +
-                                                         "arguments: %s. Error: %s. Make sure you have correct " +
-                                                         "connection properties.", connectionString,
-                                                       connectionArguments, e.getMessage()), e);
+      Properties args = (Properties) connectionArguments.clone();
+      args.setProperty("password", "*");
```
yaojiefeng: can we just remove the password from the properties?

seanzhougoogle: done :)

yaojiefeng (approved):

PR #1477, opened 2021-08-13
CDAP-18359 get error message from root cause

seanzhougoogle:


tivv (commented):
Could you please remove unrelated changes? Also, it would be great to have at least some tests for various databases we directly support. E.g. take some invalid host name and see what message do we get this way. For postgresql it should be better than now, but we'd better check others as well

seanzhougoogle:
> Could you please remove unrelated changes? 

sorry, didn't notice that I committed the submodule reference


>Also, it would be great to have at least some tests for various databases we directly support. E.g. take some invalid host name and see what message do we get this way. For postgresql it should be better than now, but we'd better check others as well

Good point :) Tested with mysql, oracle, sql server, with wrong connection info, error messages got make sense:)

tivv:
Is it easy enough to have those tests automated?

seanzhougoogle:
> Is it easy enough to have those tests automated?


not yet, those require some real database to be brought up when the test case starts. We have plan to improve it by bring up some docker images.

tivv (approved):

PR #1479, opened 2021-08-14
CDAP-18366 enhance error message

seanzhougoogle:


hydrator-common/pom.xml
```diff
@@ -34,6 +34,10 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>commons-lang</groupId>
```
yaojiefeng: it is better we do not introduce this dependency, I am unsure if it has any side effect

seanzhougoogle: extract the utils method I need.

yaojiefeng (approved):

PR #1484, opened 2021-08-19
Feature/add quote option

niceyang:
Feature/add quote option

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -38,18 +38,12 @@ JSON - is not supported, user has to manually provide the output schema.
 **Skip Header:** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
 **Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
-is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
-The first field will have '1' as its value and the second will have 'a, b, c' as its value. The quote characters will be trimmed.
-The newline delimiter cannot be within quotes. For example, rows below are invalid:
-  
-* "ab
-     
-     c", d
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like `1, "a, b, c"` will output two fields.
+The first field will have '1' as its value and the second will have `a, b, c` as its value. The quote characters will be trimmed.
+The newline delimiter cannot be within quotes. 
 
 It also assumes the quotes are well enclosed. The left quote will match the first following quote right before the delimiter. If there is an 
-unenclosed quote, it will be included as part of data. For example:
-
-* "a, "b"c,d", e" -> [a, b"c,d, e]
+unenclosed quote, an exception will be thrown.
 
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -39,13 +39,13 @@ JSON - is not supported, user has to manually provide the output schema.
 
 **Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
 is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
-The first field will have '1' as its value and the second will have 'a, b, c' as its value.
-The newline delimiter and embedded quotes cannot be within quotes. The quote characters will be trimmed. For example, rows below are invalid:
-  1. "\\"d\\""
-  2. "ab
-    
-     c", d
+The first field will have '1' as its value and the second will have 'a, b, c' as its value. The quote characters will be trimmed.
+The newline delimiter cannot be within quotes. For example, rows below are invalid:
+  
+   * "ab
      
+     c", d
+
 It also assumes the quotes are well enclosed. If there is an unenclosed quote, the data will be determined as invalid and the pipeline will fail.
 
 **Maximum Split Size:** Maximum size in bytes for each input partition.
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -42,11 +42,14 @@ is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that
 The first field will have '1' as its value and the second will have 'a, b, c' as its value. The quote characters will be trimmed.
 The newline delimiter cannot be within quotes. For example, rows below are invalid:
   
-   * "ab
+* "ab
      
      c", d
 
-It also assumes the quotes are well enclosed. If there is an unenclosed quote, the data will be determined as invalid and the pipeline will fail.
+It also assumes the quotes are well enclosed. The left quote will match the first following quote right before the delimiter. If there is an 
+unenclosed quote, it will be included as part of data. For example:
+
+* "a, "b"c,d", e" -> [a, b"c,d, e]
 
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -37,6 +37,17 @@ JSON - is not supported, user has to manually provide the output schema.
 
 **Skip Header:** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
+The first field will have '1' as its value and the second will have 'a, b, c' as its value.
+The newline delimiter and embedded quotes cannot be within quotes. The quote characters will be trimmed. For example, rows below are invalid:
+  1. "\\"d\\""
+  2. "ab
+    
+     c", d
+     
+It also assumes the quotes are well enclosed. If there is an unenclosed quote, the data will be determined as invalid and the pipeline will fail.
+
 **Maximum Split Size:** Maximum size in bytes for each input partition.
 Smaller partitions will increase the level of parallelism, but will require more resources and overhead.
 The default value is 128MB.
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -671,7 +671,8 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
         !enableQuotedValues ? split :
             StructuredRecord.builder(schema).set("id", 1L).set("val1", join.substring(1, join.length() - 1))
                 .set("file", fileText.toURI().toString()).build(),
-        StructuredRecord.builder(schema).set("id", 2L).set("val1", "sam").set("file", fileText.toURI().toString()).build()
+        StructuredRecord.builder(schema).set("id", 2L).set("val1", "sam")
+            .set("file", fileText.toURI().toString()).build()
     );
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -667,9 +667,9 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
       StructuredRecord.builder(schema).set("id", 1L).set("val1", "\"a").set("val2", "b").set("val3", "c\"")
         .set("file", fileText.toURI().toString()).build();
     Set<StructuredRecord> expected = ImmutableSet.of(
-        StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
-        StructuredRecord.builder(schema).set("id", 1L).set("file", fileText.toURI().toString()).build(),
-        StructuredRecord.builder(schema).set("id", 2L).set("name", "sam").set("file", fileText.toURI().toString()).build()
+      StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
+      StructuredRecord.builder(schema).set("id", 1L).set("file", fileText.toURI().toString()).build(),
+      StructuredRecord.builder(schema).set("id", 2L).set("name", "sam").set("file", fileText.toURI().toString()).build()
     );
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -664,12 +664,15 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
       .startAndWaitForRun(ProgramRunStatus.COMPLETED, 5, TimeUnit.MINUTES);
 
     StructuredRecord split =
-      StructuredRecord.builder(schema).set("id", 1L).set("val1", "\"a").set("val2", "b").set("val3", "c\"")
-        .set("file", fileText.toURI().toString()).build();
+        StructuredRecord.builder(schema).set("id", 1L).set("val1", "\"a").set("val2", "b").set("val3", "c\"")
+            .set("file", fileText.toURI().toString()).build();
     Set<StructuredRecord> expected = ImmutableSet.of(
-      StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 1L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 2L).set("name", "sam").set("file", fileText.toURI().toString()).build()
+        StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
+        !enableQuotedValues ? split :
+            StructuredRecord.builder(schema).set("id", 1L).set("val1", join.substring(1, join.length() - 1))
+                .set("file", fileText.toURI().toString()).build(),
+        StructuredRecord.builder(schema).set("id", 2L).set("val1", "sam")
+            .set("file", fileText.toURI().toString()).build()
     );
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -428,7 +428,7 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
 
     String uri = inputFile.toURI().toString();
     ApplicationManager appManager = createSourceAndDeployApp(appName, inputFile, format, outputDatasetName, schema,
-                                                             delimeter, true, false);
+                                                             delimeter, true, false, false);
 
     appManager.getWorkflowManager(SmartWorkflow.NAME).startAndWaitForRun(ProgramRunStatus.COMPLETED,
                                                                          5, TimeUnit.MINUTES);
@@ -622,45 +622,56 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
 
   @Test
   public void testReadCSV() throws Exception {
-    testReadDelimitedText(FileFormat.CSV.name(), ",");
+    testReadDelimitedText(FileFormat.CSV.name(), ",", false);
+    testReadDelimitedText(FileFormat.CSV.name(), ",", true);
   }
 
   @Test
   public void testReadTSV() throws Exception {
-    testReadDelimitedText(FileFormat.TSV.name(), "\t");
+    testReadDelimitedText(FileFormat.TSV.name(), "\t", false);
+    testReadDelimitedText(FileFormat.TSV.name(), "\t", true);
   }
 
   @Test
   public void testReadDelimited() throws Exception {
-    testReadDelimitedText(FileFormat.DELIMITED.name(), "\u0001");
+    testReadDelimitedText(FileFormat.DELIMITED.name(), "\u0001", false);
+    testReadDelimitedText(FileFormat.DELIMITED.name(), "\u0001", true);
   }
 
-  private void testReadDelimitedText(String format, String delimiter) throws Exception {
+  private void testReadDelimitedText(String format, String delimiter, boolean enableQuotedValues) throws Exception {
     File fileText = new File(temporaryFolder.newFolder(), "test.txt");
     String outputDatasetName = UUID.randomUUID().toString();
 
     Schema schema = Schema.recordOf("user",
-                                    Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
-                                    Schema.Field.of("name", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
-                                    Schema.Field.of("file", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+        Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
+        Schema.Field.of("val1", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+        Schema.Field.of("val2", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+        Schema.Field.of("val3", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+        Schema.Field.of("file", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
 
     String appName = UUID.randomUUID().toString();
     ApplicationManager appManager = createSourceAndDeployApp(appName, fileText, format, outputDatasetName, schema,
-                                                             delimiter, false, true);
+        delimiter, false, true, enableQuotedValues);
 
+    String join = Joiner.on(delimiter).join(new String[] {"\"a", "b", "c\""});
     String inputStr = new StringBuilder()
-      .append("0").append("\n")
-      .append("1").append(delimiter).append("\n")
-      .append("2").append(delimiter).append("sam\n").toString();
+        .append("0").append("\n")
+        .append("1").append(delimiter).append(join).append("\n")
+        .append("2").append(delimiter).append("sam\n").toString();
     FileUtils.writeStringToFile(fileText, inputStr);
 
     appManager.getWorkflowManager(SmartWorkflow.NAME)
-      .startAndWaitForRun(ProgramRunStatus.COMPLETED, 5, TimeUnit.MINUTES);
+        .startAndWaitForRun(ProgramRunStatus.COMPLETED, 5, TimeUnit.MINUTES);
 
+    StructuredRecord split =
+        StructuredRecord.builder(schema).set("id", 1L).set("val1", "\"a").set("val2", "b").set("val3", "c\"")
+            .set("file", fileText.toURI().toString()).build();
     Set<StructuredRecord> expected = ImmutableSet.of(
-      StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 1L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 2L).set("name", "sam").set("file", fileText.toURI().toString()).build()
+        StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
+        !enableQuotedValues ? split :
+            StructuredRecord.builder(schema).set("id", 1L).set("val1", join.substring(1, join.length() - 1))
+                .set("file", fileText.toURI().toString()).build(),
+        StructuredRecord.builder(schema).set("id", 2L).set("val1", "sam").set("file", fileText.toURI().toString()).build()
     );
 
     DataSetManager<Table> outputManager = getDataset(outputDatasetName);
@@ -668,6 +679,7 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
 
     Assert.assertEquals(expected, output);
   }
+
 
   @Test
   public void testTextFormatWithoutOffset() throws Exception {
@@ -1080,13 +1092,14 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
 
   private ApplicationManager createSourceAndDeployApp(String appName, File file, String format,
                                                       String outputDatasetName, Schema schema) throws Exception {
-    return createSourceAndDeployApp(appName, file, format, outputDatasetName, schema, null, false, true);
+    return createSourceAndDeployApp(appName, file, format, outputDatasetName, schema, null, false, true, false);
   }
 
   private ApplicationManager createSourceAndDeployApp(String appName, File file, String format,
                                                       String outputDatasetName, Schema schema,
                                                       @Nullable String delimiter, boolean skipHeader,
-                                                      boolean includePath) throws Exception {
+                                                      boolean includePath,
+                                                      boolean enableQuotedValues) throws Exception {
 
     ImmutableMap.Builder<String, String> sourceProperties = ImmutableMap.<String, String>builder()
       .put(Constants.Reference.REFERENCE_NAME, appName + "TestFile")
@@ -1096,6 +1109,9 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
       .put("skipHeader", String.valueOf(skipHeader));
     if (includePath) {
       sourceProperties.put("pathField", "file");
+    }
+    if (enableQuotedValues) {
+      sourceProperties.put("enableQuotedValues", String.valueOf(enableQuotedValues));
     }
     if (delimiter != null) {
       sourceProperties.put("delimiter", delimiter);
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
@@ -643,36 +643,33 @@ public class FileBatchSourceTest extends ETLBatchTestBase {
     String outputDatasetName = UUID.randomUUID().toString();
 
     Schema schema = Schema.recordOf("user",
-        Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
-        Schema.Field.of("val1", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
-        Schema.Field.of("val2", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
-        Schema.Field.of("val3", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
-        Schema.Field.of("file", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
+      Schema.Field.of("id", Schema.of(Schema.Type.LONG)),
+      Schema.Field.of("val1", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+      Schema.Field.of("val2", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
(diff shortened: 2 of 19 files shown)


core-plugins/widgets/File-batchsource.json
```diff
@@ -76,6 +76,24 @@
...
+              "value": "true",
+              "label": "True"
+            },
+            "off": {
+              "value": "false",
+              "label": "False"
+            }
+          }
+        },
+
```
yaojiefeng: nit - extra line

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -109,6 +109,13 @@
     "'delimited'. Default value is false.")
   private Boolean skipHeader;
 
+  @Macro
+  @Nullable
+  @Description("Whether to treat content between quotes as a value. This value will only be used if the format " +
+      "is 'csv', 'tsv' or 'delimited'. Default value is false.")
+  protected Boolean enableQuotedValues;
+
+
```
yaojiefeng: nit - extra line

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/CSVInputFormatProvider.java
```diff
@@ -26,21 +26,23 @@
...
 @Name(CSVInputFormatProvider.NAME)
 @Description(CSVInputFormatProvider.DESC)
 public class CSVInputFormatProvider extends PathTrackingInputFormatProvider<DelimitedConfig> {
   static final String NAME = "csv";
   static final String DESC = "Plugin for reading files in csv format.";
   public static final PluginClass PLUGIN_CLASS =
-    new PluginClass(ValidatingInputFormat.PLUGIN_TYPE, NAME, DESC, CSVInputFormatProvider.class.getName(),
-                    "conf", DelimitedConfig.DELIMITED_FIELDS);
+      new PluginClass(
+          ValidatingInputFormat.PLUGIN_TYPE,
```
yaojiefeng: this indentation should be 2 instead of 4, please update your intellij settings. Please also don't split the line on each parameter. We only need to ensure the line is < 120 characters. This applies to all the rest code

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/CSVInputFormatProvider.java
```diff
@@ -65,14 +67,20 @@ public void validate(FormatContext context) {
     Schema schema = super.getSchema(context);
     FailureCollector collector = context.getFailureCollector();
     if (!conf.containsMacro(PathTrackingConfig.NAME_SCHEMA) && schema == null) {
-      collector.addFailure("CSV format cannot be used without specifying a schema.", "Schema must be specified.")
-        .withConfigProperty("schema");
+      collector
+          .addFailure(
```
yaojiefeng: don't need to split the line here

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -43,13 +44,12 @@
 import java.util.Map;
 import javax.annotation.Nullable;
 
-/**
- * Common config for delimited related formats.
- */
+/** Common config for delimited related formats. */
```
yaojiefeng: Don't need this change

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -58,11 +58,19 @@
...
+  public static final String DESC_ENABLE_QUOTES =
+      "Whether to treat content between quotes as a " + "value. Default value is false.";
+  public static final String DESC_SKIP_HEADER =
+      "Whether to skip the first line of each file. " + "Default value is false.";
 
   static {
     Map<String, PluginPropertyField> fields = new HashMap<>(FIELDS);
-    fields.put("skipHeader", new PluginPropertyField("skipHeader", DESC_SKIP_HEADER, "boolean", false, true));
+    fields.put(
+        "skipHeader",
```
yaojiefeng: don't need to split the lines

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -28,18 +28,17 @@
...
 import io.cdap.plugin.format.input.PathTrackingInputFormatProvider;
-
 import java.util.HashMap;
 import java.util.Map;
 import javax.annotation.Nullable;
 
-/**
- * Reads delimited text into StructuredRecords.
- */
+/** Reads delimited text into StructuredRecords. */
```
yaojiefeng: don't write javadoc like this, should just use the old structure

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -67,33 +81,46 @@ public void validate(FormatContext context) {
...
+        PathTrackingDelimitedInputFormat.SKIP_HEADER, String.valueOf(conf.getSkipHeader()));
+    properties.put(
+        PathTrackingDelimitedInputFormat.ENABLE_QUOTES_VALUE,
+        String.valueOf(conf.getEnableQuotedValues()));
   }
 
-  /**
-   * Plugin config for delimited input format
-   */
+  /** Plugin config for delimited input format */
```
yaojiefeng: Please revert all javadoc format changes

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -67,33 +81,46 @@ public void validate(FormatContext context) {
...
+    private static final String DELIMITER = "delimiter";
     private static final String DELIMITER_DESC = "Delimiter to use to separate record fields.";
 
     @Macro
     @Nullable
     @Description(DELIMITER_DESC)
     private String delimiter;
   }
-
-  private static PluginClass getPluginClass() {
```
yaojiefeng: why is this removed? i believe it is used by the static variable above, does the code still compile?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+  private static final char QUOTE_CHAR = '\"';
+
+  /**
+   * Split the delimited string based on the delimiter. The delimiter should not contain any quotes.
+   * The method will behave like this: 1. if there is no quote, it will behave same as {@link
+   * String#split(String)} 2. if there are quotes in the string, the method will find pairs of
+   * quotes, content within each pair of quotes will not get splitted even if there is delimiter in
+   * that. For example, if string is a."b.c"."d.e.f" and delimiter is '.', it will get split into
+   * [a, b.c, d.e.f]. if string is "val1.val2", then it will not get splitted since the '.' is
+   * within pair of quotes. if string is "val1.val2"", the quote right after val2 will be kept as
```
yaojiefeng: should also add "val1.val2"" scenario to the docs, and more generally on what will happen if the delimited string contains odd number of quotes

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+   * part of the value, so [val1, val2"].
```
yaojiefeng: also should document the behavior if the quotes are not at front and end of a string, for example, `a, b"c,d"e`, and delimiter is `,`

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+   * @param delimitedString the string to split
+   * @param delimiter the separtor
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static List<String> splitQuotesString(String delimitedString, String delimiter) {
+    // use a tree map so we can easily look up if a delimiter is within a pair of quotes
+    // the map has key of index and value to the index of paired quote. The map key is the index of
+    // the starting and the ending quote and the value is the index of the paired quote. An
+    // un-enclosed quote will trigger the exception and fail the pipeline. Example:
```
yaojiefeng: can you also add example for the bad case, i.e, there is unenclosed quote?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+   * part of the value, so [val1, val2"].
+   *
+   * @param delimitedString the string to split
+   * @param delimiter the separtor
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static List<String> splitQuotesString(String delimitedString, String delimiter) {
+    // use a tree map so we can easily look up if a delimiter is within a pair of quotes
+    // the map has key of index and value to the index of paired quote. The map key is the index of
```
yaojiefeng: remove `the map has key of index and value to the index of paired quote` as the following sentence already explains it

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+        quotesPos.put(firstQuotePos, i);
+        firstQuotePos = -1;
+      }
+    }
+
+    List<String> result = new ArrayList<>();
+    StringBuilder split = new StringBuilder();
+    for (int i = 0; i < delimitedString.length(); i++) {
+      // if the length is not enough for the delimiter, just add it to split
+      if (i + delimiter.length() > delimitedString.length()) {
```
yaojiefeng, line 82: do it like
```
if (i + delimiter.length() > delimitedString.length()) {
  split.append(...)
  continue;
}
...
```
to avoid the huge nesting block

yaojiefeng, line 82: do it like
```
if (i + delimiter.length() > delimitedString.length()) {
  split.append(...)
  continue;
}
...
```
to avoid the huge nesting block

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
yaojiefeng: After reading the code will it behave like that, the map for this value will be 
```
0 -> 9
9 -> 0
```
The last quote will get ignored right?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,30 +32,131 @@
...
+        if (!delimitedString.startsWith(delimiter, i)) {
+          split.append(delimitedString.charAt(i));
+          continue;
+        }
+
+        // find a delimiter
+        // find the prev quote and next quote index
+        Integer prevQuotePos = quotesPos.floorKey(i);
+        Integer nextQuotePos = quotesPos.ceilingKey(i);
+        // if the they exist and they form a pair, we will not split, just add the current character
```
yaojiefeng: `if the they` -> `if they`

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
yaojiefeng: should be 2021

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
yaojiefeng: I don't see the logic checks for un-enclosed quote and throw any exception about it, how will the pipeline fail?

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
```diff
@@ -0,0 +1,107 @@
...
+    expected = ImmutableList.of("a", "b,c", "d,e,f");
+    Assert.assertEquals(expected, PathTrackingDelimitedInputFormat.splitQuotesString(test, ","));
+
+    test = "a###\"b###c\"###\"d###e###f\"";
+    expected = ImmutableList.of("a", "b###c", "d###e###f");
+    Assert.assertEquals(expected, PathTrackingDelimitedInputFormat.splitQuotesString(test, "###"));
+  }
+
+  @Test
+  public void testBadQuotes() {
```
yaojiefeng: this test should never pass as no exception is thrown in the method. Should make sure the added unit tests pass before commit

yaojiefeng (commented):
Thanks for the commit. Took a first pass, there are quite a bit code standard issues, here is the code convention for the repo:
1. The indentation should be 2 characters instead of 4.
2. Do not split line on parameter, we normally split if the line is > 120 characters
3. Java doc format should not get changed.
4. Make sure to run a `mvn clean package -DskipTests` before commit to ensure the checkstyle passes
5. Make sure the added unit test passes before commit

core-plugins/widgets/File-batchsource.json
niceyang: Thanks, fixed.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
niceyang: Thanks, fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/CSVInputFormatProvider.java
niceyang: Fixed, it was made by the g4 fix tool.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
niceyang: Fixed, it was made by the g4 fix tool.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/CSVInputFormatProvider.java
niceyang: Sure, reverted.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
niceyang: Yeah, reversed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: Yeah, reversed

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: Yeah reverted

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: It's not removed, just moved up by the formatting tool.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: No the last quote will kept, the logic flow is ->
1. When encountering the delimiter, the current buffer is 'val1.val2', due to the quote
2. The quote right after the val2 will be ignore
3. no we have not encounter the real delimiter, so keep adding the last " into the value -> val1.val2"
4. the we encounter the delimiter and there is no match for the last quote, so there is a break

niceyang: The doc is refined :)

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: Yeah, added.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: Sorry, the behavior is updated here. It was decided to remove the restriction.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: ops, updated

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: oh, yeah. Updated.

niceyang: yeah

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: updated

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
niceyang: updated

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: Sorry the doc was not updated here. The pipeline wont fail in the odd number of quotes senario.

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
niceyang: apologize, this file is not up to date

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -58,11 +60,16 @@
...
+  public static final String DESC_ENABLE_QUOTES =
+    "Whether to treat content between quotes as a " + "value. Default value is false.";
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. " + "Default value is false.";
 
   static {
     Map<String, PluginPropertyField> fields = new HashMap<>(FIELDS);
     fields.put("skipHeader", new PluginPropertyField("skipHeader", DESC_SKIP_HEADER, "boolean", false, true));
+    fields.put(NAME_ENABLE_QUOTES_VALUES,
+      new PluginPropertyField("filenameOnly", DESC_ENABLE_QUOTES, "boolean", false, true));
```
yaojiefeng: here the name should not be "filenameOnly"m should be NAME_ENABLE_QUOTES_VALUES, this is causing the unit test to fail

core-plugins/widgets/File-batchsource.json
```diff
@@ -76,6 +76,22 @@
             "plugin-method": "getSchema"
           }
         },
+        {
+          "widget-type": "toggle",
+          "name": "enableQuotedValues",
+          "label": "Enable Quoted Values",
```
albertshau, line 82: there is a way to tell the UI to only show this based on the value of the format field. See https://github.com/cdapio/hydrator-plugins/blob/aedb88784e5e9f5dbd9b87db030ef7f8c61d184c/core-plugins/widgets/File-batchsink.json#L78 for an example.

Actually, the 'delimiter' field should follow the same behavior and only be displayed when the format is 'delimited'. Do you mind also making that change?

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,20 @@ JSON - is not supported, user has to manually provide the output schema.
 
 **Skip Header:** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
+The first field will have '1' as its value and the second will have 'a, b, c' as its value. The quote characters will be trimmed.
+The newline delimiter cannot be within quotes. For example, rows below are invalid:
+  
+* "ab
```
albertshau: this makes it difficult to read, I think it's fine to just say newlines are not allowed within quotes without the example.

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,20 @@ JSON - is not supported, user has to manually provide the output schema.
...
+     
+     c", d
+
+It also assumes the quotes are well enclosed. The left quote will match the first following quote right before the delimiter. If there is an 
+unenclosed quote, it will be included as part of data. For example:
+
+* "a, "b"c,d", e" -> [a, b"c,d, e]
```
albertshau: I don't follow why this output is expected. Is it 4 output columns (`a`, `b"c`, `d`, `e`) or is it two output columns (`a, b"c`, `d, e`)?

Also, from the discussion on the jira, I thought we had decided that unenclosed quotes would result in errors? That is the most restrictive and easiest to understand behavior. Generally it's better to be more restrictive when introducing a new feature.

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,20 @@ JSON - is not supported, user has to manually provide the output schema.
 
 **Skip Header:** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
```
albertshau, line 40: Does the entire value need to be quoted or can you have a quote in the middle of a value? For example, is the following valid?
 
```
"abc"def,ghijkl
```

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
```diff
@@ -109,6 +109,12 @@
     "'delimited'. Default value is false.")
   private Boolean skipHeader;
 
+  @Macro
+  @Nullable
+  @Description("Whether to treat content between quotes as a value. This value will only be used if the format " +
+    "is 'csv', 'tsv' or 'delimited'. Default value is false.")
```
albertshau: nit: Default value -> The default value

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -58,11 +60,16 @@
   public static final Map<String, PluginPropertyField> DELIMITED_FIELDS;
 
   // description
-  public static final String DESC_SKIP_HEADER = "Whether to skip the first line of each file. Default value is false.";
+  public static final String DESC_ENABLE_QUOTES =
+    "Whether to treat content between quotes as a " + "value. Default value is false.";
```
albertshau: don't need the + to concatenate these strings.

Default value -> The default value

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -58,11 +60,16 @@
...
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. " + "Default value is false.";
```
albertshau: same comment as above

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -67,21 +66,35 @@ public void validate(FormatContext context) {
     Schema schema = super.getSchema(context);
     FailureCollector collector = context.getFailureCollector();
     if (!conf.containsMacro(PathTrackingConfig.NAME_SCHEMA) && schema == null) {
-      collector.addFailure("Delimited format cannot be used without specifying a schema.",
-                           "Schema must be specified.").withConfigProperty("schema");
+      collector
+        .addFailure(
```
albertshau: nit: the first method call should be on the same line if possible.

```
collector.addFailure(
  "Delimited ...", ...)
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -67,21 +66,35 @@ public void validate(FormatContext context) {
...
+          .withConfigProperty("schema");
+    }
+
+    if (conf.containsMacro(DelimitedConfig.NAME_ENABLE_QUOTES_VALUES) || conf.containsMacro(Conf.DELIMITER)) {
+      return;
+    }
+
+    if (conf.getEnableQuotedValues() && conf.delimiter != null && conf.delimiter.contains("\"")) {
+      throw new IllegalArgumentException(
+        String.format("The delimeter %s cannot contain \" when quotes are enabled as value.", conf.delimiter));
```
albertshau: when quotes are enabled as value -> when quoted values are enabled

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -67,21 +66,35 @@ public void validate(FormatContext context) {
...
+          "Schema must be specified.")
+          .withConfigProperty("schema");
+    }
+
+    if (conf.containsMacro(DelimitedConfig.NAME_ENABLE_QUOTES_VALUES) || conf.containsMacro(Conf.DELIMITER)) {
+      return;
+    }
+
+    if (conf.getEnableQuotedValues() && conf.delimiter != null && conf.delimiter.contains("\"")) {
+      throw new IllegalArgumentException(
```
albertshau: instead of throwing an exception, this should be collector.addFailure(...).withConfigProperty("delimiter") so that the delimiter gets highlighted in a red box in the UI.

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,20 @@ JSON - is not supported, user has to manually provide the output schema.
...
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like '1, "a, b, c"' will output two fields.
```
albertshau: using the single quotes in '1, "a, b, c"' is a bit confusing. Does using backticks \` instead format it properly?

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -622,45 +622,57 @@ public void testReadJson() throws Exception {
...
-      .append("1").append(delimiter).append("\n")
+      .append("1").append(delimiter).append(join).append("\n")
       .append("2").append(delimiter).append("sam\n").toString();
     FileUtils.writeStringToFile(fileText, inputStr);
 
     appManager.getWorkflowManager(SmartWorkflow.NAME)
       .startAndWaitForRun(ProgramRunStatus.COMPLETED, 5, TimeUnit.MINUTES);
 
+    StructuredRecord split =
+        StructuredRecord.builder(schema).set("id", 1L).set("val1", "\"a").set("val2", "b").set("val3", "c\"")
```
albertshau: fix indentation

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
```diff
@@ -622,45 +622,57 @@ public void testReadJson() throws Exception {
...
+            .set("file", fileText.toURI().toString()).build();
     Set<StructuredRecord> expected = ImmutableSet.of(
-      StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 1L).set("file", fileText.toURI().toString()).build(),
-      StructuredRecord.builder(schema).set("id", 2L).set("name", "sam").set("file", fileText.toURI().toString()).build()
+        StructuredRecord.builder(schema).set("id", 0L).set("file", fileText.toURI().toString()).build(),
+        !enableQuotedValues ? split :
```
albertshau, line 671: this is difficult to read. Would be better to create a HashSet instead of ImmutableSet and add to it using an `if(enabledQuotedValues)` check

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,25 +32,130 @@
...
+    // key is the index of the starting and the ending quote and the value is the index of the
+    // paired quote. Example:
+    // pos: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
+    //      " | a | , | b | " | , | " | c | , | d | "  |
+    // key-value pairs in the map:
+    // 0 -> 4
+    // 4 -> 0
+    // 6 -> 10
+    // 10 -> 6
+    TreeMap<Integer, Integer> quotesPos = new TreeMap<>();
```
albertshau: this map approach seems to be overly complex (I think because it wants to handle unenclosed quotes in a particular way?). It's simpler and more efficient to make a single pass through and use a boolean to keep track of whether we're in a quoted value or not. Something more like

```
boolean isWithinQuotes = false;
for (int i = 0; i < delimitedString.length(); i++) {
  ...
  if (delimitedString.charAt(i) == QUOTE_CHAR) {
    isWithinQuote = !isWithinQuote;
    continue;
  }
  if (isDelimiter && !isWithinQuote) {
    // add new String
  }
}
```

Ideally it also wouldn't add elements to a list, but instead return an Iterator (or an Iterable), where the current position is kept as state. Can extend guava's AbstractIterator to make this simpler.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,25 +32,130 @@
...
+   * quote will be read as part of value. if the quotes are not at front and end of a string, for
+   * example: a, b"c,d"e and delimiter is ',', then the delimiter inside the quote won't break the
+   * value, so [a, b"c,d"e].
+   *
+   * @param delimitedString the string to split
+   * @param delimiter the separtor
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static List<String> splitQuotesString(String delimitedString, String delimiter) {
```
albertshau: making this return a List increases the memory usage and also forces an algorithm that makes at least 2 passes through the data instead of a single pass. It's better to return an Iterable or an Iterator. See below comment for more on this.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,25 +32,130 @@
...
+        split = new StringBuilder();
+      }
+    }
+
+    // add what is remaining to the result
+    result.add(trimQuotes(split.toString()));
+    return result;
+  }
+
+  /** Trim the quotes if and only if the start and end of the string are quotes */
```
albertshau: this is odd behavior to me, it means something like `a"b"c` gets read as-is but `"abc"` gets read as `abc`? Unless we have a concrete use case for this, I think it's simpler and more understandable if we always trim the quotes.

core-plugins/widgets/File-batchsource.json
niceyang: sure, updated.

core-plugins/docs/File-batchsource.md
niceyang: got, updated.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSourceConfig.java
niceyang: Fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
niceyang: Fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
niceyang: Fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: sure, fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: Fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
niceyang: Yeah, fixed.

core-plugins/docs/File-batchsource.md
niceyang: yeah, is it more clear just removing the `?

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
niceyang: sure, updated.

core-plugins/docs/File-batchsource.md
albertshau: It's best to use backtick to enclose instead of single quote, as that should format differently.

core-plugins/docs/File-batchsource.md
niceyang: The quote is used to allow the delimiter to be in the value, if it's in the middle of the value, it would be still valid unless it's unclosed.

core-plugins/docs/File-batchsource.md
niceyang: Yeah updated.

core-plugins/src/test/java/io/cdap/plugin/batch/source/FileBatchSourceTest.java
niceyang: Fixed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
niceyang: Yeah, let's always trim the quotes

niceyang:
This PR is obsolete due to the rebase issues.

A new PR is created and merged for the feature/add quote option 

Link of new PR:
https://github.com/cdapio/hydrator-plugins/pull/1498

PR #1486, opened 2021-08-20
[CDAP-18290] ValidationException should log stacktraces

ghost:
[CDAP-18290](https://cdap.atlassian.net/browse/CDAP-18290)

This change is padding empty strings ("") at the end in case there are fewer values than required in delimited string. This the same behaviour displayed by Spark during pipeline execution.

Whenever the number of values in a row (obtained after splitting by delimiter) is less than the number of values in the file’s first row (considered as the header), an unhandled ArrayIndexOutOfBounds exception is thrown. The message of the exception is the index that was accessed. This message is visible in the UI when we press the "Get Schema" button for the source node or clicking on the Deploy button in Pipeline Studio without providing source schema.

The The “2” in the exception message refers to the index that caused the exception.
This change compares the number of elements in the header row and the current row being read from source and adds the an error to the Failure Collector if required with meaningful message before ArrayIndexOutOfBounds exception is raised.

Steps to reproduce bug are given in [Jira](https://cdap.atlassian.net/browse/CDAP-18290).

Before:

Screenshot 2021-08-20 at 8 19 41 PM

After:

No error is thrown, last column is detected as Nullable.

### DIFF
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
@@ -42,9 +42,15 @@ public class DataTypeDetectorUtils {
   public static void detectDataTypeOfRowValues(Map<String, Schema> override,
                                                DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper,
                                                String[] columnNames, String[] rowValues) {
+    ArrayList<String> rowValuesList = new ArrayList<>(Arrays.asList(rowValues));
+    // Pad empty strings at the end if fewer values than required
+    // This is the same behaviour exhibited by Spark during pipeline execution
+    while (rowValuesList.size() < columnNames.length) {
+      rowValuesList.add("");
+    }
     for (int columnIndex = 0; columnIndex < columnNames.length; columnIndex++) {
       String name = columnNames[columnIndex];
-      String value = rowValues[columnIndex];
+      String value =  rowValuesList.get(columnIndex);
       if (!override.containsKey(name)) {
         dataTypeDetectorStatusKeeper.addDataType(name, DataTypeDetectorStatusKeeper.detectValueDataType(value));
       }
--- a/format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
+++ b/format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
@@ -116,4 +116,31 @@ public class DataTypeDetectorUtilsTest {
 
     assertEquals(expectedDataTypeDetectorStatusKeeper, actualDataTypeDetectorStatusKeeper);
   }
+
+  @Test
+  public void testDetectDataTypeOfEachDatasetColumnWithMissingValues() {
+    String[] columnNames = new String[]{"column_A", "column_B", "column_C"};
+    String[] rowValues1 = new String[]{"100A", "2020", "129.99"};
+    // Row 2 has two values instead of 3, so empty string will be padded at the end
+    String[] rowValues2 = new String[]{"Happy", "17.99"};
+    DataTypeDetectorStatusKeeper actualDataTypeDetectorStatusKeeper = new DataTypeDetectorStatusKeeper();
+    DataTypeDetectorUtils.detectDataTypeOfRowValues(new HashMap<>(), actualDataTypeDetectorStatusKeeper, columnNames,
+                                                    rowValues1);
+    DataTypeDetectorUtils.detectDataTypeOfRowValues(new HashMap<>(), actualDataTypeDetectorStatusKeeper, columnNames,
+                                                    rowValues2);
+    actualDataTypeDetectorStatusKeeper.validateDataTypeDetector();
+    Map<String, Schema> override = new HashMap<>();
+
+    List<Schema> actualFields = DataTypeDetectorUtils
+      .detectDataTypeOfEachDatasetColumn(override, columnNames, actualDataTypeDetectorStatusKeeper)
+      .stream().map(Schema.Field::getSchema).collect(Collectors.toList());
+
+    List<Schema> expectedFields = new ArrayList<>(Arrays.asList(
+      Schema.Field.of(columnNames[0], Schema.of(Schema.Type.STRING)),
+      Schema.Field.of(columnNames[1], Schema.of(Schema.Type.DOUBLE)),
+      Schema.Field.of(columnNames[2], Schema.nullableOf(Schema.of(Schema.Type.DOUBLE)))
+    )).stream().map(Schema.Field::getSchema).collect(Collectors.toList());
+
+    assertEquals(expectedFields, actualFields);
+  }
 }


format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
```diff
@@ -96,7 +96,7 @@ public void testDetectDataTypeOfEachDatasetColumn() {
   }
 
   @Test
-  public void testDetectDataTypeOfRowValue() {
+  public void testDetectDataTypeOfRowValue() throws RuntimeException {
```
CuriousVini: For runtime exception there is no need to add `throws`

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -41,7 +41,12 @@
    */
   public static void detectDataTypeOfRowValues(Map<String, Schema> override,
                                                DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper,
-                                               String[] columnNames, String[] rowValues) {
+                                               String[] columnNames, String[] rowValues) throws RuntimeException {
+    if (rowValues.length < columnNames.length) {
+      throw new RuntimeException(String.format(
```
CuriousVini: could we just add this to failure collector instead?

format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
ghost: Removed

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
ghost: The default sample size for getting schema is 1000, there could potentially be many rows with missing fields. This may result in having many duplicate failures in the failure collector. 
The exception being thrown will eventually be caught by a failure collector [here](https://github.com/cdapio/hydrator-plugins/blob/ac2ae149e30c28a936f0860c633bba03ada634d8/format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingInputFormatProvider.java#L76) .
All other exceptions in the [calling function](https://github.com/cdapio/hydrator-plugins/blob/bugfix/cdap-18290/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java#L162) are also being thrown instead of using failure collector. 
Since FormatContext or FailureCollector aren't being passed to the calling functions, I will have to pass them as a new argument and probably need convert other throw statements to collector.addFailure() in [DelimitedConfig.getDefaultSchema()](https://github.com/cdapio/hydrator-plugins/blob/bugfix/cdap-18290/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java#L162).
Is this required?

ghost: Changed. The getSchema and getDefaultSchema were present in DelimitedConfig.java instead of using them to override functions in DelimitedInputFormatProvider.java. As a result it was not easy to get access to format context or failure collector. I had to move these functions over. I referred the style followed in [format-avro](https://github.com/cdapio/hydrator-plugins/blob/develop/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java) while doing this.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -38,10 +40,20 @@
    * @param dataTypeDetectorStatusKeeper Object that keeps the state of automated data type detection process.
    * @param columnNames Column names.
    * @param rowValues Row values.
+   * @param collector Failure Collector to store failures
    */
   public static void detectDataTypeOfRowValues(Map<String, Schema> override,
                                                DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper,
-                                               String[] columnNames, String[] rowValues) {
+                                               String[] columnNames, String[] rowValues,
+                                               @Nullable FailureCollector collector) {
```
CuriousVini: why do we make it nullable?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -38,10 +40,20 @@
...
+    if (rowValues.length < columnNames.length) {
+      if (collector != null) {
+        collector.addFailure(String.format(
+          "Number of values in row (%d) is less than number of values in header (%d).",
+          rowValues.length, columnNames.length), String.format("Row Values: %s", Arrays.toString(rowValues)));
```
CuriousVini: In this case corrective action will be to make sure row contains same number of values as columns.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -38,10 +40,20 @@
...
    */
   public static void detectDataTypeOfRowValues(Map<String, Schema> override,
                                                DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper,
-                                               String[] columnNames, String[] rowValues) {
+                                               String[] columnNames, String[] rowValues,
+                                               @Nullable FailureCollector collector) {
+    if (rowValues.length < columnNames.length) {
+      if (collector != null) {
+        collector.addFailure(String.format(
+          "Number of values in row (%d) is less than number of values in header (%d).",
```
CuriousVini: Its possible that `header` is not present in csv file. Should we change error message to `number of columns`?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -96,4 +109,79 @@ private static PluginClass getPluginClass() {
     return new PluginClass(ValidatingInputFormat.PLUGIN_TYPE, NAME, DESC, DelimitedInputFormatProvider.class.getName(),
                            "conf", properties);
   }
+
+  @Nullable
```
CuriousVini: when can schema be null?

format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
```diff
@@ -102,7 +102,7 @@ public void testDetectDataTypeOfRowValue() {
     String[] rowValues = new String[]{"2", "24L", "3.14", "false", "100.05", "100A"};
 
     DataTypeDetectorUtils.detectDataTypeOfRowValues(new HashMap<>(), actualDataTypeDetectorStatusKeeper, columnNames,
-                                                    rowValues);
+                                                    rowValues, null);
```
CuriousVini: Can we use MockFailureCollector?
https://github.com/cdapio/hydrator-plugins/blob/ac2ae149e30c28a936f0860c633bba03ada634d8/core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerConfigTest.java#L92

CuriousVini (commented):
A few comments

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
arjansbal: I'm not sure.  I am referring to similar function declared in [avro-format](https://github.com/cdapio/hydrator-plugins/blob/aedb88784e5e9f5dbd9b87db030ef7f8c61d184c/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java#L89-L107) and [format-parquet](https://github.com/cdapio/hydrator-plugins/blob/aedb88784e5e9f5dbd9b87db030ef7f8c61d184c/format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java#L72-L83). I think it may be a mistake here and in the files I'm referring. Can you have a look and confirm?

format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
ghost: Yes, changed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
ghost: When schema contains macros, super.getSchema(context) should return `null`.  I am referring to similar function declared in [format-avro](https://github.com/cdapio/hydrator-plugins/blob/aedb88784e5e9f5dbd9b87db030ef7f8c61d184c/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java#L89-L107) and [format-parquet](https://github.com/cdapio/hydrator-plugins/blob/aedb88784e5e9f5dbd9b87db030ef7f8c61d184c/format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java#L72-L83).
Before this change, [the function in DelimitedConfig.java](https://github.com/cdapio/hydrator-plugins/blob/develop/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java#L85-L86) used to return `null` directly.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
ghost: Changed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
ghost: You are right, changed message.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
ghost: The reason for this is that the calling function [getDefaultSchema](https://github.com/cdapio/hydrator-plugins/blob/dba265837f986b7bb95b155b07e73266a51d3560/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java#L136-L186) receives FormatContext as a nullable param. If context is null, context.getFailureCollector cannot be called, so collector will be null. 

I've referred to [this function](https://github.com/cdapio/hydrator-plugins/blob/develop/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java#L116-L148). I'm not sure why getDefaultSchema here has context as a Nullable param. The only function that I could find calling it, getSchema in the same file, has context as a non-nullable parameter. The implementation of getDefaultSchema doesn't check context for null value before accessing it. So it seems that context should never have a null value in format-avro.

I have removed `@Nullable` from context param in the calling function getDefaultSchema and from collector in the above lines. Removed null value checks checks too.

rmstar (approved):
Please squash commits

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
ghost: Change discarded

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
ghost: Change discarded.

PR #1487, opened 2021-08-24
[CDAP-15794] making the error message consistent when delimited schema is wrong

itsankit-google:
This change is for making the error message consistent when the data has more fields than the schema by using the same lib to print the length of data fields in splitted delimited string. 
Previously, splitting of every row was implemented using Splitter library of Guava but while printing the error message the length was being printed using java String.split() method.

Added a loop for counting the no of fields in the splitted data and removed the String.split() method.

Before-
Screen Shot 2021-08-20 at 6 18 57 PM

After-
Screen Shot 2021-08-23 at 9 57 13 AM

JIRA - https://cdap.atlassian.net/browse/CDAP-15794

### DIFF
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
@@ -80,10 +80,14 @@ public class PathTrackingDelimitedInputFormat extends PathTrackingInputFormat {
 
         StructuredRecord.Builder builder = StructuredRecord.builder(schema);
         Iterator<Schema.Field> fields = schema.getFields().iterator();
+        Iterable<String> splitDelimitedString = Splitter.on(delimiter).split(delimitedString);
+        int numDataFields = 0;
+        for (String part : splitDelimitedString) {
+          numDataFields++;
+        }
 
-        for (String part : Splitter.on(delimiter).split(delimitedString)) {
+        for (String part : splitDelimitedString) {
           if (!fields.hasNext()) {
-            int numDataFields = delimitedString.split(delimiter).length;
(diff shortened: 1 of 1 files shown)


format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -80,10 +80,14 @@ public NullWritable getCurrentKey() {
 
         StructuredRecord.Builder builder = StructuredRecord.builder(schema);
         Iterator<Schema.Field> fields = schema.getFields().iterator();
+        Iterable<String> splitDelimitedString = Splitter.on(delimiter).split(delimitedString);
+        int numDataFields = 0;
+        for (String part : splitDelimitedString) {
```
albertshau: This will cause an extra loop through every record. It's better to only do this in the error scenario. 

That being said, I don't understand when the Splitter would return a different number of elements than split(). What would be an example of when they are different?

itsankit-google: When splitting this string "2018,9,111,NIXON,,JESSE,02/01/1840,Database Engineer,Baseball Operations,New York Yankees,07/07/2014,,,,Male,,,White,Baseball Ops,,\"Group 3 - Professionals, Technicians, Supervisors\",Active,," the output from Splitter is 26 fields but the output from split() is 24 fields which I think is because the latter by default removed trailing empty strings.

itsankit-google: Added the numDataFields loop in the error scenario.

albertshau (approved):

PR #1488, opened 2021-08-27
Adding additional docs to the Hasher plugin : the fields must be non nullable

ashamshad:
When debugging the Hasher plugin, we found out that the `fields` must be `non nullable`, as well as being of type `string`.
This PR is to add that information in the plugin docs.

### DIFF
--- a/transform-plugins/docs/Hasher-transform.md
+++ b/transform-plugins/docs/Hasher-transform.md
@@ -4,13 +4,10 @@
 Description
 -----------
 Hashes fields using a digest algorithm such as ``MD2``, ``MD5``, ``SHA1``, ``SHA256``, ``SHA384``, or ``SHA512``.<br>
+The **fields** to be hashed must be of type `string` and `non-nullable`.
 
 Configuration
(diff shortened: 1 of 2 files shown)


transform-plugins/docs/Hasher-transform.md
```diff
@@ -3,11 +3,14 @@
...
 
 Configuration
 -------------
 **fields:** Specifies the fields to be hashed.
 
 **hash:** Specifies the hashing algorithm.
+
+Prerequisites
+-------------
+The **fields** to be hashed must be of type `string` and `non-nullable`.
```
albertshau: Put this as a sentence in the fields description instead of in a separate section.

albertshau (approved):
minor comment about moving the content, otherwise lgtm.

transform-plugins/docs/Hasher-transform.md
ashamshad: @albertshau Thanks for reviewing, made the change as per suggested.

PR #1495, opened 2021-09-15
[PLUGIN-754] Fix for certain avro schemas being parsed incorrectly

DJSagarAhire:
This is a fix for the bug [PLUGIN-754](https://cdap.atlassian.net/browse/PLUGIN-754).

On investigation, the following issues were observed:
* **Missing key for map.** The map datatype is always assumed to have a key type and a value type in the CDAP schema, but in an avro schema the [key type of a map is assumed to be a string](https://avro.apache.org/docs/current/spec.html#Maps) and therefore not explicitly mentioned. This PR adds a string key to the JSON representation of an avro map schema before converting it into a CDAP schema.
* **No support for referencing custom types.** An enum or record type mentioned in an avro schema cannot be referenced within the same schema as it is not a registered type in the CDAP schema. This PR substitutes such referenced types in the JSON representation of the avro schema before converting it into a CDAP schema.

Output for the same file mentioned in the bug report after the fix:

image

Known issue: The array type reference ("Score" in the above image) is resolved in the AvroToStructuredTransformer class as per the debugger but the UI does not receive this information, [throwing an error](https://github.com/cdapio/ui-schema-parser/blob/7fa333e0fe1208de6adc17e597ec847f5ae84124/lib/types.js#L101). This will be handled in a follow-up PR if possible.

[PLUGIN-754]: https://cdap.atlassian.net/browse/PLUGIN-754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
+++ b/format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
@@ -18,6 +18,7 @@ package io.cdap.plugin.format.avro;
 
 import com.google.common.collect.Maps;
 import com.google.gson.Gson;
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import io.cdap.cdap.api.data.format.StructuredRecord;
@@ -36,6 +37,8 @@ import javax.annotation.Nullable;
  * Create StructuredRecords from GenericRecords
  */
 public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
+
+  private static final Gson GSON = new Gson();
 
   private final Map<Integer, Schema> schemaCache = Maps.newHashMap();
   private final Map<String, JsonObject> customTypes = Maps.newHashMap();
@@ -96,7 +99,7 @@ public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, 
       structuredSchema = schemaCache.get(hashCode);
     } else {
       String strSchema = schema.toString();
-      String jsonSchema = preprocessSchema(new Gson().fromJson(strSchema, JsonObject.class)).toString();
+      String jsonSchema = preprocessSchema(GSON.fromJson(strSchema, JsonObject.class)).toString();
       structuredSchema = Schema.parseJson(jsonSchema);
       schemaCache.put(hashCode, structuredSchema);
     }
@@ -112,23 +115,17 @@ public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, 
     JsonElement type = schema.get("type");
 
     if (type.isJsonArray()) {   // Union
-      for (JsonElement subtype : type.getAsJsonArray()) {
-        if (!subtype.isJsonPrimitive()) {
-          preprocessSchema(subtype.getAsJsonObject());
-        } else if (customTypes.containsKey(subtype.getAsString())) {
-          schema.remove("type");
-          schema.add("type", customTypes.get(subtype.getAsString()));
-        }
-      }
+      JsonArray processedUnion = preprocessUnion(type.getAsJsonArray());
+      schema.remove("type");
+      schema.add("type", processedUnion);
     } else if (type.isJsonObject()) {   // Unnamed Complex type
+
       preprocessSchema(type.getAsJsonObject());
     } else {
       String typeName = type.getAsString();
+
       switch (typeName) {
         case "record":
-          if (!schema.has("fields")) {
-            return schema;
-          }
           for (JsonElement field : schema.get("fields").getAsJsonArray()) {
             preprocessSchema(field.getAsJsonObject());
           }
@@ -137,19 +134,13 @@ public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, 
 
         case "map":
           schema.addProperty("keys", "string");
-          if (!schema.get("values").isJsonPrimitive()) {
-            preprocessSchema(schema.getAsJsonObject("values"));
-          }
+          JsonElement value = schema.get("values");
+          processElementInSchema(schema, value, "values");
           break;
 
         case "array":
           JsonElement items = schema.get("items");
-          if (!items.isJsonPrimitive()) {
-            preprocessSchema(schema.getAsJsonObject("items"));
-          } else if (customTypes.containsKey(items.getAsString())) {
-            schema.remove("items");
-            schema.add("items", customTypes.get(items.getAsString()));
-          }
+          processElementInSchema(schema, items, "items");
           break;
 
         case "enum":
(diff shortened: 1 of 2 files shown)


format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +95,76 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
     if (schemaCache.containsKey(hashCode)) {
       structuredSchema = schemaCache.get(hashCode);
     } else {
-      structuredSchema = Schema.parseJson(schema.toString());
+      String strSchema = schema.toString();
+      String jsonSchema = preprocessSchema(new Gson().fromJson(strSchema, JsonObject.class)).toString();
```
tivv: Please don't create `Gson` each time, it's pretty expensive. Follow example of other places where we create it once for a class.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +95,76 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
+        }
+      }
+    } else if (type.isJsonObject()) {   // Unnamed Complex type
+      preprocessSchema(type.getAsJsonObject());
+    } else {
+      String typeName = type.getAsString();
+      switch (typeName) {
+        case "record":
+          if (!schema.has("fields")) {
+            return schema;
```
tivv: Do you still need to put it into customTypes?

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +95,76 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
+
+    JsonElement type = schema.get("type");
+
+    if (type.isJsonArray()) {   // Union
+      for (JsonElement subtype : type.getAsJsonArray()) {
+        if (!subtype.isJsonPrimitive()) {
+          preprocessSchema(subtype.getAsJsonObject());
+        } else if (customTypes.containsKey(subtype.getAsString())) {
+          schema.remove("type");
+          schema.add("type", customTypes.get(subtype.getAsString()));
```
tivv: Could you please explain what are you doing here (probably even with a comment). I see that for a union you are replacing the whole thing with just a subtype, e.g. `{type:[A,B,C]}` will be replaced with `{type:$(customTypes.get("C")}`. Why?

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +95,76 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
+          }
+          break;
+
+        case "array":
+          JsonElement items = schema.get("items");
+          if (!items.isJsonPrimitive()) {
+            preprocessSchema(schema.getAsJsonObject("items"));
+          } else if (customTypes.containsKey(items.getAsString())) {
+            schema.remove("items");
+            schema.add("items", customTypes.get(items.getAsString()));
```
tivv: So, you are replacing type references with named type definitions. Why? Could you please explain in the PR description what is it doing and why. As of now, it would potentially create tons of duplicates of named type definitions within schema that share same name.

tivv: @fernst could you please review this as well? I am not sure about duplicating records with same name

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -35,6 +38,7 @@
 public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
 
   private final Map<Integer, Schema> schemaCache = Maps.newHashMap();
+  private final Map<String, JsonObject> customTypes = Maps.newHashMap();
```
tivv, line 45: As soon as you do the resolution, please implement it with namespace support, see https://avro.apache.org/docs/current/spec.html#names

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
fernst: Agree. `GSON` should be a static instance.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
DJSagarAhire: If a record does not have the `fields` attribute, it's an error. I thought I'd immediately return it, but now I intend to just not handle this case explicitly so that an exception will be thrown in such a case.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
DJSagarAhire: That's an error - if `{type:[A,B,C]}` is found out of which `C` is a custom type, then only `C` should be replaced. Making that change now.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
DJSagarAhire: This is to handle the point "No support for referencing custom types" mentioned in the PR description above. If this is not done, the CDAP schema attempts to parse the type reference using one of the built in types, which it does not find, leading to an error.

For example, if an enum named `myenum` is already defined and referenced in another field as `{"name": "foo", "type": "myenum"}`, it throws an error `No enum constant io.cdap.cdap.api.data.schema.Schema.Type.MYENUM` as it is not present in [Schema.Type](https://github.com/cdapio/cdap/blob/develop/cdap-api-common/src/main/java/io/cdap/cdap/api/data/schema/Schema.java#L57-L72).

tivv: Should you then remove "name" when doing substitutions? Or set unique name for each substitution? Otherwise we have many copies with same name. I believe @fernst was solving similar problem

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +98,96 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
-      structuredSchema = Schema.parseJson(schema.toString());
+      String strSchema = schema.toString();
+      String jsonSchema = preprocessSchema(GSON.fromJson(strSchema, JsonObject.class)).toString();
+      structuredSchema = Schema.parseJson(jsonSchema);
       schemaCache.put(hashCode, structuredSchema);
     }
     return structuredSchema;
   }
+
+  private JsonObject preprocessSchema(JsonObject schema) {
```
tivv: In one place you are using returned JsonObject (that is always the input as of now), in others you assume it's mutating the input. Please use same style.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +98,96 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
+
+        case "enum":
+          customTypes.put(schema.getAsJsonPrimitive("name").getAsString(), schema);
+          break;
+
+        default:
+          if (customTypes.containsKey(typeName)) {
+            schema.remove("type");
+            schema.add("type", customTypes.get(typeName));
+          }
```
tivv, line 154: Are you dropping other things like primitive types? E.g. how somelhing like `"fields" : [
	      {"name": "a", "type": "long"},
	      {"name": "b", "type": "string"}
	      ]` work?

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
DJSagarAhire: I've changed it so that every method that accepts a `JsonObject` modifies it in place and returns nothing.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
DJSagarAhire: Added namespace resolution. Please have a look and let me know if it looks good.

fernst:
Is this pull request still relevant? If so, please rebase and resolve conflicts. If not, please close this PR.

PR #1496, opened 2021-09-21
[PLUGIN-754] Add String key to avro maps in AvroToStructuredTransformer

DJSagarAhire:
This is a fix for the bug [PLUGIN-754](https://cdap.atlassian.net/browse/PLUGIN-754).

The reason for the exception was that the map datatype in avro was missing a key type. The map datatype is always assumed to have a key type and a value type in the CDAP schema, but in an avro schema the [key type of a map is assumed to be a string](https://avro.apache.org/docs/current/spec.html#Maps) and therefore not explicitly mentioned. This PR adds a string key to the JSON representation of an avro map schema before converting it into a CDAP schema.

While investigating this bug, another issue was found relating to custom types defined in the schema being referenced elsewhere in the same schema. The reference is not being detected currently by CDAP. This will be fixed in [a separate PR](https://github.com/cdapio/hydrator-plugins/pull/1495).

### DIFF
--- a/format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
+++ b/format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
@@ -37,6 +37,8 @@ import javax.annotation.Nullable;
  */
 public class AvroToStructuredTransformer extends RecordConverter<GenericRecord, StructuredRecord> {
 
+  private static final Gson GSON = new Gson();
+
   private final Map<Integer, Schema> schemaCache = Maps.newHashMap();
 
   public StructuredRecord transform(GenericRecord genericRecord) throws IOException {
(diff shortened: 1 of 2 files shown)


tivv (commented):
Could you please add a test?

DJSagarAhire:
> Could you please add a test?

Done.

format-avro/src/main/java/io/cdap/plugin/format/avro/AvroToStructuredTransformer.java
```diff
@@ -91,9 +94,58 @@ public Schema convertSchema(org.apache.avro.Schema schema) throws IOException {
...
-      structuredSchema = Schema.parseJson(schema.toString());
+      String strSchema = schema.toString();
+      String jsonSchema = preprocessSchema(new Gson().fromJson(strSchema, JsonObject.class)).toString();
+      structuredSchema = Schema.parseJson(jsonSchema);
       schemaCache.put(hashCode, structuredSchema);
     }
     return structuredSchema;
   }
+
+  private JsonObject preprocessSchema(JsonObject schema) {
```
tivv: Please make it void to make the usage consistent. Other than that looks good for me

DJSagarAhire: Right, that was in one of the later commits in the other PR. Fixed.

tivv (approved):

PR #1498, opened 2021-10-01
Feature/add quote option in file source

niceyang:


format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -58,11 +59,16 @@
   public static final Map<String, PluginPropertyField> DELIMITED_FIELDS;
 
   // description
-  public static final String DESC_SKIP_HEADER = "Whether to skip the first line of each file. Default value is false.";
+  public static final String DESC_ENABLE_QUOTES =
+    "Whether to treat content between quotes as a value. The default value is false.";
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. The Default value is false.";
```
albertshau: Default -> default

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,16 +31,82 @@
...
+        split = new StringBuilder();
+        i = i + delimiter.length() - 1;
+        continue;
+      }
+
+      // delimiter within quotes
+      split.append(cur);
+    }
+    if (isWithinQuotes) {
+      throw new IOException("Quotes are not enclosed.");
```
albertshau: Found a line with an unenclosed quote. Ensure that all values are properly quoted, or disable quoted values.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -80,18 +148,27 @@ public NullWritable getCurrentKey() {
...
+        Iterable<String> splits;
+        if (!enableQuotesValue) {
+          splits = Splitter.on(delimiter).split(delimitedString);
+        } else {
+          splits = splitQuotesString(delimitedString, delimiter);
+        }
 
-        for (String part : splitDelimitedString) {
+        int numSchemaFields = schema.getFields().size();
+        int numDataFields = 0;
```
albertshau: this should be moved back to where it was before. It is only done in the error scenario in order to prevent looping an extra time for each line.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -30,16 +31,82 @@
...
+   * @param delimitedString the string to split
+   * @param delimiter the separtor
+   * @return a list of splits of the original string
+   */
+  @VisibleForTesting
+  static Iterable<String> splitQuotesString(String delimitedString, String delimiter)
+    throws IOException {
+
+    boolean isWithinQuotes = false;
+    List<String> result = new ArrayList<>();
```
albertshau: since the results are stored in a List, this actually ends up looping twice for each line. Once to build this list and then once again to go through the List iterator. It also ends up using extra memory.

It is better to create something like a SplitQuotesIterator that has mostly the same logic, except it keeps track of i and only returns the next value when needed. This way it only loops once. You can extend AbstractIterator from guava to simplify the implementation.

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,14 @@ JSON - is not supported, user has to manually provide the output schema.
 
 **Skip Header:** Whether to skip the first line of each file. Supported formats are 'text', 'csv', 'tsv', 'delimited'.
 
+**Enable Quoted Values** Whether to treat content between quotes as a value. This value will only be used if the format
+is 'csv', 'tsv' or 'delimited'. For example, if this is set to true, a line that looks like `1, "a, b, c"` will output two fields.
+The first field will have '1' as its value and the second will have `a, b, c` as its value. The quote characters will be trimmed.
```
albertshau: for consistency, use backticks instead of single quote for the '1'

core-plugins/docs/File-batchsource.md
```diff
@@ -37,6 +37,14 @@ JSON - is not supported, user has to manually provide the output schema.
...
+The newline delimiter cannot be within quotes.
+
+It also assumes the quotes are well enclosed. The left quote will match the first following quote right before the delimiter. If there is an
+unenclosed quote, an exception will be thrown.
```
albertshau: nit: many of our users don't know what an exception is. an exception will be thrown -> an error will occur.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -101,22 +111,34 @@ public NullWritable getCurrentKey() {
...
+            // Ensure if date time field, value is in correct format
             SchemaValidator.validateDateTimeField(nextField.getSchema(), fieldName, part);
             builder.convertAndSet(fieldName, part);
           }
         }
         return builder;
       }
 
+      private Iterator<String> getSplitsIterator(boolean enableQuotesValue, String delimitedString, String delimiter) {
+        if (!enableQuotesValue) {
```
albertshau: nit: a little more readable to do `if (enableQuotesValue)` instead of the opposite.

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
```diff
@@ -0,0 +1,117 @@
...
+import io.cdap.plugin.format.delimited.input.PathTrackingDelimitedInputFormat.SplitQuotesIterator;
+import org.junit.Assert;
+import org.junit.Test;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+
+public class DelimitedStringTest {
```
albertshau: nit: name this SplitQuotesIteratorTest

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+  private SplitQuotesIterator splitQuotesIterator;
```
albertshau: looks unused?

format-delimited/src/test/java/io/cdap/plugin/format/delimited/input/DelimitedStringTest.java
```diff
@@ -0,0 +1,117 @@
...
+    Assert.assertEquals(expected, getListFromIterator(splitQuotesIterator = new SplitQuotesIterator(test, ",")));
+
+    test = "a###\"b###c\"###\"d###e###f\"";
+    expected = ImmutableList.of("a", "b###c", "d###e###f");
+    Assert.assertEquals(expected, getListFromIterator(splitQuotesIterator = new SplitQuotesIterator(test, "###")));
+  }
+
+  @Test
+  public void testBadQuotes() throws Exception {
+    final String test = "Value1,value2.1 value2.2\"value2.2.1,value2.3\",val\"ue3,value4";
```
albertshau: don't need the final here.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -128,4 +150,82 @@ public void close() throws IOException {
...
+   * trimed.
+   */
+  @VisibleForTesting
+  static class SplitQuotesIterator extends AbstractIterator<String> {
+    private String delimitedString;
+    private String delimiter;
+    private int index;
+    private boolean endingWithDelimiter = false;
+
+    private SplitQuotesIterator () {}
```
albertshau: I think this breaks the checkstyle. Don't need this one, the other constructor should always be used.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -128,4 +150,82 @@ public void close() throws IOException {
...
+   * will find pairs of quotes, content within each pair of quotes will not get splitted even if there
+   * is delimiter in that. For example, if string is a."b.c"."d.e.f" and delimiter is '.', it will get
+   * split into [a, b.c, d.e.f]. if string is "val1.val2", then it will not get splitted since the '.'
+   * is within pair of quotes. If the delimited string contains odd number of quotes, which mean the
+   * quotes are not closed, an exception will be thrown. The quote within the value will always be
+   * trimed.
+   */
+  @VisibleForTesting
+  static class SplitQuotesIterator extends AbstractIterator<String> {
+    private String delimitedString;
```
albertshau: private final

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -128,4 +150,82 @@ public void close() throws IOException {
...
+    private String delimiter;
```
albertshau: private final

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -128,4 +150,82 @@ public void close() throws IOException {
...
+   * will behave same as {@link String#split(String)} 2. if there are quotes in the string, the method
+   * will find pairs of quotes, content within each pair of quotes will not get splitted even if there
+   * is delimiter in that. For example, if string is a."b.c"."d.e.f" and delimiter is '.', it will get
+   * split into [a, b.c, d.e.f]. if string is "val1.val2", then it will not get splitted since the '.'
+   * is within pair of quotes. If the delimited string contains odd number of quotes, which mean the
+   * quotes are not closed, an exception will be thrown. The quote within the value will always be
+   * trimed.
+   */
+  @VisibleForTesting
+  static class SplitQuotesIterator extends AbstractIterator<String> {
```
albertshau: nit: better to pull this out into a separate class instead of a nested class

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -128,4 +150,82 @@ public void close() throws IOException {
...
+          return split.toString();
+        }
+
+        // delimiter within quotes
+        split.append(cur);
+        index++;
+      }
+
+      if (isWithinQuotes) {
+        throw new RuntimeException("Found a line with an unenclosed quote. Ensure that all values are properly"
```
albertshau: should avoid throwing plain Exception or RuntimeException in most situations. An IllegalArgumentException would be a little more specific here (unit test should check for the specific exception type too)

albertshau (commented):
minor comments

albertshau (approved):
lgtm. Please squash commits and make sure the build runs successfully, then we can merge.

PR #1509, opened 2021-10-29
Add logic to set the most skewed stage as the first stage in a join definition.

fernst:
Added logic to reorder stages when creating a join definition. This is useful when pushing down joins into BigQuery to prevent RESOURCE_EXCEEDED errors.

See https://cdap.atlassian.net/browse/PLUGIN-918 

image

image

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.batch.joiner;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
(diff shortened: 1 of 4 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/joiner/Joiner.java
```diff
@@ -263,4 +281,24 @@ public void prepareRun(BatchJoinerContext context) {
...
     return operations;
   }
+
+  /**
+   * Reorders join stages so the supplied stage name is always first.
+   * @param stages list of input stages
+   * @param leftStage stage to move to the first position in the input list.
+   */
+  protected void reorderJoinStages(List<JoinStage> stages, String leftStage) {
+    stages.sort((js1, js2) -> {
```
DJSagarAhire, line 291: Sorting the list is an odd way to do this. Is there a reason we aren't just swapping the first stage with `leftStage` using something like `lastIndexOf(leftStage)` to find the index where the swap needs to happen?

DJSagarAhire (approved):
LGTM

PR #1512, opened 2021-11-11
Escaping regex sequences in schema column headers split.

sanjanasandeep:
**Issue:**
https://cdap.atlassian.net/browse/PLUGIN-972
https://b.corp.google.com/issues/205116841

**Context:**

java's String::split is for regex expressions. Column headers are incorrectly split on regex metacharacters, thereby throwing an error. Escaping regex sequences safely is the solution.

### DIFF
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
@@ -98,10 +98,13 @@ public class DataTypeDetectorUtils {
    */
   public static String[] setColumnNames(String rawLine, boolean skipHeader, String delimiter) {
     String[] columnNames;
+    final String quotedDelimiter = Pattern.quote(delimiter);
     if (skipHeader) {
-      columnNames = rawLine.split(delimiter);
+      // String.split uses regex. Here we safely escape regex sequences by using Pattern.quote
+      // Pattern.quote returns a literal pattern string
+      columnNames = rawLine.split(quotedDelimiter);
     } else {
-      columnNames = new String[rawLine.split(delimiter, -1).length];
+      columnNames = new String[rawLine.split(quotedDelimiter, -1).length];
       for (int j = 0; j < columnNames.length; j++) {
         columnNames[j] = String.format("%s_%s", "body", j);
       }
--- a/format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
+++ b/format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
@@ -44,6 +44,22 @@ public class DataTypeDetectorUtilsTest {
   }
 
   @Test
+  public void testExtractedColumnNamesRegexMetaCharacterDelimiter() {
+    String headerLine = "column_A|column_B|column_C";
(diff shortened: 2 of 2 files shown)


format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -66,7 +66,8 @@ public static void detectDataTypeOfRowValues(Map<String, Schema> override,
    * @return A list of detected schema fields per each column of the dataset.
    */
   public static List<Schema.Field> detectDataTypeOfEachDatasetColumn(Map<String, Schema> override, String[] columnNames,
-    DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper) {
+                                                                     DataTypeDetectorStatusKeeper
```
fernst: Move this to a new line for better legibility

```
public static List<Schema.Field> detectDataTypeOfEachDatasetColumn(
    DataTypeDetectorStatusKeeper dataTypeDetectorStatusKeeper) {
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -99,9 +100,11 @@ public static void detectDataTypeOfRowValues(Map<String, Schema> override,
   public static String[] setColumnNames(String rawLine, boolean skipHeader, String delimiter) {
     String[] columnNames;
     if (skipHeader) {
-      columnNames = rawLine.split(delimiter);
+      // String.split uses regex. Here we safely escape regex sequences by using Pattern.quote
+      // Pattern.quote returns a literal pattern string
+      columnNames = rawLine.split(Pattern.quote(delimiter));
```
fernst: nitpick: If the delimiter will always be used as `Pattern.quote(delimiter)`, consider storing this in a separate variable.

```
String quotedDelimiter = Pattern.quote(delimiter):
...
rawLine.split(quotedDelimiter);
```

fernst (changes requested):

format-delimited/src/test/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtilsTest.java
```diff
@@ -43,6 +43,22 @@ public void testExtractedColumnNames() {
     assertArrayEquals(expectedColumnNames, actualColumnNames);
   }
 
+  @Test
```
fernst, line 46: 👍 to adding tests for this

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
sanjanasandeep: Yup makes sense.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
sanjanasandeep: Yeah, this was an ide issue that I missed. Fixed it now

sanjanasandeep (commented):
Addressed comments from previous revision

fernst (approved):
LGTM

PR #1522, opened 2022-01-26
fix the bug when precision is greater than 19

seanzhougoogle:


hydrator-common/src/main/java/io/cdap/plugin/common/db/DBUtils.java
```diff
@@ -330,15 +330,10 @@ public static Schema getSchema(String typeName, int sqlType, int precision, int
...
-          if (precision >= 19) {
-            throw new SQLException(
-              new UnsupportedTypeException(String.format("Unsupported SQL Type %s. Unable to store column %s as long." +
-                                                           " Current precision is %d, it has to be less than 19.",
-                                                         typeName, columnName, precision)));
-          }
+
           // if there are no digits after the point, use integer types
-          type = scale != 0 ? Schema.Type.DOUBLE :
+          ////SQL DECIMAL can be 5 - 17 bytes, not all can be held in a long
```
yaojiefeng: nit, remove extra "//"

yaojiefeng (approved):

hydrator-common/src/main/java/io/cdap/plugin/common/db/DBUtils.java
seanzhougoogle: done :)

PR #1531, opened 2022-02-07
Added the options for sampling properties in file connectors

sechegaray:


format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -58,18 +72,55 @@
...
+    this.sampleProperties = new HashSet<>();
+  }
+
+  /**
+   * Adds available sample fields for each different entityType.
+   *
+   * @param entityType
+   * @param fileSourceClass FileSourceProperties class containing the configuration and description of the properties.
+   */
+  protected void addSampleFields(String entityType,  Class<? extends FileSourceProperties> fileSourceClass) {
```
yaojiefeng: Let's call this `initSampleFields`

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -58,18 +72,55 @@
...
+    List<Field> sourceFields = new ArrayList<>();
+    try {
+      for (String sourceFieldName: SAMPLE_FIELD_NAMES) {
+        sourceFields.add(fileSourceClass.getDeclaredField(sourceFieldName));
+      }
+    } catch (NoSuchFieldException e) {
+      // Skip sample option as it's not implemented by the source
```
yaojiefeng: Do we know what is the behavior for UI if some field is missing? What will the widget show?

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -73,6 +74,7 @@
 
   public FileConnector(FileConnectorConfig config) {
     super(config);
+    addSampleFields(FILE_TYPE, FileSourceConfig.class);
```
yaojiefeng: You will need to add/edit the widget file corresponding to this class, otherwise UI won't know how to show these properties

core-plugins/src/main/java/io/cdap/plugin/batch/connector/FileConnector.java
```diff
@@ -113,17 +116,22 @@ public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest requ
...
     super.setConnectorSpec(request, builder);
-    // not use ImmutableMap here in case any property is null
-    Map<String, String> properties = new HashMap<>();
+    Map<String, String> properties = new HashMap<>(getSourceSampleProperties(request));
     properties.put("path", request.getPath());
-    properties.put("format", FileTypeDetector.detectFileFormat(
-      FileTypeDetector.detectFileType(request.getPath())).name().toLowerCase());
+
+    // Only detect format if it has not been set by sample properties
+    if (!properties.containsKey(FileSourceConfig.NAME_FORMAT)) {
```
yaojiefeng, line 130: There is unit test for this method, I think it will break after these changes, can you update the test and also add unit test for your changes?

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -163,24 +214,75 @@ protected String getFullPath(String path) {
...
+    entity.addProperty(SAMPLE_FILE_ENCODING_KEY, BrowseEntityPropertyValue.builder(
+      SAMPLE_FILE_ENCODING_DEFAULT, BrowseEntityPropertyValue.PropertyType.SAMPLE_DEFAULT).build());
   }
 
-  private ValidatingInputFormat getValidatingInputFormat(ConnectorContext context, String path) throws IOException {
+  /**
+   *
+   * @return Set of available sample properties
+   */
+  protected Set<BrowseEntityTypeInfo> getSampleProperties() {
```
yaojiefeng, line 244: You can just make `sampleProperties` protected field so all subclasses can use it

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -163,24 +214,75 @@ protected String getFullPath(String path) {
...
+      builder.add("format", format.name());
+    }
+
     if (format.equals(FileFormat.TEXT)) {
       builder.add("schema", DEFAULT_TEXT_SCHEMA);
     }
     builder.addAll(config.getProperties().getProperties());
     builder.addAll(getFileSystemProperties(path));
-
+    builder.add("fileSystemProperties", GSON.toJson(getFileSystemProperties(path)));
```
yaojiefeng, line 291: Why do we need to add this, isn't file system properties already set in Line 284?

yaojiefeng (commented):
Some comments

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
sechegaray: done

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
sechegaray: The UI will only show the fields given by getSampleProperties(). If the property is missing then the UI will not show it.

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
sechegaray: Yes, but I wanted to provide a getter in case we have to change/add behavior in the future.

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
sechegaray: added comment

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -163,24 +214,77 @@ protected String getFullPath(String path) {
...
+  /**
+   * Receives a request and returns a map with only the valid sample properties.
+   *
+   * @param request Request containing the sample properties.
+   * @return a Map with chosen sample properties
+   */
+  protected Map<String, String> getSourceSampleProperties(ConnectorSpecRequest request) {
+    Map<String, String> requestProperties = request.getProperties();
+    Map<String, String> sourceProperties = new HashMap<>();
+    if (requestProperties != null) {
```
yaojiefeng: You don't need to perform a null check here, the property should at least be empty, did you see sometimes this is null?

format-common/src/main/java/io/cdap/plugin/format/connector/AbstractFileConnector.java
```diff
@@ -163,24 +214,77 @@ protected String getFullPath(String path) {
...
+    return sampleProperties;
+  }
+
+  /**
+   * Receives a request and returns a map with only the valid sample properties.
+   *
+   * @param request Request containing the sample properties.
+   * @return a Map with chosen sample properties
+   */
+  protected Map<String, String> getSourceSampleProperties(ConnectorSpecRequest request) {
```
yaojiefeng: So I believe this method is used to retrieve the spec for a plugin, then we shouldn't call it this name since the property returned will also be used by the sink. Just call this `getAdditionalSpecProperties`

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -93,6 +93,7 @@ public NullWritable getCurrentKey() {
             int numDataFields = 0;
             splitsIterator = getSplitsIterator(enableQuotesValue, delimitedString, delimiter);
             while (splitsIterator.hasNext()) {
+              splitsIterator.next();
```
yaojiefeng, line 96: This is interesting, if we don't have this change, the logic will be here forever. Does this mean this functionality is never used by any customer or we don't have any unit test for it?

yaojiefeng (approved):
Minor nit and a question, LGTM

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
sechegaray: I think we don't have any unit tests. As I've hit this issue while testing.

yaojiefeng:
You will need a widget file for the file connector for UI to show the properties

yaojiefeng:
The changes LGTM, but it seems like the build has unit test failures, can you fix it?

yaojiefeng (approved):

PR #1532, opened 2022-02-07
Enable connections for db source/sink

yaojiefeng:


database-plugins/src/main/java/io/cdap/plugin/DBManager.java
```diff
@@ -53,17 +53,20 @@ public DBManager(ConnectionConfig config) {
       config.connectionArguments), config.jdbcPluginType);
   }
 
-  @Nullable
-  public Class<? extends Driver> validateJDBCPluginPipeline(PipelineConfigurer pipelineConfigurer,
-                                                            String jdbcPluginId, FailureCollector collector) {
+  public void validateCredentials(FailureCollector collector) {
     if (!config.containsMacro(DBConfig.USER) && !config.containsMacro(DBConfig.PASSWORD)
-      && config.getUser() == null && config.getPassword() != null) {
+          && config.getUser() == null && config.getPassword() != null) {
```
avirmani, line 58: nit: space?

avirmani (approved):

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
```diff
@@ -144,8 +144,4 @@ public String getReferenceName() {
   public String getColumnNameCase() {
     return columnNameCase;
   }
-
-  public boolean getUseConnection() {
-    return useConnection == null ? false : useConnection;
-  }
```
sanjanasandeep, line 150: why is this removed?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
```diff
@@ -71,6 +71,7 @@
   @Name(JDBC_PLUGIN_NAME)
   @Description("Name of the JDBC plugin to use. This is the value of the 'name' key defined in the JSON file " +
     "for the JDBC plugin.")
```
sanjanasandeep, line 73: Should we have same text as in the doc file? (i.e; add (Macro-enabled))

database-plugins/src/main/java/io/cdap/plugin/db/common/DBBaseConfig.java
yaojiefeng: Because it was a placeholder to default connection to false, now since we reenable it, we don't need the default value any more

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
```diff
@@ -106,13 +106,9 @@ protected DBConnectorPath getDBConnectorPath(String path) throws IOException {
...
-    Map<String, String> rawProperties = config.getRawProperties().getProperties();
     Map<String, String> properties = new HashMap<>();
-    properties.put(DBConnectorConfig.CONNECTION_STRING, rawProperties.get(DBConnectorConfig.CONNECTION_STRING));
-    properties.put(DBConnectorConfig.JDBC_PLUGIN_NAME, rawProperties.get(DBConnectorConfig.JDBC_PLUGIN_NAME));
-    properties.put(DBConnectorConfig.USER, rawProperties.get(DBConnectorConfig.USER));
-    properties.put(DBConnectorConfig.PASSWORD, rawProperties.get(DBConnectorConfig.PASSWORD));
-    properties.put(DBConnectorConfig.CONNECTION_ARGUMENTS, rawProperties.get(DBConnectorConfig.CONNECTION_ARGUMENTS));
+    properties.put(DBBaseConfig.NAME_USE_CONNECTION, "true");
+    properties.put(DBBaseConfig.NAME_CONNECTION, request.getConnectionWithMacro());
     properties.put(DBBaseConfig.JDBC_PLUGIN_TYPE, DBUtils.PLUGIN_TYPE_JDBC);
```
sanjanasandeep, line 112: should there be unit tests for this pr?

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnectorConfig.java
yaojiefeng: Already added to source/sink doc, for connector, we don't add those since using macro is rare use case there, except for secure macro

sanjanasandeep (approved):
LGTM

seanzhougoogle (approved):

database-plugins/src/main/java/io/cdap/plugin/db/connector/DBConnector.java
yaojiefeng: We have ITN for these connectors, but it was not updated yet(it was still based on connection is true), will have a follow up pr to update those

database-plugins/src/main/java/io/cdap/plugin/DBManager.java
yaojiefeng: The original indentation was off, since this is separated for a if condition, the indentation should based on the start of the condition

PR #1533, opened 2022-02-11
Generating aggregation definition for Deduplicate plugin

sanjanasandeep:
**Testing**

unit tests added.

Need to test this end to end.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
@@ -25,17 +25,26 @@ import io.cdap.cdap.etl.api.Emitter;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.StageConfigurer;
+import io.cdap.cdap.etl.api.aggregation.DeduplicateAggregationDefinition;
 import io.cdap.cdap.etl.api.batch.BatchAggregator;
 import io.cdap.cdap.etl.api.batch.BatchAggregatorContext;
 import io.cdap.cdap.etl.api.batch.BatchRuntimeContext;
+import io.cdap.cdap.etl.api.relational.Engine;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.InvalidRelation;
+import io.cdap.cdap.etl.api.relational.LinearRelationalTransform;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
+import io.cdap.cdap.etl.api.relational.StringExpressionFactoryType;
 import io.cdap.plugin.batch.aggregator.function.SelectionFunction;
 import io.cdap.plugin.common.SchemaValidator;
 import io.cdap.plugin.common.TransformLineageRecorderUtils;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import javax.annotation.Nullable;
+
 
 /**
  * Deduplicate aggregator.
@@ -44,7 +53,7 @@ import javax.annotation.Nullable;
 @Name("Deduplicate")
 @Description("Deduplicates input records, optionally restricted to one or more fields. Takes an optional " +
   "filter function to choose one or more records based on a specific field and a selection function.")
-public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord> {
+public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord> implements LinearRelationalTransform {
   private final DedupConfig dedupConfig;
   private List<String> uniqueFields;
   private DedupConfig.DedupFunctionInfo filterFunction;
@@ -190,5 +199,24 @@ public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord>
           .withConfigProperty("filterOperation");
       }
     }
+  }
+
+  @Override
+  public boolean canUseEngine(Engine engine) {
+    Optional<ExpressionFactory<String>> expressionFactory = engine.
+            getExpressionFactory(StringExpressionFactoryType.SQL);
+    return expressionFactory.isPresent();
+  }
+
+  @Override
+  public Relation transform(RelationalTranformContext relationalTranformContext, Relation relation) {
+    DeduplicateAggregationDefinition deduplicateAggregationDefinition = DedupAggregatorUtils
+            .generateAggregationDefinition(relationalTranformContext, relation, filterFunction, uniqueFields);
+
+    if (deduplicateAggregationDefinition == null) {
+      return new InvalidRelation("filterOperation needs to be either min or max");
+    }
+
+    return relation.deduplicate(deduplicateAggregationDefinition);
   }
 }
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.batch.aggregator;
+
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.aggregation.DeduplicateAggregationDefinition;
+import io.cdap.cdap.etl.api.relational.CoreExpressionCapabilities;
+import io.cdap.cdap.etl.api.relational.Expression;
+import io.cdap.cdap.etl.api.relational.ExpressionFactory;
+import io.cdap.cdap.etl.api.relational.Relation;
+import io.cdap.cdap.etl.api.relational.RelationalTranformContext;
+import io.cdap.cdap.etl.api.relational.StringExpressionFactoryType;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Utility class for DedupAggregator.
+ */
+public class DedupAggregatorUtils {
+
+    public static DeduplicateAggregationDefinition generateAggregationDefinition
+            (RelationalTranformContext relationalTranformContext,
+             Relation relation,
+             DedupConfig.DedupFunctionInfo filterFunction,
+             List<String> uniqueFields) {
+        Optional<ExpressionFactory<String>> expressionFactory = relationalTranformContext.getEngine().
+                getExpressionFactory(StringExpressionFactoryType.SQL);
+        DeduplicateAggregationDefinition.FilterFunction aggFilterFunction;
+        if (filterFunction.getFunction() == DedupConfig.Function.MAX) {
+            aggFilterFunction = DeduplicateAggregationDefinition.FilterFunction.MAX;
+        } else if (filterFunction.getFunction() == DedupConfig.Function.MIN) {
+            aggFilterFunction = DeduplicateAggregationDefinition.FilterFunction.MIN;
+        } else {
+            return null;
+        }
+
+        Map<String, Expression> selectExpressions = new HashMap<>();
+        List<Expression> dedupExpressions = new ArrayList<>();
+        ExpressionFactory<String> stringExpressionFactory = expressionFactory.get();
+
+        Expression filterExpression = getColumnName(relation, filterFunction.getField(), stringExpressionFactory);
+
+        for (Schema.Field field : relationalTranformContext.getOutputSchema().getFields()) {
+            selectExpressions.put(field.getName(), getColumnName(relation, field.getName(), stringExpressionFactory));
+        }
+
+
+        for (String uniqueField : uniqueFields) {
+            dedupExpressions.add(getColumnName(relation, uniqueField, stringExpressionFactory));
+        }
+
+        return DeduplicateAggregationDefinition.builder()
+                .select(selectExpressions)
+                .dedupOn(dedupExpressions)
+                .filterDuplicatesBy(filterExpression, aggFilterFunction)
+                .build();
+    }
+
+    static Expression getColumnName(Relation relation,
+                                    String name, ExpressionFactory<String> stringExpressionFactory) {
+        if (stringExpressionFactory.getCapabilities()
+                .contains(CoreExpressionCapabilities.CAN_GET_QUALIFIED_COLUMN_NAME)) {
+            return stringExpressionFactory.getQualifiedColumnName(relation, name);
+        } else {
+            return stringExpressionFactory.compile(name);
+        }
+    }
+}
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtilsTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtilsTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package io.cdap.plugin.batch.aggregator;
+
+import io.cdap.cdap.api.data.schema.Schema;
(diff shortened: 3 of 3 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -28,27 +28,32 @@
...
 @Name("Deduplicate")
 @Description("Deduplicates input records, optionally restricted to one or more fields. Takes an optional " +
   "filter function to choose one or more records based on a specific field and a selection function.")
-public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord> {
+public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord> implements LinearRelationalTransform {
   private final DedupConfig dedupConfig;
   private List<String> uniqueFields;
   private DedupConfig.DedupFunctionInfo filterFunction;
   private SelectionFunction selectionFunction;
+  private Schema schema;
```
tivv: final

tivv: And move to the top along with other final fields

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -191,4 +197,15 @@ private void validateSchema(Schema inputSchema, List<String> uniqueFields,
       }
     }
   }
+
+    @Override
+    public boolean canUseEngine(Engine engine) {
+        return LinearRelationalTransform.super.canUseEngine(engine);
```
tivv: Why do you need this override?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Utility class for DedupAggregator.
+ */
+public class DedupAggregatorUtils {
+
+    public static Relation dedupTransform(RelationalTranformContext relationalTranformContext, Relation relation,
```
tivv: Is this reused? What is the reason to introduce the utils class? Also please add a detailed javadoc as this method is not obvious

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
tivv: Also generally if you need specific expression factory it's good to validate it here. This way transform won't be called if not needed

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+                                          Schema schema, DedupConfig.DedupFunctionInfo filterFunction,
+                                          List<String> uniqueFields) {
+        Optional<ExpressionFactory<String>> expressionFactory = relationalTranformContext.getEngine().
+                getExpressionFactory(StringExpressionFactoryType.SQL);
+        if (!expressionFactory.isPresent()) {
+            return new InvalidRelation("expression factory is not present");
+        }
+        DeduplicateAggregationDefinition.FilterFunction aggFilterFunction = null;
+        if (filterFunction.getFunction().equals(DedupConfig.Function.MAX)) {
```
tivv: You can use `==` for enums

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+            aggFilterFunction = DeduplicateAggregationDefinition.FilterFunction.MIN;
+        } else {
+            return new InvalidRelation("filterOperation needs to either min or max");
+        }
+
+        Map<String, Expression> selectExpressions = new HashMap<>();
+        List<Expression> dedupExpressions = new ArrayList<>();
+        ExpressionFactory<String> stringExpressionFactory = expressionFactory.get();
+
+        Expression filterExpression = stringExpressionFactory.compile(filterFunction.getField());
```
tivv: Please ensure you get properly escaped name by calling `io.cdap.cdap.etl.api.relational.ExpressionFactory#getQualifiedColumnName`

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+        for (Schema.Field field : schema.getFields()) {
+            selectExpressions.put(field.getName(), stringExpressionFactory.compile(field.getName()));
```
tivv: Please ensure you get properly escaped name by calling io.cdap.cdap.etl.api.relational.ExpressionFactory#getQualifiedColumnName

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        for (String uniqueField : uniqueFields) {
+            dedupExpressions.add(stringExpressionFactory.compile(uniqueField));
```
tivv: Please ensure you get properly escaped name by calling io.cdap.cdap.etl.api.relational.ExpressionFactory#getQualifiedColumnName

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtilsTest.java
```diff
@@ -0,0 +1,174 @@
...
+        Mockito.doReturn(engine).when(relationalTranformContext).getEngine();
+        uniqueFields = new ArrayList<>();
+        uniqueFields.add("uniqueField");
+        schema = Schema.recordOf(
+                "schema",
+                Schema.Field.of("uniqueField", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                Schema.Field.of("selectField", Schema.nullableOf(Schema.of(Schema.Type.STRING))),
+                Schema.Field.of("filterField", Schema.nullableOf(Schema.of(Schema.Type.STRING)))
+        );
+        relationOriginal = new Relation() {
```
tivv: It seems you can just use mocks for this. It would save you quite a bit of overrides

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtilsTest.java
```diff
@@ -0,0 +1,174 @@
...
+            public Relation select(Map<String, Expression> map) {
+                return null;
+            }
+
+            @Override
+            public Relation filter(Expression expression) {
+                return null;
+            }
+        };
+        expressionFactoryOriginal = new ExpressionFactory<String>() {
```
tivv: Same here. Using mocks will be more consice

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: I suggested moving this into a separate class for easier testing (without having to set up mocks).

I'm fine with using Utility classes and static functions as long as we document them.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: If we do this, we need to check if the engine supports this capability.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: If we do this, we need to check if the engine supports this capability.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: If we do this, we need to check if the engine supports this capability.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
tivv: Sure. Or leave out any engine that does not. It should return invalid expression in this case, so no additional checks needed. But given it's not user input, it's better to call. Otherwise tomorrow we'll need to update all the plugins to properly quote fields

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
tivv: Makes sense

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
sanjanasandeep: what's the reason this one must specifically be final ?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
sanjanasandeep: okay, so I can move the expression factory is present or not check to here

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
tivv: That's overall rule. Anything that can be final should be marked so. It help when looking at the code to see which values are mutable (and may have concurrency issues) and which are not

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
tivv: Correct. Overall you are guarantied to only get engines in transform you said you can use

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
sanjanasandeep: so just to confirm, we need this:

```
selectExpressions.put(field.getName(), stringExpressionFactory.compile(stringExpressionFactory.getQualifiedColumnName(relation, field.getName())))
```

tivv: Correct. With the recent discussion though, no "compile" would be needed as `getQualifiedColumnName` would return an `Expression`

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
sanjanasandeep: so I think we get `Schema` from `BatchAggregatorContext` - so it can't be final in that case.  Unless we can initialize it in the constructor

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
sanjanasandeep: yes, will make that change once that PR is merged.

fernst: I would create a function to handle this by checking capabilities exposed by the engine.

```
private String getColumnName(Relation rel,  String name) {
  if (expressionFactory.getCapabilities().contains(CoreExpressionCapabilities.CAN_GET_QUALIFIED_COLUMN_NAME)) {
    return expressionFactory.getQualifiedColumnName(relation, name)
  } else {
    return name;
}
```

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!expressionFactory.isPresent()) {
+            return new InvalidRelation("expression factory is not present");
+        }
+        DeduplicateAggregationDefinition.FilterFunction aggFilterFunction = null;
+        if (filterFunction.getFunction().equals(DedupConfig.Function.MAX)) {
+            aggFilterFunction = DeduplicateAggregationDefinition.FilterFunction.MAX;
+        } else if (filterFunction.getFunction().equals(DedupConfig.Function.MIN)) {
+            aggFilterFunction = DeduplicateAggregationDefinition.FilterFunction.MIN;
+        } else {
+            return new InvalidRelation("filterOperation needs to either min or max");
```
DJSagarAhire: nit: Grammar. filterOperation needs to **be** either min or max.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
DJSagarAhire: Since this is the output schema the variable name should indicate it's the output schema to avoid confusion with input schema.

sanjanasandeep: @DJSagarAhire we now get schema from `RelationalTranformContext`, so getting rid of this

fernst: You can remove this schema as you won't have this available in the execution. You can just use the Schema supplied in the `relationalTranformContext`

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+public class DedupAggregatorUtils {
+
+    public static DeduplicateAggregationDefinition generateAggregationDefinition
+            (RelationalTranformContext relationalTranformContext,
+             Relation relation,
+             DedupConfig.DedupFunctionInfo filterFunction,
+             List<String> uniqueFields) {
+        Optional<ExpressionFactory<String>> expressionFactory = relationalTranformContext.getEngine().
+                getExpressionFactory(StringExpressionFactoryType.SQL);
+        DeduplicateAggregationDefinition.FilterFunction aggFilterFunction = null;
```
tivv: nit: you don't actually need `= null` as you do assignment in all branches

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -0,0 +1,89 @@
...
+                .dedupOn(dedupExpressions)
+                .filterDuplicatesBy(filterExpression, aggFilterFunction)
+                .build();
+    }
+
+    static String getColumnName(Relation relation,
+                                String name, ExpressionFactory<String> stringExpressionFactory) {
+        if (stringExpressionFactory.getCapabilities()
+                .contains(CoreExpressionCapabilities.CAN_GET_QUALIFIED_COLUMN_NAME)) {
+            return stringExpressionFactory.getQualifiedColumnName(relation, name).extract();
```
tivv: How about making this return an `Expression`? Currently you will do `extract` and then `compile` most of the times. It's actually not needed.  You can still do `compile` if `CAN_GET_QUALIFIED_COLUMN_NAME` is not set

sanjanasandeep: okay so we can remove the check and have just this:

```
static ExtractableExpression<String> getColumnName(Relation relation,
                                                       String name, ExpressionFactory<String> stringExpressionFactory) {
            return stringExpressionFactory.getQualifiedColumnName(relation, name);
}
```

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
sanjanasandeep: yup right

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
tivv: What I mean is 
```
 if (stringExpressionFactory.getCapabilities()
                .contains(CoreExpressionCapabilities.CAN_GET_QUALIFIED_COLUMN_NAME)) {
            return stringExpressionFactory.getQualifiedColumnName(relation, name);
} else {
            return stringExpressionFactory.compile(name);
}
```

sanjanasandeep: so I don't think I understand why we need to use `getQualifiedColumnName` if we can just `compile` and that's valid too.

fernst: Because `getQualifiedColumnName` allows us to generate a column name that we are 100% certain is "safe" to use with the SQL engine.

In BigQuery's case, identifiers can be wrapped in backticks (`) as a way to ensure that the column name is safe to use.

So, for example:

```
select `thisísaçolumnName` from ...
```

Would be valid SQL in BQ because the column name is qualified.

tivv: Yep. For CDAP more important cases where column name is a reserved word. We don't allow tons of things, bu t we do allow column names like `select` or `from`. Those fail without quoting, see https://cdap.atlassian.net/browse/PLUGIN-1003

sanjanasandeep: so in the else case (when we don't use `getQualifiedColumnName` and simply `compile`) shouldn't we throw an exception (since we are not 100% safe here)?

fernst: It's not necessary to go that far.

sanjanasandeep: I see. Revised this in the new rev

tivv (approved):

PR #1534, opened 2022-02-15
Add relational transform to GroupByAggregator

DJSagarAhire:
This PR adds an implementation of `LinearRelationalTransform` to `GroupByAggregator`. This also adds a `GroupByAggregationDefinition` in the class.

This PR makes an assumption that conditional aggregations are not supported, but checks for further supported functions are performed by the engine.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.batch.aggregator;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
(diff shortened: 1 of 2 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -393,4 +411,64 @@ private Schema getGroupKeySchema(Schema inputSchema) {
...
+    Optional<ExpressionFactory<String>> expressionFactory = relationalTranformContext
+      .getEngine().getExpressionFactory(StringExpressionFactoryType.SQL);
+    if (!expressionFactory.isPresent()) {
+      return new InvalidRelation("Cannot find an Expression Factory");
+    }
+
+    GroupByAggregationDefinition aggregationDefinition = getAggregationDefinition(expressionFactory.get(), relation);
+
+    if (aggregationDefinition == null) {
+      return new InvalidRelation("Cannot create a relation as aggregation definition is invalid");
```
fernst: I would say "Aggregation definition is not supported." This can only be null for conditional aggregations.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -393,4 +411,64 @@ private Schema getGroupKeySchema(Schema inputSchema) {
...
+      String columnName = getColumnName(expressionFactory, relation, field);
+      Expression groupByExpression = expressionFactory.compile(columnName);
+      groupByExpressions.add(groupByExpression);
+      selectExpressions.put(field, groupByExpression);
+    }
+
+    for (GroupByConfig.FunctionInfo aggregate: functionInfos) {
+      String alias = aggregate.getName();
+      String columnName = getColumnName(expressionFactory, relation, aggregate.getField());
+      GroupByConfig.Function function = aggregate.getFunction();
```
fernst, line 467: I remember not all functions were supported in the design doc. Can you make this a Set (of supported functions) so we can immediately determine if this is supported or not in BQ?)

DJSagarAhire: I've called this out in the PR description - should the decision on whether the function is supported or not be taken in the engine rather than in the aggregate definition since the list of supported functions can be different depending on engine?

fernst: Yes, each plugin should "decide" what they support and not.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -98,12 +110,18 @@
   private List<String> groupByFields;
   private List<GroupByConfig.FunctionInfo> functionInfos;
   private Schema outputSchema;
+  private GroupByAggregationDefinition aggregationDefinition;
```
sanjanasandeep, line 132: nit: specifying that it is defined for testing would help

fernst (approved):
LGTM

PR #1537, opened 2022-02-19
initialize dedupconfig values in transform. function

sanjanasandeep:


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -210,6 +210,8 @@ public boolean canUseEngine(Engine engine) {
 
   @Override
   public Relation transform(RelationalTranformContext relationalTranformContext, Relation relation) {
+    uniqueFields = dedupConfig.getUniqueFields();
```
tivv: Why do we need this fields at all? Can we always take from `dedupConfig`

sanjanasandeep: do you mean pass `dedupConfig` directly as a parameter?

tivv: `dedupConfig` is a field of this class. You can just use it in any non-static method

sanjanasandeep: So I mean : ` DedupAggregatorUtils
            .generateAggregationDefinition(relationalTranformContext, relation, dedupConfig)`. Is this what you are talking about as well?

tivv: No, I am talking about `DedupAggregator` only. You have `dedupConfig` as a final field and this additional two fields that are part of `dedupConfig`. It's not clear why do you need the additional fields

sanjanasandeep: I am sorry I am a bit confused, do you mean 
1. why these fields `uniqueFields` and `filterFunction` are being initialized here? OR
2. why do we need to extract them at all, when we can use `dedupConfig` directly in the helper function?

Answer for 1 is: 
These fields are initialized in the `initialize()` function here : https://github.com/cdapio/hydrator-plugins/blob/develop/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java#L108 and this function is not being called (maybe because we are now implementing `LinearRelationalTransform`, that's why I initialize them here in `transform`.

For 2:

My suggestion is the previous comment at L215

fernst: I think Vitalii's point is that initialization for these variables is not really necessary. You can just pass these into the `generateAggregationDefinition` without storing them in the global variable.

```
    DeduplicateAggregationDefinition deduplicateAggregationDefinition = DedupAggregatorUtils
            .generateAggregationDefinition(relationalTranformContext, relation, dedupConfig.getFilterFunction(), dedupConfig.getUniqueFields());
```

sanjanasandeep: okay, got it. Making the change now.

fernst (approved):
LGTM

tivv (approved):

PR #1545, opened 2022-03-14
Add test case for BQ Aggregation pushdown for group by

DJSagarAhire:
This PR adds a test case for a `GroupByAggregationDefinition` with mixed validity, i.e. a definition in which some expressions are valid for pushdown and some aren't. The expected behaviour is that an invalid relation is generated as some of the expressions are not supported for pushdown.

### DIFF
--- a/core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/GroupByRelationalTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/GroupByRelationalTest.java
@@ -77,6 +77,20 @@ public class GroupByRelationalTest {
   }
 
   @Test
+  public void testMixedValidityGroupBy() {
+    GroupByConfig config = new GroupByConfig("profession",
+                                             "numEmployees: countif(*): condition(salary>100000)," +
+                                               "stddevSal: stddev(salary)," +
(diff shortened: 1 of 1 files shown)


fernst (approved):
LGTM

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/GroupByRelationalTest.java
```diff
@@ -76,6 +76,20 @@ public void testValidGroupBy() throws Exception {
     Assert.assertEquals(3, selectExpressions.size());
   }
 
+  @Test
+  public void testMixedValidityGroupBy() {
+    GroupByConfig config = new GroupByConfig("profession",
+                                             "numEmployees: countif(*): condition(salary>100000)," +
+                                               "stddevSal: stddev(salary)," +
+                                               "maxSal: max(salary)");
```
sanjanasandeep, line 84: can we specify which expressions are valid for pushdown?

DJSagarAhire: They are specified in `functionSqlMap` in `GroupByAggregator`.

PR #1547, opened 2022-03-14
Added support for Count Nulls aggregation in BQ Pushdown

fernst:


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -119,6 +119,7 @@
     put(GroupByConfig.Function.COLLECTLIST, "ARRAY_AGG(%s IGNORE NULLS)");
     put(GroupByConfig.Function.COLLECTSET, "ARRAY_AGG(DISTINCT %s IGNORE NULLS)");
     put(GroupByConfig.Function.COUNTDISTINCT, "COUNT(DISTINCT %s) + IF(COUNTIF(%<s IS NULL) > 0, 1, 0)");
+    put(GroupByConfig.Function.COUNTNULLS, "COUNTIF(%s IS NULL)");
```
sanjanasandeep, line 122: why do we need this specifically?

fernst: This is a relatively simple aggregation to add.

sanjanasandeep (approved):

PR #1551, opened 2022-03-16
Validation exception for min/max filter function on a non numeric filter field

sanjanasandeep:
Issue : https://cdap.atlassian.net/browse/CDAP-18919

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
@@ -484,10 +484,17 @@ implements LinearRelationalTransform {
   }
 
   private String getColumnName(ExpressionFactory<String> expressionFactory, Relation relation, String name) {
-    if (expressionFactory.getCapabilities().contains(CoreExpressionCapabilities.CAN_GET_QUALIFIED_COLUMN_NAME)) {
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -198,6 +198,17 @@ private void validateSchema(Schema inputSchema, List<String> uniqueFields,
                              null)
           .withConfigProperty("filterOperation");
       }
+
+      Schema.Type schemaType = field.getSchema().getUnionSchema(0).getType();
+      if ((function.getFunction().equals(DedupConfig.Function.MAX) ||
+              function.getFunction().equals(DedupConfig.Function.MIN)) &&
+              (Schema.Type.STRING.equals(schemaType) || Schema.Type.BOOLEAN.equals(schemaType)
+                      || Schema.Type.BYTES.equals(schemaType))) {
```
tivv: Please add tests, including decimal (I think it's represented as bytes in physical schema).
Also it's better to have white list here, not black list. We also have arrays and structs

sanjanasandeep: yes, agreed. Also tests are failing - need to fix that rn

PR #1552, opened 2022-03-17
Add support for count(*) operatin in BQ statements

fernst:


tivv (approved):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -484,10 +484,17 @@ private GroupByAggregationDefinition getAggregationDefinition(ExpressionFactory<
   }
 
   private String getColumnName(ExpressionFactory<String> expressionFactory, Relation relation, String name) {
+    // If the column name is *, return as such.
+    if ("*".equals(name)) {
+      return name;
+    }
+
```
sanjanasandeep, line 491: do we need a unit test later?

sanjanasandeep (approved):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
fernst: @DJSagarAhire please add unit tests for the logic that builds the definition

PR #1562, opened 2022-03-30
Allow Deduplicate Plugin to support decimal and datetime min/max comparisons in Spark

sanjanasandeep:
**Issue**

https://cdap.atlassian.net/browse/CDAP-18941

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
@@ -61,6 +61,9 @@ public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord>
   private SelectionFunction selectionFunction;
   private static final EnumSet<Schema.Type> ALLOWED_SCHEMA_TYPES = EnumSet.of(Schema.Type.INT, Schema.Type.LONG,
           Schema.Type.FLOAT, Schema.Type.DOUBLE);
+  private static final EnumSet<Schema.LogicalType> ALLOWED_LOGICAL_SCHEMA_TYPES = EnumSet.of(Schema.LogicalType.DATE,
+          Schema.LogicalType.DATETIME, Schema.LogicalType.DECIMAL, Schema.LogicalType.TIME_MICROS,
+          Schema.LogicalType.TIME_MILLIS, Schema.LogicalType.TIMESTAMP_MILLIS, Schema.LogicalType.TIMESTAMP_MICROS);
 
   public DedupAggregator(DedupConfig dedupConfig) {
     super(dedupConfig.numPartitions);
@@ -201,11 +204,13 @@ public class DedupAggregator extends RecordReducibleAggregator<StructuredRecord>
                         null)
                 .withConfigProperty("filterOperation");
       }
-      Schema.Type fieldType = field.getSchema().isNullable() ?
-              field.getSchema().getNonNullable().getType() : field.getSchema().getType();
+      Schema fieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
+      Schema.Type fieldType = fieldSchema.getType();
+      Schema.LogicalType logicalFieldType = fieldSchema.getLogicalType();
       if ((function.getFunction() == DedupConfig.Function.MAX || function.getFunction() == DedupConfig.Function.MIN)
-        && !ALLOWED_SCHEMA_TYPES.contains(fieldType)) {
-        collector.addFailure(String.format("Unsupported filter operation %s(%s): Field '%s' is non numeric ",
+        && (!ALLOWED_SCHEMA_TYPES.contains(fieldType) && !ALLOWED_LOGICAL_SCHEMA_TYPES.contains(logicalFieldType))) {
+        collector.addFailure(String.format("Unsupported filter operation %s(%s): Field has a type that is not " +
+                                        "supported for deduplication operations",
                                 function.getFunction(), function.getField(), function.getField()),
                         null)
                 .withConfigProperty("filterOperation");
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/MaxSelection.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/MaxSelection.java
@@ -18,6 +18,12 @@ package io.cdap.plugin.batch.aggregator.function;
 
 import io.cdap.cdap.api.data.schema.Schema;
 
+import java.math.BigDecimal;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZonedDateTime;
+
 /**
  * A {@link SelectionFunction} that can be used to select the record with the max value of a given field.
  */
@@ -45,5 +51,30 @@ public class MaxSelection extends NumberSelection {
   @Override
   protected int compareDouble(double val1, double val2) {
     return Double.compare(val1, val2);
+  }
+
+  @Override
+  protected int compareBigDecimal(BigDecimal val1, BigDecimal val2) {
+    return val1.compareTo(val2);
+  }
+
+  @Override
+  protected int compareLocalDateTime(LocalDateTime val1, LocalDateTime val2) {
+    return val1.compareTo(val2);
+  }
+
+  @Override
+  protected int compareLocalDate(LocalDate date1, LocalDate date2) {
+    return date1.compareTo(date2);
+  }
+
+  @Override
+  protected int compareZonedDateTime(ZonedDateTime timestamp1, ZonedDateTime timestamp2) {
+    return timestamp1.compareTo(timestamp2);
+  }
+
+  @Override
+  protected int compareLocalTime(LocalTime time1, LocalTime time2) {
+    return time1.compareTo(time2);
   }
 }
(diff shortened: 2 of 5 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -205,7 +205,7 @@ private void validateSchema(Schema inputSchema, List<String> uniqueFields,
               field.getSchema().getNonNullable().getType() : field.getSchema().getType();
       if ((function.getFunction() == DedupConfig.Function.MAX || function.getFunction() == DedupConfig.Function.MIN)
         && !ALLOWED_SCHEMA_TYPES.contains(fieldType)) {
-        collector.addFailure(String.format("Unsupported filter operation %s(%s): Field '%s' is non numeric ",
+        collector.addFailure(String.format("Unsupported filter operation %s(%s): Field '%s' type is not allowed",
```
fernst: nit: change to "Field <bla> has a type that is not supported for deduplication operations

fernst (approved):
Code looks great, but can you add some tests?

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/MaxSelection.java
```diff
@@ -46,4 +48,14 @@ protected int compareFloat(float val1, float val2) {
...
   }
+
+  @Override
+  protected int compareBytes(ByteBuffer val1, ByteBuffer val2) {
+    return val1.compareTo(val2);
+  }
+
+  @Override
+  protected int compareString(String val1, String val2) {
+    return String.CASE_INSENSITIVE_ORDER.compare(val1, val2);
```
tivv: Why case insensitive? Is it confirmed by product?

tivv: What is the date/time format used? Are we sure string comparison order will follow date value order? Could you add a test

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
```diff
@@ -67,6 +73,10 @@ public StructuredRecord select(StructuredRecord record1, StructuredRecord record
         return compareFloat((Float) val1, (Float) val2) > 0 ? record1 : record2;
       case DOUBLE:
         return compareDouble((Double) val1, (Double) val2) > 0 ? record1 : record2;
+      case BYTES:
+        return compareBytes((ByteBuffer) val1, (ByteBuffer) val2) > 0 ? record1 : record2;
```
tivv: I think it can also be `byte[]`, see `io.cdap.cdap.api.data.format.StructuredRecord#getDecimal`

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -60,7 +61,7 @@
   private DedupConfig.DedupFunctionInfo filterFunction;
   private SelectionFunction selectionFunction;
   private static final EnumSet<Schema.Type> ALLOWED_SCHEMA_TYPES = EnumSet.of(Schema.Type.INT, Schema.Type.LONG,
-          Schema.Type.FLOAT, Schema.Type.DOUBLE);
+          Schema.Type.FLOAT, Schema.Type.DOUBLE, Schema.Type.BYTES, Schema.Type.STRING);
```
fernst: We should explicitly check if the fields are either DateTime or BigDecimal, not only checking the underlying type.

Also for BigDecimals, we need to verify if the underlying object is a BigDecimal instance, or transform it from bytes to BigDecimal using https://github.com/cdapio/cdap/blob/3a90d7c854bda94d894153f7fb6d99d51ff09ce0/cdap-api-common/src/main/java/io/cdap/cdap/api/common/Bytes.java#L801

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/MaxSelection.java
```diff
@@ -46,4 +48,14 @@ protected int compareFloat(float val1, float val2) {
   protected int compareDouble(double val1, double val2) {
     return Double.compare(val1, val2);
   }
+
+  @Override
```
fernst, line 56: we should convert to BigDecimal and compare using BigDecimal utils here.

fernst (changes requested):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
tivv: Agree. For BigDecimals, byets comparison would give wrong order. We need to compare specific values

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/MaxSelection.java
sanjanasandeep: uh shouldn't it ideally be case insensitive?

sanjanasandeep: yeah I'll add a test

tivv: Well, it depends on product. As an engineer I would not expect my max function to be case-insensitive by default

tivv: And if we do only dates here, I'd not make it case insensitive as it just adds more processing

sanjanasandeep: Ah okay I get it. Thanks!

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
sanjanasandeep: got it

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
sanjanasandeep: yup agree.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/NumberSelection.java
```diff
@@ -58,18 +79,47 @@ public StructuredRecord select(StructuredRecord record1, StructuredRecord record
...
+        case TIMESTAMP_MICROS:
+          return compareZonedDateTime(record1.getTimestamp(fieldName), record2.getTimestamp(fieldName)) > 0 ?
+                  record1 : record2;
+        case TIME_MILLIS:
+          return compareLocalTime(record1.getTime(fieldName), record2.getTime(fieldName)) > 0 ?
+                  record1 : record2;
+        case TIME_MICROS:
+          return compareLocalTime(record1.getTime(fieldName), record2.getTime(fieldName)) > 0 ?
+                  record1 : record2;
+        default:
```
sanjanasandeep, line 105: I noticed that these are additional numeric fields that we were not comparing

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/DedupAggregatorValidationTest.java
```diff
@@ -75,12 +91,47 @@ public void testValidateHappyPathNonNumericFilterField() {
...
+        record  = StructuredRecord.builder(inputSchema).setDecimal("price",
+                new BigDecimal("4000.91")).build();
+
+        functionInfo = new DedupConfig.DedupFunctionInfo("price", DedupConfig.Function.MAX);
+        Mockito.doReturn(uniqueFields).when(dedupConfig).getUniqueFields();
+        Mockito.doReturn(functionInfo).when(dedupConfig).getFilter();
+        DedupAggregator dedupAggregator = new DedupAggregator(dedupConfig);
+        dedupAggregator.initialize(context);
+        StructuredRecord actualRes = dedupAggregator.mergeValues(aggValue, record);
+        Assert.assertEquals(actualRes.getDecimal("price"), new BigDecimal("4000.91"));
```
sanjanasandeep, line 158: above tests for datetime and decimal specifically.

fernst: Can you add tests to all other types (timestamps, etc)?

sanjanasandeep: added additional tests:)

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregator.java
```diff
@@ -201,11 +202,13 @@ private void validateSchema(Schema inputSchema, List<String> uniqueFields,
...
       }
-      Schema.Type fieldType = field.getSchema().isNullable() ?
-              field.getSchema().getNonNullable().getType() : field.getSchema().getType();
+      Schema fieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
+      Schema.Type fieldType = fieldSchema.getType();
+      Schema.LogicalType logicalFieldType = fieldSchema.getLogicalType();
       if ((function.getFunction() == DedupConfig.Function.MAX || function.getFunction() == DedupConfig.Function.MIN)
-        && !ALLOWED_SCHEMA_TYPES.contains(fieldType)) {
-        collector.addFailure(String.format("Unsupported filter operation %s(%s): Field '%s' is non numeric ",
+        && (!ALLOWED_SCHEMA_TYPES.contains(fieldType) && logicalFieldType == null)) {
```
fernst: We should probably restrict logical types to the types that we want to support, checking that `ALLOWED_SCHEMA_TYPES` or `ALLOWED_LOGICAL_TYPES` is what we want it to be.

fernst (approved):
LGTM

PR #1563, opened 2022-04-01
Fixed issue with variance aggregator calculating incorrect results in some circumstances.

fernst:
We've updated the formula to calculate variance.

Note that we keep a running mean and squared mean. To ensure overflow and/or catastrophic cancellation is not a problem, we divide by the count of aggregated records on each step.

We've validated that the outputs for this aggregator match calculations performed by the BigQuery engine.


sumengwang (approved):
LGTM

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Variance.java
```diff
@@ -63,20 +60,26 @@ public void mergeValue(StructuredRecord record) {
...
     mean += delta / count;
-    double delta2 = value.doubleValue() - mean;
-    squareMean += delta * delta2;
+
+    // Calculate delta of the squared value vs the squared mean and adjust the squared mean
+    double deltaSquared = valueSquared - squaredMean;
+    squaredMean += deltaSquared / count;
+
     // if we divide by count, the result will be population variance,
     // count - 1, we will get sample variance
```
sumengwang: this comment seems to be unnecessary now

sanjanasandeep:
do we need unit tests for this?

sanjanasandeep (approved):
lgtm

sechegaray (approved):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/Variance.java
fernst: You're right. I've updated this comment.

PR #1564, opened 2022-04-01
[PLUGIN-1167] Schema detection file path clarify for .parquet and .avro file

JessieYee:
Jira: https://cdap.atlassian.net/browse/PLUGIN-1167

Solve the problem about detecting wrong file path when setting format as ".parquet" or ".avro".

Updated code:

- only search the file path ending with patter ".parquet" or ".avro", instead of pattern in the middle of path.
- Ignore the empty ".parquet" file or ".avro" file. Send error message: "Could Not find non-empty file" if the directory only contain empty ".parquet" or ".avro" file or file with invalid schema.

### DIFF
--- a/core-plugins/docs/File-batchsource.md
+++ b/core-plugins/docs/File-batchsource.md
@@ -24,9 +24,18 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
 
 **Get Schema:** Auto-detects schema from file. Supported formats are: avro, parquet, csv, delimited, tsv, blob 
 and text.
+
 Blob - is set by default as field named 'body' of type bytes.
+
 Text - is set by default as two fields: 'body' of type bytes and 'offset' of type 'long'.
+
 JSON - is not supported, user has to manually provide the output schema.
+
+Parquet - If the path is a directory, the plugin will look for files ending in '.parquet' to read the schema from. 
+If no such file can be found, an error will be returned.
+
+Avro - If the path is a directory, the plugin will look for files ending in '.avro' to read the schema from. 
+If no such file can be found, an error will be returned.
 
 **Override:** A list of columns with the corresponding data types for whom the automatic data type detection gets
  skipped. 
--- a/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
+++ b/format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
@@ -41,6 +41,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapreduce.Job;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 
@@ -115,35 +116,33 @@ public class AvroInputFormatProvider extends PathTrackingInputFormatProvider<Avr
    */
   public Schema getDefaultSchema(@Nullable FormatContext context) throws IOException {
     String filePath = conf.getProperties().getProperties().getOrDefault("path", null);
-    SeekableInput seekableInput = null;
-    FileReader<GenericRecord> dataFileReader = null;
-    try {
-      Job job = JobUtils.createInstance();
-      Configuration hconf = job.getConfiguration();
-      // set entries here, before FileSystem is used
-      for (Map.Entry<String, String> entry : conf.getFileSystemProperties().entrySet()) {
-        hconf.set(entry.getKey(), entry.getValue());
-      }
-      Path file = conf.getFilePathForSchemaGeneration(filePath, ".+\\.avro", hconf, job);
+    Job job = JobUtils.createInstance();
+    Configuration hconf = job.getConfiguration();
+
+    for (Map.Entry<String, String> entry : conf.getFileSystemProperties().entrySet()) {
+      hconf.set(entry.getKey(), entry.getValue());
+    }
+
+    List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.avro$", hconf, job);
+    for (Path file : paths) {
       DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
-      seekableInput = new FsInput(file, hconf);
-      dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
-      GenericRecord firstRecord;
-      if (!dataFileReader.hasNext()) {
-        return null;
-      }
-      firstRecord = dataFileReader.next();
-      return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
-    } catch (IOException e) {
-      context.getFailureCollector().addFailure("Schema parse error", e.getMessage());
-    } finally {
-      if (dataFileReader != null) {
-        dataFileReader.close();
-      }
-      if (seekableInput != null) {
-        seekableInput.close();
+      try (SeekableInput seekableInput = new FsInput(file, hconf);
+           FileReader<GenericRecord> dataFileReader = DataFileReader.openReader(seekableInput, dataReader)) {
+        if (!dataFileReader.hasNext()) {
+          continue;
+        }
+        GenericRecord firstRecord = dataFileReader.next();
+        if (firstRecord == null) {
+          continue;
+        }
+        return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
+      } catch (IOException e) {
+        context.getFailureCollector().addFailure("Schema parse error", e.getMessage());
       }
     }
+    context.getFailureCollector().addFailure("Could not find a valid Avro file to parse schema. " +
+                                               "Expected to find non-empty Avro file with valid schema.",
+                                                 null);
     return null;
   }
 }
--- a/format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
@@ -33,8 +33,10 @@ import org.apache.hadoop.mapreduce.Job;
 
 import java.io.IOException;
 import java.lang.reflect.Type;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -112,21 +114,25 @@ public class PathTrackingConfig extends PluginConfig {
    * @param path path from config
    * @param regexPathFilter the regex used to filter the files
    * @param job job to retrieve the file system
-   * @return {@link Path}
+   * @return {@link List<Path>}
    */
-  public Path getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
-    throws IOException {
+  public List<Path> getFilePathsForSchemaGeneration(String path,
+                                                    String regexPathFilter,
+                                                    Configuration configuration,
+                                                    Job job)
+    throws IOException { 
     Path fsPath = new Path(path);
     // need this to load the extra class loader to avoid ClassNotFoundException for the file system
     FileSystem fs = JobUtils.applyWithExtraClassLoader(job, getClass().getClassLoader(),
                                                        f -> FileSystem.get(fsPath.toUri(), configuration));
-
+    List<Path> paths = new ArrayList<Path>();
     if (!fs.exists(fsPath)) {
       throw new IOException("Input path not found");
     }
 
     if (fs.isFile(fsPath)) {
-      return fsPath;
+      paths.add(fsPath);
+      return paths;
     }
 
     final FileStatus[] files = fs.listStatus(fsPath);
@@ -138,17 +144,23 @@ public class PathTrackingConfig extends PluginConfig {
     if (files.length == 0) {
       throw new IllegalArgumentException("Provided directory is empty");
     }
-
+    
+    Pattern pattern = null;
+    if (!Strings.isNullOrEmpty(regexPathFilter)) {
+      pattern = Pattern.compile(regexPathFilter);
+    }
     for (FileStatus file : files) {
-      if (Strings.isNullOrEmpty(regexPathFilter)) {
-        return file.getPath();
+      if (pattern == null) {
+        paths.add(file.getPath());
       } else {
-        Pattern pattern = Pattern.compile(regexPathFilter);
         Matcher matcher = pattern.matcher(file.getPath().toString());
         if (matcher.find()) {
-          return file.getPath();
+          paths.add(file.getPath());
         }
       }
+    }
+    if (!paths.isEmpty()) {
+      return paths;
     }
     throw new IllegalArgumentException(String.format("No file inside \"%s\" matched regex \"%s\"!", path,
                                                      regexPathFilter));
--- a/format-common/src/test/java/io/cdap/plugin/format/input/PathTrackingConfigTest.java
+++ b/format-common/src/test/java/io/cdap/plugin/format/input/PathTrackingConfigTest.java
@@ -1,0 +1,163 @@
+/*
+ *  Copyright ©2022 Cask Data, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ *  use this file except in compliance with the License. You may obtain a copy of
+ *  the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations under
+ *  the License.
+ */
+
+package io.cdap.plugin.format.input;
+
+import io.cdap.plugin.common.batch.JobUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapreduce.Job;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class PathTrackingConfigTest {
+  String pathStr;
+  Job job;
+  Configuration hconf;
+  PathTrackingConfig conf = new PathTrackingConfig();
+
+  @ClassRule
+  public static TemporaryFolder tmpFolder = new TemporaryFolder();
+
+  @Before
+  public void init() throws IOException {
+    job = JobUtils.createInstance();
+    hconf = job.getConfiguration();
+  }
+  /**
+   * Check whether two list of Path are same, ignore the order.
+   */
+  public Boolean equalPaths(List<Path> paths1, List<Path> paths2) {
+    String p1;
+    String p2;
+    int l1 = paths1.size();
+    int l2 = paths2.size();
+    if (l1 != l2) {
+      return false;
+    }
+    for (int i = 0; i < l1; i++) {
+      p1 = paths1.get(i).toString();
+      p2 = paths2.get(i).toString();
+      if (!p1.equals(p2)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Test read all parquet file from a directory
+   */
+  @Test
+  public void testGetFilePathsForSchemaGeneration1() throws IOException {
+    File testDir = tmpFolder.newFolder("d");
+    File testFile1 = File.createTempFile("part-0", ".parquet", testDir);
+    File testFile2 = File.createTempFile("part-1", ".parquet", testDir);
+    pathStr = testDir.getAbsolutePath();
+
+    // Get List of Path from getFilePathsForSchemaGeneration
+    List<Path> filePaths = conf.getFilePathsForSchemaGeneration(pathStr, ".+\\.parquet$", hconf, job);
+
+    // Created Expected List of Path
+    List<Path> expectedPaths = new ArrayList<Path>();
+    expectedPaths.add(new Path("file:" + testFile1.getAbsolutePath()));
+    expectedPaths.add(new Path("file:" + testFile2.getAbsolutePath()));
+
+    Collections.sort(expectedPaths);
+    Collections.sort(filePaths);
+    Assert.assertTrue(equalPaths(expectedPaths, filePaths));
+  }
+
+  /**
+   * Test read all parquet file from a directory ending with .parquet
+   */
+  @Test
+  public void testGetFilePathsForSchemaGeneration2() throws IOException {
+    File testDir = tmpFolder.newFolder("d.parquet");
+    File testFile1 = File.createTempFile("part-0", ".parquet", testDir);
+    File testFile2 = File.createTempFile("part-1", ".parquet", testDir);
+    File testFile3 = File.createTempFile("part-2", ".txt", testDir);
+    pathStr = testDir.getAbsolutePath();
+
+    // Get List of Path from getFilePathsForSchemaGeneration
+    List<Path> filePaths = conf.getFilePathsForSchemaGeneration(pathStr, ".+\\.parquet$", hconf, job);
+
+    // Created Expected List of Path
+    List<Path> expectedPaths = new ArrayList<Path>();
+    expectedPaths.add(new Path("file:" + testFile1.getAbsolutePath()));
+    expectedPaths.add(new Path("file:" + testFile2.getAbsolutePath()));
+
+    Collections.sort(expectedPaths);
+    Collections.sort(filePaths);
+    Assert.assertTrue(equalPaths(expectedPaths, filePaths));
+  }
+
+  /**
+   * Test read all avro file from a directory
+   */
+  @Test
+  public void testGetFilePathsForSchemaGeneration3() throws IOException {
+    File testDir = tmpFolder.newFolder("d2");
+    File testFile1 = File.createTempFile("avro-0", ".avro", testDir);
+    File testFile2 = File.createTempFile("avro-1", ".avro", testDir);
+    pathStr = testDir.getAbsolutePath();
+
+    // Get List of Path from getFilePathsForSchemaGeneration
+    List<Path> filePaths = conf.getFilePathsForSchemaGeneration(pathStr, ".+\\.avro$", hconf, job);
+
+    // Created Expected List of Path
+    List<Path> expectedPaths = new ArrayList<Path>();
+    expectedPaths.add(new Path("file:" + testFile1.getAbsolutePath()));
+    expectedPaths.add(new Path("file:" + testFile2.getAbsolutePath()));
+
+    Collections.sort(expectedPaths);
+    Collections.sort(filePaths);
+    Assert.assertTrue(equalPaths(expectedPaths, filePaths));
+  }
+
+  /**
+   * Test read all avro file from a directory ending with .avro
+   */
+  @Test
+  public void testGetFilePathsForSchemaGeneration4() throws IOException {
+    File testDir = tmpFolder.newFolder("d.avro");
+    File testFile1 = File.createTempFile("avro-0", ".avro", testDir);
+    File testFile2 = File.createTempFile("avro-1", ".avro", testDir);
+    File testFile3 = File.createTempFile("avro-2", ".txt", testDir);
+    pathStr = testDir.getAbsolutePath();
+
+    // Get List of Path from getFilePathsForSchemaGeneration
+    List<Path> filePaths = conf.getFilePathsForSchemaGeneration(pathStr, ".+\\.avro$", hconf, job);
+
+    // Created Expected List of Path
+    List<Path> expectedPaths = new ArrayList<Path>();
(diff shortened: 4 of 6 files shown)


format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,23 +118,32 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
-      Path file = conf.getFilePathForSchemaGeneration(filePath, ".+\\.avro", hconf, job);
-      DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
-      seekableInput = new FsInput(file, hconf);
-      dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
-      GenericRecord firstRecord;
-      if (!dataFileReader.hasNext()) {
-        return null;
+      // Path file = conf.getFilePathForSchemaGeneration(filePath, ".+\\.avro$", hconf, job);
+
+      ArrayList<Path> paths = conf.getFilePathForSchemaGeneration(filePath, ".+\\.avro$", hconf, job);
```
fernst: Can we add a few tests to validate this regular expression?

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,23 +118,32 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+        try {
+          dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
+          firstRecord = dataFileReader.next();
+          return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
+        } catch (Exception e) {
+          continue;
+        }
+      }
+      if (firstRecord == null) {
+        context.getFailureCollector().addFailure("Could Not find non-empty file", null);
```
fernst: nit: Change this to `Could not find a valid Avro file to parse schema.`

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,25 @@ public Schema getSchema(FormatContext context) {
...
+      ArrayList<Path> paths = conf.getFilePathForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job);
+      for (Path file : paths) {
+        reader = AvroParquetReader.builder(file).build();
+        record = (GenericData.Record) reader.read();
+        if (record != null) {
+          return Schema.parseJson(record.getSchema().toString());
+        }
+      }
+      if (record == null) {
+        context.getFailureCollector().addFailure("Could Not find non-empty file", null);
```
snehalakshmisha: Can we add more details about what is the non-empty file we were looking for?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -188,7 +189,9 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
     for (Map.Entry<String, String> entry : getFileSystemProperties().entrySet()) {
       configuration.set(entry.getKey(), entry.getValue());
     }
-    Path filePath = getFilePathForSchemaGeneration(path, regexPathFilter, configuration, job);
+//    Path filePath = getFilePathForSchemaGeneration(path, regexPathFilter, configuration, job);
```
snehalakshmisha: Remove commented code

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -114,19 +115,21 @@ public Schema getSchema() {
    * @param job job to retrieve the file system
    * @return {@link Path}
    */
-  public Path getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
+  public ArrayList<Path>
```
fernst: Change this so the function name is in the same line:

```
  public ArrayList<Path> getFilePathForSchemaGeneration(String path, 
                                                        String regexPathFilter, 
                                                        Configuration configuration, 
                                                        Job job) { ...
```

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,25 @@ public Schema getSchema(FormatContext context) {
...
       Configuration hconf = job.getConfiguration();
       // set entries here, before FileSystem is used
       for (Map.Entry<String, String> entry : conf.getFileSystemProperties().entrySet()) {
         hconf.set(entry.getKey(), entry.getValue());
       }
-      final Path file = conf.getFilePathForSchemaGeneration(filePath, ".+\\.parquet", hconf, job);
-      reader = AvroParquetReader.builder(file).build();
-      GenericData.Record record = (GenericData.Record) reader.read();
-      return Schema.parseJson(record.getSchema().toString());
+      ArrayList<Path> paths = conf.getFilePathForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job);
```
fernst: Also add tests for this regex

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
fernst: nit: Change this to `Could not find a valid Parquet file to parse schema.`

fernst (changes requested):
Added a few comments/suggestions

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -114,19 +115,21 @@ public Schema getSchema() {
...
+    getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
```
sechegaray: If its returning a List should we change this to getPaths

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
sechegaray: Does it need to be an ArrayList? If not can we change the signature to return a List instead

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -114,19 +115,21 @@ public Schema getSchema() {
    * @param job job to retrieve the file system
    * @return {@link Path}
```
sechegaray: Please update documentation

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -188,7 +189,9 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+    ArrayList<Path> paths = getFilePathForSchemaGeneration(path, regexPathFilter, configuration, job);
+    Path filePath = paths.get(0);
```
sechegaray, line 193: We are returning a list but only using the first result?  Why are we returning a list?

sechegaray (changes requested):

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
fernst: +1. We should always return the Interface

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -114,19 +115,21 @@ public Schema getSchema() {
...
-  public Path getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
+  public ArrayList<Path>
+    getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
     throws IOException {
     Path fsPath = new Path(path);
     // need this to load the extra class loader to avoid ClassNotFoundException for the file system
     FileSystem fs = JobUtils.applyWithExtraClassLoader(job, getClass().getClassLoader(),
                                                        f -> FileSystem.get(fsPath.toUri(), configuration));
-
+    final ArrayList<Path> paths = new ArrayList<>();
```
fernst: `paths` can be a `List<Path>`

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
JessieYee: Thanks for suggestion! I will try to use the List instead of ArrayList

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
JessieYee: I test them locally, use my own directory path and change the file name. If we want to add the test, where should I write the test file?

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
JessieYee: In previous code for function getFilePathForSchemaGeneration, we only return the first matched path we detected in the directory. However, we need to check if file is empty or not for ".parquet" and ".avro" format. I change the return type to list, so that we can iterate to check all of the matched file path. And return the first non-empty one.

getFilePathForSchemaGeneration function return a list, because we want to get all possible result for ".parque" or ".avro" format. 

For delimited format, we also use the function getFilePathForSchemaGeneration. Based on my understanding of previous code, we only return the first result path. So I use a get(0) to find the first result for delimited.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
JessieYee: I will update the error message for empty file.

For the parquet format, It might be possible to get an empty ".parquet" file which will return null error when we click the get schema button. And the non-empty file means the ".parquet" file which contained data.

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
JessieYee: Yep, I will update them. Thanks!

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
JessieYee: Maybe change to getFilePathsForSchemaGeneration?

core-plugins/docs/File-batchsource.md
```diff
@@ -16,18 +16,28 @@ Properties
 
 **Format:** Format of the data to read.
 The format must be one of 'avro', 'blob', 'csv', 'delimited', 'json', 'parquet', 'text', 'tsv', or the
-name of any format plugin that you have deployed to your environment.
+name of any format plugin that you have deployed to your environment.co
```
fernst: This was added by mistake

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,23 +118,31 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+          dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
+          firstRecord = dataFileReader.next();
+          return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
+        } catch (Exception e) {
+          continue;
+        }
+      }
+      if (firstRecord == null) {
+        context.getFailureCollector().addFailure("Could not find a valid Avro file to parse schema. " +
+                                                   "Expected to detect non-empty Avro file with valid schema.",
```
fernst: Change `detect` to `find`.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,27 @@ public Schema getSchema(FormatContext context) {
...
+      for (Path file : paths) {
+        reader = AvroParquetReader.builder(file).build();
+        record = (GenericData.Record) reader.read();
+        if (record != null) {
+          return Schema.parseJson(record.getSchema().toString());
+        }
+      }
+      if (record == null) {
+        context.getFailureCollector().addFailure("Could not find a valid Parquet file to parse schema. " +
+                                                   "Expected to detect non-empty Parquet file with valid schema.",
```
fernst: replace `detect` with `find`.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,23 +118,31 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
-      DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
-      seekableInput = new FsInput(file, hconf);
-      dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
-      GenericRecord firstRecord;
-      if (!dataFileReader.hasNext()) {
-        return null;
+      List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.avro$", hconf, job); 
+      for (Path file : paths) {
+        DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
+        seekableInput = new FsInput(file, hconf);
```
chtyim: This needs to be closed.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,23 +118,31 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+        try {
+          dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
```
chtyim: This needs to be closed.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
fernst: For the Parquet use case, we do check files until we find the first one that can give us a valid schema. This is not the case for Delimited and PathTracking implementations.

fernst (changes requested):
Code LGTM, but can you please add a few tests for the new logic?

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -114,19 +116,23 @@ public Schema getSchema() {
    * @param job job to retrieve the file system
    * @return {@link Path}
```
de-lan: nitpick: `Path` --> `List<Path>`

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,30 @@ public Schema getSchema(FormatContext context) {
...
-      GenericData.Record record = (GenericData.Record) reader.read();
-      return Schema.parseJson(record.getSchema().toString());
+      List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job); 
+      for (Path file : paths) {
+        reader = AvroParquetReader.builder(file).build();
+        record = (GenericData.Record) reader.read();
+        if (record != null) {
+          return Schema.parseJson(record.getSchema().toString());
+        }
+        if (reader != null) {
```
de-lan: This condition check might not be necessary? My IDE is saying `Condition 'reader != null' is always 'true' `

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,30 @@ public Schema getSchema(FormatContext context) {
...
+          reader.close();
+        }
+      }
+      if (record == null) {
```
de-lan: Here as well. `Condition 'record == null' is always 'true'`

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
JessieYee: Yes, I will remove this condition

fernst (approved):
Please squash all 18 commits into one before merging

de-lan (approved):

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,32 +118,40 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
+          firstRecord = dataFileReader.next();
+          return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
+        } catch (Exception e) {
+          continue;
+        } finally {
+          if (dataFileReader != null) {
```
chtyim: Use `try` block to close resources instead of doing it in finally (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -112,21 +114,25 @@ public Schema getSchema() {
    * @param path path from config
    * @param regexPathFilter the regex used to filter the files
    * @param job job to retrieve the file system
-   * @return {@link Path}
+   * @return {@link List<Path>}
    */
-  public Path getFilePathForSchemaGeneration(String path, String regexPathFilter, Configuration configuration, Job job)
-    throws IOException {
+  public List<Path> getFilePathsForSchemaGeneration(String path,
+                                                   String regexPathFilter,
```
chtyim: Misalign

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -141,15 +147,18 @@ public Path getFilePathForSchemaGeneration(String path, String regexPathFilter,
...
       } else {
         Pattern pattern = Pattern.compile(regexPathFilter);
         Matcher matcher = pattern.matcher(file.getPath().toString());
         if (matcher.find()) {
-          return file.getPath();
+          paths.add(file.getPath());
         }
       }
     }
+    if (paths.size() > 0) {
```
chtyim: Use `!paths.isEmpty()` is more clear about the intention.

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -141,15 +147,18 @@ public Path getFilePathForSchemaGeneration(String path, String regexPathFilter,
 
     for (FileStatus file : files) {
       if (Strings.isNullOrEmpty(regexPathFilter)) {
-        return file.getPath();
+        paths.add(file.getPath());
       } else {
         Pattern pattern = Pattern.compile(regexPathFilter);
```
chtyim: Don't compile pattern inside a loop. Compile it outside and use matcher inside the loop.

format-common/src/test/java/io/cdap/plugin/format/input/PathTrackingConfigTest.java
```diff
@@ -0,0 +1,161 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class PathTrackingConfigTest {
+  String pathStr;
+  Job job;
+  Configuration hconf;
+  List<Path> expectedPaths;
```
chtyim: Avoid using field for this as it can get accumulated across tests. Use local variable instead.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -92,17 +93,28 @@ public Schema getSchema(FormatContext context) {
...
-      GenericData.Record record = (GenericData.Record) reader.read();
-      return Schema.parseJson(record.getSchema().toString());
+      List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job); 
+      for (Path file : paths) {
+        reader = AvroParquetReader.builder(file).build();
+        record = (GenericData.Record) reader.read();
+        if (record != null) {
+          return Schema.parseJson(record.getSchema().toString());
+        }
+        reader.close();
```
chtyim: Use try with resource to make sure resource is always closed (e.g. in this case, since you have a `return` earlier, the reader actually never get closed).

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,32 +118,40 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
-        return null;
+      List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.avro$", hconf, job); 
+      for (Path file : paths) {
+        DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
+        seekableInput = new FsInput(file, hconf);
+        try {
+          dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
+          firstRecord = dataFileReader.next();
+          return new AvroToStructuredTransformer().convertSchema(firstRecord.getSchema());
+        } catch (Exception e) {
```
sechegaray: Nit: We are capturing all kinds of exception here. We should log the exception to make it easy to debug (e.g. to separate issues such as are we not loading the file because we don't have read permissions? or because it's invalid?)

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
sechegaray: +1

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
fernst: This could lead to noisy logging, but I think logging exceptions here is a good idea. In general, I think customers should be aware if some of their avro files might be formatted incorrectly, or if there are permission issues for some files in the directory.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
JessieYee: If we have return in try, reader.close() in finally will still be executed. I will update them use the try-with resource.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
JessieYee: Thank you Terence. I will update it :)

core-plugins/docs/File-batchsource.md
```diff
@@ -24,10 +24,20 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
...
 **Get Schema:** Auto-detects schema from file. Supported formats are: avro, parquet, csv, delimited, tsv, blob 
 and text.
+
 Blob - is set by default as field named 'body' of type bytes.
+
 Text - is set by default as two fields: 'body' of type bytes and 'offset' of type 'long'.
+
 JSON - is not supported, user has to manually provide the output schema.
 
+Parquet - If Path is a directory, will search all "parquet" files, and read the first valid non-empty "parquet"
```
albertshau: if the path is a directory, the plugin will look for files ending in '.parquet' to read the schema from. If no such file can be found, an error will be returned.

core-plugins/docs/File-batchsource.md
```diff
@@ -24,10 +24,20 @@ If the format is 'text', the schema must contain a field named 'body' of type 's
...
+
 Text - is set by default as two fields: 'body' of type bytes and 'offset' of type 'long'.
+
 JSON - is not supported, user has to manually provide the output schema.
 
+Parquet - If Path is a directory, will search all "parquet" files, and read the first valid non-empty "parquet" 
+file's schema under the directory. Return error message if no non-empty "parquet" file is detected or all schemas 
+are invalid.
+
+Avro - If Path is a directory, will search all "avro" files, and read the first valid non-empty "avro" file's schema
```
albertshau: if the path is a directory, the plugin will look for files ending in '.avro' to read the schema from. If no such file can be found, an error will be returned.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
albertshau: If there is an exception here, it's better to actually let it propagate up. An invalid file will cause an error when the pipeline runs, so it's better to error here also instead of pretending that it is ok.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,32 +118,40 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
...
-      GenericRecord firstRecord;
-      if (!dataFileReader.hasNext()) {
-        return null;
+      List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.avro$", hconf, job); 
+      for (Path file : paths) {
+        DatumReader<GenericRecord> dataReader = new GenericDatumReader<>();
+        seekableInput = new FsInput(file, hconf);
+        try {
+          dataFileReader = DataFileReader.openReader(seekableInput, dataReader);
+          firstRecord = dataFileReader.next();
```
albertshau: need to check dataFileReader.hasNext() before calling next() here. If there is no next, continue.

Probably also safer to check if firstRecord is null and continue if it is.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
```diff
@@ -117,32 +118,40 @@ public Schema getDefaultSchema(@Nullable FormatContext context) throws IOExcepti
     String filePath = conf.getProperties().getProperties().getOrDefault("path", null);
     SeekableInput seekableInput = null;
     FileReader<GenericRecord> dataFileReader = null;
+    GenericRecord firstRecord = null;
```
albertshau: better to declare things closer to where they are used. In this case, SeekableInput, FileReader, and GenericRecord  can all be declared within the loop instead of outside of it.

format-avro/src/main/java/io/cdap/plugin/format/avro/input/AvroInputFormatProvider.java
JessieYee: OK. So if we find an invalid file (no permission or something cause fail to read), the loop will stop and an error will be shown in the UI. And if we find an empty file, we will just skip it and check the next file.  Is that correct?

format-common/src/test/java/io/cdap/plugin/format/input/PathTrackingConfigTest.java
JessieYee: Got it :)

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
JessieYee: Got it :)

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingConfig.java
```diff
@@ -138,18 +144,20 @@ public Path getFilePathForSchemaGeneration(String path, String regexPathFilter,
     if (files.length == 0) {
       throw new IllegalArgumentException("Provided directory is empty");
     }
-
+    Pattern pattern = Pattern.compile(regexPathFilter);
     for (FileStatus file : files) {
       if (Strings.isNullOrEmpty(regexPathFilter)) {
```
chtyim: If the filter string is null, the compilation of Pattern would fail. You should move this check to outside of the loop such that you can have `pattern` set to `null` if no filter is provided and then checks for `pattern == null` here.

JessieYee: Thanks, finish checking the filter.

sechegaray (approved):

PR #1565, opened 2022-04-04
Fixed issues with Concat and Concat Distinct aggregators

fernst:
Fixed issue with Concat aggregator which caused an empty "right" side of the merge operation to appear in some cases.

Fixed issue with Concat Distinct aggregator not filtering duplicate records when merging partitions

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/ConcatDistinct.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/ConcatDistinct.java
@@ -21,6 +21,10 @@ import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.data.schema.Schema.Type;
 
+import java.util.LinkedHashSet;
+import java.util.Set;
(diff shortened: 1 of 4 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/function/ConcatDistinct.java
```diff
@@ -44,41 +47,25 @@ public ConcatDistinct(String fieldName, Schema fieldSchema) {
...
       String value = record.get(fieldName);
-      if (value != null && !concatString.contains(value)) {
-        if (firstString) {
-          concatString = value;
-          firstString = false;
-          return;
-        }
-        concatString = String.format("%s, %s", concatString, value);
-      }
+      values.add(value);
```
sumengwang, line 57: I think LinkedHashSet allows one null value. Do you want null value?

fernst: We don't want nulls. We check for nulls in line 55 of this file before entering this code block.

sumengwang (approved):
LGTM

sanjanasandeep (approved):

PR #1568, opened 2022-04-04
Enable starttls for pipeline email alerts.

sanjanasandeep:
**Issue**
https://cdap.atlassian.net/browse/PLUGIN-1140

**Description**

When `tls` is set as protocol in the ui:  `tls` is passed as configuration property protocol causing "no providers are detected for tls" thrown by `javax.mail`. The issue here is `tls` is not really a protocol. Ideally we want to use `smtp` with `starttls` command enabled.

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
@@ -78,6 +78,7 @@ public class EmailAction extends PostAction {
     Authenticator authenticator = null;
 
     Properties javaMailProperties = new Properties();
+    String protocolForTransport = config.protocol;
     javaMailProperties.put("mail.smtp.host", config.host);
     javaMailProperties.put("mail.smtp.port", config.port);
     if (!(Strings.isNullOrEmpty(config.username))) {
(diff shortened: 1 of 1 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
```diff
@@ -93,7 +93,9 @@ public PasswordAuthentication getPasswordAuthentication() {
       javaMailProperties.put("mail.smtp.ssl.enable", true);
     }
     if ("TLS".equalsIgnoreCase(config.protocol)) {
-      javaMailProperties.put("mail.smtp.starttls.enable", true);
+      javaMailProperties.put("mail.smtp.starttls.enable", "true");
+      javaMailProperties.put("mail.smtp.ssl.trust", config.host);
+      config.protocol = "smtp"; // protocol is smtp not tls.
```
dli357: The `config.protocol` field directly represents the text input for the `protocol` field displayed in the UI, so it might be a bit jarring to change it from under the user. Can we instead pass in a separate variable to line 123?

sanjanasandeep: yeah I agree

dli357 (approved):
Thanks!

PR #1570, opened 2022-04-07
CDAP-18994 fix errors with nullable map values

albertshau:
Remove schema preprocessing that was causing failures when a schema
had a map with nullable values. Manually modifying the schema json is
brittle, doesn't fix the issue for parquet, and is also not performant.

CDAP schema parsing now does the correct thing so no preprocessing
is needed.


albertshau:
cherry-pick of #1569

seanzhougoogle (approved):

PR #1571, opened 2022-04-08
[PLUGIN-1195] Added logic to cast FLOAT fields to NUMERIC when using for Window function for deduplication

fernst:


sanjanasandeep (approved):
lgtm

PR #1572, opened 2022-04-08
Added condition inside expression

sechegaray:


njbriggs (approved):
LGTM. FWIW, the client-side types for this structure are at https://github.com/cdapio/cdap-ui/blob/develop/app/cdap/components/shared/ConfigurationGroup/types.ts#L74 and the change matches the types.

PR #1575, opened 2022-04-11
Updated String Expression Factory Type to BigQuery SQL

fernst:
Depends on https://github.com/cdapio/cdap/pull/14178

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
@@ -16,8 +16,10 @@
 
 package io.cdap.plugin.batch.aggregator;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.aggregation.DeduplicateAggregationDefinition;
+import io.cdap.cdap.etl.api.engine.sql.StandardSQLCapabilities;
 import io.cdap.cdap.etl.api.relational.CoreExpressionCapabilities;
 import io.cdap.cdap.etl.api.relational.Expression;
 import io.cdap.cdap.etl.api.relational.ExpressionFactory;
@@ -40,6 +42,15 @@ public class DedupAggregatorUtils {
 
   private static final String CAST_TO_NUMERIC_FORMAT = "CAST(%s AS NUMERIC)";
 
+  /**
+   * Generates a {@link DeduplicateAggregationDefinition} based on an input relation and deduplication configuration.
+   * @param ctx transform context
+   * @param relation input relation
+   * @param filter filter function
+   * @param uniqueFields unique fields when deduplicating
+   * @return Definition for a deduplicate operation, or null if this deduplicate definition is not supported.
+   */
+  @Nullable
   public static DeduplicateAggregationDefinition generateAggregationDefinition(RelationalTranformContext ctx,
                                                                                Relation relation,
                                                                                DedupConfig.DedupFunctionInfo filter,
@@ -48,8 +59,13 @@ public class DedupAggregatorUtils {
     String inputRelationName = ctx.getInputRelationNames().stream().findFirst().orElse(null);
     Schema inputSchema = inputRelationName != null ? ctx.getInputSchema(inputRelationName) : null;
 
-    Optional<ExpressionFactory<String>> expressionFactory = ctx.getEngine().
-      getExpressionFactory(StringExpressionFactoryType.SQL);
+    Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(ctx, inputSchema);
+
+    // If the expression factory is not present, this aggregation cannot be handled by the plugin.
+    if (!expressionFactory.isPresent()) {
+      return null;
+    }
+
     DeduplicateAggregationDefinition.FilterFunction aggFilterFunction;
     switch (filter.getFunction()) {
       case MAX:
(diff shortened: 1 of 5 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -107,23 +108,49 @@ static Expression getDedupColumnName(Relation relation,
...
+      return Optional.empty();
+    }
+
+    // The BigQuery capability is required for Float or Double fields.
+    boolean requiresBigQueryCapability =
+      inputSchema.getFields().stream().anyMatch(DedupAggregatorUtils::isFloatOrDoubleField);
+
+    // If the BigQuery capability is required, ensure the SQL engine supports this capability.
+    return requiresBigQueryCapability ?
+      ctx.getEngine().getExpressionFactory(StringExpressionFactoryType.SQL, StandardSQLCapabilities.BIGQUERY) :
```
tivv, line 150: nit: you don't need to check for SQL as I suppose BigQuery would imply SQL.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -48,8 +50,7 @@ public static DeduplicateAggregationDefinition generateAggregationDefinition(Rel
     String inputRelationName = ctx.getInputRelationNames().stream().findFirst().orElse(null);
     Schema inputSchema = inputRelationName != null ? ctx.getInputSchema(inputRelationName) : null;
 
-    Optional<ExpressionFactory<String>> expressionFactory = ctx.getEngine().
-      getExpressionFactory(StringExpressionFactoryType.SQL);
+    Optional<ExpressionFactory<String>> expressionFactory = getExpressionFactory(ctx, inputSchema);
```
tivv, line 62: Please handle case when this Optional is empty

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -107,24 +108,31 @@ public class GroupByAggregator extends RecordReducibleAggregator<AggregateResult
...
+      put(GroupByConfig.Function.AVG, "AVG(%s)");
+      put(GroupByConfig.Function.MAX, "MAX(%s)");
+      put(GroupByConfig.Function.MIN, "MIN(%s)");
+      put(GroupByConfig.Function.STDDEV, "STDDEV_POP(%s)");
+      put(GroupByConfig.Function.SUM, "SUM(%s)");
+      put(GroupByConfig.Function.VARIANCE, "VAR_POP(%s)");
+      put(GroupByConfig.Function.COUNT, "COUNT(%s)");
+      put(GroupByConfig.Function.COUNTNULLS, "SUM(CASE WHEN %s IS NULL THEN 1 END)");
+      put(GroupByConfig.Function.COUNTDISTINCT,
+          "COUNT(DISTINCT %s) + IF(SUM(CASE WHEN %s IS NULL THEN 1 END) > 0, 1, 0)");
```
tivv: It's better to use `case` instead of `if` to be ansi compatible

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -107,24 +108,31 @@ public class GroupByAggregator extends RecordReducibleAggregator<AggregateResult
...
-    put(GroupByConfig.Function.COUNTNULLS, "COUNTIF(%s IS NULL)");
-    put(GroupByConfig.Function.CONCAT, "STRING_AGG(CAST(%s AS STRING), \", \")");
-    put(GroupByConfig.Function.CONCATDISTINCT, "STRING_AGG(DISTINCT CAST(%s AS STRING) , \", \")");
-    put(GroupByConfig.Function.LOGICALAND, "COALESCE(LOGICAL_AND(%s), TRUE)");
-    put(GroupByConfig.Function.LOGICALOR, "COALESCE(LOGICAL_OR(%s), FALSE)");
-  }};
+      put(GroupByConfig.Function.AVG, "AVG(%s)");
+      put(GroupByConfig.Function.MAX, "MAX(%s)");
+      put(GroupByConfig.Function.MIN, "MIN(%s)");
+      put(GroupByConfig.Function.STDDEV, "STDDEV_POP(%s)");
```
tivv, line 117: This looks BQ-specific, also VAR_POP

fernst: Per https://www.oreilly.com/library/view/sql-in-a/9780596155322/ch04s02.html the VAR_POP and STDDEV_POP functions are ANSE SQL standard

fernst: I can also find docs for them in Oracle, Teradata, MySQL, Postgres, etc.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: Good catch, I thought I had added this.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
tivv: Every day you learn something new!

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
```diff
@@ -48,8 +50,13 @@ public static DeduplicateAggregationDefinition generateAggregationDefinition(Rel
...
+
+    // If the expression factory is not present, this aggregation cannot be handled by the plugin.
+    if (!expressionFactory.isPresent()) {
+      return null;
```
tivv, line 66: Is caller ready to handle nulls?  Please mark method as `Nullable`. 
nit: It would be great to have a javadoc as this is public method

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
fernst: Updated

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/DedupAggregatorUtils.java
fernst: I've updated the annotation and added the javadoc.

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -107,24 +108,31 @@ public class GroupByAggregator extends RecordReducibleAggregator<AggregateResult
...
+      put(GroupByConfig.Function.SUM, "SUM(%s)");
+      put(GroupByConfig.Function.VARIANCE, "VAR_POP(%s)");
+      put(GroupByConfig.Function.COUNT, "COUNT(%s)");
+      put(GroupByConfig.Function.COUNTNULLS, "SUM(CASE WHEN %s IS NULL THEN 1 ELSE 0 END)");
+      put(GroupByConfig.Function.COUNTDISTINCT,
+          "COUNT(DISTINCT %s) + CASE WHEN SUM(CASE WHEN %>s IS NULL THEN 1 ELSE 0 END) > 0 THEN 1 ELSE 0 END");
```
tivv: nit: can this be `COUNT(DISTINCT %s) + MAX(CASE WHEN %>s IS NULL THEN 1 ELSE 0 END)` or it would produce a wrong result for empty set? Two nested `CASE` are hard to read

tivv (approved):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
fernst: This expression can be made safe by adding a Coalesce:

COUNT(DISTINCT %s) + COALESCE(MAX(CASE WHEN %>s IS NULL THEN 1 ELSE 0 END), 0)

PR #1582, opened 2022-04-18
[PLUGIN-1210]Added support for Shortest and Longest String operations in BigQuery Pushdown

fernst:


core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -135,6 +135,10 @@ public class GroupByAggregator extends RecordReducibleAggregator<AggregateResult
       put(GroupByConfig.Function.CONCATDISTINCT, "STRING_AGG(DISTINCT CAST(%s AS STRING) , \", \")");
       put(GroupByConfig.Function.LOGICALAND, "COALESCE(LOGICAL_AND(%s), TRUE)");
       put(GroupByConfig.Function.LOGICALOR, "COALESCE(LOGICAL_OR(%s), FALSE)");
+      put(GroupByConfig.Function.SHORTESTSTRING,
+          "STRING_AGG(CAST(%s AS STRING) ORDER BY LENGTH(CAST(%<s AS STRING)) ASC LIMIT 1)");
+      put(GroupByConfig.Function.LONGESTSTRING,
+          "COALESCE(STRING_AGG(CAST(%s AS STRING) ORDER BY LENGTH(CAST(%<s AS STRING)) DESC LIMIT 1), \"\")");
```
tivv: Why do you need coalesce just in one of the functions? According to docs `string_agg` skips nulls.

fernst: Because the Shortest String aggregator returns `null` for no input values, but the Longest String aggregator returns `""` for the same input.

sanjanasandeep (approved):

tivv (approved):

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
fernst: Fixed https://cdap.atlassian.net/browse/PLUGIN-1210 alongside this PR

PR #1586, opened 2022-04-20
[PLUGIN-1161] Add Get Schema button to Projection transform

JessieYee:
Jira: https://cdap.atlassian.net/browse/PLUGIN-1161

Add "get schema" button, enable customer to check the updated schema.

### DIFF
--- a/core-plugins/widgets/Projection-transform.json
+++ b/core-plugins/widgets/Projection-transform.json
@@ -51,9 +51,39 @@
             "key-placeholder": "Field Name",
             "value-placeholder": "New Field Name"
           }
+        },
+        {
+          "plugin-function": {
+          "method": "POST",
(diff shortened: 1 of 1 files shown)


sumengwang (approved):
LGTM

core-plugins/widgets/Projection-transform.json
```diff
@@ -51,9 +51,39 @@
...
+          "label": "Get Schema Value",
+          "required-fields": [],
+          "plugin-method": "getSchema"
+          }
         }
       ]
     }
   ],
-  "outputs": []
+  "outputs": [
```
albertshau: I don't think you need this, there doesn't need to be a default schema.

JessieYee: Update the output part.

PR #1593, opened 2022-04-26
CDAP-18878 add additional function to get path

yaojiefeng:
This is a required for https://github.com/data-integrations/amazon-s3-plugins/pull/126, the s3 source/sink need a way to retrieve whether to keep using s3n from runtime arguments for backward compat.

### DIFF
--- a/format-common/src/main/java/io/cdap/plugin/format/plugin/FileSinkProperties.java
+++ b/format-common/src/main/java/io/cdap/plugin/format/plugin/FileSinkProperties.java
@@ -18,6 +18,7 @@ package io.cdap.plugin.format.plugin;
 
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.cdap.etl.api.StageContext;
 import io.cdap.plugin.format.FileFormat;
 
 import java.util.Map;
@@ -69,8 +70,19 @@ public interface FileSinkProperties {
 
   /**
    * Get the path to write to.
+   *
+   * @deprecated use {@link #getPath(StageContext)} instead
    */
+  @Deprecated
   String getPath();
+
+  /**
+   * Get the path to write to, with provided stage context, this can be used when the path is determined by some
+   * runtime arguments.
+   */
+  default String getPath(StageContext context) {
+    return getPath();
+  }
 
   /**
    * Get the name of the format plugin to use to write the data.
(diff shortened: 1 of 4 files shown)


format-common/src/main/java/io/cdap/plugin/format/plugin/FileSinkProperties.java
```diff
@@ -72,6 +73,14 @@ default void validate(FailureCollector collector, Map<String, String> arguments)
    */
   String getPath();
 
+  /**
+   * Get the path to write to, with provided stage context, this can be used when the path is determined by some
+   * runtime arguments. The context is null at configure time
+   */
+  default String getPath(@Nullable StageContext context) {
```
albertshau: it's odd to me that the context is nullable. Where are we calling with null and can we just call getPath() in those situations instead?

yaojiefeng: I thought the method will be called in configurePipeline, but it is actually only called in prepareRun, so context should never be null, removed nullable here

format-common/src/main/java/io/cdap/plugin/format/plugin/FileSinkProperties.java
```diff
@@ -72,6 +73,14 @@ default void validate(FailureCollector collector, Map<String, String> arguments)
    */
   String getPath();
```
albertshau, line 77: do we ever call this anymore? If not, we need to annotate it as deprecated and update the javadoc. If so, we need to update the javadoc and describe how it is different than the new method.

format-common/src/main/java/io/cdap/plugin/format/plugin/FileSourceProperties.java
```diff
@@ -63,6 +64,14 @@ default void validate(FailureCollector collector) {
    */
   String getPath();
```
albertshau, line 68: do we ever call this anymore? If not, we need to annotate it as deprecated and update the javadoc. If so, we need to update the javadoc and describe how it is different than the new method.

yaojiefeng: This method is no different from the added method and is not called, with the added method you can have additional information about runtime arguments, marked it as deprecated.

format-common/src/main/java/io/cdap/plugin/format/plugin/FileSinkProperties.java
yaojiefeng: This method is no different from the added method and is not called, with the added method you can have additional information about runtime arguments, marked it as deprecated.

albertshau (approved):
lgtm

PR #1598, opened 2022-04-27
mark getOutputDir private

yaojiefeng:


format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSink.java
```diff
@@ -58,7 +58,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(AbstractFileSink.class);
   private static final String NAME_FORMAT = "format";
   private final T config;
-  private String outputPath;
+  protected String outputPath;
```
albertshau: on closer look, it's better to keep this private and keep the getOutputDir() method protected. Don't want to expose this and let children overwrite this value.

albertshau (approved):

PR #1599, opened 2022-04-27
[CDAP-18910] improve file connection markdown

sumengwang:
[CDAP-18910](https://cdap.atlassian.net/browse/CDAP-18910)

image

image

### DIFF
--- a/core-plugins/docs/File-connector.md
+++ b/core-plugins/docs/File-connector.md
@@ -1,9 +1,15 @@
-# File Connector
+# File Connection
 
 
 Description
 -----------
-File connector can be used to browse and sample the local file system
+Use this connection to browse and sample the local file system.
+
+Properties
+----------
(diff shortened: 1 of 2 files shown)


yaojiefeng:
These are here because when browsing to a file, UI needs this to show the format and encoding, so we shouldn't remove it

yaojiefeng:
It is better to have UI to not show the label if no properties to display, UI relies on the plugin spec to show the properties, so if it sees no properties under the label, better just hide the entire label

de-lan (approved):

core-plugins/docs/File-connector.md
```diff
@@ -1,9 +1,15 @@
-# File Connector
+# File Connection
```
de-lan, line 1: Is this change on purpose?

sumengwang: yes requested by Robin

PR #1600, opened 2022-04-28
[Plugin-1129] automatically fixes schema field names

sechegaray:


format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -141,11 +142,57 @@ public static void validateSchemaFieldNames(String[] fieldNames) {
...
   }
+
+  /**
+   * Cleans an array of field names to make sure they comply with avro field naming standard.
+   * It also makes sure each name is unique in the list.
+   * Field names can start with [A-Za-z_] and subsequently contain only [A-Za-z0-9_].
+   *
+   * Steps:
+   * 1) Trim surrounding spaces
+   * 2) If it starts with a number, prepend "_"
```
tivv: I'd prepend "f_". Often fields that start with _ are considered hidden. Also we use field names as DB names pretty often and I am not sure if starting with _ would bring us issues

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -141,11 +142,57 @@ public static void validateSchemaFieldNames(String[] fieldNames) {
...
+   * 3) Replace invalid characters with "_"
+   * 4) Check if the name has been found before if so add _# where # is the number of times seen before
+   * @param fieldNames an array of field names to be cleaned
+   */
+  private static String[] cleanSchemaFieldNames(String[] fieldNames) {
```
tivv, line 168: How is it done now? I think we already have some cleaning up in wrangler, no?

tivv: Also, why returning an array? Usually list is a much better data type to return

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -141,11 +142,57 @@ public static void validateSchemaFieldNames(String[] fieldNames) {
...
+    final Map<String, Integer> seenFieldNames = new HashMap<>();
+    final Pattern notValidPattern = Pattern.compile("[^A-Za-z0-9_]");
+    for (String fieldName : fieldNames) {
+      StringBuilder cleanFieldNameBuilder = new StringBuilder();
+
+      // Remove any spaces at the end of the strings
+      fieldName = fieldName.trim();
+
+      // If it starts with a number, or it's an empty string prepend the replacement char
+      if ((fieldName.length() == 0) || (Character.isDigit(fieldName.charAt(0)))) {
```
tivv: `isDigit` is not same as 0-9. You may end up appending extra replacement if it's some unicode digit

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -141,11 +142,57 @@ public static void validateSchemaFieldNames(String[] fieldNames) {
...
+   * 2) If it starts with a number, prepend "_"
+   * 3) Replace invalid characters with "_"
+   * 4) Check if the name has been found before if so add _# where # is the number of times seen before
+   * @param fieldNames an array of field names to be cleaned
+   */
+  private static String[] cleanSchemaFieldNames(String[] fieldNames) {
+    final String replacementChar = "_";
+    final List<String> cleanFieldNames = new ArrayList<>();
+    final Map<String, Integer> seenFieldNames = new HashMap<>();
+    final Pattern notValidPattern = Pattern.compile("[^A-Za-z0-9_]");
```
tivv: Please add a constant with precompiled pattern. Consider adding "+" to the pattern as usually it makes sense to replace all invalid chars with a single _. Otherwise you end up with _, __, ___. I'd prefer _1,_2,_3 much better :)

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -141,11 +142,57 @@ public static void validateSchemaFieldNames(String[] fieldNames) {
...
+
+      // Check if the field exist if so append and index at the end
+      String cleanFieldName = cleanFieldNameBuilder.toString();
+
+      while (seenFieldNames.containsKey(cleanFieldName)) {
+        cleanFieldNameBuilder.append(replacementChar).append(seenFieldNames.get(cleanFieldName));
+        seenFieldNames.put(cleanFieldName, seenFieldNames.get(cleanFieldName) + 1);
+        cleanFieldName = cleanFieldNameBuilder.toString();
+      }
+      seenFieldNames.put(cleanFieldName, 1);
```
tivv: So, it gives us "a", "a_1", "a_2", correct?
Should it be "a", "a_2", "a_3"? "a" and "a_1" can be easily confused

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
sechegaray: after talking with product we decided on col_

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -122,30 +125,81 @@ public static String[] setColumnNames(String rawLine, boolean skipHeader,
...
+      // If it's an empty string replace it with BLANK
+      if (fieldName.length() == 0) {
+        cleanFieldNameBuilder.append("BLANK");
+      } else if ((fieldName.charAt(0) >= '0') && (fieldName.charAt(0) <= '9')) {
+        // Prepend a col_ if the first character is a number
+        cleanFieldNameBuilder.append("col_");
+      }
+
+      // Replace all invalid characters with the replacement char
+      cleanFieldNameBuilder.append(NOT_VALID_PATTERN.matcher(fieldName).replaceAll(replacementChar));
```
tivv, line 187: Totally nit, but given we have StringBuilder consider java.util.regex.Matcher#appendReplacement

format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataTypeDetectorUtils.java
```diff
@@ -122,30 +125,81 @@ public static String[] setColumnNames(String rawLine, boolean skipHeader,
...
+
+      // Check if the field exist if so append and index at the end
+      // We use lowercase to match columns "A" and "a" to avoid issues with wrangler.
+      String cleanFieldName = cleanFieldNameBuilder.toString();
+      String lowerCaseCleanFieldName = cleanFieldName.toLowerCase();
```
tivv, line 192: nit: `toLowerCase(Locale.ROOT)`, otherwise it's default locale that can be dependent on OS settings

tivv (approved):

PR #1601, opened 2022-04-28
Add javax.mail property as a configuration for Email Alert Plugin

sanjanasandeep:
**Issue**
https://cdap.atlassian.net/browse/PLUGIN-1233

** Description **

Added a configuration to the Email Alert plugin that takes javax.mail properties as input from the user (https://www.educba.com/java-mail-properties/?source=leftnav). Also added the fix for the SSLv3- conscrypt issue (https://cdap.atlassian.net/browse/PLUGIN-1140)

** Testing **

https://dev651-apestel-dot-usw1.datafusion.googleusercontent.com/pipelines/ns/default/view/ephemeral_test_no_input_v3

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
@@ -27,8 +27,11 @@ import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.StageConfigurer;
 import io.cdap.cdap.etl.api.batch.BatchActionContext;
 import io.cdap.cdap.etl.api.batch.PostAction;
+import io.cdap.plugin.common.ConfigUtil;
 import io.cdap.plugin.common.batch.action.ConditionConfig;
 
+import java.util.Collections;
+import java.util.Map;
 import java.util.Properties;
 import javax.annotation.Nullable;
 import javax.mail.Authenticator;
@@ -79,6 +82,9 @@ public class EmailAction extends PostAction {
 
     Properties javaMailProperties = new Properties();
     String protocolForTransport = config.protocol;
+    Map<String, String> userJavaMailProperties = Strings.isNullOrEmpty(config.configurableJavaMailProperties) ?
+            Collections.emptyMap() : ConfigUtil.parseKeyValueConfig(config.configurableJavaMailProperties,
+            ",", "=");
     javaMailProperties.put("mail.smtp.host", config.host);
     javaMailProperties.put("mail.smtp.port", config.port);
     if (!(Strings.isNullOrEmpty(config.username))) {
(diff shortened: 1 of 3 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
```diff
@@ -94,11 +100,24 @@ public PasswordAuthentication getPasswordAuthentication() {
...
-      javaMailProperties.put("mail.smtp.starttls.enable", "true");
-      javaMailProperties.put("mail.smtp.ssl.trust", config.host);
+      javaMailProperties.put("mail.smtp.starttls.enable", true);
+      if (!userJavaMailProperties.containsKey("mail.smtp.starttls.enable")) {
+        javaMailProperties.put("mail.smtp.starttls.enable", "true");
+      }
+      if (!userJavaMailProperties.containsKey("mail.smtp.ssl.trust")) {
+        javaMailProperties.put("mail.smtp.ssl.trust", config.host);
+      }
       protocolForTransport = "smtp"; // protocol is smtp not tls.
```
dli357, line 116: nit: "Underlying protocol for startTLS is SMTP" for clarity

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
```diff
@@ -94,11 +100,24 @@ public PasswordAuthentication getPasswordAuthentication() {
...
     }
 
+     if (!userJavaMailProperties.containsKey("mail.smtps.ssl.protocols")) {
+       javaMailProperties.put("mail.smtps.ssl.protocols", "TLSv1.2");
```
dli357: Is there a reason we aren't using TLSv1.3? Is that the one which isn't available without conscrypt?

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
```diff
@@ -193,6 +212,12 @@ public static class Config extends ConditionConfig {
     @Macro
     private Boolean includeWorkflowToken;
 
+    @Nullable
+    @Description("Optional property that can be used to pass specific set of javamail properties. The syntax " +
+            "is 'property name:property value'. If this property is not set, default values are chosen.")
```
dli357: Are the properties specified in `<property_name>:<value>` format? The `keyvalue` widget used below seems to imply that it is delimited by `,` and uses `=` to differentiate key and value.

dli357: nit: Please wrap the `<property_name>` and `<property_value>` in brackets for clarity, i.e. `<property_name>:<value>`, thanks!

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
sanjanasandeep: yeah TLSv.12 is the latest version of TLS that works without conscrypt.

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
sanjanasandeep: I wanted to make it clear how it is specified in the ui, which is as a key value pair. Hence the `<property_name>:<value>` format.

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
dli357: Gotcha, since the field [accepts a space-delimited list of protocols](https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/package-summary.html), can we try adding `TLSv1.2 TLSv1.3` as default?

dli357: Discussed offline, but we might also need to set these configurations for `mail.smtp.ssl.protocols` for use in startTLS as well, thanks!

sanjanasandeep: yeah I am running that test rn.

core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java
```diff
@@ -192,6 +211,12 @@ public static class Config extends ConditionConfig {
     @Description("Whether to include the contents of the workflow token in the email message. Defaults to false.")
     @Macro
     private Boolean includeWorkflowToken;
+    
+    @Nullable
+    @Description("Optional property that can be used to pass specific set of javamail properties. The syntax " +
+            "is '<property_name>:<value>'. If this property is not set, default values are chosen.")
```
dli357, line 217: nit: Maybe we should document what the default values are?

dli357 (approved):

PR #1609, opened 2022-04-29
[PLUGIN-1156] Update schema nullable issue.

JessieYee:
Solve problem when Schema is a nullable long.
Jira: https://cdap.atlassian.net/browse/PLUGIN-1156
Update from PR: https://github.com/cdapio/hydrator-plugins/pull/1602/files

### DIFF
--- a/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
@@ -192,7 +192,9 @@ public class DBRecord implements Writable, DBWritable, Configurable, DataSizeRep
     } else if (o instanceof BigInteger) {
       Schema schema = field.getSchema();
       BigInteger bigint = ((BigInteger) o);
-
+      if (schema.isNullable()) {
+        schema = schema.getNonNullable();
+      }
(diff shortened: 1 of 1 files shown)


database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
```diff
@@ -192,7 +192,9 @@ private void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilde
     } else if (o instanceof BigInteger) {
       Schema schema = field.getSchema();
       BigInteger bigint = ((BigInteger) o);
-
+      if (schema.isNullable()) {
```
seanzhougoogle, line 195: nit : why using different code style in two place ?

seanzhougoogle (approved):

database-plugins/src/main/java/io/cdap/plugin/DBRecord.java
JessieYee: Under line 209, we use this style to check if Schema is nullable or not. So I want to stay the same style here.

seanzhougoogle: got it , sg :)

PR #1618, opened 2022-05-05
e2e DateTransform Plugin related testcases

Swati10-11:
Please review @itsankit-google @rmstar 

@sawantpritam


e2e-test/src/test/resources/pluginDataCyAttributes.properties
```diff
@@ -0,0 +1,31 @@
...
+updateTableSchema=switch-allowSchemaRelaxation
+updateTableSchemaMacroInput=allowSchemaRelaxation
+outputSchemaMacroInput=Output Schema-macro-input
+path=path
+pathSuffix=suffix
+format=select-format
+formatMacroInput=format
+sampleSize=sampleSize
+delimiter=delimiter
+skipHeader=switch-skipHeader
```
rmstar: please remove configs that are not relevant to date transform plugin

e2e-test/src/test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,18 @@
+projectId=cdf-athena
+dataset=test_automation
+csvFormat=csv
```
rmstar: please remove irrelevant configs

rmstar (commented):
Date transform plugin needs to be downloaded from the hub. How are we handling that step when running using Github actions?

e2e-test/src/test/resources/pluginDataCyAttributes.properties
Swati10-11: Removed

e2e-test/src/test/resources/pluginParameters.properties
Swati10-11: Removed

itsankit-google:
> Date transform plugin needs to be downloaded from the hub. How are we handling that step when running using Github actions?

Using the `run_e2e_test` script, we deploy the latest build plugin, why will we need to download it from hub?
https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/scripts/run_e2e_test.py#L70

itsankit-google (commented):
These tests should be added in respective plugin modules instead of the parent? @ksahilCF

Swati10-11:
New PR raised
https://github.com/data-integrations/date-transform/pull/15

PR #1619, opened 2022-05-06
Added feature flags for relational transform execution

fernst:


core-plugins/pom.xml
```diff
@@ -42,15 +42,15 @@
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
-      <artifactId>hydrator-test</artifactId>
+      <artifactId>cdap-etl-api</artifactId>
     </dependency>
```
sanjanasandeep: why this?

sanjanasandeep (approved):

sanjanasandeep (approved):

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/GroupByRelationalTest.java
```diff
@@ -54,6 +54,7 @@ public class GroupByRelationalTest {
   @Before
   public void setUp() {
     Mockito.doReturn(engine).when(relationalTranformContext).getEngine();
+    Mockito.doReturn(true).when(relationalTranformContext).isFeatureEnabled(Mockito.anyString());
     Mockito.doReturn(Optional.of(expressionFactory)).when(engine).getExpressionFactory(Mockito.any(),
```
sanjanasandeep, line 58: we don't need this for `dedup aggregator test`?

core-plugins/pom.xml
fernst: oops, fixed

core-plugins/src/test/java/io/cdap/plugin/batch/aggregator/GroupByRelationalTest.java
fernst: We are not actually testing the `transform` method for the Dedup aggregator. We only test the logic to generate the definition.

fernst:
Solved on the Google-Cloud side

PR #1623, opened 2022-05-09
Updated property used for skewed input stage adjustment

fernst:


sechegaray (approved):

core-plugins/docs/Joiner-batchjoiner.md
```diff
@@ -90,32 +90,27 @@ to the number of partitions used for the input that is not loaded into memory.
 If no value is given, or if an inner advanced join is being performed, the number of partitions will be determined
 by the value of 'spark.sql.shuffle.partitions' in the engine config, which defaults to 200. 
 
+**Skewed Input Stage:**  Name of the skewed input stage. The skewed input stage is the one that contains many rows that join
+to the same row in the non-skewed stage. Ex. If stage A has 10 rows that join on the same row in stage B, then
```
bdmogal: nit: in the example, could we make it more like a million instead of 10? 

Also, s/ex/eg/

fernst: Changed.

PR #1626, opened 2022-05-10
[CDAP-19266] tied get schema button with format instead of delimiter

JessieYee:
File Source not showing `Get Schema` button. This Button is only visible when delimited format is selected. Ideally `Get Schema` should be allowed for every format.

Related PR: https://github.com/data-integrations/google-cloud/pull/981
Jira: https://cdap.atlassian.net/browse/CDAP-19266

### DIFF
--- a/core-plugins/widgets/File-batchsource.json
+++ b/core-plugins/widgets/File-batchsource.json
@@ -28,17 +28,11 @@
           "name": "format",
           "widget-attributes": {
             "plugin-type": "validatingInputFormat"
-          },
-          "plugin-function": {
-            "method": "POST",
-            "widget": "outputSchema",
-            "label": "Get Schema Value",
-            "required-fields": [
-              "path"
-            ],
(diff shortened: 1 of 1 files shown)


sumengwang (approved):

core-plugins/widgets/File-batchsource.json
```diff
@@ -28,6 +28,16 @@
           "name": "format",
           "widget-attributes": {
             "plugin-type": "validatingInputFormat"
+          },
+          "plugin-function": {
```
seanzhougoogle: I think it should not be tied to any widget, try using format like below ?
https://github.com/data-integrations/database-plugins/blob/fabc900f9dde17b11c7d5a884f9fc52e5289f037/mysql-plugin/widgets/Mysql-batchsource.json#L106

seanzhougoogle (approved):

itsankit-google (approved):
LGTM :)

PR #1637, opened 2022-05-18
e2e send-email test scenarios

muminmmd:
@rmstar @itsankit-google Please review.

CC: @sawantpritam


e2e-test/src/test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,68 @@
...
+sendEmailCompletion=completion
+sendEmailSuccess=success
+sendEmailFailure=failure
+sendEmailSender=testsendemailalert@gmail.com
+sendEmailInvalidSender=ymlsing
+sendEmailRecipients=testsendemailalert@gmail.com
+sendEmailInvalidRecipients=incorrectrecp
+sendEmailMessage=test case
+sendEmailUsername=testsendemailalert@gmail.com
+sendEmailPassword=dummyPassword
```
rmstar: we shouldn't store passwords in plaintext

e2e-test/src/test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * GCP test hooks.
+ */
+public class TestSetupHooks {
+  private static boolean firstFileSourceTestFlag = true;
+  private static boolean firstFileSinkTestFlag = true;
+
+  @Before(order = 1, value = "@FILE_SOURCE_TEST")
+  public static void setFileSourceAbsolutePath() {
```
rmstar, line 39: are we using file source plugin in these tests? If not, please remove irrelevant code

rmstar (commented):
- How are we verifying that email gets sent successfully?
- What happens if email is not sent successfully? Is there an error?

muminmmd:
> * How are we verifying that email gets sent successfully?
> * What happens if email is not sent successfully? Is there an error?

> * We are going to validate it from the pipeline logs.  Bug has been raised <PLUGIN-1255> as currently it is not getting logged in pipeline logs.

> * Test scenario updated, when email is not sent successfully.

e2e-test/src/test/resources/pluginParameters.properties
muminmmd: Password updated in Environment variables.

e2e-test/src/test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
muminmmd: Yes, we are using file source plugin.

e2e-test/src/test/features/sendemail/source/MacrosSendEmailAlert.feature
```diff
@@ -0,0 +1,67 @@
...
+    And Enter runtime argument value "sendEmailPassword" for key "password"
+    And Enter runtime argument value "sendEmailProtocol" for key "protocol"
+    And Enter runtime argument value "sendEmailHost" for key "host"
+    And Enter runtime argument value "sendEmailPort" for key "port"
+    And Enter runtime argument value "sendEmailIncludeWorkflowToken" for key "token"
+    And Run the Pipeline in Runtime with runtime arguments
+    And Wait till pipeline is in running state
+    And Open and capture logs
+    And Verify the pipeline status is "Succeeded"
+    #And Verify email sent message in logs (Implement this step once PLUGIN-1255 is fixed.)
```
rmstar, line 65: I don't think it's expected that we would log a message when email is sent successfully. 
A RuntimeException is thrown in the failure case - does this show up in the pipeline logs when there is an error in sending email? https://github.com/cdapio/hydrator-plugins/blob/develop/core-plugins/src/main/java/io/cdap/plugin/batch/action/EmailAction.java#L154

PR #1641, opened 2022-06-16
CDAP-19078 refactor schema detection

albertshau:
Refactor schema detection so that FileSystem operations occur
in the file source instead of in the format.

This ensures that custom classes (like token providers)
that are visible only to the file source can be instantiated when
traversing the file system.

It also is a cleaner approach, as formats should only be concerned
with how data is formatted, and not with where that data comes from.
It removes the need for the format plugin to read plugin properties
that it shouldn't know anything about, such as the format name,
file path, etc.

Parquet is handled in a hacky way for the moment, as it requires
a seekable input stream in order to read schema information from
the file footer, and requires the length of the file.


albertshau:
Needed to upgrade the parquet and avro libraries, which ended up needing a bunch of other pom changes.

The important part of this change is that the part that iterates through files in the FileSystem to detect schema is moved from the formats into a SchemaDetector class that the AbstractFileSource and AbstractFileConnector both call.

core-plugins/src/test/java/io/cdap/plugin/transform/StructuredtoAvroTest.java
```diff
@@ -97,7 +97,6 @@ public void testOutputSchemaUsage() throws Exception {
     GenericRecord result = avroTransformer.transform(record);
     Assert.assertEquals(123L, result.get("id"));
     Assert.assertEquals("ABC", result.get("name"));
-    Assert.assertNull(result.get("age"));
```
yaojiefeng, line 100: is this change needed?

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
```diff
@@ -115,21 +113,27 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
...
-    }
-
     ValidatingInputFormat validatingInputFormat =
       pipelineConfigurer.usePlugin(ValidatingInputFormat.PLUGIN_TYPE, fileFormat, fileFormat, builder.build());
     FormatContext context = new FormatContext(collector, null);
     validateInputFormatProvider(context, fileFormat, validatingInputFormat);
+    validatePathField(collector, schema);
 
-    if (validatingInputFormat != null && shouldGetSchema()) {
+    if (validatingInputFormat != null && schema == null && !config.containsMacro("schema")) {
```
yaojiefeng, line 121: here we are skipping the schema detection if schema is not null. I remember UI will always have default schema with text format, so we are relying on the fact that UI does not pass in the schema for getSchema to work. Some of our db sources always get schema no matter the schema is passed in or not, is there a reason we don't always get schema here? Is it because we want to let user fix the schema if our schema detection is somehow wrong(since we only do a small sample to get the schema)?

format-common/src/main/java/io/cdap/plugin/format/SchemaDetector.java
```diff
@@ -0,0 +1,160 @@
...
+                                                     regexPathFilter));
+  }
+
+  /**
+   * Supplies an InputStream of data to detect schema from.
+   *
+   * It implements Function as a hack to pass the file length down to the Parquet format, which needs it to seek to
+   * the file footer.
+   */
+  private static class FileStreamSupplier implements Supplier<InputStream>, Function<Void, Long> {
```
yaojiefeng: It looks strange to have this class extends `Function` and the input is Void. Is it better to just make this class public, so in format-parquent, you don't need to check if the Supplier is a Function, and call `apply(null)` to get the length?

format-common/src/main/java/io/cdap/plugin/format/SchemaDetector.java
```diff
@@ -0,0 +1,160 @@
...
+    String regexPathFilter = null;
+    if ("avro".equalsIgnoreCase(formatName)) {
+      regexPathFilter = ".+\\.avro$";
+    } else if ("parquet".equalsIgnoreCase(formatName)) {
+      regexPathFilter = ".+\\.parquet$";
+    }
+
+    Path fsPath = new Path(path);
+    // need this to load the extra class loader to avoid ClassNotFoundException for the file system
+    FileSystem fs = JobUtils.applyWithExtraClassLoader(job, getClass().getClassLoader(),
```
yaojiefeng, line 78: is this still needed if this happens on the source?

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -72,49 +75,53 @@ protected void addFormatProperties(Map<String, String> properties) {
...
-    List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job);
-    for (Path file : paths) {
-      try (ParquetReader reader = AvroParquetReader.builder(file).withConf(hconf).build()) {
-        GenericData.Record record = (GenericData.Record) reader.read();
-        if (record == null) {
-          continue;
+    @Override
+    public SeekableInputStream newStream() {
+      // hack, ValidatingInputFormat.detectSchema() should be updated to provide a seekable stream
+      FSDataInputStream inputStream = (FSDataInputStream) streamSupplier.get();
```
yaojiefeng: I am confused about this cast, does this supplier suppose to give `FileStreamSupplier`? Why can it get cast to `FSDataInputStream`?

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
```diff
@@ -72,49 +75,53 @@ protected void addFormatProperties(Map<String, String> properties) {
...
 
-    List<Path> paths = conf.getFilePathsForSchemaGeneration(filePath, ".+\\.parquet$", hconf, job);
-    for (Path file : paths) {
-      try (ParquetReader reader = AvroParquetReader.builder(file).withConf(hconf).build()) {
-        GenericData.Record record = (GenericData.Record) reader.read();
-        if (record == null) {
-          continue;
+    @Override
+    public SeekableInputStream newStream() {
+      // hack, ValidatingInputFormat.detectSchema() should be updated to provide a seekable stream
```
yaojiefeng: Any reason we can't do that?

format-parquet/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
```diff
@@ -54,17 +78,33 @@ public class AvroSchemaConverter {
 
   private final boolean assumeRepeatedIsListElement;
   private final boolean writeOldListStructure;
+  private final boolean writeParquetUUID;
+  private final boolean readInt96AsFixed;
 
   public AvroSchemaConverter() {
-    this.assumeRepeatedIsListElement = ADD_LIST_ELEMENT_RECORDS_DEFAULT;
+    this(ADD_LIST_ELEMENT_RECORDS_DEFAULT);
```
yaojiefeng, line 85: This contains a lot of complex logic for the schema conversion, do we have unit tests for this?

yaojiefeng (commented):
Some questions about the parquet file schema detection.

core-plugins/src/test/java/io/cdap/plugin/transform/StructuredtoAvroTest.java
albertshau: with the change in avro version, it actually results in an exception being thrown instead of a null when you try to get a field that isn't in the schema.

Can you think of any problems with this? This is incompatible behavior in general, but I don't think it should be a problem in the formats.

format-common/src/main/java/io/cdap/plugin/format/SchemaDetector.java
albertshau: I think it doesn't work because of classloading. The format isn't able to cast it directly because it will be the FileStreamSupplier from the File plugin.

I will try it again (did this part a while ago) to confirm, and add a comment if it can't be cast directly.

format-common/src/main/java/io/cdap/plugin/format/SchemaDetector.java
albertshau: good question, probably not. Will double check

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
albertshau: I need to cherry pick this onto the release branch, so can't update the API there. Can make the change in CDAP afterwards.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
albertshau: the supplier here is just a `Supplier<InputStream>`, which is a FSDataInputStream underneath

format-parquet/src/main/java/org/apache/parquet/avro/AvroSchemaConverter.java
albertshau: this is just copy-pasted from the newer avro version, so relying on them for correctness. We do have some parquet tests that cover the basics.

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
albertshau: Is the schema not always null when the 'get schema' button is used?

If not, how is the plugin tell the difference between this is a schema that should be overridden by detection vs something the user wants to use? For example, with the delimited formats, the user will normally rename the columns to something more sensible.

albertshau: For DBs, we had similar issues where for compatibility reasons, the plugin needs to be able to read some columns as a different type. For example, before the datetime type was added, it would read those columns as strings. So we needed to add some logic that allowed somebody to specify a schema where datetime columns are strings.

albertshau: On a side note, I don't think the existing behavior works with the path field property either. Users are supposed to be able to add fields for the file path.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
albertshau: Though it's not great practice, I'm thinking it may actually be better to just add in in 6.7.1 since nothing was using it in 6.7.0. Would avoid a bunch of complications around compatibility

core-plugins/src/test/java/io/cdap/plugin/transform/StructuredtoAvroTest.java
yaojiefeng: 👍  Don't think there will be a problem as long as the schema set is correct

format-common/src/main/java/io/cdap/plugin/format/plugin/AbstractFileSource.java
yaojiefeng: 👍  I think when get schema is hit, the schema will be null from UI. I just want to understand the use case here.

format-parquet/src/main/java/io/cdap/plugin/format/parquet/input/ParquetInputFormatProvider.java
yaojiefeng: I still don't get where do we actually provide this as `FSDataInputStream`, am I missing anything here?

albertshau: The FileStreamSupplier implements this as FileSystem.open(), which returns a FSDataInputStream

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingInputFormatProvider.java
```diff
@@ -21,7 +21,9 @@
 import io.cdap.cdap.etl.api.validation.FormatContext;
 import io.cdap.cdap.etl.api.validation.ValidatingInputFormat;
 
+import java.util.ArrayList;
```
yaojiefeng: nit - unused?

format-common/src/main/java/io/cdap/plugin/format/input/PathTrackingInputFormatProvider.java
```diff
@@ -21,7 +21,9 @@
...
 import java.util.HashMap;
+import java.util.List;
```
yaojiefeng: nit - unused?

format-common/src/main/java/io/cdap/plugin/format/HadoopInputFile.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+
+import java.io.IOException;
+
+/**
+ * InputFile that comes from a Hadoop FileSystem.
+ */
+public class HadoopInputFile implements InputFile {
```
yaojiefeng: nit - to me it is more natural to call this `FileSystemInputFile` since we use `FileSystem` to get the files back, `Hadoop` might make people think this is specifically related to hdfs

format-common/src/main/java/io/cdap/plugin/format/HadoopInputFiles.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Supplies an InputStream of data to detect schema from.
+ *
+ * It implements Function as a hack to pass the file length down to the Parquet format, which needs it to seek to
+ * the file footer.
+ */
+public class HadoopInputFiles implements InputFiles {
```
yaojiefeng: Same here - feel it is better to call this `FileSystemInputFiles`

format-common/src/main/java/io/cdap/plugin/format/HadoopSeekableInputStream.java
```diff
@@ -0,0 +1,44 @@
...
+
+import io.cdap.cdap.etl.api.validation.DelegatingSeekableInputStream;
+import org.apache.hadoop.fs.FSDataInputStream;
+
+import java.io.IOException;
+
+/**
+ * InputFile that comes from a Hadoop FileSystem.
+ */
+public class HadoopSeekableInputStream extends DelegatingSeekableInputStream {
```
yaojiefeng: Same here - feel it is better to call this `FileSystemSeekableInputStream`

yaojiefeng (approved):
minor nit, LGTM

format-common/src/main/java/io/cdap/plugin/format/SchemaDetector.java
albertshau: This is still required actually, the conf's classloader still doesn't have access to something like GoogleHadoopFileSystem

PR #1645, opened 2022-06-24
Add e2e test for Joiner plugin

jockech:
add e2e test for joiner and file plugins.

Pipeline:
image

Since there is a [bug](https://cdap.atlassian.net/browse/CDAP-19382) for running CDAP on K8S, the pipeline hasn't been tested on K8S and only on 6.8.0-snapshot sandbox and the result is below:

image

### DIFF
--- a/core-plugins/src/e2e-test/features/joiner/Joiner.feature
+++ b/core-plugins/src/e2e-test/features/joiner/Joiner.feature
@@ -1,0 +1,79 @@
+#
+# Copyright © 2022 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+@Joiner
+Feature: Joiner analytics - Verify File source data transfer using Joiner analytics
+  @GCS_SOURCE_TEST @GCS_SOURCE_JOINER_TEST @GCS_SINK_TEST
+  Scenario: To verify data is getting transferred from File source to File sink plugin successfully with Joiner
+    Given Open Datafusion Project to configure pipeline
+    When Select plugin: "File" from the plugins list as: "Source"
+    When Select plugin: "File" from the plugins list as: "Source"
+    Then Move plugins: "File2" by xOffset 0 and yOffset 200
+    When Expand Plugin group in the LHS plugins list: "Analytics"
+    When Select plugin: "Joiner" from the plugins list as: "Analytics"
+    Then Connect plugins: "File" and "Joiner" to establish connection
+    Then Connect plugins: "File2" and "Joiner" to establish connection
+    When Expand Plugin group in the LHS plugins list: "Sink"
+    When Select plugin: "File" from the plugins list as: "Sink"
+    Then Connect plugins: "Joiner" and "File3" to establish connection
+    Then Navigate to the properties page of plugin: "File"
+    Then Enter input plugin property: "referenceName" with value: "firstName"
+    Then Enter input plugin property: "path" with value: "gcsSourceBucket1"
+    Then Select dropdown plugin property: "format" with option value: "csv"
+    Then Click plugin property: "skipHeader"
+    Then Click plugin property: "enableQuotedValues"
+    Then Click on the Get Schema button
+    Then Verify the Output Schema matches the Expected Schema: "firstNameOutputSchema"
+    Then Validate "File" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "File2"
+    Then Enter input plugin property: "referenceName" with value: "lastName"
+    Then Enter input plugin property: "path" with value: "gcsSourceBucket2"
+    Then Select dropdown plugin property: "format" with option value: "csv"
+    Then Click plugin property: "skipHeader"
+    Then Click plugin property: "enableQuotedValues"
+    Then Click on the Get Schema button
+    Then Verify the Output Schema matches the Expected Schema: "lastNameOutputSchema"
+    Then Validate "File" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "Joiner"
+    Then Expand fields
+    Then Uncheck plugin "File2" field "id" alias checkbox
+    Then Select joiner type "Inner"
+    Then Enter numPartitions 1
+    Then Select dropdown plugin property: "inputsToLoadMemory" with option value: "File"
+    Then Scroll to validation button and click
+    Then Validate "Joiner" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "File3"
+    Then Enter input plugin property: "referenceName" with value: "result"
+    Then Enter input plugin property: "path" with value: "gcsTargetBucket"
+    Then Select dropdown plugin property: "format" with option value: "csv"
+    Then Close the Plugin Properties page
+    Then Save the pipeline
+    Then Preview and run the pipeline
+    Then Wait till pipeline preview is in running state
+    Then Open and capture pipeline preview logs
+    Then Verify the preview run status of pipeline in the logs is "succeeded"
+    Then Close the pipeline logs
+    Then Close the preview
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Verify the CSV Output File matches the Expected Output File With Expected Partition
+    Then Close the pipeline logs
--- a/core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
+++ b/core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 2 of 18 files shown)


core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
```diff
@@ -228,7 +228,7 @@ public void testOuterJoin() throws Exception {
       Schema.Field.of("renamed_category", Schema.nullableOf(Schema.of(Schema.Type.STRING))));
 
     ETLStage joinStage =
-      new ETLStage("joiner",
+      new ETLStage("io/cdap/plugin/file/joiner",
```
rmstar: why are we changing unit tests? isn't this unrelated to e2e tests?

core-plugins/src/e2e-test/java/io/cdap/plugin/joiner/stepsdesign/Joiner.java
```diff
@@ -0,0 +1,40 @@
...
+ * the License.
+ */
+package io.cdap.plugin.joiner.stepsdesign;
+
+import io.cdap.e2e.utils.CdfHelper;
+import io.cdap.plugin.joiner.actions.JoinerActions;
+import io.cucumber.java.en.Then;
+
+/**
+ *
```
rmstar: nit: add comment text

core-plugins/src/e2e-test/java/io/cdap/plugin/joiner/locators/JoinerLocators.java
```diff
@@ -0,0 +1,40 @@
...
+package io.cdap.plugin.joiner.locators;
+
+import io.cdap.e2e.utils.SeleniumDriver;
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.support.FindBy;
+import org.openqa.selenium.support.How;
+
+/**
+ *
```
rmstar: nit: add comment text

core-plugins/src/e2e-test/java/io/cdap/plugin/joiner/actions/JoinerActions.java
```diff
@@ -0,0 +1,42 @@
...
+ * the License.
+ */
+package io.cdap.plugin.joiner.actions;
+
+import io.cdap.e2e.utils.ElementHelper;
+import io.cdap.e2e.utils.SeleniumHelper;
+import io.cdap.plugin.joiner.locators.JoinerLocators;
+
+/**
+ *
```
rmstar: nit: add comment text

core-plugins/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,11 @@
+projectId=odf-integration-test
+gcsCsvFile1=testdata/first_name.csv
+gcsCsvFile2=testdata/last_name.csv
+gcsSourceBucket1=
```
rmstar: When will bucket params get filled in? This test will with github actions on PRs, right?

core-plugins/src/e2e-test/features/joiner/Joiner.feature
```diff
@@ -0,0 +1,75 @@
...
+    Then Open and capture pipeline preview logs
+    Then Verify the preview run status of pipeline in the logs is "succeeded"
+    Then Close the pipeline logs
+    Then Close the preview
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
```
rmstar, line 79: can we verify that the output written to the file sink is correct?

rmstar (commented):
please remove changes wrangler-transform, google-cloud submodules

core-plugins/src/e2e-test/resources/pluginParameters.properties
jockech: It will be filled during the [TestSetupHooks](https://github.com/cdapio/hydrator-plugins/pull/1645/files/e6a093058ecf689af2221190cb2835d13502f8fa#diff-00ff67eba38f39d3a7e4a2c62ef6b2fd6bd9cdc31ecf10a265bf5e426cd13adfR60)

core-plugins/src/e2e-test/features/joiner/Joiner.feature
jockech: good catch. But the data in GCS will be partitioned and hard to verify even if I [set the number of partitions to 1](https://github.com/cdapio/hydrator-plugins/pull/1645/files/e6a093058ecf689af2221190cb2835d13502f8fa#diff-395a661e25a14cf563e72940dad92a122606550bae42774737d4a8bb3f89bae2R55) in `Joiner`

core-plugins/src/e2e-test/resources/pluginParameters.properties
rmstar: got it, thanks!

core-plugins/src/test/java/io/cdap/plugin/batch/joiner/JoinerTestRun.java
jockech: It's unrelated to e2e tests. I haven't changed it and I guessed it's due to running some maven command and gets changed automatically. will revert these changes.

core-plugins/src/e2e-test/features/joiner/Joiner.feature
rmstar: ah I see. is there a reason we want to write to GCS sink? If you write to BQ you it will be easier verify

jockech: The main reason is @albertshau has mentioned not to mix plugins up in the e2e test. We don't want any `google-cloud plugins` appear in e2e tests in `hydrator-plugin/core-plugin` repo. And I haven't found a suitable sink plugin for this case.

core-plugins/src/e2e-test/resources/pluginParameters.properties
itsankit-google: Can we please add a comment?

core-plugins/src/e2e-test/features/joiner/Joiner.feature
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+@Joiner
+Feature: Joiner analytics - Verify File source data transfer using Joiner analytics
+  @HDF_FILE_SOURCE_TEST @HDF_FILE_SOURCE_JOINER_TEST @HDF_FILE_SINK_TEST @HDF_FILE_SINK_JOINER_TEST
```
itsankit-google: Why do we have `HDF` prefix in these tags? Are they specific to `HDF`?

jockech: Since only `HDF` ones need to have a GCS bucket otherwise we could just use local file for testing.

itsankit-google: Got it, thanks for the info.

May be we should then create a new `Test Runner` which can run tests specific to HDF because once we start adding more tests using local file they will start failing in `HDF`.

`Test Runner` can be made configurable through `pom.xml`.
See this PR for ref - https://github.com/data-integrations/google-cloud/pull/1067

core-plugins/src/e2e-test/features/joiner/Joiner.feature
itsankit-google: Can we compose all the files in GCS under the particular destination bucket and then read and verify?
https://stackoverflow.com/questions/24187243/how-to-use-compose-on-gcs-using-the-java-client

core-plugins/src/e2e-test/features/joiner/Joiner.feature
jockech: But for local file, they can use `@FILE_SOURCE or @FILE_SINK` as the tag in scenario.

jockech: After a second thought, I think file on GCS could also be used for sandbox testing. I have changed the tag name to `GCS_xxx` in the end.

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -0,0 +1,155 @@
...
+      // ----part-r-0000
+      // ----part-r-0001....
+      // Since the number of partition is set to 1 in Joiner plugin, only part-r-0000 and _SUCCESS are expected.
+      for (Blob blob : StorageClient.listObjects(gcsTargetBucketName).iterateAll()) {
+        String name = blob.getName();
+        if (name.contains("part-r")) {
+          partitions++;
+        }
+        if (name.contains("part-r-00000")) {
+          content = new String((blob.getContent()));
```
itsankit-google: To make it more generic, does it make sense to combine the data of all `part-r-*` files? Suppose `expectedPartitions` is set to `2` then it will fail.

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -0,0 +1,154 @@
...
+      for (Blob blob : StorageClient.listObjects(gcsTargetBucketName).iterateAll()) {
+        String name = blob.getName();
+        if (name.contains("part-r")) {
+          partitions++;
+          sb.append(new String((blob.getContent())));
+        }
+      }
+      Assert.assertEquals("Output partition should match",
+                          partitions, Integer.parseInt(PluginPropertyUtils.pluginProp("expectedPartitions")));
+      Assert.assertTrue("Output content should match", Strings.equals(sb.toString(), expectedOutput));
```
itsankit-google: on a second thought, are we handling the case where data can be shuffled by spark??

jockech: I think so but I haven't taken a close look at corresponding logic and it seems the output file is ordered according to my recent test after creating 200 partitions, some of which are empty.

jockech: Added the logic for sorting the output.

itsankit-google (approved):
lgtm

PR #1651, opened 2022-06-28
CDAP-19078 refactor schema detection

albertshau:
Refactor schema detection so that FileSystem operations occur
in the file source instead of in the format.

This ensures that custom classes (like token providers)
that are visible only to the file source can be instantiated when
traversing the file system.

It also is a cleaner approach, as formats should only be concerned
with how data is formatted, and not with where that data comes from.
It removes the need for the format plugin to read plugin properties
that it shouldn't know anything about, such as the format name,
file path, etc.

Parquet is handled in a hacky way for the moment, as it requires
a seekable input stream in order to read schema information from
the file footer, and requires the length of the file.

The parquet and avro libraries needed to be upgraded in order to
use new parquet reader classes that allow reading a more generic
seekable input stream instead of a Hadoop specific stream. This
upgrade in libraries also forced several other pom changes to add
some more excludes to ensure Spark unit tests run and plugins
can be packaged properly.


albertshau:
cherry-pick of #1641

PR #1658, opened 2022-07-08
e2e tests folder structure and workflow setup for core-plugins

sawantpritam:
@itsankit-google @rmstar @tivv  Please review.

With current approach using GitHub workflows built-in path filters, we will have to duplicate the yml file for each module as this built-in path filters don't work on a level of individual steps.
https://github.com/cdapio/hydrator-plugins/blob/9d1042e8ece355111488404fba79120bdfde4fe2/.github/workflows/e2e.yml#L20

Better approach is using dorny/paths-filter https://github.com/marketplace/actions/paths-changes-filter , using it we can add  only separate steps as per module path. No need to duplicate the yml files.
But currently dorny/paths-filter@v2 is not allowed to be used in cdapio/hydrator-plugins.
Can we add this action in allowed actions list ?


sawantpritam:
> @itsankit-google @rmstar @tivv Please review.
> 
> With current approach using GitHub workflows built-in path filters, we will have to duplicate the yml file for each module as this built-in path filters don't work on a level of individual steps.
> 
> https://github.com/cdapio/hydrator-plugins/blob/9d1042e8ece355111488404fba79120bdfde4fe2/.github/workflows/e2e.yml#L20
> 
> Better approach is using dorny/paths-filter https://github.com/marketplace/actions/paths-changes-filter , using it we can add only separate steps as per module path. No need to duplicate the yml files. But currently dorny/paths-filter@v2 is not allowed to be used in cdapio/hydrator-plugins. Can we add this action in allowed actions list ?

@tivv Updated yml file to use dorny/paths-filter
Please whitelist dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721

sawantpritam:
Conflicts resolved.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
...
+      - uses: actions/checkout@v2.3.4
+        with:
+          path: plugin
+      - uses: dorny/paths-filter@b2feaf19c27470162a626bd6fa8438ae5b263721
+        id: filter
+        with:
+          working-directory: plugin
+          filters: |
+            core-plugins:
+              - 'core-plugins/**'
```
itsankit-google, line 53: We are depending on changes in the modules within the repo, that makes sense for PRs. What will be the output of this filter in case of manual trigger(workflow_dispatch)?

sawantpritam: in case of workflow_dispatch changes will be detected from last commit.

sawantpritam: Updated workflow for manual trigger. Please review.

.github/workflows/e2e_manual_run.yml
```diff
@@ -0,0 +1,72 @@
...
+#  License for the specific language governing permissions and limitations under
+#  the License.
+
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+# Note: Any changes to this workflow would be used only after merging into develop
+name: Build e2e tests
+
+on:
+  workflow_dispatch:
```
itsankit-google: instead of creating another workflow, can we add an additional step which is skipped for other type of events in the same workflow??

sawantpritam: I tried implementing conditional strategy matrix but did not get desired outcome thats why created separate workflow.

sawantpritam: I figured out solution. Added workflow_dispatch as well in same workflow file.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,123 @@
...
+    # 3) For PRs that are labeled as build and
+    #  - It's a code change
+    #  - A build label was just added
+    # A bit complex, but prevents builds when other labels are manipulated
+    if: >
+      github.event_name == 'workflow_dispatch'
+      || github.event_name == 'push'
+      || (contains(github.event.pull_request.labels.*.name, 'build')
+          && (github.event.action != 'labeled' || github.event.label.name == 'build')
+          )
```
itsankit-google, line 40: Can we use `trigger.yml` to trigger the workflow?
https://github.com/cdapio/hydrator-plugins/blob/develop/.github/workflows/trigger.yml
similar to https://github.com/cdapio/hydrator-plugins/blob/develop/.github/workflows/build.yml

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,123 @@
...
+          path: e2e
+      - name: Cache
+        uses: actions/cache@v2.1.3
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ github.workflow }}-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-${{ github.workflow }}
+      - name: Run required e2e tests
+        if: github.event_name != 'workflow_dispatch' && (!contains( github.event.head_commit.message, 'Merge pull request'))
```
itsankit-google: instead of `commit.message`,use `github.event_name == 'push'`.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,123 @@
...
+          destination: e2e-tests-cucumber-reports/${{ github.event.repository.name }}/${{ github.ref }}
+          glob: '**/target/cucumber-reports/**'
+      - name: github-status-action
+        uses: Sibz/github-status-action@67af1f4042a5a790681aad83c44008ca6cfab83d
+        if: always()
+        with:
+          authToken: ${{ secrets.GITHUB_TOKEN }}
+          state: success
+          context: Cucumber report
+          sha: ${{github.event.pull_request.head.sha || github.sha}}
```
itsankit-google: As we are not publishing cucumber reports back, can remove this step for now.

sawantpritam: Removed step.

.github/workflows/e2e.yml
sawantpritam: We can not use `github.event_name == 'push'` as it gets triggered for both normal push in open PR as well as for PR merge (PR merge pushes commit to the target branch it also triggers push event)

itsankit-google: The case which you described the event for that should be `synchronize` not `push`. Could you please test it once?

sawantpritam: Yes correct, that is `synchronize` event.
Updated condition.

Sorry I was confused  as commit gets pushed to branch from which PR is raised and on that branch's workflow it triggers push event.

itsankit-google (approved):

PR #1661, opened 2022-07-11
add e2e test for error collector plugin.

jockech:
1 the transform logic is to emit input to error collector if id is less than 10.
2 move the `verify` logic to a common place so it can be used by `Joiner` and `ErrorCollector` features.
image

### DIFF
--- a/core-plugins/src/e2e-test/features/errorcollector/ErrorCollector.feature
+++ b/core-plugins/src/e2e-test/features/errorcollector/ErrorCollector.feature
@@ -14,21 +14,22 @@
 # the License.
 #
 
-@Joiner
-Feature: Joiner analytics - Verify File source data transfer using Joiner analytics
-  @GCS_SOURCE_TEST @GCS_SOURCE_JOINER_TEST @GCS_SINK_TEST
-  Scenario: To verify data is getting transferred from File source to File sink plugin successfully with Joiner
+@ErrorCollector
+Feature: Error Collector - Verification of successful error emitted to ErrorCollector plugin.
+  @GCS_SOURCE_TEST @GCS_SINK_TEST
+  Scenario: To verify error is emitted to ErrorCollector plugin successfully.
     Given Open Datafusion Project to configure pipeline
     When Select plugin: "File" from the plugins list as: "Source"
-    When Select plugin: "File" from the plugins list as: "Source"
-    Then Move plugins: "File2" by xOffset 0 and yOffset 200
-    When Expand Plugin group in the LHS plugins list: "Analytics"
-    When Select plugin: "Joiner" from the plugins list as: "Analytics"
-    Then Connect plugins: "File" and "Joiner" to establish connection
-    Then Connect plugins: "File2" and "Joiner" to establish connection
+    When Expand Plugin group in the LHS plugins list: "Transform"
+    When Select plugin: "JavaScript" from the plugins list as: "Transform"
+    When Expand Plugin group in the LHS plugins list: "Error Handlers and Alerts"
+    When Select plugin: "Error Collector" from the plugins list as: "Error Handlers and Alerts"
+    Then Move plugins: "ErrorCollector" by xOffset 200 and yOffset 200
     When Expand Plugin group in the LHS plugins list: "Sink"
     When Select plugin: "File" from the plugins list as: "Sink"
-    Then Connect plugins: "Joiner" and "File3" to establish connection
+    Then Connect plugins: "File" and "JavaScript" to establish connection
+    Then Connect node type: "error" of plugin: "JavaScript" to plugin: "ErrorCollector"
+    Then Connect plugins: "ErrorCollector" and "File2" to establish connection
     Then Navigate to the properties page of plugin: "File"
     Then Enter input plugin property: "referenceName" with value: "firstName"
     Then Enter input plugin property: "path" with value: "gcsSourceBucket1"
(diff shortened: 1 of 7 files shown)


core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/CorePlugin.java
```diff
@@ -0,0 +1,91 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static io.cdap.plugin.common.stepsdesign.TestSetupHooks.gcsTargetBucketName;
```
itsankit-google: Is it needed? Can't we use `PluginPropertyUtils`?

core-plugins/src/e2e-test/features/errorcollector/ErrorCollector.feature
```diff
@@ -0,0 +1,74 @@
...
+    Then Open and capture pipeline preview logs
+    Then Verify the preview run status of pipeline in the logs is "succeeded"
+    Then Close the pipeline logs
+    Then Close the preview
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    # TODO: this line needs to change. where to put the code?
```
itsankit-google: Is this `TODO` still relevant?

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/CorePlugin.java
jockech: The verification logic only works for `core-plugin`. is it okay to move to PluginPropertyUtils and PipelineSteps in e2e test repo?

jockech: after a second thought, the test case are specific to these plugins and we should maintain the logic in this repo.

core-plugins/src/e2e-test/features/errorcollector/ErrorCollector.feature
jockech: any suggestion on where to include this step? I think it should be put in `common` package, but each package only allows one java class in this repo.

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/CorePlugin.java
itsankit-google: No, I didn't meant to move the verification logic.

Lemme clarify, what I mean is why do we need to import `gcsTargetBucketName`? Can't we use `PluginPropertyUtils` to retrieve the property?

jockech: Thanks for clarifying it.

core-plugins/src/e2e-test/features/joiner/Joiner.feature
```diff
@@ -75,5 +75,5 @@ Feature: Joiner analytics - Verify File source data transfer using Joiner analyt
     Then Wait till pipeline is in running state
     Then Open and capture logs
     Then Verify the pipeline status is "Succeeded"
-    Then Verify the CSV Output File matches the Expected Output File With Expected Partition
+    Then Verify the CSV Output File matches the Expected Output File: "joinerOutput" With Expected Partition: "expectedJoinerOutputPartitions"
```
rmstar: why are we making changes to joiner test here, seems unrelated to error collector?

rmstar (approved):

core-plugins/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -8,5 +8,19 @@ firstNameCsvFile=testdata/first_name.csv
...
 firstNameOutputSchema=[{"key":"id","value":"int"},{"key":"first_name","value":"string"}]
 lastNameOutputSchema=[{"key":"id","value":"int"},{"key":"last_name","value":"string"}]
+errorCollectorOutputSchema=[{"key":"id","value":"int"},{"key":"first_name","value":"string"},\
+  {"key":"msg","value":"string"},{"key":"code","value":"int"},{"key":"node","value":"string"}]
 joinerOutput=testdata/joiner_output.csv
-expectedPartitions=1
+errorCollectorOutput=testdata/error_collector_output.csv
+expectedJoinerOutputPartitions=1
+expectedErrorCollectorOutputPartitions=1
+transformScript=function transform(input, emitter, context) { \
```
itsankit-google: nit: I think it is specfic to `javascript transform`, can we rename it to `javascriptTransformScript`?

core-plugins/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -8,5 +8,19 @@ firstNameCsvFile=testdata/first_name.csv
 lastNameCsvFile=testdata/last_name.csv
 firstNameOutputSchema=[{"key":"id","value":"int"},{"key":"first_name","value":"string"}]
 lastNameOutputSchema=[{"key":"id","value":"int"},{"key":"last_name","value":"string"}]
+errorCollectorOutputSchema=[{"key":"id","value":"int"},{"key":"first_name","value":"string"},\
+  {"key":"msg","value":"string"},{"key":"code","value":"int"},{"key":"node","value":"string"}]
 joinerOutput=testdata/joiner_output.csv
-expectedPartitions=1
+errorCollectorOutput=testdata/error_collector_output.csv
+expectedJoinerOutputPartitions=1
+expectedErrorCollectorOutputPartitions=1
```
itsankit-google: Is error collector plugin also having some way to configure output partitions?

core-plugins/src/e2e-test/features/joiner/Joiner.feature
jockech: I have move the `verifyOutput` logic to a same place under [CorePlugin.java](https://github.com/cdapio/hydrator-plugins/blob/38f130df787f004360d3069286faae812395516a/core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/CorePlugin.java) so it can be reused by `ErrorCollector`. Since the partition of `ErrorCollector` is always 1 but partition of `Joiner` can change, I introduced another parameter `partition` to the function, which then changes the original `Joiner` feature step.

core-plugins/src/e2e-test/resources/pluginParameters.properties
jockech: don't think so
image

core-plugins/src/e2e-test/resources/pluginParameters.properties
jockech: nice catch!

core-plugins/src/e2e-test/resources/pluginParameters.properties
itsankit-google: For such cases, can we pass expected partitions as `0` indicating we are not sure then it will combine the data from all the output files and sort it and verify?

jockech: +1

itsankit-google (approved):
lgtm, pls squash commits before merge.

PR #1666, opened 2022-07-27
e2e-GroupBy Test Scenarios

itsmekumari:
@itsankit-google @rmstar Please review.
cc @sawantpritam


core-plugins/src/e2e-test/features/groupby/GroupByWithFile.feature
```diff
@@ -0,0 +1,154 @@
...
+    Then Close the preview
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Close the pipeline logs
+    Then Validate OUT record count of groupby is equal to IN record count of sink
+#    Enable below step once issue fixed (PLUGIN #1326)
+#    Then Validate output file generated by file sink plugin is equal to expected output file {string}
```
rmstar: We need to investigate why a large number of partitions are created - this will be done as part of PLUGIN-1326. However, your validation logic should remain the same - read contents of all the output files (part-r-xxx) and verify that the data written matches what you expect. Can you update the tests to do this?

rmstar (approved):
please squash commits

core-plugins/src/e2e-test/features/groupby/GroupByWithFile.feature
itsmekumari: Done.

PR #1670, opened 2022-08-02
CDAP-19461 fix schema detection with regex

albertshau:
Fix schema detection so that it properly uses the regex config
to filter out files that should be ignored.


albertshau:
cherry-pick of #1669

sechegaray (approved):

PR #1674, opened 2022-08-05
Change e2e tests run condition

sawantpritam:
@itsankit-google Please review.

Updated e2e tests run condition to execute e2e-tests for all modules irrespective of module changes.

If no change in e2e-tests inside module -> then execute only required tests
If change in e2e-tests inside module -> then execute all tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,8 +23,9 @@ on:
   workflow_dispatch:
 
 jobs:
-  modules:
-    runs-on: ubuntu-latest
+  build:
+    runs-on: k8s-runner-e2e
+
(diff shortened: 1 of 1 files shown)


itsankit-google (commented):
If there are changes inside tests of a particular module we need to execute all tests of that module & required tests of other module.

If there are no changes in e2e-test execute required tests of all module.

sawantpritam:
> If there are changes inside tests of a particular module we need to execute all tests of that module & required tests of other module.
> 
> If there are no changes in e2e-test execute required tests of all module.

Correct..implemented same logic.

itsankit-google (approved):

PR #1684, opened 2022-09-12
Enable adding FQN as a property for an asset

snehalakshmisha:
Jira: https://cdap.atlassian.net/browse/CDAP-19519

### DIFF
--- a/hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
+++ b/hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
@@ -15,25 +15,31 @@
  */
 package io.cdap.plugin.common;
 
+import javax.annotation.Nullable;
+
 /**
  * Represents a dataset with a FQN which is a fully-qualified unique identifier for a dataset
  * and the location of the dataset.
  */
 public class Asset {
 
-  public static final String DEFAULT_LOCATION = "unknown";
+  private static final String DEFAULT_LOCATION = "unknown";
 
+  private final String referenceName;
   private final String fqn;
   private final String location;
 
-  public Asset(String fqn) {
-      this.fqn = fqn;
-      this.location = DEFAULT_LOCATION;
+  private Asset(String referenceName, @Nullable String fqn, @Nullable String location) {
+    this.referenceName = referenceName;
+    this.fqn = fqn == null ? referenceName : fqn;
+    this.location = location == null ? DEFAULT_LOCATION : location;
   }
 
-  public Asset(String fqn, String location) {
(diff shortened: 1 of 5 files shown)


mrahanjam (approved):

hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
```diff
@@ -15,25 +15,33 @@
...
+  private static final String DEFAULT_LOCATION = "unknown";
 
+  private final String referenceName;
   private final String fqn;
   private final String location;
 
-  public Asset(String fqn) {
-      this.fqn = fqn;
-      this.location = DEFAULT_LOCATION;
+  private Asset(String referenceName, @Nullable String fqn, @Nullable String location) {
```
yaojiefeng, line 32: Can `referenceName` be null here? I think we will make it optional for some plugins right?

hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
```diff
@@ -15,25 +15,33 @@
  */
 package io.cdap.plugin.common;
 
+import java.util.HashMap;
```
yaojiefeng: are these used anywhere?

hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
snehalakshmisha: So this will be either the reference name or the normalized FQN.
The plugins will check if reference name is present or not - if yes it will send the same, if not it will send the normalized fqn.

hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
snehalakshmisha: Removed unused imports

hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
```diff
@@ -15,25 +15,31 @@
...
+    this.referenceName = referenceName;
+    this.fqn = fqn == null ? referenceName : fqn;
+    this.location = location == null ? DEFAULT_LOCATION : location;
   }
 
-  public Asset(String fqn, String location) {
-      this.fqn = fqn;
-      this.location = location;
+  /**
+   * @return the fully-qualified name of the {@link Asset}
```
albertshau: should update the javadoc, as this isn't the FQN

albertshau (approved):

PR #1690, opened 2022-09-21
Bump version to 2.9.2-SNAPSHOT

vsethi09:
Bump snapshot version to 2.9.2-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>hydrator-plugins</artifactId>
-  <version>2.9.1</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1030,7 +1030,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-javadoc-plugin</artifactId>
-            <version>2.9.1</version>
+            <version>2.9.2-SNAPSHOT</version>
```
sechegaray: I don't think we should bump this dependency. Was this a mistake?

vsethi09: Reverted

sechegaray (approved):

PR #1692, opened 2022-09-22
bump up submodule version to 2.9.2-SNAPSHOT to make it consistent with parent pom

seanzhougoogle:


pom.xml
```diff
@@ -1030,7 +1030,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-javadoc-plugin</artifactId>
-            <version>2.9.1</version>
+            <version>2.9.2-SNAPSHOT</version>
```
yaojiefeng, line 1033: Revert this change

yaojiefeng (approved):

sechegaray (approved):

pom.xml
seanzhougoogle: good catch

seanzhougoogle: Sebastian merged the PR by mistake and addressed this issue in another PR : https://github.com/cdapio/hydrator-plugins/pull/1693

seanzhougoogle: BTW, was looking into why this issue was not disclosed in my testing. it's because the plugin is configured in `release` profile,  maven won't complain if the profile is not used.

PR #1697, opened 2022-09-29
[CDAP-19665] FQN Generation in Generic DB PLugin

apoorvamk:
Changes: 
- Added FQN Generation for DB Connection Strings using [JDBC Parsing Library](https://github.com/opentracing-contrib/java-jdbc)
- Supported Databases: SQLServer, Postgresql, MySQL, Oracle
- Lineage Recording for DB Source
- Related Unit Tests
- Updated `Reference Name` label name in widget to `Table Name`.
- Jira:  https://cdap.atlassian.net/browse/CDAP-19665

### DIFF
--- a/database-plugins/pom.xml
+++ b/database-plugins/pom.xml
@@ -27,6 +27,10 @@
   <artifactId>database-plugins</artifactId>
   <modelVersion>4.0.0</modelVersion>
 
+  <properties>
+    <opentracingjdbc.version>0.2.15</opentracingjdbc.version>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
@@ -40,6 +44,11 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-jdbc</artifactId>
+      <version>${opentracingjdbc.version}</version>
     </dependency>
 
     <!-- test dependencies -->
--- a/database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/DBSink.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/DBSink.java
@@ -43,12 +43,14 @@ import io.cdap.cdap.etl.api.connector.Connector;
 import io.cdap.plugin.DBManager;
 import io.cdap.plugin.DBRecord;
 import io.cdap.plugin.FieldCase;
+import io.cdap.plugin.common.Asset;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.common.ReferenceBatchSink;
 import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.common.db.DBUtils;
 import io.cdap.plugin.db.batch.TransactionIsolationLevel;
 import io.cdap.plugin.db.common.DBBaseConfig;
+import io.cdap.plugin.db.common.FQNGenerator;
 import io.cdap.plugin.db.connector.DBConnector;
 import io.cdap.plugin.db.connector.DBConnectorConfig;
 import org.apache.hadoop.io.NullWritable;
@@ -137,7 +139,7 @@ public class DBSink extends ReferenceBatchSink<StructuredRecord, DBRecord, NullW
 
     Schema schema = context.getInputSchema();
     if (schema != null && schema.getFields() != null) {
-      recordLineage(context, dbSinkConfig.getReferenceName(), schema,
+      recordLineage(context, schema,
                     schema.getFields().stream().map(Schema.Field::getName).collect(Collectors.toList()));
     }
   }
@@ -215,8 +217,10 @@ public class DBSink extends ReferenceBatchSink<StructuredRecord, DBRecord, NullW
     }
   }
 
-  private void recordLineage(BatchSinkContext context, String outputName, Schema tableSchema, List<String> fieldNames) {
-    LineageRecorder lineageRecorder = new LineageRecorder(context, outputName);
+  private void recordLineage(BatchSinkContext context, Schema tableSchema, List<String> fieldNames) {
+    Asset asset = Asset.builder(dbSinkConfig.getReferenceName()).
+      setFqn(FQNGenerator.constructFQN(dbSinkConfig.getConnectionString(), dbSinkConfig.getReferenceName())).build();
+    LineageRecorder lineageRecorder = new LineageRecorder(context, asset);
     lineageRecorder.createExternalDataset(tableSchema);
     if (!fieldNames.isEmpty()) {
       lineageRecorder.recordWrite("Write", "Wrote to Database.", fieldNames);
--- a/database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
+++ b/database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
@@ -41,6 +41,7 @@ import io.cdap.plugin.DBManager;
 import io.cdap.plugin.DBRecord;
 import io.cdap.plugin.FieldCase;
 import io.cdap.plugin.StructuredRecordUtils;
+import io.cdap.plugin.common.Asset;
 import io.cdap.plugin.common.LineageRecorder;
 import io.cdap.plugin.common.ReferenceBatchSource;
 import io.cdap.plugin.common.ReferencePluginConfig;
@@ -49,6 +50,7 @@ import io.cdap.plugin.common.db.DBUtils;
 import io.cdap.plugin.common.db.DriverCleanup;
 import io.cdap.plugin.db.batch.TransactionIsolationLevel;
 import io.cdap.plugin.db.common.DBBaseConfig;
+import io.cdap.plugin.db.common.FQNGenerator;
 import io.cdap.plugin.db.connector.DBConnector;
 import io.cdap.plugin.db.connector.DBConnectorConfig;
 import org.apache.hadoop.conf.Configuration;
@@ -65,6 +67,7 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.List;
 import java.util.Properties;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -270,6 +273,12 @@ public class DBSource extends ReferenceBatchSource<LongWritable, DBRecord, Struc
     return DriverManager.getConnection(sourceConfig.getConnectionString(), properties);
   }
 
+  protected LineageRecorder getLineageRecorder(BatchSourceContext context) {
+    Asset asset = Asset.builder(sourceConfig.getReferenceName()).
+      setFqn(FQNGenerator.constructFQN(sourceConfig.getConnectionString(), sourceConfig.getReferenceName())).build();
+    return new LineageRecorder(context, asset);
+  }
+
   /**
    * {@link PluginConfig} for {@link DBSource}
    */
@@ -423,7 +432,7 @@ public class DBSource extends ReferenceBatchSource<LongWritable, DBRecord, Struc
     if (schema == null) {
       schema = context.getOutputSchema();
     }
-    LineageRecorder lineageRecorder = new LineageRecorder(context, sourceConfig.getReferenceName());
+    LineageRecorder lineageRecorder = getLineageRecorder(context);
     lineageRecorder.createExternalDataset(schema);
 
     if (schema != null && schema.getFields() != null) {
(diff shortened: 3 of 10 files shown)


database-plugins/widgets/Database-batchsource.json
```diff
@@ -73,7 +73,7 @@
       "properties": [
         {
           "widget-type": "textbox",
-          "label": "Reference Name",
+          "label": "Table Name",
```
mrahanjam: Did you get Bhooshan's approval on this change?

database-plugins/pom.xml
```diff
@@ -88,6 +88,11 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-jdbc</artifactId>
+      <version>0.2.15</version>
```
snehalakshmisha: extract version into a property

database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
```diff
@@ -270,6 +272,14 @@ private Connection getConnection() throws SQLException {
     return DriverManager.getConnection(sourceConfig.getConnectionString(), properties);
   }
 
+  protected LineageRecorder getLineageRecorder(BatchSourceContext context) {
+    // dbtype, host, port, db from the connection string
+    // table is the reference name
```
snehalakshmisha: this comment should be part of `constructFQN()`

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+
+import io.opentracing.contrib.jdbc.ConnectionInfo;
+import io.opentracing.contrib.jdbc.parser.URLParser;
+
+/**
+ * Generate FQN from DB URL Connection
+ */
+
+public class FQNGenerator {
+  private static final String POSTGRES_DEFAULT_SCHEMA = "public";
```
snehalakshmisha: Do we already have a constant for it?

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public static String constructFQN(String url, String tableName) {
+    ConnectionInfo connectionInfo = URLParser.parse(url);
+    // DB type as set by library after extraction
+    if (connectionInfo.getDbType() == "postgresql") {
```
snehalakshmisha: Extract `postgresql` into a constant

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+      // FQN for Postgresql
+      return String.format("%s://%s/%s.%s.%s", connectionInfo.getDbType(), connectionInfo.getDbPeer(),
+                           connectionInfo.getDbInstance(), getPostgresqlSchema(url), tableName);
+    } else {
+      // FQN for MySQL, Oracle, SQLServer
```
snehalakshmisha, line 39: Did you test the FQN for these 3 DBs?

database-plugins/src/test/java/io/cdap/plugin/db/batch/source/DBSourceTestRun.java
```diff
@@ -456,4 +457,18 @@ public void testNonExistentDBTable() throws Exception {
     assertDeploymentFailure(appId, etlConfig, "ETL Application with DB Source should have failed because of a " +
       "non-existent source database.");
   }
+
+  @Test
+  public void testFQNGeneration() throws Exception {
+    // Testcases consist of Connection URL, Table Name, Expected FQN String
+    String[][] testCases  = {{"jdbc:mysql://localhost:3306/sonoo", "table1", "mysql://localhost:3306/sonoo.table1"},
```
snehalakshmisha: nit: let's replace `sonoo` with `db` or something else like that

database-plugins/widgets/Database-batchsource.json
```diff
@@ -73,7 +73,7 @@
...
           "name": "referenceName",
           "widget-attributes": {
             "placeholder": "Name used to identify this source for lineage"
```
snehalakshmisha: This also needs to be updated

database-plugins/src/test/java/io/cdap/plugin/db/batch/source/DBSourceTestRun.java
```diff
@@ -456,4 +457,18 @@ public void testNonExistentDBTable() throws Exception {
     assertDeploymentFailure(appId, etlConfig, "ETL Application with DB Source should have failed because of a " +
       "non-existent source database.");
   }
+
+  @Test
+  public void testFQNGeneration() throws Exception {
```
snehalakshmisha: Each test should be testing a single functionality. Let's separate the tests for each of the DBs

snehalakshmisha (commented):
Let's add some testing details as well in the description. Will sync offline on this

database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
```diff
@@ -270,6 +272,14 @@ private Connection getConnection() throws SQLException {
     return DriverManager.getConnection(sourceConfig.getConnectionString(), properties);
   }
 
+  protected LineageRecorder getLineageRecorder(BatchSourceContext context) {
```
snehalakshmisha, line 276: Add similar changes to the Sink plugin as well

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
itsankit-google: qq - `Database schema` is also supported by  `oracle` and `sql server` but we are using it in generating FQN for only `postgres`. Is this intentional??

snehalakshmisha: yes, this is intentional - https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

database-plugins/widgets/Database-batchsource.json
snehalakshmisha: Let's change this to `Name used to identify this source for lineage. Typically, the name of the table/view`

database-plugins/widgets/Database-batchsource.json
snehalakshmisha: We need to revert this change

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
itsankit-google: I don't see any information about oracle and sql server in this documentation which you pointed above, why can't we have a similar naming convention for all the databases which support schema??

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
apoorvamk: Yes, I have included it in the tests.

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
apoorvamk: I didn't find anything in DButils or Constants under hydrator-common

database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/DBSink.java
```diff
@@ -216,7 +218,9 @@ private void setResultSetMetadata() throws Exception {
   }
 
   private void recordLineage(BatchSinkContext context, String outputName, Schema tableSchema, List<String> fieldNames) {
-    LineageRecorder lineageRecorder = new LineageRecorder(context, outputName);
+    String fqn = FQNGenerator.constructFQN(dbSinkConfig.getConnectionString(), dbSinkConfig.getReferenceName());
```
snehalakshmisha: nit: we can inline this variable if not being used elsewhere

snehalakshmisha: Same thing in the source as well

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
itsankit-google: discussed offline.

database-plugins/src/main/java/io/cdap/plugin/db/batch/source/DBSource.java
```diff
@@ -189,6 +192,11 @@ public void prepareRun(BatchSourceContext context) {
                               new SourceInputFormatProvider(DataDrivenETLDBInputFormat.class, hConf)));
 
     emitLineage(context);
```
itsankit-google: can modify this private method instead of introducing new method, otherwise we are emitting the lineage twice now in source.

apoorvamk: Added getLineageRecorder method and used the existing emitLineage method.

database-plugins/src/main/java/io/cdap/plugin/db/batch/sink/DBSink.java
```diff
@@ -216,7 +218,9 @@ private void setResultSetMetadata() throws Exception {
   }
 
   private void recordLineage(BatchSinkContext context, String outputName, Schema tableSchema, List<String> fieldNames) {
```
itsankit-google: nit - can remove `outputName` from input param if not being used anymore.

database-plugins/widgets/Database-batchsource.json
```diff
@@ -76,7 +76,7 @@
           "label": "Reference Name",
           "name": "referenceName",
           "widget-attributes": {
-            "placeholder": "Name used to identify this source for lineage"
+            "placeholder": "Name used to identify this source for lineage. Typically, the name of the table/view"
```
snehalakshmisha: Will need same change for the sink as well

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+public class FQNGenerator {
+
+  public static String constructFQN(String url, String tableName) {
+    // dbtype, host, port, db from the connection string
+    // table is the reference name
+    ConnectionInfo connectionInfo = URLParser.parse(url);
+    // DB type as set by library after extraction
+    if (connectionInfo.getDbType() == DBUtils.POSTGRESQL_TAG) {
```
snehalakshmisha: How is this working for postgres?
See https://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java

apoorvamk: Checked in the debugger, it is evaluating to true with ==. Attached screenshot

image

apoorvamk: Update: Took a look at the answers in the stack overflow question. Found this: 

```Sometimes == gives illusions of comparing String values, as in following cases:

String a="Test";
String b="Test";
if(a==b) ===> true

This is because when you create any String literal, the JVM first searches for that literal in the String pool, and if it finds a match, that same reference will be given to the new String. Hence we are seeing the result as true.

Comment (apoorvamk) [reply database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java]: I have updated it to use equals.

File: database-plugins/pom.xml [hunk 16]
```diff
@@ -88,6 +92,11 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentracing.contrib</groupId>
+      <artifactId>opentracing-jdbc</artifactId>
+      <version>${opentracingjdbc.version}</version>
+    </dependency>
```
itsankit-google: nit: Put this dependency before line no 49 as the comment says test dependencies start after it.

database-plugins/src/main/java/io/cdap/plugin/db/common/FQNGenerator.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+public class FQNGenerator {
+
+  public static String constructFQN(String url, String tableName) {
+    // dbtype, host, port, db from the connection string
+    // table is the reference name
+    ConnectionInfo connectionInfo = URLParser.parse(url);
+    // DB type as set by library after extraction
+    if (connectionInfo.getDbType().equals(DBUtils.POSTGRESQL_TAG)) {
```
itsankit-google: nit: generally we should use constants before to avoid NPE.
In this case, `DBUtils.POSTGRESQL_TAG.equals(connectionInfo.getDbType())`

itsankit-google (approved):
minor nits, otherwise LGTM.

fernst (approved):

snehalakshmisha (approved):

PR #1726, opened 2023-01-06
[CDAP-20238] Remove Spark2 references and update dependencies

adityagupta94:
As part of cdapio/cdap/pull/14797, the spark2 modules and references have been removed. Similar references in this repo need to be done and subsequent dependencies need to be updated.

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -50,15 +50,10 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
-      <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-unit-test-spark3_2.12</artifactId>
-    </dependency>
     <!-- Spark3.1.1 uses constant JavaVersion.JAVA_9 that is only there in newer versions of commons-lang3 -->
     <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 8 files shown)


core-plugins/pom.xml
```diff
@@ -221,6 +231,12 @@
       <artifactId>format-orc</artifactId>
       <version>${project.version}</version>
       <scope>test</scope>
+      <exclusions>
```
dli357: nit: If this conflicts with another dependency, can we add a brief comment here on which one?

hbase-plugins/pom.xml
```diff
@@ -109,6 +109,14 @@
           <groupId>log4j</groupId>
           <artifactId>log4j</artifactId>
         </exclusion>
+        <exclusion>
```
dli357, line 117: nit: If these conflict with other dependencies, can we add a brief comments here on which ones?

pom.xml
```diff
@@ -805,6 +815,19 @@
     </dependencies>
   </dependencyManagement>
 
+  <dependencies>
```
dli357, line 820: Should these be in the parent module scope? Seems like they're primarily used by child modules.

dli357 (approved):
LGTM, but would be good to have comments on what each section of the PR fixes. Thanks!

PR #1731, opened 2023-01-12
Bump to next snapshot version

katukota-google:


cassandra-plugins/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>io.cdap.plugin</groupId>
     <artifactId>hydrator-plugins</artifactId>
-    <version>2.9.2</version>
+    <version>2.9.4-SNAPSHOT</version>
```
fernst, line 24: The only version that should have changed was for Database plugins. No other artifact was released.

PR #1732, opened 2023-01-12
Bump to next snapshot version

katukota-google:


database-plugins/pom.xml
```diff
@@ -20,7 +20,7 @@
   <parent>
     <artifactId>hydrator-plugins</artifactId>
     <groupId>io.cdap.plugin</groupId>
-    <version>2.10.0</version>
+    <version>2.10.2-SNAPSHOT</version>
```
fernst, line 23: This is the only version that should have changed. No other artifact was released.

PR #1736, opened 2023-01-16
Added the case for reading blank cells as empty strings

Nikitapaliwal123:
Added the case for reading blank cells as empty strings in place of 'NULL' strings.


core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
```diff
@@ -90,7 +90,7 @@ public class ExcelInputReader extends BatchSource<LongWritable, Object, Structur
   private static final String RECORD = "record";
   private static final String EXIT_ON_ERROR = "Exit on error";
   private static final String WRITE_ERROR_DATASET = "Write to error dataset";
-  private static final String NULL = "NULL";
+  private static final String NULL = "";
```
MrRahulSharma, line 93: Please add a Unit Test to validate the null scenario.

MrRahulSharma:
Excel file change is not required in this PR.

core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputReader.java
MrRahulSharma: Discussed offline the UT for this does not run today, hence no change required here.

PR #1738, opened 2023-01-17
try just running http plugins on github actions to see if it hangs

albertshau:


albertshau:
please don't review, this is for testing only

PR #1739, opened 2023-01-17
CDAP-20274 switch to spark and fix dependencies

albertshau:
Use Spark instead of MapReduce for the unit test and fix the test dependencies

### DIFF
--- a/http-plugins/pom.xml
+++ b/http-plugins/pom.xml
@@ -41,16 +41,12 @@
       <artifactId>netty-http</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-unit-test</artifactId>
-    </dependency>
-    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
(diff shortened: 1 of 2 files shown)


albertshau:
build failed due to existing issues with database-plugins and transform-plugins, others passed

http-plugins/pom.xml
```diff
@@ -60,6 +56,28 @@
     <dependency>
       <groupId>io.cdap.cdap</groupId>
       <artifactId>hydrator-test</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>io.cdap.cdap</groupId>
+          <artifactId>cdap-unit-test</artifactId>
+        </exclusion>
+        <!-- pulls in incompatible jackson versions due to avro -->
+        <exclusion>
```
dli357: Is this exclusion and inclusion only for overwriting the jackson-databind version? If so, can we use dependencyManagement instead to set a desired version?

albertshau: oh I think I can remove this now since I had to add a direct dependency on it (tried doing this first). Otherwise it's a bit better to just exclude it, in case spark gets updated later and jackson-databind needs to be updated again.

fernst (approved):

albertshau:
build is red due to existing test issues. this fixes the issues with http-plugins, so merging.

PR #1747, opened 2023-01-19
[CDAP-20223] Override commons-collections to use 3.2.2

dli357:


pom.xml
```diff
@@ -286,6 +287,11 @@
         <artifactId>commons-lang3</artifactId>
         <version>${commons-lang3-version}</version>
       </dependency>
+      <dependency>
```
albertshau, line 290: not sure what it does to only add this in dependency management. Was a submodule using this without specifying the version somewhere?

dli357: `commons-collections` is brought in by a transitive dependency, so using dependencyManagement forces that transitive dependency to use a newer version.

albertshau (approved):

PR #1758, opened 2023-02-08
Replacing local files used for data validation with gcs buckets.

Vipinofficial11:


pom.xml
```diff
@@ -102,7 +102,7 @@
     <commons-lang-version>2.6</commons-lang-version>
     <commons-lang3-version>3.5</commons-lang3-version>
     <commons-collections.version>3.2.2</commons-collections.version>
-    <cdap.version>6.9.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.8.0-SNAPSHOT</cdap.version>
```
rmstar: why are we changing cdap.version?

Vipinofficial11: reverted the change, Also we don't need sl4j dependency in e2e profile so removed it.

core-plugins/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -74,7 +74,7 @@ overrideSchema=[{"key":"id","value":"float"},{"key":"firstname","value":"string"
   {"key":"lastname","value":"string"},{"key":"photo","value":"string"},{"key":"outputfield","value":"string"}]
 invalidOverrideField=invalid
 overrideFields=[{"key":"id","value":"float"}]
-readRecursivePath=testdata/file/RECURSIVE_TEST
+readRecursivePath=gs://testdata-core/RECURSIVE_TEST
```
rmstar: This will break if we run tests in a project that doesn't have permissions to access bucket `testdata-core`. It would be better to store input files in this repo and copy them to a temp GCS bucket (created for each test run, deleted at the end) which is used in the test

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
```diff
@@ -0,0 +1,85 @@
...
+    // Fetching the data for all the part-r Files.
+    for (String objectName : bucketObjectNames) {
+      if (objectName.contains("part-r")) {
+        Collections.addAll(fileData, fetchObjectData(projectId, bucketName, objectName));
+      }
+    }
+     validateBothFilesData(fileData, expectedOutputFilePath);
+  }
+
+  public static String[] fetchObjectData(String projectId, String bucketName, String objectName) {
```
rmstar: Can this be private?

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
```diff
@@ -0,0 +1,85 @@
...
+    Storage storage = StorageOptions.newBuilder().setProjectId(projectId).build().getService();
+    byte[] objectData = storage.readAllBytes(bucketName, objectName);
+    String objectDataAsString = new String(objectData, StandardCharsets.UTF_8);
+
+    // Splitting using the delimiter as a File can have more than one record.
+    return objectDataAsString.split("\n");
+  }
+
+  public static void validateBothFilesData(List<String> outputFileData, String expectedOutputFilePath) {
```
rmstar: private

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright © 2022 Cask Data, Inc.
```
rmstar: nit: 2023

core-plugins/src/e2e-test/resources/pluginParameters.properties
Vipinofficial11: Sure, added a test setup hook to create and delete a GCS bucket for this Folder. (181bad4e22257f41713fc27864db524b85146da0)

Vipinofficial11: Also the files which we are using in data validation are also stored in GCS bucket testdata-core. So I believe we should follow the above approach for those files as well and create temp buckets for them.

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
```diff
@@ -0,0 +1,85 @@
...
+  public static void validateBothFilesData(List<String> outputFileData, String expectedOutputFilePath) {
+    String[] expectedOutputFileData = fetchObjectData(projectId, expectedOutputFilesBucketName, expectedOutputFilePath);
+    int counter = 0;
+
+    for (String line : expectedOutputFileData) {
+      if (outputFileData.contains(line)) {
+        counter++;
+      }
+    }
+    Assert.assertTrue(counter == expectedOutputFileData.length);
```
rmstar, line 90: validate that the actual contents are equal, not just the size. Expected and actual data might be in different order, so sort before comparing

Vipinofficial11: I have made some changes here so instead of simply updating the counter I have added an assertion which will validate the data line by line that way we don't have to sort the data and we will be able to validate the actual data. (Also asserting the count of records in both files)  What do you think ?

 (https://github.com/cdapio/hydrator-plugins/pull/1758/commits/5523d6961d0458f27214609b99557ba401425c58#diff-d2b99a02e2d3267cdf424d99e449643d4f67ddeb40737290d94f348d0a22297aR79)

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
Vipinofficial11: updated (5523d6961d0458f27214609b99557ba401425c58)

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
Vipinofficial11: updated (5523d6961d0458f27214609b99557ba401425c58)

core-plugins/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/ValidationHelper.java
Vipinofficial11: updated (5523d6961d0458f27214609b99557ba401425c58)

rmstar (approved):
please squash commits

rmstar:
Please take a look at e2e test failures

Vipinofficial11:
> Please take a look at e2e test failures

The failures are due to timeout issue, transformation plugins are taking more time than expected to transfer the records.

rmstar:
> > Please take a look at e2e test failures
> 
> The failures are due to timeout issue, transformation plugins are taking more time than expected to transfer the records.

Is this a transient issue? Do we need to bump up the timeout?

rmstar (approved):
please squash commits

Vipinofficial11:
> please squash commits

squashed

PR #1759, opened 2023-02-14
Refactored the common database handling code into hydrator plugin code.

MrRahulSharma:
Consolidating and refactoring single location to have the Database datatype handling logic in Hydrator-plugin.


itsankit-google (commented):
Afaik, classes related to specific DB plugins (oracle, postgres, sqlserver) should be in data-integrations/database-plugins repo.

@albertshau @rmstar wdyt??

PR #1761, opened 2023-02-23
Remove snapshot from release/2.10 and update gitmodules

vsethi09:
Remove snapshot from release/2.10 and update gitmodules


iam-divya (approved):
LGTM

PR #1764, opened 2023-03-11
Add tag property to the asset

snehalakshmisha:
To support field-level lineage for multi-source/multi-sink plugins, we need a tag that can be used to match the right source to its target.

Jira: https://cdap.atlassian.net/browse/CDAP-20439

### DIFF
--- a/hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
+++ b/hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
@@ -30,12 +30,17 @@ public class Asset {
   private final String fqn;
   private final String location;
   private final String projectId;
+  // may hold information about the source/sink asset
+  // only used in the case of multi-source/multi-sink
+  private final String marker;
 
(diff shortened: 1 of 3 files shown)


hydrator-common/src/main/java/io/cdap/plugin/common/Asset.java
```diff
@@ -30,12 +30,16 @@ public class Asset {
   private final String fqn;
   private final String location;
   private final String projectId;
+  // may hold information about the source/sink asset
+  private final String tag;
```
yaojiefeng: I think per yesterday's discussion, we want to call this `marker`

yaojiefeng: Please add description on when this will be used, currently only use case is multi source/sink

snehalakshmisha: Done

yaojiefeng (approved):

PR #1767, opened 2023-03-29
update submodule branches and corresponding base commits

CuriousVini:


.gitmodules
```diff
@@ -18,24 +18,24 @@
...
 [submodule "condition-plugins"]
 	path = condition-plugins
 	url = ../../data-integrations/condition-plugins.git
-	branch = develop
+	branch = release/1.12
 [submodule "azure"]
 	path = azure
 	url = ../../data-integrations/azure
-	branch = develop
+	branch =
```
vsethi09: branch is missing

CuriousVini: Ah thanks for the catch

vsethi09 (approved):

PR #1794, opened 2023-06-27
Changes for SFTP multiline support for csv

vikasrathee-cs:
Changes for SFTP multiline support for csv

### DIFF
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
@@ -39,6 +39,7 @@ public class DelimitedConfig extends PathTrackingConfig {
   public static final String NAME_ENABLE_QUOTES_VALUES = "enableQuotedValues";
   public static final String NAME_OVERRIDE = "override";
   public static final String NAME_SAMPLE_SIZE = "sampleSize";
+  public static final String NAME_ENABLE_MULTILINE_SUPPORT = "enableMultilineSupport";
   public static final Map<String, PluginPropertyField> DELIMITED_FIELDS;
 
   // description
@@ -46,24 +47,37 @@ public class DelimitedConfig extends PathTrackingConfig {
     "Whether to treat content between quotes as a value. The default value is false.";
   public static final String DESC_SKIP_HEADER =
     "Whether to skip the first line of each file. The default value is false.";
+  public static final String DESC_ENABLE_MULTILINE =
+    "Whether to support content spread over multiple lines if it is between quotes. The default value is false";
 
   static {
     Map<String, PluginPropertyField> fields = new HashMap<>(FIELDS);
     fields.put("skipHeader", new PluginPropertyField("skipHeader", DESC_SKIP_HEADER, "boolean", false, true));
     fields.put(NAME_ENABLE_QUOTES_VALUES,
-      new PluginPropertyField(NAME_ENABLE_QUOTES_VALUES, DESC_ENABLE_QUOTES, "boolean", false, true));
+               new PluginPropertyField(NAME_ENABLE_QUOTES_VALUES, DESC_ENABLE_QUOTES, "boolean", false, true));
+    fields.put(NAME_ENABLE_MULTILINE_SUPPORT,
+               new PluginPropertyField(NAME_ENABLE_MULTILINE_SUPPORT, DESC_ENABLE_MULTILINE, "boolean", false, true));
     DELIMITED_FIELDS = Collections.unmodifiableMap(fields);
   }
+
+  @Macro
+  @Nullable
+  @Description(DESC_ENABLE_QUOTES)
+  protected Boolean enableQuotedValues;
+
+  @Macro
+  @Nullable
+  @Description(DESC_ENABLE_MULTILINE)
+  protected Boolean enableMultilineSupport;
 
   @Macro
   @Nullable
   @Description(DESC_SKIP_HEADER)
   private Boolean skipHeader;
 
-  @Macro
-  @Nullable
-  @Description(DESC_ENABLE_QUOTES)
-  protected Boolean enableQuotedValues;
+  public DelimitedConfig() {
+    super();
+  }
 
   public boolean getSkipHeader() {
     return skipHeader != null && skipHeader;
@@ -73,12 +87,12 @@ public class DelimitedConfig extends PathTrackingConfig {
     return enableQuotedValues != null && enableQuotedValues;
   }
 
-  public long getSampleSize() {
-    return Long.parseLong(getProperties().getProperties().getOrDefault(NAME_SAMPLE_SIZE, "1000"));
+  public Boolean getEnableMultilineSupport() {
+    return enableMultilineSupport != null && enableMultilineSupport;
   }
 
-  public DelimitedConfig() {
-    super();
+  public long getSampleSize() {
+    return Long.parseLong(getProperties().getProperties().getOrDefault(NAME_SAMPLE_SIZE, "1000"));
   }
 
   /**
--- a/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
+++ b/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
@@ -33,6 +33,7 @@ import io.cdap.plugin.format.delimited.common.DataTypeDetectorStatusKeeper;
 import io.cdap.plugin.format.delimited.common.DataTypeDetectorUtils;
 import io.cdap.plugin.format.input.PathTrackingConfig;
 import io.cdap.plugin.format.input.PathTrackingInputFormatProvider;
+import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -49,9 +50,9 @@ import javax.annotation.Nullable;
 @Name(DelimitedInputFormatProvider.NAME)
 @Description(DelimitedInputFormatProvider.DESC)
 public class DelimitedInputFormatProvider extends PathTrackingInputFormatProvider<DelimitedInputFormatProvider.Conf> {
-  public static final PluginClass PLUGIN_CLASS = getPluginClass();
   static final String NAME = "delimited";
   static final String DESC = "Plugin for reading files in delimited format.";
+  public static final PluginClass PLUGIN_CLASS = getPluginClass();
   private final Conf conf;
 
   public DelimitedInputFormatProvider(Conf conf) {
@@ -77,8 +78,8 @@ public class DelimitedInputFormatProvider extends PathTrackingInputFormatProvide
     FailureCollector collector = context.getFailureCollector();
     if (!conf.containsMacro(PathTrackingConfig.NAME_SCHEMA) && schema == null && context.getInputSchema() == null) {
       collector.addFailure(
-        "Delimited format cannot be used without specifying a schema.",
-        "Schema must be specified.")
+          "Delimited format cannot be used without specifying a schema.",
+          "Schema must be specified.")
         .withConfigProperty("schema");
     }
 
@@ -88,8 +89,8 @@ public class DelimitedInputFormatProvider extends PathTrackingInputFormatProvide
 
     if (conf.getEnableQuotedValues() && conf.delimiter != null && conf.delimiter.contains("\"")) {
       collector.addFailure(
-        "The delimiter %s cannot contain \" when quoted values are enabled.",
-        "Check the delimiter.")
+          "The delimiter %s cannot contain \" when quoted values are enabled.",
+          "Check the delimiter.")
         .withConfigProperty("delimiter");
     }
   }
@@ -99,6 +100,11 @@ public class DelimitedInputFormatProvider extends PathTrackingInputFormatProvide
     properties.put(PathTrackingDelimitedInputFormat.DELIMITER, conf.delimiter == null ? "," : conf.delimiter);
     properties.put(PathTrackingDelimitedInputFormat.SKIP_HEADER, String.valueOf(conf.getSkipHeader()));
     properties.put(PathTrackingDelimitedInputFormat.ENABLE_QUOTES_VALUE, String.valueOf(conf.getEnableQuotedValues()));
+    properties.put(PathTrackingDelimitedInputFormat.ENABLE_MULTILINE_SUPPORT,
+                   String.valueOf(conf.getEnableMultilineSupport()));
+    if (conf.getEnableMultilineSupport()) {
+      properties.put(FileInputFormat.SPLIT_MINSIZE, Long.toString(Long.MAX_VALUE));
+    }
   }
 
   @Nullable
@@ -168,6 +174,6 @@ public class DelimitedInputFormatProvider extends PathTrackingInputFormatProvide
     Map<String, PluginPropertyField> properties = new HashMap<>(DelimitedConfig.DELIMITED_FIELDS);
     properties.put("delimiter", new PluginPropertyField("delimiter", Conf.DELIMITER_DESC, "string", false, true));
     return new PluginClass(ValidatingInputFormat.PLUGIN_TYPE, NAME, DESC, DelimitedInputFormatProvider.class.getName(),
-      "conf", properties);
+                           "conf", properties);
   }
 }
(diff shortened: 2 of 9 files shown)


format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,7 +69,48 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
         if (delegate.nextKeyValue()) {
           // skip to next if the current record is header
           if (skipHeader && delegate.getCurrentKey().get() == 0L) {
-            return delegate.nextKeyValue();
+            delegate.nextKeyValue();
```
albertshau: need to return false if this returns false. Otherwise it will error on files that only contain a header (would be good to add a unit test for this scenario)

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,7 +69,48 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+          // getCurrentValue logic moved to nextKeyValue to avoid discrepancies in the data if there are multiple
+          // calls to getCurrentValue method.
+          String delimitedString = delegate.getCurrentValue().toString();
+          builder = StructuredRecord.builder(schema);
+          Iterator<Schema.Field> fields = schema.getFields().iterator();
+          Iterator<String> splitsIterator = getSplitsIterator(enableQuotesValue, delimitedString, delimiter);
+
+          while (splitsIterator.hasNext()) {
+            String part = splitsIterator.next();
+            if (!fields.hasNext()) {
```
albertshau, line 86: this is a comment on existing code, but it would be good to move everything in this `if` block into some other method like `handleBadLine()` or something similar. That would make the logic here easier to follow

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,7 +69,48 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+              int numDataFields = 0;
+              splitsIterator = getSplitsIterator(enableQuotesValue, delimitedString, delimiter);
```
albertshau: this is no longer safe to do with multi-line, as we could have advanced the delegate already. It doesn't look like this was required to create this in the first place, we could have just kept a counter instead of recreating the iterator.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitMultilineIterator.java
```diff
@@ -0,0 +1,125 @@
...
+ * should not contain any quotes. The spliterator will behave like this: 1. if there is no quote, it
+ * will behave same as {@link String#split(String)} 2. if there are quotes in the string, the method
+ * will find pairs of quotes, content within each pair of quotes will not get split even if there
+ * is delimiter in that. For example, if string is a."b.c"."d.e.f" and delimiter is '.', it will get
+ * split into [a, b.c, d.e.f]. if string is "val1.val2", then it will not get split since the '.'
+ * is within a pair of quotes. If data is spread across multiple lines and quote is started within first line, it
+ * will iterate over next lines until closing quote is there.If the delimited string contains odd number of quotes,
+ * which mean the quotes are not closed, an exception will be thrown. The quote within the value will always be trimmed.
+ */
+public class SplitMultilineIterator extends AbstractIterator<String> {
```
albertshau: this class looks like it has a good amount of duplicate code with the SplitQuotesIterator. There should be a way to enhance the current iterator, or create a common base class instead of creating a completely separate class.

The concern here is if we add more options in the future (ex: ability to configure the quote character, allow escaping quotes, etc.), the two classes can easily get out of sync.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitMultilineIterator.java
```diff
@@ -0,0 +1,125 @@
...
+        appendNextLineToDelimitedString(isWithinQuotes);
+        continue;
+      }
+
+      // if the length is not enough for the delimiter or it's not a delimiter, just add it to split
+      if (index + delimiter.length() > delimitedString.length() ||
+        !delimitedString.startsWith(delimiter, index)) {
+        split.append(cur);
+        index++;
+        appendNextLineToDelimitedString(isWithinQuotes);
```
albertshau: strings are immutable, so appending like this is actually creating a new string and can be relatively costly. It also means the string grows ever longer and uses more memory. 

It would be more efficient to append a newline to the split, set delimitedString to the record reader's next line, and reset the index to 0.

albertshau (commented):
also need to make sure files are not broken up into multiple splits when this option is set. This can be done by setting to minimum split size to a max value in the input format configs (see https://github.com/cdapio/hydrator-plugins/blob/94ec33fccdecd636da245d6c91474e2eb19ee428/format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java#L101).

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitMultilineIterator.java
vikasrathee-cs: done.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitMultilineIterator.java
vikasrathee-cs: added to a single class.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
vikasrathee-cs: used a counter here.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
vikasrathee-cs: moved.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
vikasrathee-cs: condition added.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -39,31 +39,43 @@ public class DelimitedConfig extends PathTrackingConfig {
...
+               new PluginPropertyField(NAME_ENABLE_QUOTES_VALUES, DESC_ENABLE_QUOTES, "boolean", false, true));
+    fields.put(NAME_ENABLE_MULTILINE_SUPPORT,
+               new PluginPropertyField(NAME_ENABLE_MULTILINE_SUPPORT, DESC_ENABLE_MULTILINE, "boolean", false, true));
     DELIMITED_FIELDS = Collections.unmodifiableMap(fields);
   }
 
+  @Macro
+  @Nullable
+  @Description(DESC_ENABLE_QUOTES)
+  protected Boolean enableQuotedValues;
```
albertshau, line 66: nit: add a newline between variables

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedConfig.java
```diff
@@ -39,31 +39,43 @@ public class DelimitedConfig extends PathTrackingConfig {
...
+  @Macro
+  @Nullable
+  @Description(DESC_ENABLE_MULTILINE)
+  protected Boolean enableMultilineSupport;
```
albertshau, line 71: nit: add a line between variables

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -49,16 +50,53 @@
...
   static final String DESC = "Plugin for reading files in delimited format.";
+  public static final PluginClass PLUGIN_CLASS = getPluginClass();
   private final Conf conf;
 
   public DelimitedInputFormatProvider(Conf conf) {
     super(conf);
     this.conf = conf;
   }
 
+  static Schema detectSchema(DelimitedConfig conf, String delimiter,
```
albertshau: did the logic here change or did it just get moved? It would be better not to move the method around so that it is easy to see what changed

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/DelimitedInputFormatProvider.java
```diff
@@ -49,16 +50,53 @@
...
+      }
+      List<Schema.Field> fields = DataTypeDetectorUtils.detectDataTypeOfEachDatasetColumn(
+        conf.getOverride(), columnNames, dataTypeDetectorStatusKeeper);
+      Schema schema = Schema.recordOf("text", fields);
+      return PathTrackingInputFormatProvider.addPathField(context.getFailureCollector(), schema, conf.getPathField());
+    }
+    return null;
+  }
+
+  private static PluginClass getPluginClass() {
```
albertshau: similar comment here, avoid moving methods around, it makes it hard to see what changed

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,66 +69,68 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
         if (delegate.nextKeyValue()) {
           // skip to next if the current record is header
           if (skipHeader && delegate.getCurrentKey().get() == 0L) {
-            return delegate.nextKeyValue();
+            if (!delegate.nextKeyValue()) {
+              return false;
+            }
+          }
+          // getCurrentValue logic moved to nextKeyValue to avoid discrepancies in the data if there are multiple
```
albertshau: nit: Whoever is looking at this in the future won't know what was in getCurrentValue before. So let's change this comment to be something like `this logic must be in nextKeyValue to prevent multiple calls to getCurrentValue from advancing the delegate reader`

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,66 +69,68 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
         }
         return false;
       }
 
+      private void handleImproperString(boolean containsQuote, int numDataFields) throws IOException {
+        int numSchemaFields = schema.getFields().size();
+        String message =
+          String.format(
+            "Found a row with %d fields when the schema only contains %d field%s.",
+            numDataFields, numSchemaFields, numSchemaFields == 1 ? "" : "s");
```
albertshau, line 107: this gets called when the schema has fewer fields than the line, so numDataFields will be equal to numSchemaFields here. Before this, need to finish iterating through the splitsIterator in order to get the real number of data fields.

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitQuotesIterator.java
```diff
@@ -31,14 +34,20 @@
...
+  private static final String LINE_SEPARATOR = "\n";
   private final String delimiter;
+  private final RecordReader recordReader;
+  private final boolean enableMultilineSupport;
+  private String delimitedString;
   private int index;
   private boolean endingWithDelimiter = false;
 
-  public SplitQuotesIterator(String delimitedString, String delimiter) {
+  public SplitQuotesIterator(String delimitedString, String delimiter, RecordReader recordReader,
```
albertshau: need to annotate recordReader as `@Nullable` and handle the case when it's null (it's null when used in DataTypeDetectorUtils)

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitQuotesIterator.java
```diff
@@ -94,4 +106,30 @@ protected String computeNext() {
...
+  /**
+   * This method will fetch the next line from Record Reader if split is still within quotes which means data is also
+   * there in next line.
+   *
+   * @param isWithinQuotes whether current split value is within quotes or not
+   */
+  private void appendNextLineToDelimitedString(StringBuilder split, boolean isWithinQuotes) {
+    try {
+      if (enableMultilineSupport && index == delimitedString.length() && isWithinQuotes) {
+        if (recordReader.nextKeyValue()) {
```
albertshau: nit: in these situations it's a little more readable to check the error condition first:

```
if (!recordReader.nextKeyValue()) {
  throw ...
}
split.append(LINE_SEPARATOR);
...
```

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
vikasrathee-cs: Added

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitQuotesIterator.java
vikasrathee-cs: Annotated. and it is null when enableMultiLine support flag is false. and this will used only when flag is true

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
```diff
@@ -69,66 +69,72 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
+          String delimitedString = delegate.getCurrentValue().toString();
+          builder = StructuredRecord.builder(schema);
+          Iterator<Schema.Field> fields = schema.getFields().iterator();
+          Iterator<String> splitsIterator = getSplitsIterator(enableQuotesValue, delimitedString, delimiter);
+          int dataFieldsCount = 0;
+          while (splitsIterator.hasNext()) {
+            dataFieldsCount++;
+            String part = splitsIterator.next();
+            if (!fields.hasNext()) {
+              while (splitsIterator.hasNext()) {
```
albertshau, line 87: thinking about this a bit more, there's a bad data scenario where there is an unenclosed quote. In that case, this will loop through the entire file, which can take a long time.

It would be a little better to pass this iterator to the handlerImproperString() method. In there, we can add some limit on the number of iterations (ex: 100), after which we throw an exception with a more specific message about there likely being an unenclosed quote.

albertshau: this can also be done in a follow up PR if we just want to get the core logic merged first

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/SplitQuotesIterator.java
```diff
@@ -61,6 +71,7 @@ protected String computeNext() {
       if (cur == QUOTE_CHAR) {
         isWithinQuotes = !isWithinQuotes;
         index++;
+        appendNextLineToDelimitedString(split, isWithinQuotes);
```
albertshau: nit: this is a bit confusing since it doesn't always append. let's rename it to appendNextLineIfNeeded()

albertshau (commented):
minor comments

format-delimited/src/main/java/io/cdap/plugin/format/delimited/input/PathTrackingDelimitedInputFormat.java
vikasrathee-cs: sure, that can be done later. But in that case we have to change logic of appendNextLineIfNeeded method, because iterator.next is working on basis of this method. And the number(e.g. 100) that we are taking, we will have this limitation that it can support a multiline spread across multiple lines within that limit only. (e.g. a multiline record with 101 or more lines will fail at both places.)

albertshau (approved):

albertshau:
@vikasrathee-cs the automated build is failing, though many of those tests don't look related. Do the tests pass for you locally?

vikasrathee-cs:
> related

I am not experiencing the above issues in my local. 
Getting passed for all plugins, adding results for core-plugins.
Tests run: 323, Failures: 0, Errors: 0, Skipped: 2

PR #1795, opened 2023-07-03
[CDAP-19665] Fix PSQL schema generation in FQN

itsankit-google:
Schema can be specified in JDBC URL for PSQL with the help of `currentSchema` parameter.

ref: https://stackoverflow.com/questions/4168689/is-it-possible-to-specify-the-schema-when-connecting-to-postgres-with-jdbc


snehalakshmisha (approved):
LGTM

PR #1806, opened 2023-08-28
Small refactor

MrRahulSharma:
JIRA : https://cdap.atlassian.net/browse/PLUGIN-1673

Validation:
Validated that the Wrangler is working.

### DIFF
--- a/hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
+++ b/hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
@@ -106,6 +106,10 @@ public abstract class AbstractDBConnector<T extends PluginConfig & DBConnectorPr
 
   @Override
   public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest request) throws IOException {
+    return this.browse(request);
+  }
+
+  public BrowseDetail browse(BrowseRequest request) throws IOException {
     DBConnectorPath dbConnectorPath = getDBConnectorPath(request.getPath());
     try (Connection connection = getConnection(dbConnectorPath)) {
       int limit = request.getLimit() == null || request.getLimit() <= 0 ? Integer.MAX_VALUE : request.getLimit();


Aryan-Verma (approved):

itsankit-google (commented):
Please add a JIRA in the PR description.

itsankit-google (commented):
And also add a section in the PR description, how this change was tested.

We should test if browsing connections from wrangler are working as expected.

MrRahulSharma:
There is no JIRA available for this change.

hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java
```diff
@@ -106,6 +106,10 @@ public void test(ConnectorContext context) throws ValidationException {
 
   @Override
   public BrowseDetail browse(ConnectorContext connectorContext, BrowseRequest request) throws IOException {
+    return this.browse(request);
```
itsankit-google, line 109: I think it might not be correct and we should be using `ConnectorContext` and store the errors in `FailureCollector` as done in `test(ConnectorContext context)`: https://github.com/cdapio/hydrator-plugins/blob/339807e901b82ad1605d3fbc1c90bec76f14481f/hydrator-common/src/main/java/io/cdap/plugin/common/db/AbstractDBConnector.java#L102

@albertshau and @yaojiefeng wdyt??

albertshau: FailureCollector is used primarily when validating, to highlight multiple errors at the same time, or to include more information in the response like the stack trace. 

In this particular case, it doesn't look like it matters much since there would only be one generic error. It's probably still better to use the FailureCollector since we can add the stack trace to the failure and it doesn't really complicate the logic much.

albertshau: As Rahul pointed out offline, the ConnectionHandler in CDAP doesn't actually do anything with the failure collector, so we would have to fix that before actually using it anywhere

itsankit-google: created a JIRA for improvement: https://cdap.atlassian.net/browse/CDAP-20789

PR #1835, opened 2023-12-11
[Plugin-1730] Added format xls

psainics:
## Added format xls to support files with `.xls` , `.xlsx` format

Jira : [Plugin-1730](https://cdap.atlassian.net/browse/PLUGIN-1730)

## UI Fields
image

- Sample Size: The maximum number of rows that will get investigated for automatic data type detection. The default value is 1000.

- Override: A list of columns with the corresponding data types for whom the automatic data type detection gets skipped.

- Terminate If Empty Row: Whether to terminate the file reading if an empty row is encountered. The default value is false.

- Select Sheet Using: Select the sheet by name or number. Default is ‘Sheet Number’.

- Sheet Value: The name/number of the sheet to read from. If not specified, the first sheet will be read. Sheet Number are 0 based, ie first sheet is 0.

- Use First Row as Header: Whether to use first row as header. 
    - Default A,B,C,D....AA,AB,AC are used

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -279,6 +279,12 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.cdap.plugin</groupId>
+      <artifactId>format-xls</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/format-xls/pom.xml
+++ b/format-xls/pom.xml
@@ -1,0 +1,99 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <!--
+   ~ Copyright © 2024 Cask Data, Inc.
+   ~
+   ~ Licensed under the Apache License, Version 2.0 (the "License"); you may not
+   ~ use this file except in compliance with the License. You may obtain a copy of
+   ~ the License at
+   ~
+   ~ http://www.apache.org/licenses/LICENSE-2.0
+   ~
+   ~ Unless required by applicable law or agreed to in writing, software
+   ~ distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+   ~ WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+   ~ License for the specific language governing permissions and limitations under
+   ~ the License.
+   -->
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.cdap.plugin</groupId>
+    <artifactId>hydrator-plugins</artifactId>
+    <version>2.13.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>format-xls</artifactId>
+  <name>XLS format plugins</name>
+  <packaging>jar</packaging>
+  <properties>
+    <poi.version>5.2.4</poi.version>
+    <log4j-core.version>2.20.0</log4j-core.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi</artifactId>
+      <version>${poi.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi-ooxml</artifactId>
+      <version>${poi.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <scope>compile</scope>
+        <version>${log4j-core.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-etl-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-formats</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>hydrator-test</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.plugin</groupId>
+      <artifactId>format-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+  </dependencies>
+
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <_exportcontents>
+              io.cdap.plugin.format.xls.*
+            </_exportcontents>
+            <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
+            <Embed-Transitive>true</Embed-Transitive>
+            <Embed-Directory>lib</Embed-Directory>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>io.cdap</groupId>
+        <artifactId>cdap-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
+++ b/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright © 2024 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.xls.input;
+
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.format.input.PathTrackingInputFormat;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapreduce.InputSplit;
+import org.apache.hadoop.mapreduce.RecordReader;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
+import org.apache.poi.ss.usermodel.FormulaEvaluator;
+import org.apache.poi.ss.usermodel.Row;
+import org.apache.poi.ss.usermodel.Sheet;
+import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.ss.usermodel.WorkbookFactory;
+
+import java.io.IOException;
+import javax.annotation.Nullable;
+
+
+/**
+ * {@link XlsInputFormat} is {@link TextInputFormat} implementation for reading Excel files.
+ * <p>
+ * The {@link XlsInputFormat.XlsRecordReader} reads a given sheet, and within a sheet reads
+ * all columns and all rows.
+ */
+public class XlsInputFormat extends PathTrackingInputFormat {
+
+  public static final String SHEET_NUM = "Sheet Number";
+  public static final String SHEET_VALUE = "sheetValue";
+  public static final String NAME_SKIP_HEADER = "skipHeader";
+  public static final String TERMINATE_IF_EMPTY_ROW = "terminateIfEmptyRow";
+
+  @Override
+  protected RecordReader<NullWritable, StructuredRecord.Builder> createRecordReader(
+          FileSplit split, TaskAttemptContext context, @Nullable String pathField,
+          @Nullable Schema schema) throws IOException {
+    Configuration jobConf = context.getConfiguration();
+    boolean skipFirstRow = jobConf.getBoolean(NAME_SKIP_HEADER, false);
+    boolean terminateIfEmptyRow = jobConf.getBoolean(TERMINATE_IF_EMPTY_ROW, false);
+    Schema outputSchema = schema != null ? Schema.parseJson(context.getConfiguration().get("schema")) : null;
+    String sheet = jobConf.get(SHEET_NUM);
+    String sheetValue = jobConf.get(SHEET_VALUE, "0");
+    return new XlsRecordReader(sheet, sheetValue, outputSchema, terminateIfEmptyRow, skipFirstRow);
+  }
+
+  /**
+   * Reads Excel sheet, where each row is a {@link StructuredRecord} and each cell is a field in the record.
+   */
+  public static class XlsRecordReader extends RecordReader<NullWritable, StructuredRecord.Builder> {
+    // Converter for converting xls row to structured record
+    XlsRowConverter rowConverter;
+    FormulaEvaluator formulaEvaluator;
+    // Builder for building structured record
+    private StructuredRecord.Builder valueBuilder;
+    private Sheet workSheet;
+    // InputStream handler for Excel files.
+    private FSDataInputStream fileIn;
+    // Specifies the row index.
+    private int rowIndex;
+    // Specifies last row num.
+    private int lastRowNum;
+    private boolean isRowNull;
+    private final String sheet;
+    private final String sheetValue;
+    private final Schema outputSchema;
+    private final boolean terminateIfEmptyRow;
+    private final boolean skipFirstRow;
+
+    /**
+     * Constructor for XlsRecordReader.
+     */
+    public XlsRecordReader(String sheet, String sheetValue, Schema outputSchema, boolean terminateIfEmptyRow,
+                           boolean skipFirstRow) {
+      this.sheet = sheet;
+      this.sheetValue = sheetValue;
+      this.outputSchema = outputSchema;
+      this.terminateIfEmptyRow = terminateIfEmptyRow;
+      this.skipFirstRow = skipFirstRow;
+    }
+
+    @Override
+    public void initialize(InputSplit split, TaskAttemptContext context) throws IOException {
+
+      if (!(split instanceof FileSplit)) {
+        // should never happen
+        throw new IllegalStateException("Input split is not a FileSplit.");
+      }
+      FileSplit fileSplit = (FileSplit) split;
+      Configuration jobConf = context.getConfiguration();
+      // Path of input file.
+      Path file = fileSplit.getPath();
+      FileSystem fs = file.getFileSystem(jobConf);
+      fileIn = fs.open(file);
+
+      try (Workbook workbook = WorkbookFactory.create(fileIn)) {
+        formulaEvaluator = workbook.getCreationHelper().createFormulaEvaluator();
+        formulaEvaluator.setIgnoreMissingWorkbooks(true);
+        // Check if user wants to access with name or number
+        if (sheet.equals(XlsInputFormatConfig.SHEET_NUMBER)) {
+          workSheet = workbook.getSheetAt(Integer.parseInt(sheetValue));
+        } else {
+          workSheet = workbook.getSheet(sheetValue);
+        }
+        rowConverter = new XlsRowConverter(formulaEvaluator);
+      } catch (Exception e) {
+        throw new IOException("Exception while reading excel sheet. " + e.getMessage(), e);
+      }
+
+      lastRowNum = workSheet.getLastRowNum();
+      isRowNull = false;
+      rowIndex = skipFirstRow ? 1 : 0;
+      valueBuilder = StructuredRecord.builder(outputSchema);
+    }
+
+    @Override
+    public boolean nextKeyValue() {
+      // If any is true, then we stop processing.
+      if (rowIndex > lastRowNum || lastRowNum == -1 || (isRowNull && terminateIfEmptyRow)) {
+        return false;
+      }
+      // Get the next row.
+      Row row = workSheet.getRow(rowIndex);
+      valueBuilder = rowConverter.convert(row, outputSchema);
+      if (row == null || valueBuilder == null) {
+        isRowNull = true;
+        // set valueBuilder to a new builder with all fields set to null
+        valueBuilder = StructuredRecord.builder(outputSchema);
+      }
+      // if all fields are null, then the row is null
+      rowIndex++;
+
+      // Stop processing if the row is null and terminateIfEmptyRow is true.
+      return !isRowNull || !terminateIfEmptyRow;
+    }
+
+    @Override
+    public float getProgress() {
+      return (float) rowIndex / lastRowNum;
+    }
+
+    @Override
+    public void close() throws IOException {
+      if (fileIn != null) {
+        fileIn.close();
+      }
+    }
+
+    @Override
+    public NullWritable getCurrentKey() {
+      return NullWritable.get();
+    }
+
+    @Override
+    public StructuredRecord.Builder getCurrentValue() {
+      return valueBuilder;
+    }
+  }
+}
--- a/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
+++ b/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright © 2024 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.format.xls.input;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import io.cdap.cdap.api.annotation.Description;
+import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.api.dataset.lib.KeyValue;
+import io.cdap.cdap.api.plugin.PluginPropertyField;
+import io.cdap.plugin.common.KeyValueListParser;
+import io.cdap.plugin.format.input.PathTrackingConfig;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Common config for Excel related formats.
+ */
+public class XlsInputFormatConfig extends PathTrackingConfig {
+  public static final String SHEET_NUMBER = "Sheet Number";
+  private static final String NAME_OVERRIDE = "override";
+  private static final String NAME_SHEET = "sheet";
+  public static final String NAME_SHEET_VALUE = "sheetValue";
+  private static final String NAME_SKIP_HEADER = "skipHeader";
+  private static final String NAME_TERMINATE_IF_EMPTY_ROW = "terminateIfEmptyRow";
+
+  // properties
+  public static final String NAME_SAMPLE_SIZE = "sampleSize";
+
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each sheet. The default value is false.";
+  public static final String DESC_SHEET = "Select the sheet by name or number. Default is 'Sheet Number'.";
+  public static final String DESC_SHEET_VALUE = "Specifies the value corresponding to 'sheet' input. " +
+    "Can be either sheet name or sheet no; for example: 'Sheet1' or '0' in case user selects 'Sheet Name' or " +
+    "'Sheet Number' as 'sheet' input respectively. Sheet number starts with 0. Default is 'Sheet Number' 0.";
+  public static final String DESC_TERMINATE_ROW = "Specify whether to stop reading after " +
+          "encountering the first empty row. Defaults to false.";
+  public static final Map<String, PluginPropertyField> XLS_FIELDS;
+
+  static {
+    Map<String, PluginPropertyField> fields = new HashMap<>(FIELDS);
+    fields.put(NAME_SKIP_HEADER,
+            new PluginPropertyField(NAME_SKIP_HEADER, DESC_SKIP_HEADER, "boolean", false, true));
+    // Add fields specific for excel format handling.
+    fields.put(NAME_SHEET, new PluginPropertyField(NAME_SHEET, DESC_SHEET, "string", false, true));
+    fields.put(NAME_SHEET_VALUE, new PluginPropertyField(NAME_SHEET_VALUE, DESC_SHEET_VALUE, "string", false, true));
+    fields.put(NAME_TERMINATE_IF_EMPTY_ROW, new PluginPropertyField(
+                       NAME_TERMINATE_IF_EMPTY_ROW, DESC_TERMINATE_ROW, "boolean", false, true));
+    XLS_FIELDS = Collections.unmodifiableMap(fields);
+  }
+
+  @Macro
+  @Nullable
+  @Name(NAME_SHEET)
+  @Description(DESC_SHEET)
+  private String sheet;
+
+  @Macro
+  @Nullable
+  @Name(NAME_SHEET_VALUE)
+  @Description(DESC_SHEET_VALUE)
+  private String sheetValue;
+
+  @Macro
+  @Nullable
+  @Name(NAME_SKIP_HEADER)
+  @Description(DESC_SKIP_HEADER)
+  private Boolean skipHeader;
+
+  @Macro
+  @Nullable
+  @Name(NAME_TERMINATE_IF_EMPTY_ROW)
+  @Description(DESC_TERMINATE_ROW)
+  private Boolean terminateIfEmptyRow;
+
+  public XlsInputFormatConfig() {
+    super();
+  }
+
+  @VisibleForTesting
+  public XlsInputFormatConfig(@Nullable String schema, @Nullable String sheet, @Nullable String sheetValue,
+                              @Nullable Boolean skipHeader, @Nullable Boolean terminateIfEmptyRow) {
+    super();
+    this.schema = schema;
+    this.sheet = sheet;
+    this.sheetValue =  sheetValue;
+    this.skipHeader = skipHeader;
+    this.terminateIfEmptyRow = terminateIfEmptyRow;
+  }
+
+  public int getSampleSize() {
+    String sampleSize = getProperties().getProperties().getOrDefault(NAME_SAMPLE_SIZE, "1000");
+    try {
+      return Integer.parseInt(sampleSize);
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException(String.format("Invalid sample size '%s'.", sampleSize));
+    }
+  }
+
+  public String getSheet() {
+    return sheet == null ? SHEET_NUMBER : sheet;
+  }
+
+  @Nullable
+  public String getSheetValue() {
+    return sheetValue;
+  }
+
+  public boolean getSkipHeader() {
+    return skipHeader != null ? skipHeader : false;
+  }
+
+  public boolean getTerminateIfEmptyRow() {
+    return terminateIfEmptyRow != null ? terminateIfEmptyRow : false;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * Builder for building a {@link XlsInputFormatConfig}.
+   */
+  public static class Builder {
+    private String schema;
+    private String sheet;
+    private String sheetValue;
+    private Boolean skipHeader;
+    private Boolean terminateIfEmptyRow;
+
+    public Builder setSchema(String schema) {
+      this.schema = schema;
+      return this;
+    }
+
+    public Builder setSheet(String sheet) {
+      this.sheet = sheet;
+      return this;
+    }
+
+    public Builder setSheetValue(String sheetValue) {
+      this.sheetValue = sheetValue;
+      return this;
+    }
+
+    public Builder setSkipHeader(Boolean skipHeader) {
+      this.skipHeader = skipHeader;
+      return this;
+    }
+
+    public Builder setTerminateIfEmptyRow(Boolean terminateIfEmptyRow) {
+      this.terminateIfEmptyRow = terminateIfEmptyRow;
+      return this;
+    }
+
+    public XlsInputFormatConfig build() {
+     return new XlsInputFormatConfig(schema, sheet, sheetValue, skipHeader, terminateIfEmptyRow);
+    }
+  }
+
+}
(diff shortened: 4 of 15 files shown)


format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+        overrideDataTypes.put(name, schema);
+      }
+    }
+    return overrideDataTypes;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
```
albertshau, line 138: nit: add a newline after this

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+  @Name(NAME_SHEET)
+  @Description(DESC_SHEET)
+  private String sheet;
+
+  @Macro
+  @Nullable
+  @Name(NAME_SHEET_VALUE)
+  @Description(DESC_SHEET_VALUE)
+  private String sheetValue;
+
```
albertshau, line 82: nit: remove extra newline

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Reads XLS(X) into StructuredRecords.
+ */
+@Plugin(type = ValidatingInputFormat.PLUGIN_TYPE)
+@Name(XlsInputFormatProvider.NAME)
+@Description(XlsInputFormatProvider.DESC)
+public class XlsInputFormatProvider extends PathTrackingInputFormatProvider<XlsInputFormatConfig> {
+  static final String NAME = "xls";
+  static final String DESC = "Plugin for reading files in xls(x) format.";
+  public static final PluginClass PLUGIN_CLASS = PluginClass.builder()
```
albertshau, line 58: this looks like it's unused. Should remove it if so, otherwise please point out where it is being used.

If unused, please also remove the related variables, like XlsInputFormatConfig.XLS_FIELDS

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+  // properties
+  public static final String NAME_SAMPLE_SIZE = "sampleSize";
+
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. The default value is false.";
+  public static final String DESC_SHEET = "Select the sheet by name or number. Default is 'Sheet Number'.";
+  public static final String DESC_SHEET_VALUE = "Specifies the value corresponding to 'sheet' input. " +
+    "Can be either sheet name or sheet no; for example: 'Sheet1' or '0' in case user selects 'Sheet Name' or " +
+    "'Sheet Number' as 'sheet' input respectively. Sheet number starts with 0.";
+  public static final String DESC_TERMINATE_ROW = "Specify whether processing needs to be terminated in case an" +
```
albertshau: It doesn't looks like XLS_FIELDS is needed, so these should just be specified inline with the Description annotation, so the description is close to the actual variable.

albertshau: This seems like an odd option to have. I assume it's being carried over from the existing excel source?

If we don't know of an important use case for this, we should remove it.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+  private static final String NAME_TERMINATE_IF_EMPTY_ROW = "terminateIfEmptyRow";
+
+  // properties
+  public static final String NAME_SAMPLE_SIZE = "sampleSize";
+
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. The default value is false.";
+  public static final String DESC_SHEET = "Select the sheet by name or number. Default is 'Sheet Number'.";
+  public static final String DESC_SHEET_VALUE = "Specifies the value corresponding to 'sheet' input. " +
+    "Can be either sheet name or sheet no; for example: 'Sheet1' or '0' in case user selects 'Sheet Name' or " +
```
albertshau, line 53: What is the behavior if this is not specified? Does it read every sheet? Does it fail?

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+  private static final String NAME_SHEET = "sheet";
+  public static final String NAME_SHEET_VALUE = "sheetValue";
+  private static final String NAME_SKIP_HEADER = "skipHeader";
+  private static final String NAME_TERMINATE_IF_EMPTY_ROW = "terminateIfEmptyRow";
+
+  // properties
+  public static final String NAME_SAMPLE_SIZE = "sampleSize";
+
+  public static final String DESC_SKIP_HEADER =
+    "Whether to skip the first line of each file. The default value is false.";
```
albertshau: file -> sheet

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+  public boolean getTerminateIfEmptyRow() {
+    return terminateIfEmptyRow != null ? terminateIfEmptyRow : false;
+  }
+
+  /**
+   * Parses a list of key-value items of column names and their corresponding data types, manually set by the user.
+   *
+   * @return A hashmap of column names and their manually set schemas.
+   */
+  public Map<String, Schema> getOverride() throws IllegalArgumentException {
```
albertshau: I'm having trouble understanding how this is being used. Do you know what is setting this override property?

In any case, it looks like this is copied directly from DelimitedConfig. It should be defined in a single place and not duplicated.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+
+    public Builder setSheet(String sheet) {
+      this.sheet = sheet;
+      return this;
+    }
+
+    public Builder setSheetValue(String sheetValue) {
+      this.sheetValue = sheetValue;
+      return this;
+    }
```
albertshau, line 163: newline after this

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+    public Builder setSkipHeader(Boolean skipHeader) {
+      this.skipHeader = skipHeader;
+      return this;
+    }
+
+    public Builder setTerminateIfEmptyRow(Boolean terminateIfEmptyRow) {
+      this.terminateIfEmptyRow = terminateIfEmptyRow;
+      return this;
+    }
```
albertshau, line 173: newline after this

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatUtils.java
```diff
@@ -0,0 +1,89 @@
...
+   * <p>
+   * Steps:
+   * 1) Trim surrounding spaces
+   * 2) If its empty replace it with BLANK
+   * 3) If it starts with a number, prepend "col_"
+   * 4) Replace invalid characters with "_" (multiple invalid characters gets replaced with one symbol)
+   * 5) Check if the name has been found before (without considering case)
+   * if so add _# where # is the number of times seen before + 1
+   */
+  public static List<String> getSafeColumnNames(List<String> columnNames) {
```
albertshau, line 44: it doesn't look like the other formats do this column name manipulation. XLS should be consistent and avoid this logic. Leave it to Wrangler to cleanse column names.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.List;
+
+
+/**
+ * {@link XlsInputFormat} is {@link TextInputFormat} implementation for reading Excel files.
+ * <p>
+ * The {@link XlsInputFormat.XlsRecordReader} reads a given sheet, and within a sheet reads
+ * all columns and all rows.
+ */
+public class XlsInputFormat extends CombineFileInputFormat<LongWritable, StructuredRecord> {
```
albertshau: this should extend PathTrackingInputFormat so that it can handle the path related properties like other formats. It should not be a CombineFileInputFormat, as that can generate multiple splits for a single file.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  @Override
+  public void validate(FormatContext context) {
+    Schema schema = super.getSchema(context);
+    FailureCollector collector = context.getFailureCollector();
+    // When the sheet is specified by number, the sheet value must be a number
+    if (!conf.containsMacro(XlsInputFormatConfig.NAME_SHEET_VALUE)
+            && conf.getSheet().equals(XlsInputFormatConfig.SHEET_NUMBER)
+                    && (conf.getSheetValue() == null || !conf.getSheetValue().matches("[0-9]+"))) {
```
albertshau: shouldn't be checking numbers using regexes, just try to parse it into an int.

That being said, why are we overloading this single property to be either a name or a number? Seems simpler to have separate properties, one for sheet name and one for sheet number. Would be easier to describe what they each are and validate properly.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
```diff
@@ -0,0 +1,211 @@
...
+    super();
+    this.schema = schema;
+    this.sheet = sheet;
+    this.sheetValue =  sheetValue;
+    this.skipHeader = skipHeader;
+    this.terminateIfEmptyRow = terminateIfEmptyRow;
+  }
+
+  public long getSampleSize() {
+    return Long.parseLong(getProperties().getProperties().getOrDefault(NAME_SAMPLE_SIZE, "1000"));
```
albertshau: return an int here instead of a long

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -0,0 +1,192 @@
...
+        Schema schema = Schema.recordOf("xls", schemaDetector.getFields(
+                XlsInputFormatUtils.getSafeColumnNames(columnNames), conf.getOverride()));
+        return PathTrackingInputFormatProvider.addPathField(context.getFailureCollector(), schema, conf.getPathField());
+      }
+    }
+    return null;
+  }
+
+  private int getSampleSizeInt() {
+    // Note: Sample size is long, but we are casting it to int here. This is because the POI API uses int
```
albertshau: the type is defined by the XlsInputFormatConfig. The config class should return an int there instead of a long.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
albertshau: please make sure this behavior is covered by unit tests

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+        formulaEvaluator.setIgnoreMissingWorkbooks(true);
+        formatter = new XlsInputFormatDataFormatter(formulaEvaluator);
+        // Check if user wants to access with name or number
+        if (sheet.equals(XlsInputFormatConfig.SHEET_NUMBER)) {
+          workSheet = workbook.getSheetAt(Integer.parseInt(sheetValue));
+        } else {
+          workSheet = workbook.getSheet(sheetValue);
+        }
+      } catch (Exception e) {
+        throw new IllegalArgumentException("Exception while reading excel sheet. " + e.getMessage(), e);
```
albertshau: IllegalArgumentException -> IOException

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      lastRowNum = workSheet.getLastRowNum();
+      rowIndex = 0;
+      isRowNull = false;
+
+      boolean skipFirstRow = jobConf.getBoolean(NAME_SKIP_HEADER, true);
+      if (skipFirstRow) {
+        Preconditions.checkArgument(lastRowNum != -1, "No rows found on sheet %s", sheetValue);
```
albertshau: why is this check needed? This would result in weird behavior where the pipeline fails on empty sheets if skip header is true, but succeeds on empty sheets when it is false.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+    private boolean terminateIfEmptyRow;
+    private boolean isRowNull;
+
+    @Override
+    public void initialize(InputSplit genericSplit, TaskAttemptContext context) throws IOException {
+
+      CombineFileSplit split = (CombineFileSplit) genericSplit;
+      Configuration jobConf = context.getConfiguration();
+      // Path of input file.
+      Path file = split.getPath(0);
```
albertshau: this will result in duplicate data if there are multiple splits for a single file. Seems like we're setting the min split size to long max value to prevent this, but it's simpler to not use the CombineFileInputFormat to prevent this possibility entirely.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
albertshau: After reading the implementation, it looks like this is used to stop reading when an empty row is found. Let's change the wording here, I thought 'terminated' meant it would throw an error.

`Whether to stop reading after encountering the first empty row. Defaults to false.`

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+    // Map value that represents an excel row
+    private StructuredRecord value;
+    private Sheet workSheet;
+    // InputStream handler for Excel files.
+    private FSDataInputStream fileIn;
+    // Specifies the row index.
+    private int rowIndex;
+    // Specifies last row num.
+    private int lastRowNum;
+    private Schema outputSchema;
```
albertshau: things that will never change like outputSchema, terminateIfEmptyRow should be final. Basically anything read from the Configuration is static and should be read in createRecordReader() and passed into the constructor instead of read in the initialize() method.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Reads excel spread sheet, where the keys are the offset in the excel file and the text is the complete record.
+   */
+  public static class XlsRecordReader extends RecordReader<LongWritable, StructuredRecord> {
+    // DataFormatter to format and get each cell's value as String
+    XlsInputFormatDataFormatter formatter;
+    FormulaEvaluator formulaEvaluator;
+    // Map key that represents the row index.
+    private LongWritable key;
```
albertshau: Is the key used anywhere? If not, it's better to use a NullWritable instead of LongWritable.

If it is used, please point out where.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+        } else {
+          // set empty row in the output
+          value = StructuredRecord.builder(outputSchema).build();
+          rowIndex++;
+          return true;
+        }
+      }
+      key = new LongWritable(rowIndex);
+
+      StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
```
albertshau: would be more readable to move the Row -> StructuredRecord conversion logic to a separate method or class. Also would make it easier to unit test the conversion logic.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+      List<Schema.Field> fields = outputSchema.getFields();
+
+      isRowNull = true;
+      for (int cellIndex = 0; cellIndex < row.getLastCellNum(); cellIndex++) {
+        if (cellIndex >= fields.size()) {
+         throw new IllegalArgumentException(
+           String.format("Schema contains less fields than the number of columns in the excel file. " +
```
albertshau: less -> fewer

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+        Cell cell = row.getCell(cellIndex, Row.MissingCellPolicy.RETURN_BLANK_AS_NULL);
+        if (cell == null) {
+          // Blank cells are skipped, builder will set null for the field, no processing needed.
+          continue;
+        }
+        Schema.Field field = fields.get(cellIndex);
+        Schema.Type type = field.getSchema().isNullable() ?
+                field.getSchema().getNonNullable().getType() : field.getSchema().getType();
+        String result = formatter.formatCellValue(cell, type);
```
albertshau: this is wasteful and error prone. We shouldn't be converting to a String and then using convertAndSet to change back to the actual type. Should be setting the value directly on the builder based on the cell type.

albertshau: Instead of the XlsInputFormatDataFormatter, it would be better to have a XlsRowConverter class that contains all the logic for converting a Row to a StructuredRecord.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatDataFormatter.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.CellType;
+import org.apache.poi.ss.usermodel.DataFormatter;
+import org.apache.poi.ss.usermodel.DateUtil;
+import org.apache.poi.ss.usermodel.FormulaEvaluator;
+
+/**
+ * Formats the cell value of an Excel file.
+ */
+public class XlsInputFormatDataFormatter {
```
albertshau: nit: remove 'InputFormat' from the class name, it doesn't need to be used with Hadoop InputFormats

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
albertshau: Is the intent to use this in a future PR? The pattern for other formats is to add the format to ETLBatchTestBase.setupTest() and have test cases in FileBatchSourceTest

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -0,0 +1,192 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
```
albertshau: update to 2024 everywhere

pom.xml
```diff
@@ -45,6 +45,7 @@
     <module>solrsearch-plugins</module>
     <module>spark-plugins</module>
     <module>transform-plugins</module>
+    <module>format-xls</module>
```
albertshau: modules are sorted alphabetically, move this up to where the other formats are

albertshau (commented):
is anything using the testdata.xlsx file in the resources directory?

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Newline Added 👍

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Newline Removed 👍

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Updated !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Newline Added !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Newline Added !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Using `Integer.parseInt` now, config class now returns int

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics: Using `Integer.parseInt` now, config class now returns int

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Updated !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Updated !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatDataFormatter.java
psainics: Removed !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics: Updated 2024 🎆

pom.xml
psainics: Moved up !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Using NullWritable Key now !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Now extending `PathTrackingInputFormat`

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Using `PathTrackingInputFormat` instead of `CombineFileInputFormat` now

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics: Yes this will be used, in `FileBatchSourceTest` to create Test Cases.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: > I assume it's being carried over from the existing excel source

Yes
 
- Updated the description for `DESC_TERMINATE_ROW`

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: Updated the description.
By default it will read the 1st sheet.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatConfig.java
psainics: I had a discussing with my senior, this feature is not required in XLS source.
This has been removed !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatUtils.java
psainics: This idea was taken from csv format, this is required to get unique column names from xls when doing schema detection.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: This check is not required now, removed !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Made values from config final and move to constructor.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Updated logic to use `XlsRowConverter`

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Added a `XlsRowConverter` class

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics: The int validation is updated, now using parse int.

This behavior of having one drop down and overloading the other field is taken from the existing Excel source plugin.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatUtils.java
albertshau: I don't see where the csv format is doing this, can you point it out? 

If existing formats are doing this, then they should all share the same code instead of it being duplicated.

albertshau:
can you squash the commits that took place after the first review? That way it will be easier to review just the changes instead of all 1.5k lines again.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatUtils.java
psainics: You can find it under `package io.cdap.plugin.format.delimited.common`
class `DataTypeDetectorUtils` 
Method `setColumnNames` / `cleanSchemaFieldNames`

albertshau: was cleanSchemaColumnNames copied directly from there except modified to use Lists instead of arrays?

If the logic is the same, can you move that class to format-common to re-use most of the logic?

format-xls/pom.xml
```diff
@@ -0,0 +1,102 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <!--
+   ~ Copyright © 2023 Cask Data, Inc.
```
albertshau: all the years should be 2024

core-plugins/pom.xml
```diff
@@ -279,6 +279,18 @@
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
     </dependency>
+      <dependency>
```
albertshau: this is a duplicate, should remove it

format-xls/pom.xml
```diff
@@ -27,19 +27,25 @@
...
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi-ooxml</artifactId>
+      <version>${poi.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-core</artifactId>
-      <scope>test</scope>
-      <version>2.17.2</version>
+      <scope>compile</scope>
```
albertshau, line 47: why is this changed to compile?

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -50,79 +46,93 @@
...
   public static final String TERMINATE_IF_EMPTY_ROW = "terminateIfEmptyRow";
 
   @Override
-  public RecordReader<LongWritable, StructuredRecord> createRecordReader(InputSplit split, TaskAttemptContext context) {
-    return new XlsRecordReader();
+  protected RecordReader<NullWritable, StructuredRecord.Builder> createRecordReader(
+          FileSplit split, TaskAttemptContext context, @Nullable String pathField,
+          @Nullable Schema schema) throws IOException {
+    Configuration jobConf = context.getConfiguration();
+    boolean skipFirstRow = jobConf.getBoolean(NAME_SKIP_HEADER, true);
```
albertshau: thought the default for this was false?

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -50,79 +46,93 @@
...
+    boolean terminateIfEmptyRow = jobConf.getBoolean(TERMINATE_IF_EMPTY_ROW, false);
+    Schema outputSchema =  schema != null ? Schema.parseJson(context.getConfiguration().get("schema")) : null;
```
albertshau: extra space after `=`

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
```diff
@@ -50,79 +46,93 @@
  * The {@link XlsInputFormat.XlsRecordReader} reads a given sheet, and within a sheet reads
  * all columns and all rows.
  */
-public class XlsInputFormat extends CombineFileInputFormat<LongWritable, StructuredRecord> {
+public class XlsInputFormat extends PathTrackingInputFormat {
 
   public static final String SHEET_NO = "Sheet Number";
```
albertshau: nit: NO -> NUM is more descriptive

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -80,11 +81,15 @@ public void validate(FormatContext context) {
...
-    if (!conf.containsMacro(XlsInputFormatConfig.NAME_SHEET_VALUE)
-            && conf.getSheet().equals(XlsInputFormatConfig.SHEET_NUMBER)
-                    && (conf.getSheetValue() == null || !conf.getSheetValue().matches("[0-9]+"))) {
-      collector.addFailure("Sheet number must be a number.", null)
-              .withConfigProperty(XlsInputFormatConfig.NAME_SHEET_VALUE);
+    if (!conf.containsMacro(XlsInputFormatConfig.NAME_SHEET_VALUE) &&
+            conf.getSheet().equals(XlsInputFormatConfig.SHEET_NUMBER) &&
+            !Strings.isNullOrEmpty(conf.getSheetValue())) {
+      try {
+        Integer.parseInt(conf.getSheetValue());
```
albertshau: while we're doing this, can also check that it's not negative

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
```diff
@@ -115,8 +122,23 @@ public Schema detectSchema(FormatContext context, InputFiles inputFiles) throws
         Sheet workSheet;
         // Check if user wants to access with name or number
         if (conf.getSheet() != null && conf.getSheet().equals(XlsInputFormatConfig.SHEET_NUMBER)) {
-          workSheet = workbook.getSheetAt(Integer.parseInt(Objects.requireNonNull(conf.getSheetValue())));
+          int sheetValue = 0;
+          if (!Strings.isNullOrEmpty(conf.getSheetValue())) {
+            try {
+              sheetValue = Integer.parseInt(conf.getSheetValue());
+            } catch (NumberFormatException e) {
+              failureCollector.addFailure("Sheet number must be a number.", null)
```
albertshau: this is code is duplicated in the validate() method, it would be better to have a method like getSheetAsNumber() that is called in both places.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
```diff
@@ -0,0 +1,167 @@
...
+  public StructuredRecord.Builder convert(Row row, Schema outputSchema) {
+    if (row == null) {
+      return null;
+    }
+    boolean isRowEmpty = true;
+    StructuredRecord.Builder builder = StructuredRecord.builder(outputSchema);
+    List<Schema.Field> fields = outputSchema.getFields();
+    for (int cellIndex = 0; cellIndex < row.getLastCellNum(); cellIndex++) {
+      if (cellIndex >= fields.size()) {
+        throw new IllegalArgumentException(
```
albertshau: this seems overly restrictive, it seems like people should be able to read a subset of the columns. Or if there are extra cells on the side it seems like it shouldn't cause a failure. Is this how the existing excel source behaves?

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
```diff
@@ -0,0 +1,167 @@
...
+          cellValue = getCellAsString(cell);
+          break;
+        case DOUBLE:
+          cellValue = getCellAsDouble(cell);
+          break;
+        case BOOLEAN:
+          cellValue = getCellAsBoolean(cell);
+          break;
+        default:
+          throw new IllegalArgumentException(
```
albertshau: this should be checked during plugin validation if it is not already happening. For cases like this when we don't expect the situation to ever happen, we usually add a comment that it's not expected and throw an IllegalStateException

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
```diff
@@ -0,0 +1,167 @@
...
+                          field.getName(), type, "string, double, boolean"));
+      }
+      if (cellValue == null) {
+        continue;
+      }
+      builder.set(field.getName(), cellValue);
+      isRowEmpty = false;
+    }
+    if (isRowEmpty) {
+      return null;
```
albertshau, line 87: method should be annotated with `@Nullable` so it is clear the caller needs to handle nulls.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
```diff
@@ -0,0 +1,167 @@
...
+        if (DateUtil.isCellDateFormatted(cell)) {
+          return dataFormatter.formatCellValue(cell);
+        }
+        return Double.toString(cell.getNumericCellValue());
+      case STRING:
+        return cell.getRichStringCellValue().getString();
+      case BOOLEAN:
+        return cell.getBooleanCellValue() ? "TRUE" : "FALSE";
+      case BLANK:
+      case ERROR:
```
albertshau, line 118: are there any other CellType values (like date?)

core-plugins/pom.xml
psainics: Removed duplicate !

format-xls/pom.xml
psainics: The version of `apache poi` being used was having compatibility issue with the log4j library being used, i had to change this to compile to make it work.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Fixed, default is now false. Thanks!

format-xls/pom.xml
psainics: Looks like a bug on GitHub, it's already 2024 🤔

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: Removed space.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormat.java
psainics: `SHEET_NUM` is better 👍

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics:
```java
 try {
        int sheetValue = Integer.parseInt(conf.getSheetValue());
        if (sheetValue >= 0) {
          return sheetValue;
        }
        failureCollector.addFailure("Sheet number must be a positive number.", null)
                    .withConfigProperty(XlsInputFormatConfig.NAME_SHEET_VALUE);
      }
```

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsInputFormatProvider.java
psainics:
```java
private Integer getSheetAsNumber(FailureCollector failureCollector) {
    if (!Strings.isNullOrEmpty(conf.getSheetValue())) {
      try {
        int sheetValue = Integer.parseInt(conf.getSheetValue());
        if (sheetValue >= 0) {
          return sheetValue;
        }
        failureCollector.addFailure("Sheet number must be a positive number.", null)
                    .withConfigProperty(XlsInputFormatConfig.NAME_SHEET_VALUE);
      } catch (NumberFormatException e) {
        failureCollector.addFailure("Sheet number must be a number.", null)
                .withConfigProperty(XlsInputFormatConfig.NAME_SHEET_VALUE);
      }
    }
    return null;
  }
```

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
psainics: Yup, this is not suppose to happen.
I have added comments for that.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
psainics: Marked !

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
psainics: No , dates are stored as Numeric, there is a util function to check if the numeric cell is formatted as date.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
psainics: Hmm, i have added `&& cellIndex < fields.size()` as a condition in the for loop.
this won't throw any errors now.

The existing plugin does not have this issue as it uses fields like `Columns To Be Extracted` and `Column Label Mapping ` to have a well defined area . 
But this forces user to manually enter all the details, so this was not included in this plugin.

format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
albertshau: nit: throw IllegalStateException instead of IllegalArgumentException, since hitting this is a problem in the code and not a problem with the input argument

albertshau (approved):
lgtm, please squash all commits and we can merge

albertshau:
@psainics there are e2e and unit test failures, please comment on whether these are new or expected

psainics:
E2E failure has been resolved.

PR #1837, opened 2023-12-22
[PLUGIN-1722][Projection Transform]Changes done for converting double to float.

vikasrathee-cs:
This change is done for a customer requesting to convert double to float in Transform Plugin.
As double can have much larger values than float, in that case it will return null, otherwise will convert from double to float.


vikasrathee-cs:
Closing pull request as we only need to add validation to stop user from downcasting

PR #1856, opened 2024-04-29
[PLUGIN-1723] Fix JavaScript Decimal Values

psainics:
## Fix JavaScript Decimal Values

Jira : [PLUGIN-1723](https://cdap.atlassian.net/browse/PLUGIN-1723)

### Description

When the output of the JS transform plugin is of Decimal data type, all the result records become null.

image

### After Fix
image
image

### Code change

- Modified `JavaScriptTransform.java`
  -  Add case to handle decimal type. 

### Unit Tests

- Modified `JavaScriptTransformTest.java`
  - Added Test `testDecimalTransform`
  -  image

[PLUGIN-1723]: https://cdap.atlassian.net/browse/PLUGIN-1723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
@@ -53,6 +53,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -270,6 +272,21 @@ public class JavaScriptTransform extends Transform<StructuredRecord, StructuredR
 
   private Object decode(Object object, Schema schema) {
     Schema.Type type = schema.getType();
+
+    Schema.LogicalType logicalType = schema.getLogicalType();
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DECIMAL:
+          BigDecimal bigDecimal = null;
+          if (object instanceof Number) {
+            double doubleValue = ((Number) object).doubleValue();
+            bigDecimal = BigDecimal.valueOf(doubleValue).setScale(schema.getScale(), RoundingMode.HALF_EVEN);
+          }
+          if (bigDecimal != null) {
+            return bigDecimal.unscaledValue().toByteArray();
+          }
+      }
+    }
 
     switch (type) {
       case NULL:
--- a/core-plugins/docs/JavaScript-transform.md
+++ b/core-plugins/docs/JavaScript-transform.md
@@ -63,6 +63,9 @@ or else scale the ``count`` field by 1024.
 **schema:** The schema of output objects. If no schema is given, it is assumed that the output
 schema is the same as the input schema.
 
+> for `Decimal` type the value is rounded using the `RoundingMode.HALF_EVEN` method if it does not fit within 
+the schema. This ensures that the value adheres to the precision and scale defined in the schema.
+
 **lookup:** The configuration of the lookup tables to be used in your script.
 For example, if lookup table "purchases" is configured, then you will be able to perform
 operations with that lookup table in your script: ``context.getLookup('purchases').lookup('key')``
--- a/core-plugins/src/test/java/io/cdap/plugin/transform/JavaScriptTransformTest.java
+++ b/core-plugins/src/test/java/io/cdap/plugin/transform/JavaScriptTransformTest.java
@@ -34,6 +34,7 @@ import io.cdap.cdap.etl.mock.transform.MockTransformContext;
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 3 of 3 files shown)


core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
```diff
@@ -271,6 +273,19 @@ private InvalidEntry<StructuredRecord> getErrorObject(Map result, StructuredReco
...
+      BigDecimal bigDecimal = null;
+      if (object instanceof Double) {
+        bigDecimal = BigDecimal.valueOf((Double) object).setScale(schema.getScale(), RoundingMode.HALF_DOWN);
+      } else if (object instanceof Integer) {
+        bigDecimal = BigDecimal.valueOf((Integer) object).setScale(schema.getScale(), RoundingMode.HALF_DOWN);
+      }
+      if (bigDecimal != null) {
+        return bigDecimal.unscaledValue().toByteArray();
+      }
+    }
```
itsankit-google: Please use case switch case like below to make the code extensible in future.

itsankit-google:
Please also add the screenshot on how the output looks while running preview with the change.

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
psainics: Updated !

```java
Schema.LogicalType logicalType = schema.getLogicalType();
if (logicalType != null) {
  switch (logicalType) {
    case DECIMAL:
      BigDecimal bigDecimal = null;
      if (object instanceof Double) {
        bigDecimal = BigDecimal.valueOf((Double) object).setScale(schema.getScale(), RoundingMode.HALF_DOWN);
      } else if (object instanceof Integer) {
        bigDecimal = BigDecimal.valueOf((Integer) object).setScale(schema.getScale(), RoundingMode.HALF_DOWN);
      }
      if (bigDecimal != null) {
        return bigDecimal.unscaledValue().toByteArray();
      }
  }
}
```

psainics:
> Please also add the screenshot on how the output looks while running preview with the change.

Added screenshot after fix !

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
```diff
@@ -271,6 +273,22 @@ private InvalidEntry<StructuredRecord> getErrorObject(Map result, StructuredReco
   private Object decode(Object object, Schema schema) {
     Schema.Type type = schema.getType();
 
+    Schema.LogicalType logicalType = schema.getLogicalType();
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DECIMAL:
+          BigDecimal bigDecimal = null;
+          if (object instanceof Double) {
+            bigDecimal = BigDecimal.valueOf((Double) object).setScale(schema.getScale(), RoundingMode.HALF_DOWN);
```
itsankit-google: why `HALF_DOWN` and not `HALF_EVEN`?

psainics: This is an open question in the design, i have no preference here !
Let me know what works the best !

itsankit-google: I think `HALF_EVEN` is the most commonly used one in applications where fairness is important, as it tends to distribute rounding errors more evenly across a large number of calculations.

itsankit-google:
> > Please also add the screenshot on how the output looks while running preview with the change.
> 
> Added screenshot after fix !

Did we also run the pipeline after deployment and was it successful with the data correctly updated in the destination?

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
```diff
@@ -271,6 +273,22 @@ private InvalidEntry<StructuredRecord> getErrorObject(Map result, StructuredReco
   private Object decode(Object object, Schema schema) {
     Schema.Type type = schema.getType();
 
+    Schema.LogicalType logicalType = schema.getLogicalType();
+    if (logicalType != null) {
+      switch (logicalType) {
+        case DECIMAL:
+          BigDecimal bigDecimal = null;
+          if (object instanceof Double) {
```
itsankit-google: Can object also be `instanceof float`?

psainics:
> > > Please also add the screenshot on how the output looks while running preview with the change.
> > 
> > 
> > Added screenshot after fix !
> 
> Did we also run the pipeline after deployment and was it successful with the data correctly updated in the destination?

Yes This has been tested by the QA team , they ran a BQ - JS - BQ pipeline !

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
psainics: I doubt, did not get a float while testing , js has all numbers 64 bits, won't fit in a float i think ...

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
psainics: Updated to `HALF_EVEN` !

core-plugins/src/main/java/io/cdap/plugin/transform/JavaScriptTransform.java
psainics: From [stack-overflow](https://stackoverflow.com/questions/25989642/why-does-java-8-nashorn-javascript-modulo-returns-0-0-double-instead-of-0-i) comments and docs i get it's only `Double or Integer`, I have even tested with int max, int max + 1 and JS native float 32. I got these two always.

Anyway as per  [docs](https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/prog_guide/javascript.html#A1147390) under `3.10 Mapping Data Types` it's suggested we should expect the `Number` type. 

Adding this code snippet should cover all the cases !

```java
if (object instanceof Number) {
  double doubleValue = ((Number) object).doubleValue();
  bigDecimal = BigDecimal.valueOf(doubleValue).setScale(schema.getScale(), RoundingMode.HALF_EVEN);
 }
```

tivv (approved):
LGTM, please clarify in documentation the rounding for decimal schema when converting from javascript. Also fix typo in "Deciaml" in PR title & description

psainics:
Added a note in the documentation !

image

PR #1868, opened 2024-05-27
e2e-RowDenormalizer-ITN

AnkitCLI:
https://github.com/cdapio/cdap-integration-tests/blob/release/6.8/integration-test-remote/src/test/java/io/cdap/cdap/app/etl/batch/RowDenormalizerTest.java


core-plugins/src/e2e-test/resources/testdata/file/expected_outputs/OUTPUT_FOR_ROWDENO_TEST.csv
```diff
@@ -0,0 +1,3 @@
+2,San Francisco
+3,Los Angeles
+1,New York
```
itsankit-google, line 4: why does the output file not have column name `newCity`?

AnkitCLI: added the column name .Previously we are skipping the header now it is enabled.

itsankit-google (approved):

PR #1873, opened 2024-07-08
[PLUGIN-1795] Override Array Max (Apache Poi) for large files

psainics:
## Override Array Max (Apache Poi) for large files

Jira : [PLUGIN-1795](https://cdap.atlassian.net/browse/PLUGIN-1795)

### Description

A soft limit in set in apache poi to avoid OOM errors cause issue when working with large files.
This PR overrides the array max.

[PLUGIN-1795]: https://cdap.atlassian.net/browse/PLUGIN-1795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -186,12 +186,12 @@
     <dependency>
       <groupId>org.apache.poi</groupId>
       <artifactId>poi</artifactId>
-      <version>5.2.4</version>
+      <version>5.2.5</version>
     </dependency>
     <dependency>
       <groupId>org.apache.poi</groupId>
       <artifactId>poi-ooxml</artifactId>
-      <version>5.2.4</version>
+      <version>5.2.5</version>
     </dependency>
     <dependency>
       <groupId>com.github.pjfanning</groupId>
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputFormat.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/source/ExcelInputFormat.java
@@ -41,6 +41,7 @@ import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
 import org.apache.poi.ss.usermodel.WorkbookFactory;
 import org.apache.poi.ss.util.CellReference;
+import org.apache.poi.util.IOUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -66,6 +67,8 @@ public class ExcelInputFormat extends TextInputFormat {
   public static final String FILE_PATTERN = "filePattern";
   public static final String SHEET = "sheet";
   public static final String SHEET_VALUE = "sheetValue";
+  public static final String EXCEL_BYTE_ARRAY_MAX_OVERRIDE = "excel.byteArrayMaxOverride";
+  public static final int EXCEL_BYTE_ARRAY_MAX_OVERRIDE_DEFAULT = Integer.MAX_VALUE / 2;
 
   @Override
   public RecordReader<LongWritable, Text> createRecordReader(InputSplit split, TaskAttemptContext context) {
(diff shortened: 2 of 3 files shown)


itsankit-google (changes requested):
What is the soft limit in apache POI by default? 

How did we came up with this limit?

psainics:
> What is the soft limit in apache POI by default?

Apache poi uses `IOUtils.safelyAllocate` to allocate array of different size, the max len for each array is different based on the type of object.

[Check the code](https://github.com/search?q=repo%3Aapache%2Fpoi%20IOUtils.safelyAllocate&type=code)

int the logs i got `maximum length for this record type is 5,000,000`

 
> How did we came up with this limit?

As we cannot predict the size of file given by user i just wanted to set a big enough number.
INT_MAX / 2 is randomly selected, it looks big enough.

Info on the function [setByteArrayMaxOverride](https://poi.apache.org/components/configuration.html)

itsankit-google:
> > What is the soft limit in apache POI by default?
> 
> Apache poi uses `IOUtils.safelyAllocate` to allocate array of different size, the max len for each array is different based on the type of object.
> 
> [Check the code](https://github.com/search?q=repo%3Aapache%2Fpoi%20IOUtils.safelyAllocate&type=code)
> 
> int the logs i got `maximum length for this record type is 5,000,000`
> 
> > How did we came up with this limit?
> 
> As we cannot predict the size of file given by user i just wanted to set a big enough number. INT_MAX / 2 is randomly selected, it looks big enough.
> 
> Info on the function [setByteArrayMaxOverride](https://poi.apache.org/components/configuration.html)

What tests were performed to decide on this value?

psainics:
> 
> What tests were performed to decide on this value?

We wanted a value bigger than the current default, INT_MAX / 2 is taken as this is a much larger number than the default and it's unlikely any user will hit this limit.

A sanity test was performed by testing a file with similar size used by the user, which failed on the old build with the same error as expected but with the current patch applied the pipeline ran successfully !

itsankit-google:
> > What tests were performed to decide on this value?
> 
> We wanted a value bigger than the current default, INT_MAX / 2 is taken as this is a much larger number than the default and it's unlikely any user will hit this limit.
> 
> A sanity test was performed by testing a file with similar size used by the user, which failed on the old build with the same error as expected but with the current patch applied the pipeline ran successfully !

What was the file size used by the user? And did we test with files of size larger than that? Why not `INT_MAX` then?

psainics:
> > > What tests were performed to decide on this value?
> > 
> > 
> > We wanted a value bigger than the current default, INT_MAX / 2 is taken as this is a much larger number than the default and it's unlikely any user will hit this limit.
> > A sanity test was performed by testing a file with similar size used by the user, which failed on the old build with the same error as expected but with the current patch applied the pipeline ran successfully !
> 
> What was the file size used by the user? And did we test with files of size larger than that? Why not `INT_MAX` then?

The file size used by user was `150MB` 
We have tested for files around that size, not more than this. 

I did not take INT_MAX as per [this post](https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size) , INT_MAX is more than the limit of array you can allocate, there could be more ` + padding ` when allocating array , so it's better to take a value less than INT_MAX, i settled on INT_MAX / 2

itsankit-google:
> > > > What tests were performed to decide on this value?
> > > 
> > > 
> > > We wanted a value bigger than the current default, INT_MAX / 2 is taken as this is a much larger number than the default and it's unlikely any user will hit this limit.
> > > A sanity test was performed by testing a file with similar size used by the user, which failed on the old build with the same error as expected but with the current patch applied the pipeline ran successfully !
> > 
> > 
> > What was the file size used by the user? And did we test with files of size larger than that? Why not `INT_MAX` then?
> 
> The file size used by user was `150MB` We have tested for files around that size, not more than this.
> 
> I did not take INT_MAX as per [this post](https://stackoverflow.com/questions/3038392/do-java-arrays-have-a-maximum-size) , INT_MAX is more than the limit of array you can allocate, there could be more `+ padding` when allocating array , so it's better to take a value less than INT_MAX, i settled on INT_MAX / 2

It might be okay to use `INT_MAX/2` by default if it works with files around 1GB size. Please test this once, also provide a way for users to change this via runtime argument in future.

psainics:
Add ability to override the `byteArrayMaxOverride` with the runtime argument `excel.byteArrayMaxOverride`

itsankit-google (approved):

PR #1904, opened 2025-01-13
[PLUGIN-1828] Validate schema fields during preparerun

psainics:
## Validate schema fields during preparerun

Jira : [PLUGIN-1828](https://cdap.atlassian.net/browse/PLUGIN-1828)

### Description

Validate schema fields during preparerun else it may lead to NPE when a field is passed does not exist.

### Code change

- Modified `GroupByAggregator.java`

[PLUGIN-1828]: https://cdap.atlassian.net/browse/PLUGIN-1828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
@@ -234,6 +234,8 @@ public class GroupByAggregator extends RecordReducibleAggregator<AggregateResult
   @Override
   public void prepareRun(BatchAggregatorContext context) throws Exception {
     super.prepareRun(context);
+    validate(context.getInputSchema(), conf.getGroupByFields(), conf.getAggregates(), context.getFailureCollector());
+    context.getFailureCollector().getOrThrowException();
     LinkedList<FieldOperation> fllOperations = new LinkedList<>();
     // in configurePipeline all the necessary checks have been performed already to set output schema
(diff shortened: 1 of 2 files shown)


psainics:
Unrelated E2E failing with `org.openqa.selenium.StaleElementReferenceException: stale element reference: stale element not found in the current frame` , all tests for group by  plugin are green.

cc : @AnkitCLI 

image

core-plugins/src/main/java/io/cdap/plugin/batch/aggregator/GroupByAggregator.java
```diff
@@ -234,6 +234,8 @@ private void validateCountDistinct(Schema.Field inputField, FailureCollector col
   @Override
   public void prepareRun(BatchAggregatorContext context) throws Exception {
     super.prepareRun(context);
+    validate(context.getInputSchema(), conf.getGroupByFields(), conf.getAggregates(), context.getFailureCollector());
```
itsankit-google, line 237: This will make `validateConditionalFunctions` get called twice in `prepareRun`, can you please remove the second call?

psainics: Removed !

itsankit-google (approved):

PR #1926, opened 2025-01-28
build submodules in github action only

itsankit-google:
There is already a profile for building submodules: 

https://github.com/cdapio/hydrator-plugins/blob/c4526a25483935a2920f752231fa39568ad69d4f/pom.xml#L1006-L1019

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,8 +29,6 @@
   <modules>
     <module>hydrator-common</module>
     <module>amazon-s3-plugins</module>
-    <module>azure</module>
-    <module>condition-plugins</module>
     <module>core-plugins</module>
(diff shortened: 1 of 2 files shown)


masivesky (approved):

pom.xml
```diff
@@ -29,8 +29,6 @@
   <modules>
     <module>hydrator-common</module>
     <module>amazon-s3-plugins</module>
```
psainics, line 31: why did we leave `amazon-s3-plugins` ?

itsankit-google: ah, sorry missed that. will fix it in follow up CL. Thanks for catching that.

PR #1941, opened 2025-02-13
Update azure to fix unit build

psainics:
Update azure to fix unit build


psainics:
https://github.com/cdapio/hydrator-plugins/pull/1942

PR #1945, opened 2025-02-20
[PLUGIN-1867] Error management changes for CSV parser

psainics:
Description
Error Management for CSVParser transform plugin exceptions
https://cdap.atlassian.net/browse/PLUGIN-1867

Code change
Modified CSVParser.java

Tests
Test Case - tested with valid and invalid scenarios

###  1. Success
Screenshot 2025-02-20 at 4 25 00 PM

### 2. Missing CSV values
Screenshot 2025-02-20 at 7 13 34 PM
Screenshot 2025-02-20 at 7 13 45 PM

### 3. Empty CSV values
Screenshot 2025-02-20 at 4 22 03 PM
Screenshot 2025-02-20 at 4 22 16 PM
Screenshot 2025-02-20 at 4 22 42 PM

### DIFF
--- a/transform-plugins/src/main/java/io/cdap/plugin/CSVParser.java
+++ b/transform-plugins/src/main/java/io/cdap/plugin/CSVParser.java
@@ -26,6 +26,9 @@ import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
 import io.cdap.cdap.api.data.schema.Schema.Field;
+import io.cdap.cdap.api.exception.ErrorCategory;
+import io.cdap.cdap.api.exception.ErrorType;
+import io.cdap.cdap.api.exception.ErrorUtils;
 import io.cdap.cdap.api.plugin.PluginConfig;
 import io.cdap.cdap.etl.api.Emitter;
 import io.cdap.cdap.etl.api.FailureCollector;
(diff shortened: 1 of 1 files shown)


transform-plugins/src/main/java/io/cdap/plugin/CSVParser.java
```diff
@@ -224,6 +231,12 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
       } else if (record == null) {
         builder.set(name, null);
       } else {
+        if (i >= record.size()) {
+          String error = String.format("Number of values are not as per output schema. " +
+                                         "Expected %d values, but found only %d values", fields.size(), record.size());
+          throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN),
```
itsankit-google: It will be great if we can log the whole record here.

psainics: Updated !

transform-plugins/src/main/java/io/cdap/plugin/CSVParser.java
```diff
@@ -224,6 +231,13 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
       } else if (record == null) {
         builder.set(name, null);
       } else {
+        if (i >= record.size()) {
+          String error = String.format("Number of values in the record : %s are not as per output schema. " +
+                                         "Expected %d values, but found only %d values",
+                                       record, fields.size(), record.size());
+          throw ErrorUtils.getProgramFailureException(new ErrorCategory(ErrorCategory.ErrorCategoryEnum.PLUGIN),
+                                                      error, error, ErrorType.USER, false, null);
+        }
```
itsankit-google, line 240: why is this an issue? We can ignore the extra columns as expected?

dheeraj-kholia-cs: Updated code to ignore extra column

transform-plugins/src/main/java/io/cdap/plugin/CSVParser.java
```diff
@@ -224,7 +231,7 @@ private StructuredRecord createStructuredRecord(@Nullable CSVRecord record, Stru
       } else if (record == null) {
         builder.set(name, null);
       } else {
-        String val = record.get(i);
+        String val = i < record.size() ? record.get(i) : "";
```
itsankit-google: why do we need this change at all?

What was the error seen?

dheeraj-kholia-cs: Without this check, record.get(i) throws ArrayIndexOutOfBoundsException.

itsankit-google: Then we should catch `ArrayIndexOutofBoundsException` and throw `ProgramFailureException`. Let's keep the behavior same as before.

dheeraj-kholia-cs: done

itsankit-google (approved):
Please squash commits before merge.

PR #1958, opened 2025-04-10
[PLUGIN-1882]Changes done for handling numeric values properly

vikasrathee-cs:
[PLUGIN-1882]
Handle numeric values properly for XLS format. 

Incorrect value is appearing for a column that is expecting integer number.
'.0' is getting appended to each number.

[PLUGIN-1882]: https://cdap.atlassian.net/browse/PLUGIN-1882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
+++ b/format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
@@ -21,7 +21,6 @@ import io.cdap.cdap.api.data.schema.Schema;
 import org.apache.poi.ss.usermodel.Cell;
 import org.apache.poi.ss.usermodel.CellType;
 import org.apache.poi.ss.usermodel.DataFormatter;
-import org.apache.poi.ss.usermodel.DateUtil;
 import org.apache.poi.ss.usermodel.FormulaEvaluator;
 import org.apache.poi.ss.usermodel.Row;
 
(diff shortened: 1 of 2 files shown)


format-xls/src/main/java/io/cdap/plugin/format/xls/input/XlsRowConverter.java
```diff
@@ -111,6 +114,15 @@ private String getCellAsString(Cell cell) {
...
         }
+        if (cell instanceof HSSFCell) {
+          ((HSSFCell) cell).setCellType(CellType.STRING);
+          return cell.getStringCellValue();
+        } else if (cell instanceof StreamingCell) {
+          return cell.getStringCellValue();
+        } else if (cell instanceof XSSFCell) {
+          ((XSSFCell) cell).setCellType(CellType.STRING);
+          return cell.getStringCellValue();
+        }
```
itsankit-google: Calling setCellType(CellType.STRING) mutates the cell — which might not be ideal if you want to preserve original formatting/types. In newer POI versions, this method is deprecated and discouraged. 

Instead, they suggest using:
```
DataFormatter formatter = new DataFormatter();
String value = formatter.formatCellValue(cell);
```

This gives you a string representation without changing the original cell type.

vikasrathee-cs: Changes done and tested for all 3 types.
Screenshot 2025-04-22 at 9 41 49 PM

itsankit-google (approved):

PR #1997, opened 2026-07-20
[PLUGIN-1960] Add SMBv2/v3 support to WindowsShareCopy with legacy SMBv1 fallback

vishwasvaidya-cloudsufi:
Kept the legacy jcifs (SMBv1) library to maintain backward compatibility. The plugin dynamically routes to either SMBv1 or SMBv2/v3, defaulting to the newer protocol for all new pipelines.

Key Changes:

- Backward Compatibility / UI: Added an SMB Protocol Version dropdown to the UI. New pipelines default to SMBv2/v3. Crucially, if the configuration receives a null value (which happens when existing pipelines are upgraded), the plugin safely falls back to SMBv1 to prevent pipeline breakage.

- Dependency Management: Retained jcifs and added smbj along with its explicitly required transitive dependencies (mbassador, asn-one, bcprov) to the pom.xml to ensure CDAP classloader compatibility.

- Modern SMB Architecture: Implemented a secure, hierarchical connection lifecycle (Client -> Session -> Share) for the new smbj execution path.

- Path Normalization: Added logic to convert UNIX-style forward slashes (/) to Windows-native backslashes (\) for the smbj share traversal.

- Directory Traversal: Implemented explicit directory vs. file traversal logic for SMBv2/v3, including filtering for Windows system pointer directories (. and ..).

- Safe File Operations: Applied AccessMask.GENERIC_READ during smbj file operations to prevent accidental file locking on the destination Windows server.

### DIFF
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -111,13 +111,33 @@
       <groupId>commons-lang</groupId>
       <artifactId>commons-lang</artifactId>
     </dependency>
-    <!-- Windows share copy action -->
+    <!-- Legacy SMBv1 Support -->
     <dependency>
       <groupId>jcifs</groupId>
       <artifactId>jcifs</artifactId>
       <version>1.3.17</version>
     </dependency>
-    <!-- End for windows share copy action -->
+    <!-- Modern SMBv2/v3 Support -->
+    <dependency>
+      <groupId>com.hierynomus</groupId>
+      <artifactId>smbj</artifactId>
+      <version>0.11.5</version>
+    </dependency>
+    <dependency>
+      <groupId>net.engio</groupId>
+      <artifactId>mbassador</artifactId>
+      <version>1.3.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.hierynomus</groupId>
+      <artifactId>asn-one</artifactId>
+      <version>0.6.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk15to18</artifactId>
+      <version>1.84</version>
+    </dependency>
     <!-- EMail action -->
     <dependency>
       <groupId>dumbster</groupId>
--- a/core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
+++ b/core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
@@ -20,6 +20,16 @@ import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import com.google.common.io.ByteStreams;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.hierynomus.msdtyp.AccessMask;
+import com.hierynomus.msfscc.fileinformation.FileIdBothDirectoryInformation;
+import com.hierynomus.mssmb2.SMB2CreateDisposition;
+import com.hierynomus.mssmb2.SMB2ShareAccess;
+import com.hierynomus.smbj.SMBClient;
+import com.hierynomus.smbj.auth.AuthenticationContext;
+import com.hierynomus.smbj.connection.Connection;
+import com.hierynomus.smbj.session.Session;
+import com.hierynomus.smbj.share.DiskShare;
+import com.hierynomus.smbj.share.File;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
@@ -41,6 +51,9 @@ import org.slf4j.LoggerFactory;
 import java.io.BufferedOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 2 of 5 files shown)


core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -183,6 +211,122 @@ private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest,
     return name;
   }
 
+  /**
+   * The new SMBv2/v3 logic
+   */
+  private void executeSMBv2(final Path hdfsDir, final FileSystem hdfs) throws Exception {
+    String normalizedSourcePath = config.sourceDirectory;
+    if (normalizedSourcePath.startsWith("/") || normalizedSourcePath.startsWith("\\")) {
+      normalizedSourcePath = normalizedSourcePath.substring(1);
```
vikasrathee-cs: in case of \\, it will remove just one slash and move the normalization logic into config.getSourceDirectory() method

core-plugins/pom.xml
```diff
@@ -111,13 +111,33 @@
...
+      <version>1.3.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.hierynomus</groupId>
+      <artifactId>asn-one</artifactId>
+      <version>0.6.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk15to18</artifactId>
```
vikasrathee-cs, line 138: we are having jdk 8, will there be any concern while using this?

core-plugins/pom.xml
```diff
@@ -111,13 +111,33 @@
...
+    <!-- Modern SMBv2/v3 Support -->
+    <dependency>
+      <groupId>com.hierynomus</groupId>
+      <artifactId>smbj</artifactId>
+      <version>0.11.5</version>
+    </dependency>
+    <dependency>
+      <groupId>net.engio</groupId>
+      <artifactId>mbassador</artifactId>
+      <version>1.3.0</version>
```
vikasrathee-cs, line 129: make sure these libs doesn't have any vulnerabilities

core-plugins/pom.xml
vishwasvaidya-cloudsufi: Bouncy Castle uses a specific naming convention where the suffix indicates the supported Java versions. jdk15to18 explicitly means the artifact is compiled for and fully supports Java 1.5 through Java 1.8. Since we are on JDK 8, this is 100% compatible and exactly the intended artifact for our environment.

core-plugins/pom.xml
vishwasvaidya-cloudsufi: Resolved to remove vulnerabilities in the added dependencies.

core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
vishwasvaidya-cloudsufi: I have moved the normalization logic into a new getSourceDirectory() method in the Config class as you suggested. I also updated the logic to strip all consecutive leading slashes.

core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -183,6 +211,118 @@ private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest,
...
+
+    SMBClient client = new SMBClient();
+    try (Connection connection = client.connect(config.netBiosHostname)) {
+      AuthenticationContext authContext = new AuthenticationContext(
+              config.netBiosUsername,
+              config.netBiosPassword.toCharArray(),
+              config.netBiosDomainName == null ? "" : config.netBiosDomainName
+      );
+
+      Session session = connection.authenticate(authContext);
```
vikasrathee-cs: is this closeable, if so, close it or write in try with resource block

vishwasvaidya-cloudsufi: I updated Session and DiskShare to use the try-with-resources block as suggested. While looking at it, I noticed that SMBClient also holds background network threads and implements AutoCloseable, so I moved SMBClient inside the try block alongside Connection

core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -81,100 +93,119 @@ public void run(ActionContext context) throws Exception {
...
+
+      try (DiskShare share = (DiskShare) session.connectShare(config.netBiosSharename)) {
```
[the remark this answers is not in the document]
vishwasvaidya-cloudsufi: Have handled them now with try with resources.

core-plugins/src/main/java/io/cdap/plugin/batch/action/WindowsShareCopy.java
```diff
@@ -183,6 +211,117 @@ private String copyFileToHDFS(FileSystem hdfs, String smbSourceFile, Path dest,
...
+        try {
+          for (final String filePath : filesToCopy) {
+            completionService.submit(new Callable<String>() {
+              @Override
+              public String call() throws Exception {
+                try {
+                  return copyFileToHDFS_SMBv2(hdfs, share, filePath, hdfsDir);
+                } catch (Exception e) {
+                  LOG.warn("Exception while copying the file {}", filePath, e);
+                  return null;
```
shruthi713: here we are just returning null. don't we have to propagate this exception? Otherwise we might be succeeding it, when it has to fail?   i guess we can remove the try-catch ? pl correct me if I'm wrong

vishwasvaidya-cloudsufi: I've removed the try-catch block so that any network or I/O exceptions will properly propagate up and fail the pipeline, rather than returning null and causing silent data loss.

shruthi713 (approved):

PR #1999, opened 2026-07-29
[🍒] Add SMBv2/v3 support to WindowsShareCopy with legacy SMBv1 fallback for release/2.12

vishwasvaidya-cloudsufi:
[🍒]

🍒 [cherrypick ](https://github.com/cloudsufi/hydrator-plugins/commit/1d559d5e7b918cb9066f63bbade5b51acad1a914)

PR : (https://github.com/cdapio/hydrator-plugins/pull/1997)

Description:

Kept the legacy jcifs (SMBv1) library to maintain backward compatibility. The plugin dynamically routes to either SMBv1 or SMBv2/v3, defaulting to the newer protocol for all new pipelines.

Key Changes:

Backward Compatibility / UI: Added an SMB Protocol Version dropdown to the UI. New pipelines default to SMBv2/v3. Crucially, if the configuration receives a null value (which happens when existing pipelines are upgraded), the plugin safely falls back to SMBv1 to prevent pipeline breakage.

Dependency Management: Retained jcifs and added smbj along with its explicitly required transitive dependencies (mbassador, asn-one, bcprov) to the pom.xml to ensure CDAP classloader compatibility.

Modern SMB Architecture: Implemented a secure, hierarchical connection lifecycle (Client -> Session -> Share) for the new smbj execution path.

Path Normalization: Added logic to convert UNIX-style forward slashes (/) to Windows-native backslashes () for the smbj share traversal.

Directory Traversal: Implemented explicit directory vs. file traversal logic for SMBv2/v3, including filtering for Windows system pointer directories (. and ..).

Safe File Operations: Applied AccessMask.GENERIC_READ during smbj file operations to prevent accidental file locking on the destination Windows server.

### DIFF
--- a/cassandra-plugins/pom.xml
+++ b/cassandra-plugins/pom.xml
@@ -21,7 +21,7 @@
   <parent>
     <groupId>io.cdap.plugin</groupId>
     <artifactId>hydrator-plugins</artifactId>
-    <version>2.12.4</version>
+    <version>2.12.5-SNAPSHOT</version>
   </parent>
 
   <name>Cassandra Plugins</name>
--- a/core-plugins/pom.xml
+++ b/core-plugins/pom.xml
@@ -111,13 +111,33 @@
       <groupId>commons-lang</groupId>
       <artifactId>commons-lang</artifactId>
     </dependency>
-    <!-- Windows share copy action -->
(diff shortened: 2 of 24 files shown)


vikasrathee-cs (approved):
LGTM

